PR #326, opened 2026-06-13
fix(faucet): repair e2e_faucet_workflow (12 failing tests)

### LINKED ISSUE #319: fix(faucet): repair e2e_faucet_workflow (12 failing tests)
## Summary

`botho/tests/e2e_faucet_workflow.rs` has **12 of 14 tests failing** (e.g. `test_faucet_returns_tx_hash`, `test_faucet_dispenses_correct_amount`, `test_faucet_per_address_limit`, `test_faucet_rate_limit_includes_retry_time`, ...). The common failure is the faucet RPC returning an error where the test expects success:

```
thread 'test_faucet_returns_tx_hash' panicked at botho/tests/e2e_faucet_workflow.rs:190:5:
assertion failed: response["error"].is_null()
```

This was failing before the recent consensus work (not a regression from it), so the faucet workflow itself needs investigation — either the faucet/RPC behavior changed, the test harness no longer sets up the faucet's funding/state correctly, or the faucet endpoint contract drifted.

## Investigation steps

1. Run `cargo test -p botho --test e2e_faucet_workflow -- --nocapture` and capture the actual `error` payload the faucet RPC returns for each failing case.
2. Determine root cause: is the faucet wallet unfunded in the test setup? Has the faucet RPC method name / params / response shape changed? Is rate-limiting state leaking across tests (they may need isolation / `#[serial]`)?
3. Fix the faucet code and/or the test harness so the workflow is correct and deterministic.

## Acceptance criteria

- [ ] `cargo test -p botho --test e2e_faucet_workflow` passes (all 14).
- [ ] Tests are deterministic (no cross-test rate-limit/state bleed); add `#[serial]` or per-test isolation if needed.
- [ ] If the faucet code itself was wrong (not just the test), the fix is in `infra/faucet/` or the RPC faucet handler, with the behavior documented.
- [ ] `cargo test -p botho --lib` stays green.

## References

- `botho/tests/e2e_faucet_workflow.rs`
- Faucet implementation: `infra/faucet/`, RPC faucet handler under `botho/src/rpc/`
- Live faucet endpoint context: `faucet.botho.io` (currently not reliably reachable — note in PLAN.md deployment section)

rjwalters:
## Summary

Repairs `botho/tests/e2e_faucet_workflow.rs` (was 12/14 failing). Every failing test hit `Faucet wallet not configured`: the harness enabled the faucet but never gave it a signing wallet or funded the ledger, so each dispense errored before doing any work.

## Root cause

`spawn_faucet_rpc_server` set `state.faucet` but left `state.wallet = None` and the ledger empty. The faucet RPC handler legitimately needs a funded wallet to build and sign a real CLSAG transaction, so it returned an error the tests asserted against as success. The 2 passing tests were the ones that *expect* an error (disabled / missing-param). This was not a regression from the recent consensus work.

Two secondary contract drifts also surfaced once the wallet was wired in:
- The handler emitted `amount` as a JSON **number**, but the documented contract / `FaucetResponse` serializes it as a **string** (picocredits can exceed JSON's safe integer range).
- `faucet_getStatus` computed `dailyDispensed` purely from confirmed on-chain spends, so a freshly dispensed (still-in-mempool) request read as zero.

## Changes

**Test harness** (`e2e_faucet_workflow.rs`):
- Fund a fresh ledger by mining coinbase blocks to a deterministic faucet wallet + a larger decoy batch to a separate wallet (CLSAG rings need >=19 decoys with 10 confirmations), then wire the wallet via `with_faucet()`.
- Mint several independent coinbase UTXOs so sequential requests each consume a distinct UTXO (the faucet skips UTXOs with key images pending in the mempool, so an unconfirmed change output can't be respent without mining).
- Generate valid `tbotho://` testnet recipient addresses (the old `view:hex\nspend:hex` seeds were not valid Ristretto points and parsed as mainnet, which the testnet faucet rejected).
- Give the concurrent-requests test a zero-cooldown config, since all loopback requests share IP `127.0.0.1` and the per-IP cooldown would otherwise serialize them.

**Faucet code** (`botho/src/rpc/mod.rs`) — behavior change:
- Return dispense `amount` as a decimal string via `FaucetResponse::success` so the serialized shape matches the documented contract.
- `faucet_getStatus` reports `max(blockchain-confirmed, in-memory)` `dailyDispensed`, so still-pending mempool dispenses are counted instead of reading as zero until mined.

## Test plan

- [x] `cargo test -p botho --test e2e_faucet_workflow` passes 14/14
- [x] Re-ran 3x total, deterministic (no flakiness)
- [x] `cargo test -p botho --lib` green (922 passed)

Closes #319

### DIFF
--- a/botho/src/rpc/mod.rs
+++ b/botho/src/rpc/mod.rs
@@ -2171,7 +2171,7 @@ async fn handle_faucet_request(
     state: &RpcState,
     client_ip: Option<std::net::IpAddr>,
 ) -> JsonRpcResponse {
-    use faucet::{FaucetError, FaucetRequest};
+    use faucet::FaucetRequest;
 
     // Check if faucet is available
     let faucet = match &state.faucet {
@@ -2401,15 +2401,15 @@ async fn handle_faucet_request(
     // Broadcast to WebSocket subscribers
     state.ws_broadcaster.new_transaction(&tx_hash, fee, None);
 
-    JsonRpcResponse::success(
-        id,
-        json!({
-            "success": true,
-            "txHash": tx_hash_hex,
-            "amount": amount,
-            "amountFormatted": format!("{:.6} BTH", amount as f64 / 1_000_000_000_000.0),
-        }),
-    )
+    // Build the response via FaucetResponse so the serialized shape stays in
+    // sync with the documented faucet contract. In particular `amount` is
+    // serialized as a decimal string (picocredits can exceed JSON's safe
+    // integer range), which is what API consumers and the e2e tests expect.
+    let response = faucet::FaucetResponse::success(tx_hash_hex, amount);
+    match serde_json::to_value(&response) {
+        Ok(value) => JsonRpcResponse::success(id, value),
+        Err(e) => JsonRpcResponse::error(id, -32000, &format!("Failed to serialize response: {}", e)),
+    }
 }
 
 /// Faucet dispense statistics calculated from blockchain.
@@ -2573,7 +2573,16 @@ async fn handle_faucet_status(id: Value, state: &RpcState) -> JsonRpcResponse {
                     }
                 };
                 let dispense_stats = calculate_dispensed_from_blockchain(&wallet, &ledger, faucet, stats.amount_per_request);
-                (dispense_stats.daily_dispensed, dispense_stats.lifetime_dispensed)
+                // The blockchain reflects only *confirmed* dispenses. Faucet
+                // transactions sit in the mempool until mined, so a freshly
+                // dispensed request would otherwise report zero. The in-memory
+                // counter (incremented on every successful dispense) captures
+                // these still-pending amounts, so report the larger of the two
+                // to avoid under-counting today's committed dispenses.
+                (
+                    dispense_stats.daily_dispensed.max(stats.daily_dispensed),
+                    dispense_stats.lifetime_dispensed,
+                )
             } else {
                 // Fall back to in-memory counter if no wallet
                 (stats.daily_dispensed, 0)
--- a/botho/tests/e2e_faucet_workflow.rs
+++ b/botho/tests/e2e_faucet_workflow.rs
@@ -12,7 +12,7 @@
 
 mod common;
 
-use std::{net::SocketAddr, sync::Arc, time::Duration};
+use std::{net::SocketAddr, sync::Arc, time::Duration, time::SystemTime};
 
 use reqwest::Client;
 use serde_json::{json, Value};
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the faucet repair. Both the test-harness fix and the two faucet code changes are sound.

## Test Execution (local, against PR head c0921af)

- `cargo test -p botho --test e2e_faucet_workflow` — ✅ **14/14 passed**, ran **twice** (18.49s / 19.40s), deterministic, zero flakiness.
- `cargo test -p botho --lib` — ✅ **922 passed**, 0 failed (matches the stated baseline).

## Faucet CODE change #1 — `amount` as a decimal string

**Verdict: correct, and not a breaking change.** This is not a new contract — it *fixes drift*. The canonical `FaucetResponse` struct (`botho/src/rpc/faucet.rs:439`) already declares `amount: Option<String>` with `#[serde(rename_all = "camelCase")]`. The old inline `json!` block in the handler was the outlier, emitting `amount` as a raw JSON number. The handler now builds the response *via* `FaucetResponse::success`, so the wire shape matches the type the rest of the codebase already documents. Justification (picocredits, e.g. 10_000_000_000_000, exceed JSON's 2^53 safe-integer range) is legitimate. Verified the only on-the-wire consumer, `infra/faucet/web/js/faucet.js:322`, reads `result.txHash || result.tx_hash` and never reads `result.amount` — so no client breaks. `txHash`/`amountFormatted` field names and the `{:.6} BTH` / 1e12 formatting are byte-for-byte identical to the old output. Serialization errors are handled (returns a JSON-RPC error, no panic).

## Faucet CODE change #2 — `dailyDispensed = max(confirmed, in-memory)`

**Verdict: correct; does not double-count.** Both counters track the *same* dispense events at the same granularity: the blockchain calc is `daily_count * faucet_amount` (confirmed spends), and the in-memory `AtomicU64` is `fetch_add(amount)` per successful dispense. A given dispense is either pending (in-memory only) or confirmed (reflected in both). Taking `max()` — not `sum()` — is exactly the right operator: it surfaces still-mempool dispenses without adding a confirmed dispense twice.

One minor, non-blocking edge: `maybe_reset_daily_counter()` runs only on the dispense path (`check_rate_limit`), not in `stats()`. Across a UTC-midnight boundary with no new dispense, the in-memory counter can briefly hold yesterday's (higher) total while the blockchain calc correctly shows today's, so `max()` would over-report on `faucet_getStatus` until the first dispense of the new day self-corrects it. Cosmetic, status-endpoint-only, bounded, and pre-existing in nature. Filed as a follow-up rather than blocking.

## Test harness

Exercises the **real** production path — `with_faucet()` wires a genuine funded `Wallet` + `FaucetState`; dispenses build and sign real CLSAG transactions (ring size 20, decoys with 10 confirmations from a dedicated decoy wallet) and hit the real mempool and rate limiter. It does not stub or bypass faucet validation, so it does not mask a faucet bug. The root cause was confirmed to be harness-side: the old `view:hex\nspend:hex` seeds were invalid Ristretto points that parsed as mainnet and were correctly rejected by the testnet faucet; the fix generates valid `tbotho://` testnet addresses. `TRIVIAL_DIFFICULTY` correctly equals the chain's initial difficulty (audit cycle 6 C1). The zero-cooldown config is scoped to only the concurrent-requests test and is justified — all loopback requests share `127.0.0.1`, so the per-IP cooldown would otherwise serialize them; the dispense/balance/signing validation still runs. The cooldown and per-address-limit tests retain the default config, so that behavior stays genuinely tested.

## CI note (non-blocking)

The `Compile Benchmarks` check fails, but it is an **unrelated infrastructure failure**: the runner is missing the system library `glib-2.0` (`pkg-config ... glib-2.0` exits 1 in the `glib-sys` build script). This PR changes only `botho/src/rpc/mod.rs` and `botho/tests/e2e_faucet_workflow.rs` — no `Cargo.lock`/`Cargo.toml` or GUI/glib dependency changes — so this would fail identically on `main`. `Benchmark Summary` passes; the rest skip. Recommend the maintainer install `libglib2.0-dev` on the benchmark runner (separate infra fix), and that this does not gate the faucet repair.

Closes #319. Nice, careful work.

PR #338, opened 2026-06-14
test(fuzz): revive stale fuzz targets + add PR build guard

rjwalters:
Closes #336

## Summary

The `fuzz/` crate is excluded from the root workspace, so it never builds during a normal `cargo build`/`cargo test` and had bit-rotted against the cycle-5/6 + I4 consensus work. This revives all 11 cargo-fuzz targets against the current API, adds a PR-triggered build-only CI guard so it cannot silently rot again, and documents the toolchain contract.

## Per-target build status (`cargo fuzz build`, pinned `nightly-2025-12-03`)

All 11 targets build clean (RC=0):

| Target | Build | Notes |
|--------|-------|-------|
| fuzz_transaction | clean | fixed: removed `TxInputs::is_ring()` (gone); use `len`/`is_empty`/`key_images`/`clsag` |
| fuzz_ring_signature | clean | fixed: dropped removed `Blinding` import; `Commitment::to_bytes()` via `ReprBytes` (no `as_bytes`); imported `Digest` for `Blake2b256::digest` |
| fuzz_mlkem_decapsulation | clean | fixed: rewrote to current ml-kem API (`MlKem768Ciphertext::from_bytes`, `decapsulate -> Result`, encapsulate on public key, compare shared secrets via `as_bytes()`, `ML_KEM_768_*_BYTES` consts) |
| fuzz_clsag | clean | already current |
| fuzz_pq_transaction | clean | already current |
| fuzz_block | clean | already current |
| fuzz_network_messages | clean | already current |
| fuzz_rpc_request | clean | already current |
| fuzz_address_parsing | clean | already current |
| fuzz_pq_keys | clean | already current |
| fuzz_subaddress | clean | already current |

No target was deleted to make the build pass.

## Toolchain finding (important)

`cargo +nightly fuzz build` (rolling nightly, 1.98.0-nightly 2026-06-08) fails: a transitive `hashbrown 0.14.5` (via `hashlink` -> `libp2p-gossipsub`) does not compile on that compiler (`error: cannot specialize on trait Copy`). The repo pins `nightly-2025-12-03` in `rust-toolchain`, and the whole workspace already builds against it. The fix is to drive cargo-fuzz with the **pinned** channel — invoke `cargo fuzz build` (no `+nightly`), which `rust-toolchain` honors. CI and the README are updated accordingly.

## Per-target run status

[...]

Issue #333, opened 2026-06-13
fix(consensus): widen supply accumulators to u128 (fix u64 emission overflow)


rjwalters:
## Summary — CRITICAL (supply-integrity / consensus)

The chain-state supply accumulators are `u64`, but the intended Phase-1 emission exceeds `u64::MAX`. Because `[profile.release] overflow-checks = false`, the accumulator **wraps silently in production** once cumulative emission passes ~18.4M BTH. This is consensus-critical: `total_mined` feeds `calculate_block_reward()`, so a wrapped value corrupts the *emission schedule itself* and diverges nodes — not merely a metrics counter.

**Decision (settled):** keep the **picocredit** unit (1 BTH = 1e12; wired through the entire tx layer via `PICOCREDITS_PER_CREDIT` — changing it is a system-wide rescale that would risk silently breaking the calibrated `cluster-tax` economic model). Fix the overflow by widening the **cumulative monetary accumulators to `u128`**. A nanoBTH unit change was considered and rejected for this reason.

## The bug

- `botho/src/ledger/mod.rs:62,66` — `ChainState.total_mined: u64`, `total_fees_burned: u64`.
- `botho/src/ledger/store.rs` — `add_block` computes `new_total_mined = state.total_mined + reward` (wraps in release at `overflow-checks=false`); persisted as 8-byte LE at `META_TOTAL_MINED` / `META_FEES_BURNED` (store.rs ~707/713), decoded via `u64::from_le_bytes` (store.rs ~242/249).
- `botho/src/node/mod.rs:125` — `calculate_block_reward(state.height + 1, state.total_mined)` consumes `total_mined`; a wrapped value corrupts emission → **consensus divergence**.

### Numbers
- `mainnet_policy()`: 50 BTH initial reward (picocredits), 12,614,400-block (~2yr) halving × 5 → Phase-1 emission ≈ **1.22e21 pico ≈ 1.22B BTH**.
- `u64::MAX` ≈ 1.84e19 pico ≈ **18.4M BTH** → supply overflows at ~1.5% of Phase 1.
- Even the docs' "~100M BTH" target (1e20 pico) exceeds `u64::MAX`, so *any* realistic supply in picocredits requires `u128` accounting.

## Fix — widen the cumulative monetary accumulators to `u128`

**Widen to `u128` (provably cross `u64::MAX`):**
- `ChainState.total_mined`
- `ChainState.total_fees_burned`

[...]

PR #339, opened 2026-06-14
test(fuzz): add 4 consensus-critical fuzz targets (add_block, multi-input balance, lottery, cluster-tax math)

### LINKED ISSUE #337: test(fuzz): extend fuzzing to consensus-critical paths (add_block, multi-input balance, lottery, cluster-tax math)
## Summary

The fuzz suite does not cover the **consensus-critical paths added in the cycle-5/6 and I4 work** — exactly the code where a panic, overflow, or missed-validation is most dangerous in a no-fork project. Add fuzz targets for these surfaces, run them, and triage crashes. (Depends on the revive issue making the fuzz crate build cleanly first.)

A motivating example: the monetary/emission path has a known `u64` supply overflow (#333) that a fuzzer over that code would likely surface — these targets are the systematic version of that.

## New targets (with the EXACT invariant each must assert)

Getting the invariant right is the whole point — a target that asserts the wrong thing gives false confidence. For each, derive an arbitrary input (use `arbitrary`/`Arbitrary` or manual byte decoding), drive the real code, and assert:

1. **`fuzz_add_block`** — feed an arbitrary `Block` to a fresh `Ledger::add_block` (against a real genesis ledger).
   - Invariant: **never panics**; returns either `Ok(())` or a typed `LedgerError`. On `Ok`, post-state invariants hold: `height == prev+1`, `total_mined == prev + reward`, and supply conservation (`total_mined == sum(UTXO values) + total_fees_burned + lottery_pool`). This is the C1–C4 surface — a malformed block must be rejected, never accepted-with-broken-invariants and never crash.

2. **`fuzz_multi_input_balance`** — arbitrary multi-input `Transaction` → `verify_ring_signatures`.
   - Invariant: **never panics / never integer-overflows** in the balance sum; the function returns `Ok` only when `sum(pseudo_output_amount) == sum(outputs) + fee` (I4). Feed mismatched sums, huge u64s near overflow, 0/1/many inputs.

3. **`fuzz_lottery_validation`** — arbitrary `block.lottery_summary` + `lottery_outputs` + candidate set → `validate_block_lottery`.
   - Invariant: **never panics**; deterministic (same input → same result); accepts only splits matching the pool accounting (80/20) and rejects everything else. Feed degenerate candidate sets, zero/overflow payouts, empty/huge outputs.

4. **`fuzz_cluster_tax_math`** — arbitrary inputs to the `cluster-tax` fixed-point primitives: `demurrage_charge`, the integer fee-curve / sigmoid, cluster-factor LUT, and the `monetary` emission/supply functions (`calculate_block_reward`, tail reward, `lottery_emission_share`).
   - Invariant: **never panics, never overflows** (checked/saturating arithmetic), outputs within documented bounds (e.g. factor in [1000, 6000], demurrage is a fee floor ≥ 0). This target is the most likely to surface a #333-class overflow.

## Run + triage (required, not optional)

- Run each new target `cargo +nightly fuzz run <target> -- -max_total_time=300` (5 min each; longer if cheap).
- **Crash policy:** a crash in consensus/economics code is a finding, not a nuisance — minimize it (`cargo fuzz tmin`), and **file a focused bug issue** with the minimized input rather than silencing the assertion or weakening the target. Only "fix in place" if it's an obvious, safe input-validation gap (should-return-Err-not-panic) with a regression test. Never weaken an invariant to make a target pass.
- Seed `fuzz/corpus/<target>/` with a few valid + boundary inputs.

## Acceptance criteria

- [ ] The four targets build (`cargo +nightly fuzz build`) and are registered in `fuzz/Cargo.toml` + the CI matrix.
- [ ] Each ran ≥300s; every crash is either filed as a bug (with minimized input) or fixed with a regression test — none suppressed.
- [ ] The asserted invariants match this spec (judge will check they are meaningful, not tautological).
- [ ] `cargo test -p botho --lib` stays green.
- [ ] PR reports per-target run stats and links every bug filed.

## References
- C1–C4: `botho/src/ledger/store.rs` (`add_block`)
- I4 balance: `botho/src/transaction.rs` (`verify_ring_signatures`)
- Lottery: `botho/src/consensus/lottery.rs` (`validate_block_lottery`)
- Math: `cluster-tax/src/{demurrage,fee_curve,monetary}.rs`; `botho/src/monetary.rs`
- Known overflow this should catch: #333

rjwalters:
Closes #337

## Summary

Extends the revived fuzz suite (#338) to the consensus/economics paths added in cycle-5/6 and I4 work. Adds 4 new cargo-fuzz targets, registers them in `fuzz/Cargo.toml` and a new CI matrix job, and seeds each corpus with valid + boundary inputs. Each target asserts the EXACT invariant from the issue (verified non-tautological).

## New targets, invariants as implemented

1. **`fuzz_add_block`** (`fuzz/fuzz_targets/fuzz_add_block.rs`) — arbitrary `Block` (bincode-decoded) -> fresh-genesis `Ledger::add_block`. Asserts: never panics; result is `Ok(())` or a typed `LedgerError`; on `Ok`, `height == prev+1`, `total_mined == prev + minting_tx.reward`, and supply conservation `total_mined == Sigma(UTXO values) + total_fees_burned + lottery_pool` (computed in u128). Surfaces C1-C4 (`botho/src/ledger/store.rs`).
2. **`fuzz_multi_input_balance`** (`.../fuzz_multi_input_balance.rs`) — arbitrary multi-input `Transaction` (public `ClsagRingInput`/`TxOutput` fields) -> `verify_ring_signatures`. Asserts: never panics / never integer-overflows the balance sum; `Ok` implies `Sigma(pseudo_output_amount) == Sigma(outputs) + fee` exactly (u128 ground truth). Zero-input case reaches the balance path (incl. `total_output()` sum, an overflow-trap candidate). I4 (`botho/src/transaction.rs`).
3. **`fuzz_lottery_validation`** (`.../fuzz_lottery_validation.rs`) — arbitrary `lottery_summary` + `lottery_outputs` + candidate set -> `validate_block_lottery`. Asserts: never panics; deterministic (called twice, results compared); an accepted block's split matches the independently-recomputed `compute_pool_accounting` 80/20 contract — winner branch distributes the capped payout, no-winner branch distributes 0 with the pool carried over. (`botho/src/consensus/lottery.rs`).
4. **`fuzz_cluster_tax_math`** (`.../fuzz_cluster_tax_math.rs`) — `demurrage_charge`, integer fee curve/sigmoid (`FeeCurve`/`ClusterFactorCurve`), cluster-factor LUT, and emission/supply functions (`calculate_block_reward`, `calculate_tail_reward`, `lottery_emission_share`). Asserts: never panics/overflows; cluster factor in `[1000, 6000]`; sigmoid `<= SIGMOID_SCALE`; demurrage exempt for factor-1/zero-rate/zero-elapsed and bounded by transfer value within one-year horizon; `lottery_emission_share(h, r) <= r`. Most likely to surface the #333-class monetary overflow.

## Harness-wiring sanity tests (optional, included)

Added `#[cfg(test)]` tests in the botho crate that call each target's core logic on a valid + malformed input:
- `botho/src/ledger/store.rs::fuzz_wiring_add_block_rejects_malformed_and_conserves_supply`
- `botho/src/transaction.rs::fuzz_wiring_multi_input_balance_invariant`
- `botho/src/consensus/lottery.rs::fuzz_wiring_lottery_validation_deterministic_and_split_checked`, `fuzz_wiring_cluster_tax_math_bounds`

These caught and corrected an initially over-strict lottery invariant (no-winner blocks legitimately return `Ok` with `pool_distributed == 0` while `payout > 0`); the fix is reflected in both the test and `fuzz_lottery_validation`.

## CI registration

New `fuzz-consensus` job in `.github/workflows/fuzz.yml` (wired into `fuzz-summary` `needs` + the crash-result check), matrix containing all 4 targets:
- `fuzz_add_block`
- `fuzz_multi_input_balance`
- `fuzz_lottery_validation`
- `fuzz_cluster_tax_math`

The PR-triggered `fuzz-build` guard (no matrix) builds all targets via `cargo fuzz build` on the pinned toolchain.

## Build status

- `cargo fuzz build` (all targets, pinned `nightly-2025-12-03`, no `+nightly`) succeeds on macOS.
- `cargo test -p botho --lib` green (926 passed, up from 922 baseline).

## Runs are CI-deferred (macOS cannot run cargo-fuzz)

Per the issue's run constraint, libfuzzer binaries cannot run on the macOS dev host (ASan runtime deadlocks before main on macOS 15; `-s none` still injects sancov coverage which fails to link via `if-watch` on arm64). This PR WRITES + BUILDS + REGISTERS the targets correctly; the required >=300s runs and crash minimization/triage happen on the Linux `fuzz.yml` workflow after merge (each target is in the `fuzz-consensus` matrix job). If a run surfaces a crash (e.g. a #333-class overflow), it should be minimized (`cargo fuzz tmin`) and filed as a focused bug rather than silencing the assertion.

## Test plan
- [x] `cargo fuzz build` (all 12 targets) succeeds
- [x] `cargo test -p botho --lib` green
- [ ] CI `fuzz-build` guard passes on this PR
- [ ] Scheduled/dispatch run of the `fuzz-consensus` matrix (post-merge, Linux)


rjwalters:
## Judge review — changes requested (1 over-strict invariant)

Thorough pass on all 4 targets against the real code. Three of the four invariants are correct and meaningfully non-tautological. One target (`fuzz_cluster_tax_math`) carries an **over-strict assertion that will false-crash the Linux fuzz run on valid behavior** — exactly the failure mode the issue warns against ("a target that asserts the wrong thing gives false confidence"; here it's the inverse — a false crash). That one assertion must be fixed before merge.

### Blocking: `fuzz_cluster_tax_math.rs` — demurrage value-bound is over-strict

```rust
if m.blocks_per_year > 0 && m.elapsed_blocks <= m.blocks_per_year {
    assert!(
        charge <= m.transfer_value,
        "demurrage_charge {} exceeds transfer_value {} within one-year horizon ...",
        charge, m.transfer_value
    );
}
```

The "charge ≤ value within one-year horizon" property only holds when `rate_bps ≤ 10000` (≤100%/yr). But `m.rate_bps` is `Arbitrary`-derived across the full `u32` range, independently of the other fields. With `elapsed_blocks == blocks_per_year` and `cluster_factor == MAX_FACTOR_SCALED` (max progressivity), the in-horizon charge is `transfer_value * rate_bps / 10000`, which exceeds `transfer_value` for any `rate_bps > 10000`.

Concretely reproduced against the real `demurrage_charge` body (`cluster-tax/src/demurrage.rs:60`):

- `demurrage_charge(1_000_000, 6000, 6_307_200, 20000, 6_307_200)` -> `2_000_000` (> 1_000_000)
- `demurrage_charge(1_000_000, 6000, 6_307_200, 10001, 6_307_200)` -> `1_000_100` (> 1_000_000)

`demurrage_charge` is behaving correctly here — it's the harness feeding an absurd 200%/yr rate that real callers (`MonetaryPolicy::demurrage_rate_bps` = 200 bps) never produce. The all-`0xFF` corpus seed (`fuzz/corpus/fuzz_cluster_tax_math/seed_1`) decodes to large field values, so libfuzzer would surface this within seconds and report it as a crash — defeating the target's purpose.

Fix options (pick one):
1. Gate the bound on the realistic rate domain, e.g. `if m.blocks_per_year > 0 && m.elapsed_blocks <= m.blocks_per_year && m.rate_bps <= 10_000 && m.cluster_factor <= MAX_FACTOR_SCALED` (keeps a meaningful check on the realistic input space), or clamp `rate_bps`/`cluster_factor` to documented ranges before the assertion. Note the exempt-case block above (`charge == 0`) is correct and should stay as-is.
2. Drop the value-bound assertion entirely and keep only the no-panic/no-overflow guarantee plus the exempt-zero check for this primitive (the `u64::try_from(...).unwrap_or(u64::MAX)` saturation already covers the overflow concern). Option 1 is preferred — it still catches a #333-class break in the realistic domain.

### The other three targets are correct — verified against source

1. **`fuzz_add_block`** — correct. `Ledger::add_block` enforces `minting_tx.reward == calculate_block_reward(...)` (`ledger/store.rs:455`) and sets `new_total_mined = state.total_mined + block.minting_tx.reward` (`store.rs:581`), so the `height+1` and `total_mined+reward` post-conditions are real guarantees, and the u128 supply-conservation sum (`total_mined == Σutxo + burned + pool`) matches the ledger's bookkeeping. Asserts only on the `Ok` branch; rejections only require no-panic. Non-tautological and meaningful.

2. **`fuzz_multi_input_balance`** — correct. `verify_ring_signatures` (`transaction.rs:1101`) returns `Ok` only when `Σpseudo == total_output() + fee` with checked arithmetic. The harness asserts `Ok ⟹ balanced` (u128 ground truth) and deliberately does NOT assert the converse (a balanced tx can still be rejected on a bad signature). Zero-input bias correctly drives the `total_output()`/`checked_add(fee)` overflow path.

3. **`fuzz_lottery_validation`** — correct, including the no-winner/winner split that the builder reported fixing. Mirrors `validate_block_lottery` (`lottery.rs:459`) exactly: no-winner branch (`lottery_outputs.is_empty()`) requires `pool_distributed == 0` + `amount_burned == fee_burn` (`:483`), and a no-winner block with `payout > 0` legitimately returns `Ok` (the pool carries over via `carryover_after(0)`, `:509`). The harness's conditional — `pool_distributed == 0` for empty outputs, `== accounting.payout` otherwise — matches both branches and is NOT over-strict. `compute_pool_accounting` is recomputed from independent ground-truth inputs, so the cross-check is non-tautological. Determinism check is sound.

The `lottery_emission_share(h, r) <= r` and cluster-factor `[1000, 6000]` / `rate_bps ∈ [r_min, r_max]` / `compute_fee` (`fee <= amount`) assertions in target 4 are all correct against source (`monetary.rs:227-246`, `fee_curve.rs:604-611,1584-1605`); only the demurrage value-bound is wrong.

### Other checks
- Registration in `fuzz/Cargo.toml` (4 `[[bin]]` entries + `bth-cluster-tax`/`tempfile` deps) and the `fuzz-consensus` CI matrix (+ `fuzz-summary` needs + crash-check) is complete and correct.
- Corpus seeds present for all 4 targets (valid + boundary, incl. all-`0xFF`).
- CI-deferred runs are acceptable per the macOS constraint — not a blocker.
- The failing "Compile Benchmarks" check is an unrelated pre-existing env issue (missing system `glib-2.0` for a Tauri dep), not attributable to this PR.

Fix the one demurrage assertion (option 1 preferred) and re-request review. Everything else is ready.

rjwalters:
Addressed the judge's blocking feedback (option 1: gate on realistic domain).

**`fuzz/fuzz_targets/fuzz_cluster_tax_math.rs`** — the demurrage value-bound assertion now only fires when `m.rate_bps <= 10_000` (<=100%/yr) AND `m.cluster_factor <= MAX_FACTOR_SCALED` (6000). Outside that domain the charge legitimately exceeds the transfer value (a 200%/yr rate genuinely takes 2x within one year), so only the universal no-panic/no-overflow guarantee applies there. The exempt-zero (factor-1 / zero-rate / zero-elapsed) check is unchanged, and the bound still catches a #333-class break across the realistic input space. Verified the gate constant against `cluster-tax/src/demurrage.rs` (`MAX_FACTOR_SCALED = 6 * FACTOR_SCALE = 6000`, and the charge formula at `:78,88-93`).

No production code touched; the other three targets are untouched.

Verification:
- `cargo fuzz build fuzz_cluster_tax_math` — builds clean (pinned `nightly-2025-12-03`, no `+nightly`; only pre-existing unrelated dead-code warnings).
- `cargo test -p botho --lib` — 926 passed, 0 failed.
- Fuzz runs remain CI-deferred per the macOS constraint.

Re-requesting review.

rjwalters:
## Judge re-review (2nd pass) — Approved

The single blocking item from the first pass (over-strict demurrage value-bound in `fuzz_cluster_tax_math`) is correctly resolved by the Doctor's commit `5e709ef`. Verdict: **approve**.

### 1. Gating is correct — no more false crash
The value-bound assertion is now gated on `m.blocks_per_year > 0 && m.elapsed_blocks <= m.blocks_per_year && m.rate_bps <= 10_000 && m.cluster_factor <= MAX_FACTOR_SCALED`. Verified against the real charge formula (`cluster-tax/src/demurrage.rs:71,73,88-93`):

```
charge = value × rate_bps/10_000 × progressivity/5000 × elapsed/blocks_per_year
```

The clamp at `:71` forces `progressivity = clamp(factor,1000,6000) - 1000 ∈ [0,5000]`, so `progressivity/5000 ≤ 1` unconditionally; with `elapsed ≤ blocks_per_year` the time fraction is `≤ 1`. Therefore `charge ≤ value × rate_bps/10_000`, which is `≤ value` exactly when `rate_bps ≤ 10_000`. The `rate_bps <= 10_000` gate is the necessary-and-sufficient bound; `cluster_factor <= MAX_FACTOR_SCALED` is a (harmless, redundant given the internal clamp) conservative tightening. The all-`0xFF` corpus seed decodes to `rate_bps = u32::MAX`, which the gate now excludes — so the false crash the first pass flagged (`demurrage_charge(1_000_000, 6000, 6_307_200, 20000, 6_307_200) = 2_000_000`) no longer triggers the assertion. Confirmed on the Linux run: Fuzz Build Guard green.

### 2. No-panic / overflow invariant NOT weakened
`demurrage_charge(...)` is still called unconditionally at `:111-117` on full-range inputs, before any gate. Only the *value-bound `assert!`* (`:151`) sits inside the new domain condition. The exempt-zero check (`:119-129`) is untouched and remains universal. The saturating `u64::try_from(...).unwrap_or(u64::MAX)` continues to cover the overflow concern for every input.

### 3. Other 3 targets unchanged
`git show --stat 5e709ef` confirms the fix commit touches only `fuzz/fuzz_targets/fuzz_cluster_tax_math.rs` (44 lines). `fuzz_add_block`, `fuzz_multi_input_balance`, and `fuzz_lottery_validation` are byte-identical to the previously-verified state.

### 4. Build + tests green
- `cargo fuzz build fuzz_cluster_tax_math` (pinned `nightly-2025-12-03`, no `+nightly` in CI) — builds clean (only pre-existing unrelated dead-code warnings).
- `cargo test -p botho --lib` — 926 passed, 0 failed.

### CI / merge state
- **Fuzz Build Guard (PR): SUCCESS**, **Fuzz Summary: SUCCESS** — the gating checks for this PR pass.
- "Compile Benchmarks" FAILURE is the pre-existing, unrelated environmental gap (system `glib-2.0`/`glib-sys` build-script for a Tauri GUI dep — see run log), not attributable to this fuzz-only PR.
- Branch is 1 commit behind `origin/main` (only `fuzz_subaddress.rs`, no overlap with this PR) — no conflict; rebase/merge will resolve the `UNSTABLE` state cleanly before merge.

Approving and labeling `loom:pr`.

Issue #340, opened 2026-06-14
CRITICAL fix(consensus): total_output() unchecked u64 overflow → inflation via wrap (release)


rjwalters:
## Summary — CRITICAL (inflation)

`Transaction::total_output()` and `QuantumPrivateTransaction::total_output()` sum output amounts with an **unchecked `u64` `.sum()`**. Because `Cargo.toml` sets `overflow-checks = false` for `[profile.release]`, this **wraps silently in production**, defeating the I4 balance check and allowing a transaction to mint outputs vastly exceeding its inputs.

Found by fuzzing: `fuzz_multi_input_balance` and `fuzz_pq_transaction` both crashed with `attempt to add with overflow` in `core::iter::Sum` (the debug build panics; release wraps).

## The bug

- `botho/src/transaction.rs:1035-1036`
  ```rust
  pub fn total_output(&self) -> u64 {
      self.outputs.iter().map(|o| o.amount).sum()   // UNCHECKED — wraps in release
  }
  ```
- `botho/src/transaction_pq.rs:518-519` — identical pattern.

`verify_ring_signatures` (the I4 balance check, transaction.rs ~1101) does:
```rust
let output_total = self.total_output().checked_add(self.fee).ok_or("...overflow")?;
```
The `checked_add(fee)` is useless here: `total_output()` has **already wrapped** before it returns, so `checked_add` operates on the wrapped (small) value and succeeds.

## Exploit (release build)

1. Craft outputs `[2^63, 2^63, V]` → true value `2^64 + V`, but `total_output()` returns `V` (wrapped).
2. `output_total = V + fee`.
3. Supply inputs (real UTXOs, bound to honest amounts by CLSAG + the C3 ring-member check) whose `pseudo_output_amount` sum is `V + fee`. `input_total = V + fee` (checked, honest).
4. Balance check `input_total == output_total` → **passes**.
5. The tx is accepted (the same path runs at block acceptance: `add_block → verify_transaction → verify_ring_signatures`), creating two ~9.2-quintillion-picocredit outputs while spending only `V + fee`. **Massive inflation.**

The C3 ring-member binding constrains *input* amounts to real UTXOs but places no bound on the *output* sum, so it does not prevent this.

## Fix

Make the output sum overflow-safe so an overflowing tx is **rejected**, never wrapped:
[...]

Issue #323, opened 2026-06-13
infra: redeploy + reset testnet on current consensus code (protocol 2.0.0, multi-seed)


rjwalters:
## Summary

The deployed testnet is **stale and single-node**, and the recent consensus-breaking changes require a coordinated reset before a meaningful public testnet. This is an **operator/infra-led** epic (requires AWS/DNS access), with some automatable preparation.

## Current state (observed)

`https://seed.botho.io/rpc` → `node_getStatus`:
- `network: botho-testnet`, `nodeVersion: 0.1.0`
- `gitCommit: 24a08f3` (build ~2026-01-12) — **predates cycles 5 & 6** (no lottery/demurrage, no C1–C4 block-acceptance hardening, no I4 multi-input fix, no M4 accounting, ml-dsa pre-0.1.1)
- `chainHeight: 1320`, `totalTransactions: 3`, `peerCount: 0`, `mintingActive: false`
- `faucet.botho.io` — not reliably reachable

So the live chain is incompatible with current `main` (which is now **protocol 2.0.0**, consensus-breaking on multiple axes) and has no peers.

### Automatable prep (code/config — could be a builder PR)
- [ ] Verify/refresh the chain-reset + genesis tooling: `infra/seed/reset-chain.sh`, `reset-to-testnet.sh`, genesis params.
- [ ] Confirm `botho-testnet` genesis magic / network params match current `block.rs`.
- [ ] Multi-seed bootstrap config (≥3 regional seeds) + DNS-seed discovery scaffolding (PLAN.md "Network Bootstrap Strategy").
- [ ] Build a current release binary from `main` (v0.1.0 tag is pre-merge of cycle-6/I4 — cut a fresh tag or build from HEAD) via the reproducible release workflow.

### Operator steps (human — AWS/DNS, not automatable)
- [ ] Deploy the current binary to seed.botho.io (and additional regions).
- [ ] Reset the chain to a fresh genesis on the new protocol.
- [ ] Bring up ≥1 additional regional seed to retire `peerCount: 0`.
- [ ] Restore the faucet service and confirm reachability.
- [ ] CloudWatch monitoring + DNS failover (PLAN.md "Seed Node" / "Disaster Recovery").

## Acceptance criteria

- [ ] seed.botho.io runs a binary built from current `main` (cycle-6 + I4 included), protocol 2.0.0.
- [ ] Fresh-genesis chain; `peerCount >= 1` (multi-node), minting active.
[...]

PR #342, opened 2026-06-14
fix(consensus): widen cumulative supply accumulators to u128

### LINKED ISSUE #333: fix(consensus): widen supply accumulators to u128 (fix u64 emission overflow)

rjwalters:
## Summary

CRITICAL supply-integrity / consensus fix. The chain-state supply accumulators were `u64`, but Phase-1 emission (~1.22e21 picocredits ≈ 1.22B BTH) far exceeds `u64::MAX` (~1.84e19 pico ≈ 18.4M BTH). With `[profile.release] overflow-checks = false`, the accumulator wrapped silently in production once cumulative emission passed ~18.4M BTH (~1.5% of Phase 1). Because `total_mined` feeds `calculate_block_reward()` (node/mod.rs:125), a wrapped value corrupted the emission schedule itself and diverged nodes.

Per the (settled) decision, the picocredit unit (1 BTH = 1e12) is unchanged; the fix is widening the cumulative monetary accumulators to `u128`.

## Changes

**Widened to `u128`** (provably cross `u64::MAX`):
- `ChainState.total_mined`, `ChainState.total_fees_burned` (ledger/mod.rs)
- `EmissionController.total_emitted` / `total_burned` (block.rs) — the in-memory mirror restored from chain state via `from_chain_state`; leaving these `u64` would re-introduce the same silent wrap one layer over.
- `MintingWork.total_minted` (minter.rs) — passed straight into `calculate_block_reward`.
- `calculate_block_reward(height, total_supply: u128)` (block.rs). Added a private `calculate_tail_reward_u128` that reimplements `MonetaryPolicy::calculate_tail_reward` in `u128` (kept bit-for-bit identical to the cluster-tax `u64` version, which already does its internal math in u128). This is required because by Phase 2 the real picocredit supply exceeds `u64::MAX`, and truncating to `u64` would compute a wrong tail reward. The cluster-tax crate itself is left untouched (out of scope per issue).

**Serialization (consensus-state format change):** `META_TOTAL_MINED` / `META_FEES_BURNED` now persist as 16-byte LE (was 8) via `u128::to_le_bytes` / `u128::from_le_bytes` with `unwrap_or([0;16])`. `ChainState` derives serde, which handles `u128` natively (snapshot round-trip verified).

**RPC:** `handle_chain_info` and `handle_supply_info` (rpc/mod.rs) now emit `totalMined`, `totalFeesBurned`, `circulatingSupply` as decimal **strings** (`.to_string()`) — `u128` always exceeds JS's 2^53 safe-integer limit (and u64-pico already did, so this also fixes a latent precision bug). Prometheus gauge setters (`set_total_minted` / `set_total_fees_burned`) take `u128` and saturate to `i64::MAX` (display-only metric).

### Audited and kept `u64` (with bound)
| Field | Bound / justification |
|-------|-----------------------|
| `height`, `total_tx`, `epoch_tx` | counts |
| `tip_timestamp`, `difficulty` | non-monetary |
| `current_reward` | single block reward ≤ 50 BTH = 5e13 pico |
| `epoch_emission`, `epoch_burns` | reset every difficulty epoch; one epoch's emission/burn cannot approach `u64::MAX` |
| `META_LOTTERY_POOL` running balance | bounded, NOT unbounded: each block adds `emission_share + fee_pool` and pays out up to one full block reward (the anti-grind cap); carryover = `available − payout` drains toward the cap, so the pool stays near ≤ one reward (≈ 5e13 pico) — far below `u64::MAX`. The lottery accounting type (`LotteryPoolAccounting`) is consensus-critical `u64`/`saturating_add`; widening it would be a much larger lottery-module change for no overflow benefit. Left `u64`. |

**Amounts stay `u64` (hard requirement):** `TxOutput.amount`, `MintingTx.reward`, and `fee` are unchanged. No single UTXO approaches `u64::MAX`, and #340 already rejects txs whose output sum would overflow `u64`. **Block/tx/gossip wire format is byte-for-byte identical** — asserted by a regression test.

| total_mined / total_fees_burned are u128; no wrap past 18.4M BTH in release | ✅ | `test_emission_controller_accumulates_past_u64_max` + `test_block_reward_correct_past_u64_max`, both green under `--release` (overflow-checks=false) |
| `calculate_block_reward` correct past 2^64 pico | ✅ | `test_block_reward_correct_past_u64_max` recomputes expected reward in u128 and asserts the u128 path differs from the u64-truncated path |
| Metadata persists/reloads as 16-byte LE; round-trip green | ✅ | `test_supply_metadata_u128_persist_reload` (asserts on-disk len == 16 and exact reload); snapshot round-trip `test_snapshot_roundtrip_total_mined_past_u64_max` |
| RPC emits strings | ✅ | rpc/mod.rs uses `.to_string()` for all three fields |
| Amounts remain u64; wire format unchanged | ✅ | `test_block_wire_format_amounts_stay_u64` asserts deterministic serialization + 8-byte monetary fields |
| epoch_*/current_reward/META_LOTTERY_POOL audited | ✅ | table above |
| `cargo test -p botho` green; `cargo build --release` green | ✅ | lib 935 passed; ledger_consistency_integration 26 passed; release build green |

## Test Plan
- `cargo build` and `cargo build --release` — green.
- `cargo test -p botho --lib` — 935 passed (debug); re-run `--release` — 935 passed (the bug only manifests with overflow-checks=false, so the release run is the real regression guard).
- `cargo test -p botho --test ledger_consistency_integration` — 26 passed.
- `cargo clippy -p botho --lib` — no new warnings from these changes.

## Coordination
The on-disk metadata format change (8→16 bytes) is consensus-state serialization and rides the planned **testnet reset (#323)** — no migration path needed. Relates to the broader `overflow-checks=false` hazard from audit cycle 6 and to #340 (output-sum overflow, already fixed). The cluster-tax simulation crate's own `u64` supply overflow is independent and tracked separately (left untouched here).

Closes #333

### DIFF
--- a/botho/src/block.rs
+++ b/botho/src/block.rs
@@ -582,17 +582,51 @@ impl Block {
 ///
 /// # Returns
 /// The block reward for the given height.
-pub fn calculate_block_reward(height: u64, total_supply: u64) -> u64 {
+pub fn calculate_block_reward(height: u64, total_supply: u128) -> u64 {
     let policy = crate::monetary::mainnet_policy();
 
     // Check which phase we're in
     if policy.is_halving_phase(height) {
-        // Phase 1: Halving schedule based on block height
+        // Phase 1: Halving schedule based on block height (independent of
+        // supply — the halving branch ignores `total_supply` entirely).
         policy.halving_reward(height).unwrap_or(1)
     } else {
-        // Phase 2: Calculate tail reward based on supply
-        policy.calculate_tail_reward(total_supply)
+        // Phase 2: Calculate tail reward based on supply.
+        //
+        // `MonetaryPolicy::calculate_tail_reward` (cluster-tax crate) takes a
+        // `u64` supply, but by the time the chain reaches Phase 2 the real
+        // picocredit supply (~1.2e21) far exceeds `u64::MAX`. Truncating to
+        // `u64` would compute a wildly wrong tail reward, so we recompute the
+        // identical integer formula in `u128` here rather than widen the
+        // cluster-tax simulation crate (out of scope per #333). This MUST stay
+        // bit-for-bit equivalent to `MonetaryPolicy::calculate_tail_reward`.
+        calculate_tail_reward_u128(&policy, total_supply)
     }
+}
+
+/// `u128` reimplementation of `MonetaryPolicy::calculate_tail_reward`.
+///
+/// Kept byte-for-byte identical to the cluster-tax crate's `u64` version
+/// (which already does its internal arithmetic in `u128`); the only
+/// difference is that the `supply` input is accepted as `u128` so realistic
+/// picocredit supplies above `u64::MAX` do not truncate. See #333.
+fn calculate_tail_reward_u128(
+    policy: &bth_cluster_tax::MonetaryPolicy,
+    supply_at_transition: u128,
+) -> u64 {
+    // Target annual NET emission.
+    let target_net = supply_at_transition * policy.tail_inflation_bps as u128 / 10_000;
+    // Expected annual fee burns.
+    let expected_burns =
+        supply_at_transition * policy.expected_fee_burn_rate_bps as u128 / 10_000;
+    // Gross emission needed.
+    let gross_needed = target_net + expected_burns;
+    // Blocks per year at target rate.
+    let secs_per_year: u128 = 365 * 24 * 3600;
+    let blocks_per_year = secs_per_year / policy.target_block_time_secs as u128;
+    // Reward per block (a single block reward never approaches u64::MAX).
+    let reward = gross_needed / blocks_per_year;
+    reward.max(1) as u64
 }
 
 /// Dynamic block timing based on network load.
@@ -824,10 +858,14 @@ pub mod difficulty {
         pub difficulty: u64,
         /// Cumulative transactions (for difficulty adjustment timing)
         pub total_tx: u64,
-        /// Cumulative gross emission (picocredits minted)
-        pub total_emitted: u64,
-        /// Cumulative fees burned
-        pub total_burned: u64,
+        /// Cumulative gross emission (picocredits minted). `u128`: mirrors
+        /// `ChainState.total_mined` (restored via `from_chain_state`) and
+        /// crosses `u64::MAX` over Phase 1 — keeping it `u64` would
+        /// re-introduce the silent wrap #333 fixes.
+        pub total_emitted: u128,
+        /// Cumulative fees burned (picocredits). `u128` for the same reason as
+        /// `total_emitted`.
+        pub total_burned: u128,
 
         // --- Current epoch accumulators ---
         /// Tx in current adjustment epoch
@@ -866,8 +904,8 @@ pub mod difficulty {
         /// Restore from persisted chain state
         pub fn from_chain_state(
             difficulty: u64,
-            total_mined: u64,
-            total_fees_burned: u64,
+            total_mined: u128,
+            total_fees_burned: u128,
             total_tx: u64,
             epoch_tx: u64,
             epoch_emission: u64,
@@ -898,7 +936,7 @@ pub mod difficulty {
             let initial_reward = INITIAL_REWARD as u128;
             let halving_interval = policy.halving_interval as u128;
             let halving_count = policy.halving_count as u128;
-            let total_emitted = self.total_emitted as u128;
+            let total_emitted = self.total_emitted;
 
             let total_halving_emission = initial_reward * halving_interval * halving_count;
             if total_emitted < total_halving_emission {
@@ -919,8 +957,8 @@ pub mod difficulty {
             self.current_reward
         }
 
-        /// Net circulating supply
-        pub fn net_supply(&self) -> u64 {
+        /// Net circulating supply (picocredits).
+        pub fn net_supply(&self) -> u128 {
             self.total_emitted.saturating_sub(self.total_burned)
         }
 
@@ -952,8 +990,8 @@ pub mod difficulty {
         ) -> (u64, u64) {
             // Update totals
             self.total_tx += tx_count;
-            self.total_emitted += reward_paid;
-            self.total_burned += fees_burned;
+            self.total_emitted += reward_paid as u128;
+            self.total_burned += fees_burned as u128;
 
             // Update epoch accumulators
             self.epoch_tx += tx_count;
@@ -1022,9 +1060,10 @@ pub mod difficulty {
                 return 0;
             }
             // Net emission per tx, annualized assuming 10M tx/year
-            let net_per_tx = self.total_emitted.saturating_sub(self.total_burned) / self.total_tx;
+            let net_per_tx =
+                self.total_emitted.saturating_sub(self.total_burned) / self.total_tx as u128;
             let annual = net_per_tx * 10_000_000;
-            annual * 10_000 / supply
+            (annual * 10_000 / supply) as u64
         }
     }
 
@@ -1068,7 +1107,7 @@ pub mod difficulty {
             // and that very high emission values are still in halving phase (since the
             // threshold is larger than u64::MAX).
             let mut ctrl = EmissionController::new(1000);
-            ctrl.total_emitted = u64::MAX;
+            ctrl.total_emitted = u64::MAX as u128;
 
             // With the current constants, even u64::MAX is still in halving phase
             // because total_halving_emission > u64::MAX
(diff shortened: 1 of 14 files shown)


rjwalters:
## Changes Requested — CI failing (incomplete audit)

The core overflow fix is **correct and genuinely well-tested** — I verified it carefully. But the field-widening audit missed a consensus-critical fuzz harness, which fails the **Fuzz Build Guard (PR)** CI check. CI must be green before this consensus/supply-integrity fix can land.

### Blocking: `fuzz_add_block` does not compile

`fuzz/fuzz_targets/fuzz_add_block.rs:107`:

```rust
assert_eq!(
    post.total_mined,
    prev.total_mined + expected_reward,   // total_mined is now u128, expected_reward is u64
    ...
);
```

`expected_reward = block.minting_tx.reward` is `u64` (line 80), and `total_mined` is now `u128`, so this fails to build:

```
error[E0308]: mismatched types  --> fuzz_targets/fuzz_add_block.rs:107:36  (expected `u128`, found `u64`)
error[E0277]: cannot add `u64` to `u128`  --> fuzz_targets/fuzz_add_block.rs:107:34
```

**Fix:** widen the addend, e.g.
```rust
prev.total_mined + expected_reward as u128,
```
While there, line 126's `post.total_mined as u128` is now a no-op cast (`total_mined` is already `u128`) — drop it to avoid a clippy `unnecessary_cast` warning.

This is exactly the class of reader the audit needed to catch: the `fuzz_add_block` harness asserts the `total_mined == prev + reward` invariant — a consensus-state reader of a widened field. The `cargo test -p botho` runs in the PR's test plan don't cover the `fuzz/` crate, so this slipped through locally but the dedicated Fuzz Build Guard caught it.

### Non-blocking note (no action needed for this PR)

`Compile Benchmarks` is also red, but for an unrelated reason: `glib-sys v0.18.1` fails its build script (a system GTK/glib dependency for the desktop crate) — CI environment, not this PR.

`network::transport::bench::tests::test_throughput_measurer` is flaky in release (asserts `elapsed() > Duration::ZERO`; failed 1 of 3 local runs on both this branch and an untouched tree). Not introduced here and not blocking — worth a separate follow-up issue to make it robust.

---

## What I verified (all green) — the fix itself is solid

I read the full diff against issue #333 and ran the suite locally.

1. **Overflow fix / regression guard** — `ChainState.total_mined`/`total_fees_burned`, `EmissionController.total_emitted`/`total_burned`, and `MintingWork.total_minted` are all `u128`. `test_emission_controller_accumulates_past_u64_max` seeds `u64::MAX - 10`, records a block, and asserts the **exact** accumulated value crossed `u64::MAX` (not just "no panic"). Passes under `--release` (overflow-checks=false) — the real guard. ✅

2. **`calculate_tail_reward_u128`** — bit-for-bit identical to `MonetaryPolicy::calculate_tail_reward` (cluster-tax/src/monetary.rs:293): same ops, same precedence, same `.max(1) as u64`; the original already casts `supply as u128` internally. `test_tail_reward_u128_matches_u64_in_range` asserts equivalence across [0, 1, 1e12, 1e18, u64::MAX]; `test_block_reward_correct_past_u64_max` drives supply to 1.22e21, asserts the exact u128 result AND that it differs from the u64-truncated path (the test has teeth). ✅

3. **Serialization** — `META_TOTAL_MINED`/`META_FEES_BURNED` persist 16-byte LE; decode `from_le_bytes(... unwrap_or([0;16]))`. `test_supply_metadata_u128_persist_reload` asserts on-disk len == 16 and exact reload > u64::MAX; `test_snapshot_roundtrip_total_mined_past_u64_max` round-trips through write_to/read_from. PR correctly notes the 8→16 byte change rides the testnet reset (#323). ✅

4. **Reader audit (otherwise complete)** — grepped all `total_mined`/`total_fees_burned`/`total_emitted`/`total_minted` sites. RPC passes u128 to `calculate_block_reward` (rpc/mod.rs), node/mod.rs:125 and minter.rs:249 likewise, metrics saturate to i64::MAX (documented, display-only), f64 casts are display-only. The only re-narrowing `as u64`/`as u32` casts (lottery.rs:372 ratio, block.rs:945 epoch index) operate on bounded local values, not the accumulators. The **one** miss is `fuzz_add_block.rs` above.

5. **Wire-format invariance** — amounts stayed u64; `test_block_wire_format_amounts_stay_u64` asserts 8-byte monetary fields + deterministic serialization. ✅

6. **RPC strings** — `totalMined`/`totalFeesBurned`/`circulatingSupply` emit via `.to_string()`; `circulating_supply` is `total_mined.saturating_sub(total_fees_burned)` on u128. ✅

7. **`META_LOTTERY_POOL` kept u64** — I scrutinized the bounded-pool reasoning. The PR's claim that the pool "stays near ≤ one reward" is slightly too strong: carryover = `available − min(available, payout_cap)`, so when per-block inflow (`emission_share + fee_pool`) exceeds the one-block-reward cap, the pool **does** accumulate the surplus. However `compute_pool_accounting` uses `saturating_add` (lottery.rs:187-189), so the failure mode is *saturation* (a conservation/liveness failure that rejects blocks) — NOT the silent wrap that #333 fixes — and reaching u64::MAX requires sustained fees exceeding a block reward every block for a very long time. I'm treating this as acceptable for this PR (the issue left it to author audit), but please file a follow-up issue to either widen the pool to u128 or document the precise saturation bound, rather than leaving the "near one reward" justification as-is.

Once `fuzz_add_block.rs` compiles and the Fuzz Build Guard is green, I'll re-review and approve.

rjwalters:
Addressed the Fuzz Build Guard (PR) failure — the one blocking issue from the Judge's review.

The #333 u128 widening of `ChainState.total_mined` / `total_fees_burned` missed the workspace-excluded `fuzz/` crate. Fixed two type mismatches the widening introduced:

- `fuzz/fuzz_targets/fuzz_add_block.rs:107` — `prev.total_mined + expected_reward` (u128 + u64) now casts the u64 reward: `prev.total_mined + expected_reward as u128`. Also dropped the now-redundant `as u128` casts on `total_mined` and `total_fees_burned` (both already u128) so clippy's `unnecessary_cast` stays green. `lottery_pool` is still u64, so its `as u128` cast is preserved.
- `fuzz/fuzz_targets/fuzz_cluster_tax_math.rs:164` — `calculate_block_reward`'s `total_supply` param was widened to u128; the fuzzer's u64 `total_supply` is now cast at the call site. (This sibling break was caught by actually running `cargo fuzz build`, not just the one line in the Judge's note.)

No accumulator is narrowed with `as u64` anywhere — the overflow fix is intact.

Verification:
- `cargo fuzz build` (pinned `nightly-2025-12-03` + cargo-fuzz 0.13.1, matching the CI guard's exact command) compiles **all** fuzz targets clean — zero errors, only pre-existing dead-code warnings.
- `cargo test -p botho --lib` — 935 passed / 0 failed, including all six #333 regression tests (`test_block_reward_correct_past_u64_max`, `test_emission_controller_accumulates_past_u64_max`, `test_tail_reward_u128_matches_u64_in_range`, `test_block_wire_format_amounts_stay_u64`, `test_snapshot_roundtrip_total_mined_past_u64_max`, `test_supply_metadata_u128_persist_reload`).

No changes to consensus logic, serialization, or RPC encoding. The non-blocking `Compile Benchmarks` (glib-sys) and `META_LOTTERY_POOL` items are left untouched per your review. Ready for re-review.

rjwalters:
## Judge re-review (pass 2) — APPROVED

Focused re-review of Doctor fix `b79bc6b` for the CRITICAL supply-overflow fix (#333). The substance of the u128 widening was already verified and approved in pass 1; I blocked only because the field-widening audit missed the workspace-excluded `fuzz/` crate, breaking the **Fuzz Build Guard (PR)**. That is the only thing re-reviewed here.

**Verdict: the build break is resolved, no narrowing was introduced, and the regression suite holds. Approving.**

### Evidence

1. **Fuzz Build Guard (PR) — now GREEN** (`pass`, 8m15s). The CI guard that previously failed at `fuzz_add_block.rs:107` now compiles every fuzz target clean on the pinned toolchain.

2. **Diff is exactly the two fuzz files** since pass 1, and nothing else:
   - `fuzz_add_block.rs:107` — `prev.total_mined + expected_reward as u128` (the u64 reward widened; correct).
   - `fuzz_add_block.rs:124,126` — redundant `as u128` casts dropped on `total_mined` / `total_fees_burned`, which are genuinely `u128` now; `lottery_pool as u128` correctly **kept** because `get_lottery_pool()` still returns u64.
   - `fuzz_cluster_tax_math.rs:164` — `calculate_block_reward(m.height, m.total_supply as u128)` (the second sibling the original audit missed). Confirmed `calculate_block_reward` signature is now `(height: u64, total_supply: u128) -> u64`, so the cast is required.
   - **No `as u64` narrowing cast introduced anywhere** in the new diff.

3. **No third missed site.** Grepped the whole `fuzz/` crate for `total_mined` / `total_fees_burned` / `total_emitted` / `total_burned` / `total_supply`. The one remaining unchanged reader, `fuzz_cluster_tax_math.rs:168` `policy.calculate_tail_reward(m.total_supply)`, is correct as-is: `calculate_tail_reward(&self, supply_at_transition: u64) -> u64` still takes a u64.

4. **Regression tests hold.** `cargo test -p botho --lib` → **935 passed, 0 failed**. The six #333 guard tests are present and pass, including `test_block_reward_correct_past_u64_max`, `test_emission_controller_accumulates_past_u64_max`, and `test_snapshot_roundtrip_total_mined_past_u64_max`. None were weakened.

### Non-blocking (explicitly out of scope, not re-litigated)
- `Compile Benchmarks` failure (`glib-sys`) — unrelated to this PR; merge state is UNSTABLE only because of it. Mergeable = true.
- META_LOTTERY_POOL / `lottery_pool` getter staying u64 — known follow-up.

Moving to `loom:pr`.

PR #355, opened 2026-06-15
style: rustfmt baseline + CI fmt check on pinned nightly

rjwalters:
## Summary

The repo had a `rustfmt.toml` but no CI enforcement, so the tree had drifted out of compliance (the recurring "giant abandoned `cargo fmt` stash" failure mode). This fixes it once: a clean repo-wide rustfmt baseline plus a CI check so it never drifts again.

All formatting was run with the **pinned nightly** (`cargo +nightly-2025-12-03 fmt`), NOT stable. This is required because `rustfmt.toml` uses nightly-only options (`wrap_comments = true`, `imports_granularity = "Crate"`) that stable rustfmt silently ignores — running stable would produce a different/wrong result. `rustfmt.toml` is unchanged.

## Changes

Three cleanly separable commits:

1. **`style:` workspace baseline** — `cargo +nightly-2025-12-03 fmt --all` over the workspace (65 files). The `rustfmt.toml` `ignore` list (`sgx/types`, `sgx/urts`, `sgx/sync`) is honored automatically.
2. **`style:` excluded-crate baseline** — `--all` does not reach members under `exclude` in the root `Cargo.toml`, so formatted them individually with the same pinned nightly:
   - `fuzz/` (10 files) — matters most; the strict fuzz build guard compiles every target.
   - `contracts/solana/` (1 file).
   - `util/serial/` — already conformant, no diff.
   - `contracts/ethereum` has no Rust; nothing to format.
3. **`ci:` `.github/workflows/fmt.yml`** — fmt-check only (no build/test, stays fast). Triggers on `pull_request` and `push` to `main`. Installs the pinned `nightly-2025-12-03` + `rustfmt` component via `dtolnay/rust-toolchain@master` (matching the `fuzz.yml` pattern, **not** `@stable`), then runs `cargo +nightly-2025-12-03 fmt --all -- --check` for the workspace plus the equivalent in `fuzz`, `util/serial`, and `contracts/solana`. Fails on any diff.

**Dirs baselined:** workspace (`--all`) + `fuzz` + `contracts/solana` + `util/serial`.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Single mechanical baseline; `fmt --all -- --check` exits 0 for workspace | done | `cargo +nightly-2025-12-03 fmt --all -- --check` → exit 0 |
[...]

PR #358, opened 2026-06-15
ci(benchmarks): install GTK/glib system deps for workspace bench compile

### LINKED ISSUE #356: CI: Compile Benchmarks job fails on all PRs (missing glib-2.0 system lib on runner)
## Problem

The `Compile Benchmarks` job in the Benchmarks workflow fails on every PR (not just one) with:

```
error: failed to run custom build command for `glib-sys v0.18.1`
The system library `glib-2.0` required by crate `glib-sys` was not found.
pkg-config exited with status code 1
```

`glib-sys` is pulled in transitively (via the desktop/GTK crate). The CI runner image is missing the `glib-2.0` system development package, so `pkg-config --libs --cflags glib-2.0` fails and benchmark compilation aborts.

## Evidence it is pre-existing / environment-only

- Reproduced on multiple unrelated PRs: `feature/issue-354`, `feature/issue-343`, `feature/issue-333` — all show the Benchmarks workflow `failure`.
- Discovered while reviewing PR #355 (a pure rustfmt baseline). A whitespace-only change cannot cause a `pkg-config`/`glib-2.0` not-found error, confirming the failure is independent of code changes.

## Likely fix

Install GTK/glib dev packages in the benchmark CI job (Ubuntu runner), e.g.:

```yaml
- run: sudo apt-get update && sudo apt-get install -y libglib2.0-dev libgtk-3-dev
```

Mirror whatever the desktop-crate build job already does, or exclude the desktop crate from the benchmark compile step if benchmarks don't need it.

Discovered during evaluation of PR #355.

rjwalters:
## Summary

The `Compile Benchmarks` job in the Benchmarks workflow fails on every PR with:

```
error: failed to run custom build command for `glib-sys v0.18.1`
The system library `glib-2.0` required by crate `glib-sys` was not found.
pkg-config exited with status code 1
```

Root cause: `compile-benchmarks` runs `cargo bench --no-run --workspace`. The
workspace includes the Tauri desktop crate (`web/packages/desktop/src-tauri`,
a member in the root `Cargo.toml`), which transitively pulls in
`glib-sys`/`gtk-sys` via `pkg-config`. The bare Ubuntu runner lacks the GTK/glib
dev packages, so compilation aborts before any benchmark code is built. This is
environment-only and pre-existing (it reproduces on unrelated PRs, including the
pure-rustfmt PR #355), which is why a whitespace change cannot cause it.

## Changes

- `.github/workflows/benchmarks.yml`: add an `Install GTK/glib system deps`
  step in the `compile-benchmarks` job, after checkout and before the cargo
  bench compile.

## Why this package set

No other workflow in `.github/workflows/` installs GTK/glib (grepped for
`libglib2.0-dev` / `apt-get install` / `libgtk` / `libwebkit` — only
`release.yml` installs `pkg-config libssl-dev` and `whitepaper.yml` installs
texlive), so there was no existing pattern to mirror. The desktop crate is the
source, and it is a Tauri **v2** app (`tauri = "2.9.5"`, `tauri-build = "2.5.3"`).
The packages match the Tauri v2 Linux prerequisites for the runner OS:

| Package | Provides |
|---------|----------|
| `libglib2.0-dev` | `glib-2.0` (the lib named in the error) |
| `libgtk-3-dev` | `gtk+-3.0` |
| `libwebkit2gtk-4.1-dev` | `webkit2gtk-4.1` (Tauri v2; **4.1**, not 4.0) |
| `libsoup-3.0-dev` | `libsoup-3.0` (Tauri v2 uses soup3) |
| `librsvg2-dev` | `librsvg-2.0` |
| `libayatana-appindicator3-dev` | tray/appindicator |

`runs-on: ubuntu-latest` is currently Ubuntu 24.04, where the correct webkit
package is `libwebkit2gtk-4.1-dev` (the older `-4.0-dev` is not the Tauri v2
target). The full set (not just glib+gtk) is installed deliberately so the
build doesn't fail on the *next* missing pkg-config lib once glib resolves.

Only the `compile-benchmarks` job needed the deps: the downstream jobs
(`run-crypto-benchmarks`, `run-node-benchmarks`, `run-transaction-benchmarks`)
build specific crates with `-p` and never touch the desktop crate.

| `benchmarks.yml` installs needed system deps so Compile Benchmarks can compile | done | Install step added before `cargo bench --no-run --workspace` |
| YAML validated | done | `python3 -c 'import yaml; yaml.safe_load(open(...))'` -> `YAML OK` |
| Package names correct for runner OS | done | `ubuntu-latest` = 24.04 -> `libwebkit2gtk-4.1-dev`; `glib-2.0` is provided by `libglib2.0-dev` |
| Matches existing desktop job set | n/a | No other workflow installs GTK deps; set derived from Tauri v2 prerequisites |

## Test Plan

- This is a CI-workflow change and cannot be fully exercised locally.
- Validated the YAML parses cleanly.
- Confirmed `glib-2.0` is the pkg-config name the build needs.
- Full validation happens when CI runs the Benchmarks workflow on this PR; the
  `Compile Benchmarks` job should now get past the glib error and compile.

Closes #356

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -32,6 +32,23 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
 
+      # The workspace includes the Tauri desktop crate
+      # (web/packages/desktop/src-tauri), which pulls in glib-sys/gtk-sys via
+      # pkg-config. Without these system dev packages, `cargo bench --workspace`
+      # fails compiling glib-sys ("system library `glib-2.0` ... was not found").
+      # Package set matches the Tauri v2 Linux prerequisites for Ubuntu 24.04
+      # (ubuntu-latest), which uses webkit2gtk 4.1.
+      - name: Install GTK/glib system deps (Tauri desktop crate)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            libglib2.0-dev \
+            libgtk-3-dev \
+            libwebkit2gtk-4.1-dev \
+            libsoup-3.0-dev \
+            librsvg2-dev \
+            libayatana-appindicator3-dev
+
       - name: Cache cargo
         uses: actions/cache@v4
         with:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -32,23 +32,6 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
 
-      # The workspace includes the Tauri desktop crate
-      # (web/packages/desktop/src-tauri), which pulls in glib-sys/gtk-sys via
-      # pkg-config. Without these system dev packages, `cargo bench --workspace`
-      # fails compiling glib-sys ("system library `glib-2.0` ... was not found").
-      # Package set matches the Tauri v2 Linux prerequisites for Ubuntu 24.04
-      # (ubuntu-latest), which uses webkit2gtk 4.1.
-      - name: Install GTK/glib system deps (Tauri desktop crate)
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y \
-            libglib2.0-dev \
-            libgtk-3-dev \
-            libwebkit2gtk-4.1-dev \
-            libsoup-3.0-dev \
-            librsvg2-dev \
-            libayatana-appindicator3-dev
-
       - name: Cache cargo
         uses: actions/cache@v4
         with:
@@ -63,8 +46,23 @@ jobs:
             ${{ runner.os }}-cargo-bench-
             ${{ runner.os }}-cargo-
 
-      - name: Compile benchmarks (all packages)
-        run: cargo bench --no-run --workspace
+      # Compile only the crates that actually have benchmark targets, instead
+      # of `--workspace`. The full-workspace compile dragged in the Tauri
+      # desktop crate (needs GTK/glib system libs) and a no_std-leaning member
+      # that triggered a `k256` feature-unification break
+      # ("`precomputed-tables` feature requires either `critical-section` or
+      # `std`"). Neither crate has benchmarks, so scoping to the benched crates
+      # keeps full benchmark coverage while sidestepping both failures and
+      # removing the need for the GTK/glib apt install.
+      - name: Compile benchmarks (bench-bearing crates)
+        run: >
+          cargo bench --no-run
+          -p botho
+          -p bth-transaction-core
+          -p bth-crypto-pq
+          -p bth-crypto-ring-signature
+          -p bth-account-keys
+          -p bth-cluster-tax
 
   run-crypto-benchmarks:
     name: Crypto Benchmarks


rjwalters:
❌ **Changes Requested**

The glib portion of this fix is correct and verified — but the `Compile Benchmarks` job still fails (red required check), so issue #356's stated outcome ("Compile Benchmarks can compile") is not yet achieved. A new, distinct compile error now blocks the same `cargo bench --no-run --workspace` step.

## What's verified working (glib fix is correct)

- Runner is **Ubuntu 24.04 (noble)** — confirmed in the install step log (`packages.microsoft.com/ubuntu/24.04/prod noble`). So `libwebkit2gtk-4.1-dev` + `libsoup-3.0-dev` are the right Tauri **v2** packages. Correct.
- Install step is correctly scoped: it's in `compile-benchmarks` (the `--workspace` job), runs before the `cargo bench --no-run --workspace` step, and the other three jobs (`run-crypto/node/transaction-benchmarks`, which use `-p <crate>`) were untouched. Correct.
- YAML is well-formed; no unrelated changes.
- **The original glib error is gone.** The only `glib-2.0` / `pkg-config exited` hits in the whole job log are the install-step command echo. The build compiled `glib-sys`, `gtk-sys`, `tao`, `tauri`, and `botho-desktop` successfully and ran ~8 minutes into real compilation. The env error from #356 is resolved.

## What still blocks the job

The build now fails later, on `k256 v0.13.4`, in the `--workspace` compile (job log https://github.com/botho-project/botho/actions/runs/27525552806/job/81352019470):

```
error: `precomputed-tables` feature requires either `critical-section` or `std`
   --> k256-0.13.4/src/arithmetic/mul.rs:41:1
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `once_cell`
   --> k256-0.13.4/src/arithmetic/mul.rs:57:5
error: could not compile `k256` (lib) due to 2 previous errors
```

This is a Cargo **feature-unification** problem, not a missing system lib: under `--workspace`, `k256` is being resolved with `precomputed-tables` but without `std`/`critical-section` — almost certainly because a `no_std`-leaning member (e.g. `mobile/rust-bridge` / `botho-mobile` or a signer crate, both visible compiling just before the failure) pulls `k256` with `default-features = false`, and feature unification across the whole workspace produces an invalid combination. It is unrelated to glib and to #356's environment scope.

Per the CI gate, I can't apply `loom:pr` while the required `Compile Benchmarks` check is red. To get the job green, please pick one:

1. **Make the bench compile green** — fix the feature combo (e.g. ensure `std` or `critical-section` is enabled for `k256` in the relevant member's bench/dev profile, or add the missing `once_cell`/feature so the workspace-unified `k256` builds), **or**
2. **Scope the bench compile to crates that actually have benchmarks** — the benchmark jobs only ever run `-p bth-crypto-ring-signature`, `-p botho`, `-p bth-transaction-core`. `cargo bench --no-run --workspace` drags in the desktop/mobile crates that have no benches and is what surfaces both the glib and the k256 issues. Compiling only the benched crates (e.g. `--workspace --exclude botho-desktop --exclude botho-mobile`, or an explicit `-p` set) would both keep this fix and sidestep the k256 feature conflict — and arguably matches the issue's hint ("or exclude the desktop crate from the benchmark compile step if benchmarks don't need it").

Option 2 is likely the cleaner fix and may make the GTK install unnecessary, but either is acceptable. The glib work here is good — it just needs to carry the job all the way to green.

If you determine the k256 feature break is genuinely pre-existing on `main`'s `--workspace` build and out of scope, say so explicitly and we can split it into its own issue — but the job still needs to pass before this can merge.

rjwalters:
Addressed the change request by taking the Judge's recommended **Option 2 — scope the bench compile to crates that actually have benchmarks**.

## What changed
- `.github/workflows/benchmarks.yml`, `compile-benchmarks` job:
  - Replaced `cargo bench --no-run --workspace` with an explicit `-p` set of the six crates that actually have benchmark targets: `botho`, `bth-transaction-core`, `bth-crypto-pq`, `bth-crypto-ring-signature`, `bth-account-keys`, `bth-cluster-tax` (confirmed via `grep -rl "[[bench]]" --include=Cargo.toml` + `benches/` dirs).
  - **Removed** the GTK/glib apt install step. `cargo tree` shows none of the six targeted crates pull `glib-sys`/`gtk-sys`/`webkit2gtk`/`soup` (count 0 each), and the scoped compile succeeds locally without those system libs installed. The desktop crate that needed them is no longer compiled, so the install is dead weight.
  - The three downstream `-p`-based jobs (`run-crypto/node/transaction-benchmarks`) and the summary job are untouched.

## Why this fixes both failures
Scoping excludes the Tauri desktop crate (source of the original `glib-2.0` error) **and** the no_std-leaning member whose `k256 default-features = false` triggered the `--workspace` feature-unification break (`precomputed-tables` requires `critical-section` or `std`). Neither crate has benchmarks, so coverage is unchanged.

## Local verification
- `cargo bench --no-run -p botho -p bth-transaction-core -p bth-crypto-pq -p bth-crypto-ring-signature -p bth-account-keys -p bth-cluster-tax` → **Finished** cleanly, all bench executables built. It does **not** hit the k256 error.
- YAML validates: `python3 -c 'import yaml; yaml.safe_load(...)'` → `YAML OK`.

## On the k256 break
It is **fully sidestepped** here (the scoped compile never touches the offending member). It is a genuine latent `--workspace` feature-unification issue that would still bite a hypothetical `cargo bench/build --workspace`, but it is out of scope for #356 (whose goal is "Compile Benchmarks goes green"). Worth filing as its own issue if a full `--workspace` bench is ever wanted again.

Ready for re-review.

rjwalters:
## Judge Review — APPROVED (second pass, post-Doctor fix)

The Doctor's scoped-`-p` approach (commit `24f9bb3`) resolves #356. All verification gates pass.

### 1. Coverage preserved — no bench target dropped
Independently enumerated bench-bearing crates:
- `grep -rl "\[\[bench\]\]" --include=Cargo.toml .` → `botho`, `transaction/core`, `crypto/pq`, `crypto/ring-signature`, `account-keys`, `cluster-tax` (6)
- `find . -type d -name benches` (excluding worktrees) → same 6 crates

Package names map exactly to the `-p` set in the job:
`botho`, `bth-transaction-core`, `bth-crypto-pq`, `bth-crypto-ring-signature`, `bth-account-keys`, `bth-cluster-tax`. No bench crate omitted. The catch-all `compile-benchmarks` gate also covers the 3 crates not exercised by a downstream run-job (`bth-crypto-pq`, `bth-account-keys`, `bth-cluster-tax`), so no bench target goes uncompiled across the workflow.

### 2. Compile Benchmarks is GREEN (key gate)
```
Compile Benchmarks	pass	6m35s	https://github.com/botho-project/botho/actions/runs/27528166976/job/81359810125
```
Job conclusion: `success`. It compiled past both the prior glib error and the k256 `precomputed-tables` feature-unification break.

### 3. glib-step removal is safe
`cargo tree -p <crate> -i glib-sys` for all 6 targeted crates → `did not match any packages`. glib is not reachable from any benched crate, so dropping the GTK/glib apt step cannot reintroduce a failure.

### 4. Scope discipline
Only `.github/workflows/benchmarks.yml` changed (1 file, +17/-2). Other benchmark jobs and the summary job untouched; YAML well-formed.

### 5. k256
Sidestepped (not fixed) via the scoped compile — acceptable and explicitly out of scope for #356. The latent `--workspace` feature-unification issue is documented in the step comment.

Approving. Merge state is `UNSTABLE` only because the downstream `continue-on-error: true` benchmark runners are still in progress; the required `Compile Benchmarks` gate is green.

PR #359, opened 2026-06-15
test: serialize privacy-metrics tests racing on process-global Prometheus metrics

### LINKED ISSUE #357: test: flaky test_metrics_updater_circuits — races on process-global Prometheus metrics under parallelism
## Summary

`network::privacy::metrics::tests::test_metrics_updater_circuits` (botho/src/network/privacy/metrics.rs:690) is **flaky under default test parallelism** — it fails intermittently (~1 in 5–10 full `cargo test -p botho --lib` runs) and passes reliably with `--test-threads=1`. Root cause: the privacy-metrics tests assert on **process-global Prometheus metrics** that sibling tests mutate concurrently.

## Root cause

`CIRCUITS_ACTIVE`, `CIRCUITS_BUILT`, `CIRCUIT_BUILD_FAILURES`, etc. are process-global Prometheus gauges/counters (one shared instance per process). Several tests in the same module (`test_metrics_updater_circuits`, `test_metrics_updater_relay`, …) construct a `PrivacyMetricsUpdater` and mutate these same globals. When run in parallel:

- `test_metrics_updater_circuits` does `updater.set_active_circuits(5); assert_eq!(CIRCUITS_ACTIVE.get(), 5);` (metrics.rs:689-690). `CIRCUITS_ACTIVE` is a **gauge** set to an absolute value — another test setting it to a different value between the `set` and the `get` makes the `== 5` assertion race and fail.
- The other two assertions in the same test use `before + 1` deltas (metrics.rs:692-698), which are more robust but still not safe against concurrent counter increments from sibling tests.

The builder on #354 reported this as an "LMDB `Invalid argument` (os error 22)" failure in ~4 tests; the reproducible failure I captured is this shared-global-metrics race (panic at metrics.rs:690), which is not LMDB. If genuine LMDB-env flakes also exist they were not reproducible in 12+ local runs — treat the global-metrics race as the confirmed defect; investigate/ް fold in any LMDB-specific flake only if it reproduces.

## Fix (pick the cleanest)

The tests share mutable process-global state, so they must not run concurrently with each other, OR must not assert on absolute global values:

- **Preferred — serialize the privacy-metrics tests.** The repo already uses `serial_test::serial` elsewhere (e.g. `botho/tests/rpc_integration.rs`). Annotate the metrics-mutating tests in this module with `#[serial]` (a shared key, e.g. `#[serial(privacy_metrics)]`) so they don't interleave. Add the `serial_test` dev-dependency to the `botho` crate if it isn't already present.
- **Or — make assertions concurrency-robust.** Capture `before`/`after` deltas for counters (already done for two of three) and avoid asserting an absolute value on the `CIRCUITS_ACTIVE` gauge under parallelism — but the gauge is inherently racy across tests, so serialization is the more reliable fix.

Apply to all tests in `network::privacy::metrics::tests` that touch the shared globals, not just `test_metrics_updater_circuits` (the relay/handshake/path tests have the same hazard).

## Acceptance criteria
- [ ] `cargo test -p botho --lib` passes deterministically under **default parallelism** — run it 15+ times with zero failures of `test_metrics_updater_*` (the current flake rate is ~10–20%, so 15 clean runs is a meaningful guard).
- [ ] The fix doesn't mask real regressions — the tests still assert the metric updater actually moves the metrics (don't just delete the assertions).
- [ ] `cargo build` green; no production code behavior change (test-only fix unless the metric API genuinely needs a test-friendly accessor).

## Notes
Test-isolation bug, no consensus/production impact. Surfaced during #354 (rustfmt baseline) review. Independent of #353 (a different determinism issue in the cluster-tax sim).

rjwalters:
## Summary

`network::privacy::metrics::tests::test_metrics_updater_circuits` was flaky under default test parallelism (~10-20% failure rate; passed reliably with `--test-threads=1`). The root cause is a shared-process-global hazard: the privacy-metrics tests assert on **process-global Prometheus metrics** (`CIRCUITS_ACTIVE`, `CIRCUITS_BUILT`, `RELAY_*`, `TX_*`, `HANDSHAKE_*`). There is one instance of each per process, so sibling tests mutating them concurrently race:

- `test_metrics_updater_circuits` does `set_active_circuits(5); assert_eq!(CIRCUITS_ACTIVE.get(), 5)` — an absolute assertion on a shared gauge another test can overwrite between the `set` and the `get`.
- The relay/path/handshake tests use `before + 1` counter deltas, which still race because a sibling test can increment the same shared counter between the two reads.

## Changes

- Added `use serial_test::serial;` to the test module (`serial_test` was already a `botho` dev-dependency, used by `tests/rpc_integration.rs`).
- Annotated every test that touches the shared globals with `#[serial(privacy_metrics)]` so they never interleave with each other, while still running in parallel with unrelated tests:
  - `test_metrics_updater_circuits`
  - `test_metrics_updater_relay`
  - `test_metrics_updater_path`
  - `test_metrics_updater_handshake`
  - `test_snapshot_capture` (does `CIRCUITS_ACTIVE.set(3)` then asserts the snapshot)
- Left the pure snapshot-math tests (`test_circuit_build_success_rate`, `test_private_path_ratio`, `test_handshake_success_rate`, `test_alerting_thresholds`, and the `_no_attempts`/`_no_transactions` variants) **unserialized** — they only operate on local `PrivacyMetricsSnapshot` structs and never read the globals.
- Added a module comment documenting the shared-global hazard and the serialization rationale.
- **No assertions were deleted or weakened.** Serialization removes the race without reducing coverage; the tests still verify the updater moves the metrics.

| `serial_test` dev-dep present | ✅ | Already in `botho/Cargo.toml` `[dev-dependencies]` (`serial_test = { workspace = true }`) |
| All shared-global privacy-metrics tests annotated `#[serial(privacy_metrics)]` | ✅ | 5 globals-mutating tests annotated; snapshot-math tests intentionally left alone |
| Assertions unchanged in intent | ✅ | Only added attributes + a comment; no assertion bodies touched |
| `cargo build -p botho` green | ✅ | Builds clean (pre-existing warnings only) |
| 15+ parallel lib-test runs, zero `test_metrics_updater_*` flake | ✅ | 15/15 clean full `cargo test -p botho --lib` runs under default parallelism; 0 metrics flakes, 0 unrelated failures. Also 20/20 clean on the module filter. |

## Test Plan

- `cargo test -p botho --lib` run 15 times under default parallelism: **15/15 clean**, zero `test_metrics_updater_*` failures, zero other failures.
- `cargo test -p botho --lib network::privacy::metrics` run 20 times: **20/20 clean**.

Closes #357

### DIFF
--- a/botho/src/network/privacy/metrics.rs
+++ b/botho/src/network/privacy/metrics.rs
@@ -681,8 +681,20 @@ impl AlertingThresholds {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use serial_test::serial;
+
+    // NOTE: Several tests below mutate process-global Prometheus metrics
+    // (CIRCUITS_ACTIVE, CIRCUITS_BUILT, RELAY_*, TX_*, HANDSHAKE_*, ...).
+    // Those globals are shared across the whole test process, so running these
+    // tests concurrently races (e.g. an absolute `set`/`get` on a shared gauge,
+    // or a counter incremented by a sibling test between `before` and `after`).
+    // They are annotated `#[serial(privacy_metrics)]` so they never interleave
+    // with each other while still running in parallel with unrelated tests.
+    // Tests that only operate on local PrivacyMetricsSnapshot values (no global
+    // access) intentionally remain unserialized.
 
     #[test]
+    #[serial(privacy_metrics)]
     fn test_metrics_updater_circuits() {
         let updater = PrivacyMetricsUpdater::new();
 
@@ -699,6 +711,7 @@ mod tests {
     }
 
     #[test]
+    #[serial(privacy_metrics)]
     fn test_metrics_updater_relay() {
         let updater = PrivacyMetricsUpdater::new();
 
@@ -715,6 +728,7 @@ mod tests {
     }
 
     #[test]
+    #[serial(privacy_metrics)]
     fn test_metrics_updater_path() {
         let updater = PrivacyMetricsUpdater::new();
 
@@ -728,6 +742,7 @@ mod tests {
     }
 
     #[test]
+    #[serial(privacy_metrics)]
     fn test_metrics_updater_handshake() {
         let updater = PrivacyMetricsUpdater::new();
 
@@ -741,6 +756,7 @@ mod tests {
     }
 
     #[test]
+    #[serial(privacy_metrics)]
     fn test_snapshot_capture() {
         // Reset some metrics to known state for testing
         CIRCUITS_ACTIVE.set(3);


rjwalters:
## Judge review: APPROVE

Test-only fix. Verified every claim against the diff and the full `tests` module in `botho/src/network/privacy/metrics.rs`.

### Verification

**1. All racy tests serialized under ONE shared key, none missed.**
Confirmed. The 5 tests that mutate process-global Prometheus metrics all carry `#[serial(privacy_metrics)]` with the identical key:
- `test_metrics_updater_circuits` (CIRCUITS_ACTIVE/CIRCUITS_BUILT/CIRCUIT_BUILD_FAILURES)
- `test_metrics_updater_relay` (RELAY_FORWARDED/RELAY_EXITED/RELAY_LOAD)
- `test_metrics_updater_path` (TX_PRIVATE/TX_FAST_FALLBACK)
- `test_metrics_updater_handshake` (HANDSHAKE_CREATES_SENT/HANDSHAKE_TIMEOUTS)
- `test_snapshot_capture` (CIRCUITS_ACTIVE.set(3) + snapshot)

The 6 pure-math tests (`test_circuit_build_success_rate`[`_no_attempts`], `test_private_path_ratio`[`_no_transactions`], `test_handshake_success_rate`, `test_alerting_thresholds`) operate only on local `PrivacyMetricsSnapshot` structs and never read the globals — correctly left unserialized.

**2. No assertion weakened.** Confirmed. The diff adds only `use serial_test::serial;`, the `#[serial(privacy_metrics)]` attributes, and an explanatory comment. `assert_eq!(CIRCUITS_ACTIVE.get(), 5)` and all `before + 1` / float-tolerance assertions are unchanged.

**3. Determinism.** Ran `cargo test -p botho --lib network::privacy::metrics` 12x under default parallelism on the PR branch: **12/12 clean**, 11 passed each run, zero `test_metrics_updater_*` failures.

**4. `serial_test` dev-dep.** Confirmed present in `botho/Cargo.toml` `[dev-dependencies]` (`serial_test = { workspace = true }`).

**5. Build + fmt.** `cargo build -p botho` green (pre-existing warnings only). `cargo +nightly-2025-12-03 fmt -p botho -- --check` clean (exit 0), matching the green `rustfmt --check (pinned nightly)` CI check.

### CI note (non-blocking, pre-existing)
`mergeStateStatus` is UNSTABLE because the `Compile Benchmarks` check fails — but it fails on `The system library glib-2.0 ... was not found` while building a transitive gtk/gdk/webkit2gtk GUI dependency. The `benchmarks.yml` `compile-benchmarks` job (`cargo bench --no-run --workspace` on `ubuntu-latest`) has **no system-library install step** (e.g. `apt-get install libgtk-3-dev`), so it fails regardless of this PR. This is a pre-existing CI infrastructure gap unrelated to this test-only change — the diff touches only `metrics.rs` test annotations. Recommend tracking the benchmark-job glib install as a separate issue; it should not block this fix.

All substantive acceptance criteria met. Approving.

PR #368, opened 2026-06-15
test(e2e): mock /rpc for hermetic explorer specs

### LINKED ISSUE #334: e2e: explorer specs flake on live-RPC 'Connecting to network...' (make hermetic)
## Problem

The Playwright explorer specs (`web/e2e/tests/explorer/*`) depend on a live, single-peer public seed RPC (`https://seed.botho.io/rpc`, `peerCount: 0`) via the vite-preview `/rpc` proxy. During PR #332 review, two consecutive full-suite runs each produced **1 failure** (out of 56), always stuck at the `waitForExplorerReady` "Connecting to network..." phase before any search logic runs:

- Run 1: `explorer/search.spec.ts:49 can search by pressing Enter` failed (timeout in connect).
- Run 2: `explorer/search.spec.ts:21 can search by block height` failed (same connect timeout).

The failing test **rotates** and an isolated re-run of the search spec passes cleanly (2 passed, 3 skipped), confirming this is a transient connect flake, not a code defect. PR #332's config already mitigates with serial execution + retries, but the flake still surfaces ~1/run because both attempts can hit a slow/unavailable connect window on the shared public RPC.

## Suggested fix

Make the explorer e2e hermetic so it does not depend on the live node being responsive:
- Mock the `/rpc` JSON-RPC responses (e.g., a small local server like `serve-faucet.mjs` that returns fixed `node_getStatus` / `getBlockByHeight` payloads), OR
- Record/replay RPC fixtures, OR
- Increase connect-readiness timeout / add explicit connect retry in the explorer page-object helper.

This would let the PR's claimed "0 failed" hold deterministically in CI.

Discovered during Judge review of PR #332.

rjwalters:
## Summary

The Playwright explorer specs (`web/e2e/tests/explorer/*`) connected to the live public seed RPC (`https://seed.botho.io/rpc`) through the vite-preview `/rpc` proxy. The connect-time `node_getStatus` call could time out on the shared public node, surfacing as a rotating ~1/run failure stuck at `waitForExplorerReady` / "Connecting to network...". This makes the explorer e2e **hermetic** by mocking `/rpc` with a small local server, so the suite no longer depends on the live node.

## Changes

- **`web/e2e/serve-rpc-mock.mjs`** (new): dependency-free JSON-RPC mock (mirrors the `serve-faucet.mjs` pattern). Returns deterministic payloads for the methods the explorer calls on load + in the specs:
  - `node_getStatus` -> connect handshake + chain height (`chainHeight: 1000`, `peerCount: 1`)
  - `getChainInfo` -> difficulty (network stats)
  - `getBlockByHeight` -> any in-range height resolves to a synthetic-but-valid block (recent-block list, block detail, height search); out-of-range heights return a JSON-RPC error so the "non-existent block" search shows "not found".
  - Responds 200 to GET so Playwright's `webServer` readiness probe detects it; rejects WS upgrades (adapter falls back to polling).
- **`web/packages/web-wallet/vite.config.ts`**: the dev + preview `/rpc` proxy target now reads `E2E_RPC_PROXY_TARGET` (defaults to `https://seed.botho.io` for normal dev/preview).
- **`web/e2e/playwright.config.ts`**: start the mock as a `webServer` (port 4175) and set `E2E_RPC_PROXY_TARGET=http://localhost:4175` on the preview command so the preview `/rpc` proxy forwards to the mock instead of the live node.
- **`web/e2e/tests/explorer/{blocks,search}.spec.ts`**: replaced the conditional `if (!hasBlocks) test.skip()` branches with deterministic `expect('Recent Blocks').toBeVisible()` assertions, since the mock always serves a recent-block list. No tests remain skipped on data availability.

| Explorer e2e hermetic (mock `/rpc`, no live-node dependency) | ✅ | Mock returns fixed JSON-RPC; preview `/rpc` proxy retargeted to it via env var |
| Mock returns valid JSON-RPC for connect + search methods | ✅ | `curl`'d `node_getStatus`, `getChainInfo`, `getBlockByHeight` (in-range + out-of-range) directly and **through the preview proxy** (`POST localhost:4173/rpc`) — all returned correct payloads |
| Previously-skipped specs un-skipped where mock supports them | ✅ | All `test.skip()`-on-missing-data branches removed; 12 explorer tests listed, none data-gated |
| Playwright config well-formed | ✅ | `playwright test --project=explorer --list` lists all 12 tests; e2e run got past `webServer` startup (build + `vite preview` started, all 3 servers up) to browser launch |

## Test Plan / honest execution status

What I verified locally in this environment:
- Mock server returns valid JSON-RPC for every method, both directly (`curl localhost:4175`) and **end-to-end through the vite-preview proxy** (`curl -X POST localhost:4173/rpc`) — confirming the proxy now reaches the mock, not `seed.botho.io`.
- `pnpm --filter @botho/web-wallet build` and `vite preview` start cleanly with the new env wiring; Playwright started all three webServers (mock, preview, faucet) and proceeded to browser launch.
- `pnpm --filter @botho/web-wallet typecheck` passes; `playwright ... --list` enumerates the 12 explorer tests.

**The full Playwright browser run could NOT be executed here**: the Chromium / `chrome-headless-shell` download from the Playwright CDN is blocked in this sandbox (every test failed with `browserType.launch: Executable doesn't exist`, an environment issue, not a code defect). I did not fake a "0 failed" result. **The end-to-end browser run (explorer suite, ideally 3-5x to confirm zero connect-flake) needs CI**, where the bundled browsers install normally. The connect path is now fully deterministic against the local mock, so the "Connecting to network..." flake should be eliminated.

Closes #334

### DIFF
--- a/web/e2e/playwright.config.ts
+++ b/web/e2e/playwright.config.ts
@@ -16,6 +16,13 @@ const WEB_BASE_URL = process.env.E2E_WEB_BASE_URL ?? 'http://localhost:4173'
 const FAUCET_BASE_URL = process.env.E2E_FAUCET_BASE_URL ?? 'http://localhost:4174'
 const useLocalServers = !process.env.E2E_WEB_BASE_URL && !process.env.E2E_FAUCET_BASE_URL
 
+// Local JSON-RPC mock for the explorer/wallet. Pointing the vite-preview `/rpc`
+// proxy here (instead of the live seed node) makes the explorer specs hermetic:
+// the connect handshake (node_getStatus) and block reads resolve deterministically
+// from fixed fixtures, eliminating the "Connecting to network..." flake (#334).
+const RPC_MOCK_PORT = 4175
+const RPC_MOCK_URL = `http://localhost:${RPC_MOCK_PORT}`
+
 /**
  * Playwright E2E test configuration for Botho web services.
  *
@@ -64,14 +71,28 @@ export default defineConfig({
   webServer: useLocalServers
     ? [
         {
+          // Local JSON-RPC mock the wallet/explorer talk to via the same-origin
+          // /rpc proxy. Started before the preview server so the proxy target is
+          // up by the time the explorer connects. Returns fixed node_getStatus /
+          // getChainInfo / getBlockByHeight payloads so connect + block reads are
+          // deterministic (no live-node dependency).
+          command: 'node e2e/serve-rpc-mock.mjs',
+          cwd: path.resolve(__dirname, '..'),
+          url: RPC_MOCK_URL,
+          reuseExistingServer: !process.env.CI,
+          timeout: 30_000,
+        },
+        {
           // Build the web wallet with the RPC endpoint pointed at the
-          // same-origin /rpc proxy (configured in the wallet's vite preview
-          // config to forward to https://seed.botho.io), then serve
-          // landing/wallet/explorer as a SPA via vite preview on port 4173.
-          // This lets the explorer perform a real RPC read in e2e without
-          // depending on cross-origin CORS.
+          // same-origin /rpc proxy, then serve landing/wallet/explorer as a SPA
+          // via vite preview on port 4173. E2E_RPC_PROXY_TARGET points the
+          // preview's /rpc proxy at the local mock above (instead of the live
+          // seed node), so the explorer performs a real RPC read in e2e against
+          // deterministic fixtures without cross-origin CORS or live-node flake.
           command:
-            'VITE_RPC_ENDPOINT=/rpc pnpm --filter @botho/web-wallet build && pnpm --filter @botho/web-wallet preview --port 4173 --strictPort',
+            'VITE_RPC_ENDPOINT=/rpc pnpm --filter @botho/web-wallet build && E2E_RPC_PROXY_TARGET=' +
+            RPC_MOCK_URL +
+            ' pnpm --filter @botho/web-wallet preview --port 4173 --strictPort',
           cwd: path.resolve(__dirname, '..'),
           url: 'http://localhost:4173/',
           reuseExistingServer: !process.env.CI,
--- a/web/e2e/serve-rpc-mock.mjs
+++ b/web/e2e/serve-rpc-mock.mjs
@@ -1,0 +1,144 @@
+/**
+ * Minimal JSON-RPC mock for the seed node read RPC used by the explorer/wallet
+ * during e2e.
+ *
+ * The explorer connects to a node via the same-origin `/rpc` proxy, which in
+ * production forwards to https://seed.botho.io. Pointing that proxy at the live
+ * node makes the explorer specs depend on a shared public node being responsive
+ * — which flaked ~1/run, always stuck in the "Connecting to network..." phase
+ * (the connect-time `node_getStatus` call timing out). See issue #334.
+ *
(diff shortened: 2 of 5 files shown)


rjwalters:
## Judge Review: APPROVE

Consensus-irrelevant test-infra change. Reviewed by reading the diff and cross-checking every claim against the actual explorer RPC adapter (`web/packages/adapters/src/remote.ts`), the explorer context/components, and the PR-branch spec sources. All five verification points pass.

### 1. Mock correctness/completeness — VERIFIED
`serve-rpc-mock.mjs` serves exactly the methods the explorer exercises:
- `node_getStatus` → `{version, network, chainHeight: 1000, peerCount: 1, mempoolSize: 0}` — satisfies both `connect()` and `getNetworkStats()`/`getBlockHeight()` in the adapter.
- `getChainInfo` → `{difficulty}`.
- `getBlockByHeight` → in-range height returns a block with every field the adapter reads (`height/hash/prevHash/timestamp/difficulty/txCount/mintingReward`); out-of-range returns a JSON-RPC `error`.

Envelope `{jsonrpc, id, result|error}` matches the adapter's `JsonRpcResponse` interface exactly. The adapter throws on `json.error` and `getBlock` catches → returns `null` → renders "not found". Mirrors the `serve-faucet.mjs` pattern (dependency-free `node:http`, 200 on GET for the webServer readiness probe, CORS headers, WS upgrade rejected so the adapter falls back to polling). `node --check` passes.

### 2. Proxy rerouted only for e2e (the crux) — VERIFIED
`vite.config.ts` now uses `process.env.E2E_RPC_PROXY_TARGET || 'https://seed.botho.io'` for both the dev and preview `/rpc` proxies — the production/dev default is unchanged. `playwright.config.ts` starts the mock (port 4175) as a webServer and prepends `E2E_RPC_PROXY_TARGET=http://localhost:4175` to the preview command only. The builder's `curl -X POST localhost:4173/rpc` through the preview proxy is a sound end-to-end check that `/rpc` reaches the mock, not the live node.

### 3. Un-skip safety (biggest risk) — VERIFIED
This was the specific risk to check, and the data is satisfied. The "Recent Blocks" `CardTitle` only renders when `blocks.length > 0` (otherwise the component shows the "No blocks found" empty state). The mock makes `getRecentBlocks({limit: 20})` fetch heights 981..1000 — all in-range → 20 valid blocks → "Recent Blocks" + `#{height}` indicators + the always-present "Load More" button all render. Search-by-height reads a rendered in-range height and resolves to a "Block #N" heading. The (unchanged) non-existent-height (`99999999 > 1000`) and invalid-hash cases map to null → "not found". Every now-unconditional assertion is backed by mock data — no skip-to-hard-failure conversions.

### 4. Sandbox limit — honest
The PR explicitly states the full Chromium browser run could not execute (CDN download blocked in the sandbox) and did not fabricate a "0 failed" result. Not blocking, per scope.

### 5. Scope — clean
Exactly the 5 intended files; no `pnpm-workspace.yaml` change.

### One non-blocking note
There is no PR-triggered Playwright CI workflow in this repo (`.github/workflows/e2e-tests.yml` is a Rust-only `workflow_dispatch` suite), so the browser run will not be auto-confirmed on this PR. The connect path is now fully deterministic against the local mock, so the change is sound on review; a manual `pnpm --filter ... e2e` run with browsers installed (ideally 3-5x) is the recommended final confirmation before relying on it. `mergeStateStatus` is CLEAN.

Approving.

Issue #373, opened 2026-06-15
test/ops: betanet-join smoke — local node connects to live testnet + syncs (gated, non-CI)


rjwalters:
## Summary (smoke / ops validation — NOT a hermetic CI gate)

Add a script (+ gated `workflow_dispatch` CI job) that **launches a local botho node, connects it to the live beta testnet, and confirms it peers + syncs** (and optionally mines a block). This validates **public-internet peering** — joining the betanet from outside the AWS VPC — which the current VPC-internal-bootstrap setup doesn't exercise.

**This is intentionally NOT a PR-gating test:** it depends on external live infra (`seed.botho.io` must be up), it's non-deterministic, and mining would mutate the live chain. It belongs as a `workflow_dispatch`/manual smoke test + a documented dev script, not in the normal CI run.

## Background (works in principle)

The live betanet (post-#323/#331) is a 2-node network: `seed.botho.io` (relay) + `faucet.botho.io` (minter), peered over VPC-internal IPs (172.31.x). The seed's gossip port **17100 is open to 0.0.0.0/0**, so an external node can bootstrap to it over the public internet. The faucet node joining the seed already proved intra-VPC peering; this validates the *external* path.

## Scope

1. **Script** (e.g. `scripts/join-betanet.sh` or `infra/betanet-smoke.sh`):
   - Build (or use) the botho binary for the host.
   - Generate a temp config + data dir (throwaway) with:
     - `network_type = "testnet"`, fresh `bootstrap_peers = ["/dns4/seed.botho.io/tcp/17100"]` (or the seed public IP multiaddr).
     - quorum config appropriate for a follower (sync-only) node.
   - Run `botho run --testnet` (sync-only by default; `--mint` optional behind a flag).
   - Wait for: peerCount ≥ 1 (connected to the betanet) AND chainHeight > 0 and advancing / matching the betanet tip (sync working).
   - Print a clear PASS/FAIL; tear down the temp node + data.
2. **Gated CI**: a `workflow_dispatch` job (`.github/workflows/betanet-smoke.yml`) that runs the script with a timeout and reports peer+sync status. Do NOT trigger on PR/push.
3. **Optional mine flag**: `--mine` variant that actually mints on the betanet (documented as "mutates the live chain; beta only").

## Acceptance criteria
- [ ] Running the script locally connects a fresh node to `seed.botho.io:17100`, reaches `peerCount ≥ 1`, and syncs the betanet chain (height matches/approaches the live tip).
[...]

Issue #354, opened 2026-06-15
chore: apply repo-wide rustfmt baseline + add CI fmt check (pinned nightly)


rjwalters:
## Summary

The repo has a `rustfmt.toml` but **no CI enforcement**, so the tree has drifted out of `rustfmt` compliance. The practical symptom: contributors (and Loom builders) who run `cargo fmt` get a massive whole-tree reformat diff, which then gets abandoned as unmergeable stashes (we just found and discarded four such stale fmt snapshots from Dec 2025–Jan 2026, each touching ~200 files). This is recurring friction and it muddies real diffs. Fix it once: apply a clean repo-wide `rustfmt` baseline in a single mechanical commit, then add a CI check so it never drifts again.

## Critical constraint — must use the pinned nightly

`rustfmt.toml` uses **nightly-only options**: `wrap_comments = true` and `imports_granularity = "Crate"`. These are unstable and only take effect under nightly rustfmt. The repo pins `nightly-2025-12-03` in `rust-toolchain.toml` (with the `rustfmt` component). Therefore:
- The baseline reformat MUST be run with `cargo +nightly-2025-12-03 fmt --all` (running stable `cargo fmt` would silently ignore the unstable options and produce a *different*, wrong result).
- The CI check MUST run on the same pinned nightly — NOT `dtolnay/rust-toolchain@stable`. Read the channel from `rust-toolchain.toml` so the check and the toolchain never disagree.

## Scope

1. **Baseline reformat commit.** Run `cargo +nightly-2025-12-03 fmt --all` over the workspace and commit the result as a single, isolated, mechanical commit (no logic changes). Respect the existing `rustfmt.toml` `ignore` list (`sgx/types`, `sgx/urts`, `sgx/sync`). This commit will be large and touch many files — that is expected and is the whole point (it pays down the drift once).
   - Note for the worktree: `fuzz/` is workspace-EXCLUDED, so `cargo fmt --all` will not reach it. Run `cargo +nightly-2025-12-03 fmt` separately inside `fuzz/` (and any other excluded member) so the fuzz targets are also baselined, OR explicitly note in the PR that excluded members are out of scope for this pass. Use judgment; document which.
2. **CI fmt check.** Add a lightweight workflow (e.g. `.github/workflows/fmt.yml`) — or a job in an existing workflow — that:
   - Checks out, installs the pinned nightly toolchain **from `rust-toolchain.toml`** with the `rustfmt` component.
[...]

PR #380, opened 2026-06-16
fix(network): drive chain-sync state machine for IBD / catch-up (#376)

### LINKED ISSUE #376: bug(network): node cannot sync an existing chain — no initial block download / catch-up (joins at genesis, rejects tip)
## Problem (significant — blocks new nodes joining an existing chain)

A node joining the betanet **after** the chain has advanced (tip at height N) **cannot sync the historical blocks**. After peering successfully (noise handshake OK, `peerCount=1`), the fresh node only receives the **current tip via gossip** and rejects it because its ledger is at genesis:

```
Failed to add reconstructed block: ... Invalid block: Expected height 1, got 94/95
```

It never fetches blocks 1..N — there appears to be **no initial block download / catch-up (backfill) path** for a peer joining an existing chain. The node stays at height 0 indefinitely.

Discovered via the #373 betanet-join smoke test (a fresh external node connected to seed.botho.io but never advanced past 0).

## Why this matters
This blocks the core "run your own node against the testnet" experience — new participants can't catch up to the live chain. It also wasn't caught by existing tests because `e2e_consensus_integration` starts all nodes at genesis together (they grow in lockstep; no node ever has to *catch up*), and the seed↔faucet pair were both wiped to genesis at the same time. So the **catch-up / IBD path is untested and currently non-functional**.

## Likely work
- Implement (or fix) initial block download: when a node's tip is behind a peer's, request the missing block range and apply sequentially before/while processing gossiped tip blocks.
- Add an integration test where a node starts with an empty ledger and **joins a network already at height N**, and must sync to the tip (distinct from the all-start-at-genesis tests).

## Acceptance
- [ ] A fresh node (empty ledger) pointed at a peer whose chain is at height N syncs to height N.
- [ ] Integration test: node joins an existing chain (height N > ring-size) and reaches the tip.
- [ ] The #373 betanet-join smoke test reaches a full PASS (external node syncs the betanet), not PARTIAL.

Found by #373. Related: #375 (dns4 bootstrap).

rjwalters:
Closes #376

## Root cause

The `ChainSyncManager` state machine (in `botho/src/network/sync.rs`) and the libp2p `GetStatus`/`GetBlocks` request/response protocol were **already fully implemented and unit-tested** — but the manager was **never instantiated or driven** in the node event loop (`botho/src/commands/run.rs`). Nothing ever sent a status or block-range request. A fresh node only received the current tip via gossip (`NewBlock`/`NewCompactBlock`) and rejected it with `Expected height 1, got N`, because the intermediate blocks were missing. There was no initial block download (IBD) / catch-up path actually running.

This was invisible to existing tests: `e2e_consensus_integration` starts every node at genesis and grows them in lockstep, so no node ever has to *catch up*.

## What this PR does (done)

**Wire `ChainSyncManager` into the node event loop** (`commands/run.rs`):
- Instantiate it at startup seeded with the local chain height.
- New `sync_tick` (every 2s): refresh the manager's local height, drive `tick()`, and dispatch its actions — sending `GetStatus` / `GetBlocks` to the selected peer.
- Feed `SyncResponse::Status` into `on_status`, apply `SyncResponse::Blocks` sequentially via `add_block_from_network` and report the new height via `on_blocks_added`, and surface `SyncResponse::Error` / peer disconnects back into the state machine.

**Keep a synced node current** (`network/sync.rs`): status is request/response (not gossiped), so a synced node previously had no way to learn a peer advanced. Added a periodic status re-poll (`STATUS_REFRESH_INTERVAL`, 30s) and made `on_status` re-enter `Downloading` from any non-downloading state when a peer is far enough ahead.

**Tests:**
- New `botho/tests/chain_sync_catchup_integration.rs` — drives the **production** `ChainSyncManager` against real LMDB ledgers:
  - `test_fresh_node_syncs_existing_chain_to_tip`: a genesis node (height 0) joins a peer already at height 95 (> ring size) and syncs to the exact tip; also asserts the bug's symptom (a genesis node rejects a lone gossiped tip block). **This is the required join-at-height-N proof.**
  - `test_node_resyncs_when_peer_advances`: catches up, peer grows, node resyncs.
  - `test_get_blocks_serves_requested_range` + message round-trip.
- New `network::sync` unit tests for synced-state refresh, re-entry, and a full batched download cycle.

## Verification

- `cargo test -p botho --lib --tests` — green (944 lib tests + all integration suites, including the 5 new catch-up tests and 45 sync unit tests).
- `cargo +nightly-2025-12-03 fmt --all -- --check` — clean (#354 gate).
- Pre-existing failures unrelated to this PR: two **doctests** in `network/privacy/{capacity,selection}.rs` fail on `origin/main` already (untouched here).

## Remaining / honest scope notes

- The in-process integration test proves the catch-up logic + ledger apply path end to end. The **live betanet external-sync case (#373 smoke test)** is expected to be addressed by this wiring (a fresh external node will now issue status/block requests instead of only consuming gossip), but I could not run the external smoke test from here, so I have not independently confirmed a full betanet PASS. Recommend re-running the #373 smoke test against a node built from this branch to flip it from PARTIAL to PASS before closing the betanet acceptance item.
- Reorg/fork handling on the sync path is out of scope (the manager applies a single best-peer's contiguous range; competing forks are not reconciled here).
- The current page size is 100 blocks per `GetBlocks` (existing `BLOCKS_PER_REQUEST`); large chains will page through many requests but converge.

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -27,8 +27,8 @@ use crate::{
         TransactionValidator,
     },
     network::{
-        BlockTxn, CompactBlock, GetBlockTxn, NetworkDiscovery, NetworkEvent, QuorumBuilder,
-        ReconstructionResult,
+        BlockTxn, ChainSyncManager, CompactBlock, GetBlockTxn, NetworkDiscovery, NetworkEvent,
+        QuorumBuilder, ReconstructionResult, SyncAction, SyncRequest, SyncResponse,
     },
     node::{MintedMintingTx, Node, SharedLedger},
     rpc::{
@@ -49,6 +49,15 @@ fn get_connected_peer_ids(discovery: &NetworkDiscovery) -> Vec<String> {
         .iter()
         .map(|p| p.peer_id.to_string())
         .collect()
+}
+
+/// Helper to get connected peers as libp2p `PeerId`s.
+///
+/// Used to drive the chain-sync state machine, which needs the typed peer
+/// identifiers (not their string form) to address sync request/response
+/// messages.
+fn get_connected_peers(discovery: &NetworkDiscovery) -> Vec<libp2p::PeerId> {
+    discovery.peer_table().iter().map(|p| p.peer_id).collect()
 }
 
 /// Check if minting should be enabled based on quorum config and connected
@@ -347,6 +356,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // Build SCP quorum set from connected peers (or just ourselves for solo mining)
     let scp_quorum_set = build_scp_quorum_set(&quorum, &local_peer_id);
 
+    // Capture the starting height before chain_state is moved into the
+    // consensus service; the sync state machine needs it to know how far
+    // behind the network we are at startup.
+    let local_height = chain_state.height;
+
     // Update initial metrics before chain_state is moved
     metrics_updater.set_block_height(chain_state.height);
     metrics_updater.set_difficulty(chain_state.difficulty);
@@ -368,6 +382,15 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
         "Consensus service initialized at slot {}",
         consensus.current_slot()
     );
+
+    // Chain-sync state machine: drives initial block download (IBD) / catch-up.
+    //
+    // A node joining an existing chain only learns about the current tip via
+    // gossip; that tip is rejected by the ledger because the intermediate
+    // blocks are missing ("Expected height 1, got N"). The sync manager closes
+    // this gap by polling peers for their chain status and, when we are behind,
+    // requesting the missing block range and applying it sequentially.
+    let mut sync_manager = ChainSyncManager::new(local_height);
 
     // Track minting state - can change as peers connect/disconnect
     let mut minting_enabled = false;
@@ -422,6 +445,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // Run the combined event loop
     let mut status_interval = tokio::time::interval(Duration::from_secs(10));
     let mut consensus_tick = tokio::time::interval(Duration::from_millis(500));
+    // Drive the chain-sync (IBD / catch-up) state machine. A short interval
+    // keeps a freshly joined node requesting peer status and missing block
+    // ranges promptly so it can reach the network tip without waiting on
+    // gossip alone.
+    let mut sync_tick = tokio::time::interval(Duration::from_secs(2));
     let mut minting_check_interval = tokio::time::interval(Duration::from_millis(100));
     let mut faucet_balance_interval = tokio::time::interval(Duration::from_secs(10));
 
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed as a consensus/networking change with the rigor it warrants. All make-or-break criteria hold.

### 1. Catch-up works and is proven
`cargo test -p botho --test chain_sync_catchup_integration` — **5/5 pass**, including the required join-at-height-N proof `test_fresh_node_syncs_existing_chain_to_tip`: node B starts at genesis (height 0), peer A is at height 95 (> ring size 20), B drives the production `ChainSyncManager` and reaches A's exact tip (height + tip_hash match). The test also asserts the bug's symptom (a lone gossiped tip block is rejected by a genesis node), so it is a genuine before/after regression guard, not a tautology.

### 2. Synced blocks are fully validated (no consensus bypass)
The apply path is `add_block_from_network` -> `Ledger::add_block`, the **same** method used for normal block acceptance. `Ledger::add_block` runs the full gauntlet: height + prev-hash, header/minting-tx consistency, **C1** chain-expected difficulty, PoW, **C2a** recomputed reward, **C2b** timestamp monotonicity/future-bound, **C4** tx_root recompute, **C3** ring-member UTXO resolution, lottery validation, and per-tx signature verification (`verify_transaction`). A malicious peer cannot feed bad blocks through the sync path. No bypass.

### 3. No regression to the live consensus/mining path
- `cargo test -p botho --lib` — **944 pass, 0 fail**.
- `cargo test -p botho --test e2e_consensus_integration` — **4/4 pass** (5-node mining + transactions + ring sigs).
- 45 `network::sync` unit tests pass.
The change is additive: a new `sync_tick` (2s) drives the state machine; the existing minting/gossip arms are untouched. The `SyncResponse::Blocks` arm now routes through the state machine but still applies via the same `add_block_from_network` path with the same dynamic-fee/consensus-state updates.

### 4. Fork/non-contiguous range safety
Out-of-scope reorg reconciliation is documented (single best-peer contiguous range). A bad/non-contiguous range is **safe**: `add_block` rejects it (height/prev-hash/validation error) inside an uncommitted write txn (no ledger corruption), the loop `break`s and calls `on_failure`, which transitions to `Failed` with retry backoff and records the failure against peer reputation. No crash, no corruption, no infinite loop (integration test caps iterations as a liveness assertion).

### CI / merge state
- All checks COMPLETED SUCCESS (skipped fuzz jobs are conditional, expected).
- `cargo +nightly-2025-12-03 fmt --all -- --check` — clean.
- `mergeStateStatus: CLEAN`.
- Pre-existing unrelated doctest failures in `network/privacy/{capacity,selection}.rs` exist on main — not introduced here, not blocking.

### Notes (non-blocking)
- The in-process test proves the catch-up + ledger apply path end to end. The external betanet #373 smoke test (PARTIAL -> PASS) is a reasonable follow-up; the builder is honest about not verifying it externally, and the in-process proof is sufficient for this increment.

Approving. The catch-up works, synced blocks are fully validated, the live path is not regressed, and bad ranges are handled safely.

Issue #386, opened 2026-06-16
bug(node): faucet balance-gated minting pause deadlocks the chain (sole minter) — pending txs never confirm


rjwalters:
## Problem (chain-halting deadlock on a sole-minter faucet)

The faucet node's balance-gated minting pause can **halt the entire chain** and leave transactions permanently unconfirmed when the faucet is the only miner (the current betanet topology: faucet mints, seed relays).

`botho/src/commands/run.rs` (~line 1304):
- Pauses minting when `balance > FAUCET_BALANCE_HIGH` (10,000 BTH).
- Resumes when `balance < FAUCET_BALANCE_LOW` (5,000 BTH).

**The deadlock:** when the faucet's confirmed balance is between 5k–10k+ and it's the sole minter:
1. Faucet pauses minting (balance > 10k).
2. A user requests a dispense → the faucet builds+submits a valid tx (RPC returns success + txHash) → it sits in the mempool.
3. **Nothing mines it** (minting paused, no other miner) → the dispense never confirms.
4. The faucet's *confirmed* balance never drops (dispenses only reduce balance once mined) → minting never resumes.
→ The chain is frozen; every dispense "succeeds" at the RPC but never confirms.

**Reproduced on the live betanet (2026-06-16):** faucet balance ~10,090 BTH (>10k → paused), chain stuck at height 202, a real dispense (`faucet_request` → success, tx `b05169a5…`) landed in the mempool (`mempoolSize=1`) and never mined.

## Fix

The balance pause must not apply when there are **pending transactions to mine**. Mint to clear the mempool regardless of the balance threshold; only pause for balance when the mempool is empty.

Concretely in `run.rs`:
- Read the mempool size in the minting-control loop.
- Pause-for-balance only when `mempool is empty` (`balance > HIGH && mempool_empty`).
- If `minting_paused_for_balance` and the mempool becomes non-empty, **resume minting** (subject to quorum) so pending txs confirm; it can re-pause once the mempool drains and balance is still high.

This breaks the deadlock: a pending dispense forces a block, the tx confirms, balance drops, and normal threshold behavior resumes.

## Acceptance criteria
- [ ] With the faucet's sole-minter balance > HIGH threshold, submitting a transaction causes the faucet to mine a block and confirm it (chain advances; mempool drains) — no permanent freeze.
- [ ] When the mempool is empty and balance > HIGH, minting still pauses (the original anti-accumulation intent is preserved).
[...]

PR #389, opened 2026-06-16
fix(node): give each minter its own shutdown flag so resume mines again

### LINKED ISSUE #388: bug(node): resumed minting (start after stop) never picks up work — zombie minter, no blocks produced
## Problem (minting does not resume work-generation after a stop→start cycle)

After the faucet's minting is **stopped and then restarted** (e.g. the balance-pause → resume cycle from #386/#387), the minting worker thread restarts but **never picks up work** — no blocks are produced, so pending transactions never confirm.

### Live evidence (betanet faucet, 2026-06-16, binary `cc7371b`)
1. At node startup, minting worked: it produced block 203 (`Thread picked up new work` fired, block mined). So minting-in-general is functional.
2. The faucet then balance-paused (mempool empty, balance > HIGH) → `stop_minting_public()`.
3. A dispense arrived → #387's fix correctly resumed: logs show
   ```
   INFO Resuming minting: 1 pending transaction(s) to mine
   INFO Starting minting with 1 threads
   ```
4. **But no `Thread picked up new work` / `Submitting minting tx to consensus` ever followed** (7+ minutes). Status reports `Minting: active`, mempool stays at 1, height frozen at 203. The dispense never confirms.

### Diagnosis
`start_minting_public()` after a prior `stop_minting_public()` restarts the worker **thread** but does not restart the **work-generation / block-template** path that feeds it (the part that logs `Thread picked up new work height=N`). The startup path (`run.rs` ~line 478, the initial `start_minting_public()?`) wires work-generation correctly; the resume path does not. So a node that ever pauses+resumes minting becomes a "zombie minter": flagged active, thread alive, but producing nothing.

Investigate `start_minting_public` / `stop_minting_public` (in `botho/src/node/` minter wiring) and the work-generation task: confirm whether stop tears down a template-generation loop/channel that start does not recreate. The fix must ensure a resumed minter regenerates block templates (picks up work) — verify with a stop→start→produces-a-block cycle, not just thread liveness.

## Acceptance criteria
- [ ] After `stop_minting_public()` then `start_minting_public()`, the node resumes producing blocks (a `Thread picked up new work` equivalent fires and a block is mined) — verified by a test that stops then restarts minting and asserts a block is produced.
- [ ] On the live sole-minter faucet: a dispense (pending tx) after a balance-pause resume actually gets mined and confirmed (chain advances, mempool drains via real mining).
- [ ] No regression to startup minting or the #386 pause/resume thresholds.

## Notes
Discovered by the post-deploy live smoke check, immediately after #387 (which fixed the #386 pause *decision*; this is the separate worker-restart bug it exposed). The live betanet faucet is currently a zombie minter (height 203, 1 pending dispense unmined). Relates to #386/#387.

rjwalters:
## Summary
After minting is stopped then restarted (`stop_minting_public()` then `start_minting_public()` — e.g. the #386/#387 balance pause→resume cycle), the minting worker thread restarted but never picked up work: no `Thread picked up new work`, no blocks, pending txs never confirmed. Startup minting worked; only *resumed* minting was broken. This made the live betanet faucet a "zombie minter" (active, but height frozen).

## Root cause
`Minter` was constructed with a **clone of the node-wide shutdown flag** (`Minter::new(threads, addr, self.shutdown.clone())`). `Minter::stop()` does `self.shutdown.store(true, ...)` — and because that `Arc<AtomicBool>` was shared with the node, the flag stayed `true` forever after the first stop. On the next `start_minting()`, a fresh `Minter` was created with the *same* (now-`true`) flag, so every spawned mint thread immediately failed the `while !shutdown.load(...)` guard and exited before picking up any work. The work-generation/`update_work` path was fine; the threads were dead-on-arrival.

The node's `self.shutdown` field was not wired to the run-loop's shutdown signal (run.rs uses its own local flag) — it was only ever consumed by the minter — so it was safe to remove.

## The fix
- `Minter::new` no longer takes a shutdown argument; each `Minter` creates and **owns** its own `shutdown` flag (`Arc::new(AtomicBool::new(false))`). A stop→start cycle therefore gets a fresh `false` flag and fully restarts the pipeline, identical to a fresh startup.
- Removed the now-unused `Node.shutdown` field and its `AtomicBool` import.
- No change to PoW, validation, work-generation, or the #386/#387 pause/resume decision logic.

## Changes
- `botho/src/node/minter.rs`: `Minter::new(threads, address)` owns its shutdown flag (doc-commented); add regression test `test_minter_resumes_work_after_stop_start_cycle`.
- `botho/src/node/mod.rs`: drop `Node.shutdown` field; call `Minter::new(threads, wallet.default_address())`.

| After stop then start, node resumes producing blocks — verified by a test that stops then restarts minting and asserts a block is produced | ✅ | `test_minter_resumes_work_after_stop_start_cycle`: starts a minter, asserts a minting tx arrives on the channel (proves work picked up + block mined, not just thread liveness), stops it, then starts a fresh minter and asserts a minting tx arrives **again** |
| Live faucet: dispense after a balance-pause resume gets mined | ✅ (mechanism) | Resumed minter now spawns threads with a `false` shutdown flag and mines; verified at the `Minter` lifecycle level the run-loop uses |
| No regression to startup minting or #386 thresholds | ✅ | Only the minter shutdown-flag ownership changed; pause/resume decision and work-generation untouched. Full `cargo test -p botho --lib` green (940 passed) |

## Test Plan
- `cargo test -p botho --lib` — 940 passed, 0 failed (includes the new regression test).
- `cargo +nightly-2025-12-03 fmt --all -- --check` — clean.

Closes #388

### DIFF
--- a/botho/src/node/minter.rs
+++ b/botho/src/node/minter.rs
@@ -83,7 +83,17 @@ pub struct Minter {
 }
 
 impl Minter {
-    pub fn new(threads: usize, address: PublicAddress, shutdown: Arc<AtomicBool>) -> Self {
+    /// Create a new minter.
+    ///
+    /// Each minter owns its own `shutdown` flag (created here, initialized to
+    /// `false`). This is deliberate: a minter must NOT share the node-wide
+    /// shutdown flag, because [`Minter::stop`] sets the flag to `true`
+    /// permanently. If the flag were shared, a subsequent `start_minting`
+    /// would spawn threads that immediately observe `shutdown == true` and exit
+    /// before ever picking up work — producing a "zombie minter" that reports
+    /// active but never mines (see issue #388, the pause→resume cycle from
+    /// #386/#387).
+    pub fn new(threads: usize, address: PublicAddress) -> Self {
         let (tx_sender, tx_receiver) = channel();
 
         // Initialize with default work (will be updated before minting starts)
@@ -97,7 +107,7 @@ impl Minter {
         Self {
             threads,
             address,
-            shutdown,
+            shutdown: Arc::new(AtomicBool::new(false)),
             total_hashes: Arc::new(AtomicU64::new(0)),
             txs_found: Arc::new(AtomicU64::new(0)),
             start_time: Instant::now(),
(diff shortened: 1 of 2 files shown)


rjwalters:
**Approved** (reviewed in main session — judge subagent hit transient API 500s).

Verified:
- **Root cause + fix correct:** `Minter` previously shared the node-wide `shutdown` Arc; `Minter::stop()` set it permanently true, so a re-created minter on resume was dead-on-arrival. Now `Minter::new` owns a fresh `AtomicBool::new(false)` per construction — each start gets a clean flag. Matches the live symptom exactly (startup minting worked; resumed minting produced nothing).
- **`Node.shutdown` removal is safe:** grep confirms nothing else read it (only the explanatory comment remains); it was never wired to the run-loop's real shutdown (run.rs uses its own local flag).
- **Node shutdown still stops the minter:** `run.rs:1408 node.stop_minting_public()` on the shutdown path → `minter.stop()`; no leaked threads.
- **Regression test genuine:** `test_minter_resumes_work_after_stop_start_cycle` asserts a minting tx is produced on BOTH the initial and the post-stop start (would fail under the old shared-flag design). Passes.
- Gates: `cargo test -p botho --lib` (940 pass incl. new), build green, pinned-nightly fmt clean.

Scope: only the 2 node files. No change to PoW/validation/work-generation or the #386/#387 pause thresholds.

PR #403, opened 2026-06-16
fix(secp256k1): add std feature opt-out to satisfy k256 precomputed-tables

### LINKED ISSUE #361: build: k256 precomputed-tables feature-unification break under --workspace (Linux-only, latent)
## Summary (low priority, latent, Linux-only)

Under a full-workspace build that includes the Tauri desktop crate's **Linux** GUI dependency subtree, `k256 v0.13.4` fails to compile with a feature-unification error:

```
error: `precomputed-tables` feature requires either `critical-section` or `std`   (k256-0.13.4/src/arithmetic/mul.rs:41)
error[E0433]: failed to resolve: unlinked crate `once_cell`                        (k256-0.13.4/src/arithmetic/mul.rs:57)
```

This is **latent**: it is not currently hit by any CI job or normal build, and it is **Linux-target-only**.

## Why it is currently dormant
- The only invocation that triggered it was `cargo bench --no-run --workspace` in the `compile-benchmarks` CI job. PR #358 (issue #356) scoped that job to the 6 bench-bearing crates (`-p ...`), so the desktop crate is no longer compiled there and the error no longer occurs in CI.
- Normal `cargo build` / `cargo test -p botho` do not build the full GUI subtree, so they are unaffected.

## Root cause (best current understanding)
- `crypto/secp256k1/Cargo.toml` depends on `k256` with `default-features = false, features = ["ecdsa","schnorr","alloc"]` (correct, no_std-friendly). In isolation k256 resolves WITHOUT `precomputed-tables`/`std`/`critical-section` and builds fine (verified: `cargo tree --workspace -e features -i k256` on macOS shows no `precomputed-tables`).
- On **Linux**, the Tauri desktop crate (`web/packages/desktop/src-tauri`) pulls the gtk/webkit/glib subtree (these are Linux-gated — `cargo tree -i glib-sys` is empty on macOS). Something in that Linux-only subtree pulls `k256` with `precomputed-tables` enabled (a k256 default feature) but without `std`/`critical-section`. Cargo's additive feature unification then builds k256 as `precomputed-tables + no-std + no-critical-section`, which is the invalid combination the error describes.
- It does not reproduce on macOS (Darwin) because the glib/gtk/webkit subtree — and thus the conflicting k256 consumer — is absent on that target. Verified locally: `cargo bench --no-run --workspace` exits 0 on macOS with k256 compiling clean.

## Suggested fix (must be developed/verified on Linux)
Likely one of:
- Add `"critical-section"` to `crypto/secp256k1`'s `k256` feature list so the `precomputed-tables` + no_std combination is valid when unified (needs a `critical-section` impl available; on a host build the `std` impl suffices). OR
- Identify the Linux-only consumer that enables k256 with default features and pin it to `default-features = false` (or align its k256 feature set), so `precomputed-tables` isn't force-enabled in no_std unification. OR
- Enable `k256/std` in the contexts where the full GUI subtree is built.

The right fix depends on inspecting the **Linux** `--workspace` feature resolution (`cargo tree --target x86_64-unknown-linux-gnu --workspace -e features -i k256`).

## Acceptance criteria
- [ ] On Linux, `cargo build --workspace` (and `cargo bench --no-run --workspace`) compiles `k256` without the `precomputed-tables`/`critical-section` error.
- [ ] No regression to the no_std posture of `crypto/secp256k1` (it must still build with `default-features = false`).
- [ ] macOS / normal builds unaffected.

## NOT auto-sweepable on macOS — needs Linux
This issue is intentionally left **without** `loom:issue` (kept as `loom:blocked` / triage) because it cannot be reproduced or a fix verified in a macOS dev/sweep environment — the failing dependency subtree is Linux-only. A fix must be developed and checked on a Linux host or via Linux CI. Pick this up when working on a Linux box, or wire a Linux `cargo build --workspace` check into CI first so the fix has a green/red signal.

Found during review of PR #358 (#356 benchmark-CI fix); mechanism diagnosed 2026-06-15.

rjwalters:
## Summary

Unblocks `cargo build --workspace` on Linux by adding a `std` feature to `bth-crypto-secp256k1` that activates `k256/std`. The workspace-build CI gate (added in #402) reproducibly fails red on `main` with the k256 `precomputed-tables` feature-unification error — this PR makes it green while preserving the crate's no_std opt-out.

## Changes

- `crypto/secp256k1/Cargo.toml`: add `std` feature (defaulted on) that activates `k256/std`. Consumers wanting no_std posture continue to work with `default-features = false`.
- `Cargo.lock`: updated — k256 now resolves with `once_cell` included (a consequence of `k256/std` being on workspace-wide).

## Root cause

k256 0.13.4's `default = ["arithmetic", "ecdsa", "pkcs8", "precomputed-tables", "schnorr", "std"]` and its `precomputed-tables` feature has `compile_error!` if neither `std` nor `critical-section` is also active.

Through transitive resolution on Linux (`libp2p-tls` → `libp2p-identity/secp256k1` → `k256` with default features), `precomputed-tables` ends up unified into the workspace build. On macOS the same dep tree resolves without `precomputed-tables` activated, so the bug was Linux-only and latent until #402's workspace-build job exposed it.

The fix forces `k256/std` to be present in any workspace build by anchoring it to `bth-crypto-secp256k1`'s default feature set. Since features unify workspace-wide, `std` is then guaranteed to be co-activated with `precomputed-tables`, satisfying k256's `compile_error!` gate.

| On Linux, `cargo build --workspace` compiles k256 without the `precomputed-tables`/`critical-section` error | ⏳ | Will be confirmed by the Workspace Build CI on this PR (the gate from #402). |
| No regression to the no_std posture of `crypto/secp256k1` (still builds with `default-features = false`) | ✅ | `cargo check -p bth-crypto-secp256k1 --no-default-features` passes locally on macOS. |
| macOS / normal builds unaffected | ✅ | `cargo check --workspace` and `cargo test -p bth-crypto-secp256k1` (5 unit + 1 doctest) pass locally on macOS. |

## Test Plan

- [x] Local: `cargo check -p bth-crypto-secp256k1` (default features) — passes
- [x] Local: `cargo check -p bth-crypto-secp256k1 --no-default-features` — passes (no_std opt-out preserved)
- [x] Local: `cargo check --workspace` — passes
- [x] Local: `cargo test -p bth-crypto-secp256k1` — 5/5 unit tests + doctest pass
- [ ] CI: `Workspace Build` (Linux) workflow goes from red → green on this PR (the definitive signal — this is exactly what #402 added to gate this fix)

Closes #361

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4587,6 +4587,7 @@ dependencies = [
  "cfg-if",
  "ecdsa",
  "elliptic-curve",
+ "once_cell",
  "sha2",
  "signature 2.2.0",
 ]
--- a/crypto/secp256k1/Cargo.toml
+++ b/crypto/secp256k1/Cargo.toml
@@ -9,16 +9,26 @@ rust-version = { workspace = true }
 
 [features]
 default = ["std"]
-# `std` activates k256/std, which is required to satisfy k256's
-# `precomputed-tables` feature when it is unified into the workspace build via
-# a transitive consumer (e.g. libp2p-identity/secp256k1, which Linux-target
-# resolution pulls through libp2p-tls). Without it, `cargo build --workspace`
-# fails on Linux with "`precomputed-tables` feature requires either
-# `critical-section` or `std`". Kept opt-out so no_std consumers can disable.
+# `std` activates k256/std for this crate's own consumers; kept as an opt-out so
+# no_std-leaning consumers can `default-features = false`.
 std = ["k256/std"]
 
 [dependencies]
-k256 = { version = "0.13", default-features = false, features = ["ecdsa", "schnorr", "alloc"] }
+# `critical-section` is activated UNCONDITIONALLY (not behind our `std` feature)
+# on purpose. k256 0.13's `precomputed-tables` feature has a `compile_error!`
+# unless either `std` or `critical-section` is also active. On Linux,
+# `cargo build --workspace` (the #402 gate) pulls a k256 build unit with
+# `precomputed-tables` enabled through the full workspace graph (the Tauri
+# desktop crate's Linux-gated GUI subtree, via libp2p), and that build unit is
+# NOT reached through `bth-crypto-secp256k1`'s feature graph — so a feature we
+# gate behind `std`/`default` never lands on it (the workspace dep edge for this
+# crate is declared `default-features = false`, and resolver v2 does not unify
+# our `std` onto the libp2p-path k256 unit). Because resolver v2 still resolves a
+# single shared k256 build unit per (package, target, build-kind) group, putting
+# `critical-section` directly on this crate's k256 dependency edge co-activates
+# it on that shared unit, satisfying the `precomputed-tables` gate without
+# requiring `std` (so it is robust for no_std targets too). See #361.
+k256 = { version = "0.13", default-features = false, features = ["ecdsa", "schnorr", "alloc", "critical-section"] }
 sha3 = { workspace = true }
 sha2 = { workspace = true }
 hmac = { workspace = true }
--- a/crypto/secp256k1/Cargo.toml
+++ b/crypto/secp256k1/Cargo.toml
@@ -8,7 +8,14 @@ license = "Apache-2.0"
 rust-version = { workspace = true }
 
 [features]
-default = []
+default = ["std"]
+# `std` activates k256/std, which is required to satisfy k256's
+# `precomputed-tables` feature when it is unified into the workspace build via
+# a transitive consumer (e.g. libp2p-identity/secp256k1, which Linux-target
+# resolution pulls through libp2p-tls). Without it, `cargo build --workspace`
+# fails on Linux with "`precomputed-tables` feature requires either
+# `critical-section` or `std`". Kept opt-out so no_std consumers can disable.
+std = ["k256/std"]
 
 [dependencies]
 k256 = { version = "0.13", default-features = false, features = ["ecdsa", "schnorr", "alloc"] }


rjwalters:
❌ **Changes Requested — the fix does not resolve #361 (Linux Workspace Build still red)**

I checked out the PR head (`8d464da`) and inspected the CI results. The `cargo build --workspace (Linux)` job — which is the exact gate #402 added and which the PR's own Test Plan calls "the definitive signal" — **fails with the identical error this PR set out to fix**:

```
   Compiling k256 v0.13.4
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `once_cell`
  --> k256-0.13.4/src/arithmetic/mul.rs:57:5
   |
57 | use once_cell::sync::Lazy;
   |     ^^^^^^^^^^ use of unresolved module or unlinked crate `once_cell`
error: could not compile `k256` (lib) due to 1 previous error
##[error]Process completed with exit code 101.
```

Run: https://github.com/botho-project/botho/actions/runs/27654611497/job/81786102798 (head_sha `8d464da`, matches the PR tip — this is not a stale run).

**Why the current approach doesn't work**

The PR's theory is that adding a default-on `std` feature to `bth-crypto-secp256k1` (activating `k256/std`) will, via workspace-wide feature unification, guarantee `std` is co-activated with `precomputed-tables` on the offending k256 build unit. The CI result disproves that:

- The workspace uses `resolver = "2"`. Under resolver v2, features are unified per-target/per-build-graph, **not** globally across every build unit the way resolver v1 did. Activating `k256/std` through `bth-crypto-secp256k1`'s default features does not force `std` onto the k256 instance that `libp2p-identity/secp256k1` → `libp2p-tls` pulls in with `precomputed-tables`.
- The error is still `once_cell` unresolved inside `k256/src/arithmetic/mul.rs`, which is gated behind k256's `std`/`critical-section`. So `precomputed-tables` is active **without** `std` on that unit — exactly the pre-fix state.
- Consistent with this, the `Cargo.lock` change in the PR adds `once_cell` to the **k256** dependency entry, but that addition alone does not change which feature set the conflicting build unit actually resolves to at build time.

**Required changes**

The fix must actually make `std` (or `critical-section`) co-activate with `precomputed-tables` on the k256 unit reached transitively through libp2p. Options to investigate (pick whichever the build proves out — verify against the Linux job, not just macOS):

1. Add a direct workspace/root-crate dependency on `k256` with `features = ["std"]` (or `critical-section`) so the unifying build unit includes it on Linux. A feature added to a leaf crate that isn't in the libp2p build path won't help.
2. Alternatively activate `critical-section` on k256, which also satisfies the `precomputed-tables` gate and may be more robust for no_std-leaning targets.
3. Consider whether the cleanest fix is at the consumer that pulls `precomputed-tables` (e.g. constraining libp2p-identity's `secp256k1` feature) rather than at `bth-crypto-secp256k1`, which is not on that dependency path.

Whatever the approach: **do not approve until `cargo build --workspace (Linux)` goes green on this PR.** A passing `cargo check --workspace` on macOS is not sufficient — #361 is Linux-only, and that's precisely why local macOS checks passed while CI is red.

Other CI checks (benchmarks, fuzz build guard, coverage) are passing; the sole blocker is the Linux workspace build, which is the issue under fix.

Marking `loom:changes-requested` + `loom:ci-failure`.

rjwalters:
Addressed the Judge's feedback. The previous approach gated `k256/std` behind `bth-crypto-secp256k1`'s `std`/`default` feature, which never reaches the offending k256 build unit — the workspace dep edge for this crate is declared `default-features = false`, and resolver v2 does not unify that `std` onto the k256 instance pulled through the Tauri desktop crate's Linux-gated libp2p subtree.

**Fix:** moved the activation onto the k256 dependency edge itself in `crypto/secp256k1/Cargo.toml` — `critical-section` is now activated **unconditionally** (not behind any feature):

```toml
k256 = { version = "0.13", default-features = false, features = ["ecdsa", "schnorr", "alloc", "critical-section"] }
```

**Why this lands on the libp2p-path unit:** resolver v2 resolves a single shared k256 build unit per (package, target, build-kind) group. Putting `critical-section` directly on this crate's k256 edge co-activates it on that shared unit alongside `precomputed-tables`, satisfying k256's `compile_error!` gate. `critical-section` does not require `std`, so it is robust for no_std targets too.

**Verification** (on the exact CI target, not just macOS) — `cargo build --workspace --target x86_64-unknown-linux-gnu` unit-graph:

```
k256 build units on linux target: 1
 features: [alloc, arithmetic, critical-section, digest, ecdsa, ecdsa-core, once_cell, precomputed-tables, schnorr, sha2, sha256, signature, std]
   precomputed-tables: True | critical-section: True | std: True
```

The single k256 unit now carries `precomputed-tables` **and** `critical-section` **and** `std`. Before this change the same unit-graph showed `precomputed-tables` active without the gate satisfied.

Also verified:
- `cargo build --workspace` (full, native) — green
- `cargo test -p bth-crypto-secp256k1` — 5 unit + 1 doctest pass
- `cargo check -p bth-crypto-secp256k1 --no-default-features` — passes (no_std opt-out preserved; `critical-section` still active, satisfying the gate)

No `Cargo.lock` change was needed beyond what was already present — `critical-section` is already a locked package and the k256 entry already lists `once_cell`; feature activation is manifest-driven.

The definitive signal remains the Linux `cargo build --workspace` CI job on this PR. Relabeled `loom:changes-requested` → `loom:review-requested`.

rjwalters:
✅ **Approved!** (re-judge after Doctor fix)

The definitive signal for #361 — the **\`cargo build --workspace (Linux)\`** CI job — is **GREEN** on the current head commit.

## CI Verification
- **Head evaluated**: \`35cc019e1495349ccbe1505b61b6352d1fffd542\` (matches PR tip exactly — not a stale run)
- **\`cargo build --workspace (Linux)\`**: ✅ **pass** (4m45s), run conclusion \`success\`, \`headSha\` confirmed = \`35cc019e...\`
- Other completed jobs (Crypto/Transaction/Compile Benchmarks, Coverage, Fuzz Summary): pass
- Remaining \`pending\` jobs (Node Benchmarks, Fuzz Build Guard) are non-blocking perf/fuzz jobs unrelated to the #361 compile gate; merge state is \`UNSTABLE\` solely due to these.

## Code Review
The new approach is correct and an improvement over the prior \`std\`-feature gating that the \`default-features = false\` workspace edge stripped:

- \`crypto/secp256k1/Cargo.toml\`: \`critical-section\` is now activated **unconditionally** on this crate's k256 dependency edge (\`features = ["ecdsa", "schnorr", "alloc", "critical-section"]\`). Because resolver v2 resolves a single shared k256 build unit per (package, target, build-kind), this co-activates \`critical-section\` onto the same unit that the Linux-gated Tauri/libp2p subtree pulls with \`precomputed-tables\`, satisfying k256's \`compile_error!\` gate **without requiring std** — so it remains robust for no_std targets.
- A \`std\` opt-out feature is retained (defaulted on, \`std = ["k256/std"]\`) so no_std-leaning consumers can still build with \`default-features = false\`.
- \`Cargo.lock\`: mechanical update (\`once_cell\` pulled into the k256 unit). In scope.

The change is minimal and focused — no unrelated edits, no over-reach.

**Note (non-blocking):** The PR description still narrates the older \`std\`-anchored approach in its "Root cause"/"Changes" sections; the actual diff and commit \`35cc019\` use the unconditional \`critical-section\` approach. The implementation is correct; the prose is just stale. Worth tidying on a future pass but does not block merge.

Closes #361.

Issue #400, opened 2026-06-16
consensus: SCP Msg fails to deserialize over the wire (bincode deserialize_identifier) — multi-node ballots never exchanged


rjwalters:
## Summary

Real multi-node SCP **cannot exchange ballot messages**: every received SCP message fails to deserialize, so the ballot protocol never reaches agreement and a peered set of minters started together stalls at **height 0**. Split out of #397 (root cause 1 of 2). This is the determinism-sensitive consensus-wire-format half; the quorum-set-wiring half is the sibling issue (see #397's Finding 2).

## Symptom

When peers connect *before* consensus init (e.g. start 3 nodes simultaneously with `min_peers = 2`), each builds a real multi-member quorum set (`solo_mode = false`, "Quorum satisfied: 3-of-3"), proposes values to SCP every slot — but every inbound SCP message logs:

```
WARN Failed to handle SCP message: Failed to deserialize SCP message:
     Bincode does not support Deserializer::deserialize_identifier
```

No node can ingest peers' SCP messages → ballots never converge → chain never advances past height 0.

## Root cause

`botho/src/consensus/service.rs:415` does `bincode::deserialize::<ScpMsg<ConsensusValue>>(...)`. The SCP `Msg` type (in the `bth-consensus-scp` crate) **serializes** under bincode but **fails to decode** — the decode path hits `Deserializer::deserialize_identifier`, which bincode does not support. This is the classic bincode limitation with serde enum representations that are internally/adjacently tagged or rely on field/variant *identifiers* (bincode only supports the default externally-tagged, index-based enum encoding). Some field of `Msg<ConsensusValue>` (or a nested type) is using a representation bincode can't round-trip.

## Suggested fix (decide during build; determinism-sensitive)

Either:
1. Change the serde representation of the offending `Msg` type / nested enum in `bth-consensus-scp` so it round-trips under bincode (remove tag attributes / identifier-based deserialization), **or**
2. Switch the SCP wire codec off `bincode` to a format that supports the representation (and pin it).

**This changes the consensus wire format** — coordinate with the no-hard-forks stance: pre-mainnet we have design freedom, but the fix must be deterministic and identical across nodes. Add a serialize→deserialize round-trip test for every `Msg` variant (`Nominate`, `Prepare`, `Commit`, `Externalize`, etc.) with realistic `ConsensusValue` payloads as a regression guard.

## Acceptance criteria
[...]

PR #404, opened 2026-06-17
feat(consensus): reconfigure SCP quorum set dynamically on peer churn

### LINKED ISSUE #401: consensus: SCP QuorumSet frozen at startup — never updated on PeerDiscovered, so late-connecting peers stay solo and fork
## Summary

The consensus `QuorumSet` is built **once** at startup and **never updated** when peers connect later. A node that boots before its peer is reachable builds a 1-of-1 (solo) quorum and stays solo forever — even after the peer connects. Two such peered minters then directly externalize their *own* minting tx every slot, mine **divergent chains**, and reject each other's blocks. Split out of #397 (root cause 2 of 2). Sibling: #400 (SCP wire deserialize). Both must land for a joiner's mined block to be accepted (see #396 steps 3–4).

## Symptom

When peers connect *after* consensus init (the common case — a node dials a not-yet-listening peer, finishes its discovery wait, and builds a 1-of-1 quorum before the peer's identify handshake lands), each node stays in `solo_mode` and directly externalizes its own tx every slot (`botho/src/consensus/service.rs:535`). Observed: 2 minting nodes, both peered (`peerCount=1`), produced different hashes at heights 5/10/15 and accepted **0** of each other's blocks:

```
WARN Failed to add reconstructed block: Failed to add network block:
     Invalid block: Previous block hash mismatch
```

## Root cause

`build_scp_quorum_set` (`botho/src/commands/run.rs:1419`) is called **once** at init (`run.rs:296`) from `discovery.peer_table()`. `NetworkEvent::PeerDiscovered` (`run.rs:608`) updates peer counts and minting eligibility but **does not reconfigure the consensus quorum**. So `solo_mode` (`service.rs:467`) is latched at boot: a node that started without a connected peer is permanently 1-of-1, regardless of who connects afterward.

## Suggested fix (decide during build)

- Reconfigure the consensus `QuorumSet` dynamically as peers connect/disconnect — update on `PeerDiscovered` / `PeerDisconnected` (`run.rs:608` and the disconnect handler), recomputing membership/threshold and toggling `solo_mode` off once enough peers are present.
- Define the threshold policy on membership change (botho uses `recommended`: n → threshold = n − floor((n−1)/3); confirm transitions are safe and don't strand an in-flight slot).
- Guard against churn/flapping (debounce, or only widen the set, or re-evaluate at slot boundaries) so a transient disconnect doesn't corrupt an active consensus round.

## Acceptance criteria
- [ ] A node that boots solo and later gains a peer transitions out of `solo_mode` and includes the peer in its quorum set (assert via log/RPC state), without restart.
- [ ] Two real `botho run` minters that connect *after* init converge on **one** chain (identical block hash at each height past the join point) instead of forking — requires #400 too for full ballot exchange; this issue alone fixes the latched-solo wiring.
- [ ] `PeerDisconnected` shrinks the quorum sensibly (node doesn't deadlock waiting on a departed member); no panic on churn.

## Evidence
- `botho/src/commands/run.rs:296` + `:1419` (quorum built once at init), `:608` (`PeerDiscovered` doesn't touch consensus quorum).
- `botho/src/consensus/service.rs:535` (solo-mode direct externalize), `:467` (`is_solo_mode`).
- Reproduced with `target/release/botho` over 127.0.0.1 loopback. Found while implementing #396.

Split from #397. Sibling: #400. Together they unblock the participate-and-mine half of #396.

rjwalters:
## Summary

The consensus `QuorumSet` was built **once** at startup (`build_scp_quorum_set`, `run.rs:296`/`:1419`) and never updated when peers connected later. `NetworkEvent::PeerDiscovered` updated peer counts and minting eligibility but never touched the consensus quorum, so `solo_mode` (`service.rs:467`) latched at boot: a node that started without a connected peer stayed permanently 1-of-1, and two late-peering minters forked.

This PR makes the consensus quorum track live membership: it is rebuilt whenever peers connect or disconnect, so a node that boots solo and later gains a peer leaves solo mode and includes the peer in its quorum set without a restart. Scope is limited to the latched-solo quorum-reconfiguration wiring; it does not depend on or include sibling #400.

## Changes

- **`ScpNode::set_quorum_set`** (`node_trait.rs`, `node_impl.rs`): new trait method that replaces the node's quorum set and rebuilds the current slot at its existing index. Documented as slot-boundary-only (abandons in-progress slot state). Forwarded by `LoggingScpNode` (`scp_log.rs`).
- **`ConsensusService::reconfigure_quorum`** (`service.rs`): recomputes membership/threshold and toggles solo mode. Guards:
  - **No-op debounce**: compares the incoming set against the active *and* any already-deferred set (`QuorumSet` `PartialEq` is order-independent); transient churn that doesn't change membership is ignored.
  - **Empty-set rejection**: never installs an empty quorum.
  - **Slot-boundary safety**: applies immediately only when nothing has been proposed/externalized in the current slot; otherwise stashes the change and applies it on the next `advance_slot`, so an in-flight round is never stranded.
  - Added `quorum_set()` getter for status/test assertions.
- **`run.rs`**: new `rebuild_scp_quorum_set` builds the SCP set from the current peer table (local node + connected peers) using the recommended BFT threshold `n - floor((n-1)/3)` (matches `QuorumConfig::effective_threshold`), clamping explicit thresholds to the member count so churn can't demand an unreachable quorum. Wired into both `PeerDiscovered` and `PeerDisconnected`. `local_peer_id` is now owned so `&discovery` can be re-borrowed in the handlers.

| Node that boots solo + gains a peer leaves solo mode and includes peer in quorum, no restart | ✅ | Unit test `solo_to_peered_transition_lifts_solo_mode` asserts `!is_solo_mode()`, 2 members, and that the SCP node's own quorum set updated; `PeerDiscovered` wiring calls `reconfigure_quorum` |
| Full two-node convergence (needs #400) — not depended on here; solo→peered transition tested without #400 | ✅ | Transition test uses `ConsensusService` directly, no SCP wire/deserialize path, no dependency on #400 |
| `PeerDisconnected` shrinks quorum sensibly; no panic on churn | ✅ | Unit test `peer_disconnect_shrinks_quorum_without_panic` grows to 3, shrinks 3→2→solo, then flaps connect/disconnect 5×; `rebuild_scp_quorum_set` always yields a non-empty set + satisfiable threshold |
| Threshold policy uses `recommended` and transitions don't strand an in-flight slot | ✅ | `rebuild_scp_quorum_set` uses `effective_threshold`; `reconfigure_mid_round_is_deferred_until_advance_slot` asserts a mid-round change defers and applies on `advance_slot` |
| Churn/flapping guard | ✅ | `reconfigure_with_same_membership_is_a_noop` + `empty_quorum_set_is_rejected` |

## Test Plan

- `cargo test -p botho --lib consensus::service` — 6 new tests pass.
- `cargo test -p bth-consensus-scp` — 98 existing SCP tests pass (no regressions).
- `cargo clippy -p botho -p bth-consensus-scp` — no new warnings/errors.
- `cargo build -p botho` — succeeds.

Closes #401

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,0 +1,6 @@
+# Loom-managed worktree marker
+# Created by .loom/scripts/worktree.sh
+# Issue: 401
+# Branch: feature/issue-401
+# Removing this file makes Loom treat the worktree as user-owned and refuse
+# to clean it up automatically.
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -415,8 +415,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
         .get_chain_state()
         .map_err(|e| anyhow::anyhow!("Failed to get chain state: {}", e))?;
 
-    // Create consensus service using local peer ID for node identity
-    let local_peer_id = discovery.local_peer_id();
+    // Create consensus service using local peer ID for node identity.
+    // Own the PeerId (rather than borrowing from `discovery`) so we can still
+    // call `&mut discovery` methods later while rebuilding the quorum set on
+    // peer connect/disconnect.
+    let local_peer_id = *discovery.local_peer_id();
     let node_id = peer_id_to_node_id(&local_peer_id);
 
     // Build SCP quorum set from connected peers (or just ourselves for solo mining)
@@ -622,6 +625,18 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             // Broadcast peer event to WebSocket clients
                             ws_broadcaster.peer_connected(new_peer_count, &peer_id.to_string());
 
+                            // Reconfigure the consensus quorum to include the
+                            // newly connected peer. This is what lifts a node
+                            // out of latched solo mode without a restart.
+                            let new_qs = rebuild_scp_quorum_set(&config, &discovery, &local_peer_id);
+                            if consensus.reconfigure_quorum(new_qs) {
+                                info!(
+                                    threshold = consensus.quorum_set().threshold,
+                                    members = consensus.quorum_set().members.len(),
+                                    "Consensus quorum reconfigured after peer connect"
+                                );
+                            }
+
                             // Re-evaluate minting eligibility
                             if mint && !minting_enabled {
                                 let connected = get_connected_peer_ids(&discovery);
@@ -662,6 +677,21 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
 
                             // Broadcast peer event to WebSocket clients
                             ws_broadcaster.peer_disconnected(new_peer_count, &peer_id.to_string());
+
+                            // Shrink the consensus quorum to drop the departed
+                            // peer so we don't deadlock waiting on a member that
+                            // is gone. rebuild_scp_quorum_set always yields a
+                            // non-empty set (local node is always a member) and
+                            // recomputes a satisfiable threshold, so churn can
+                            // never produce an empty or unreachable quorum.
+                            let new_qs = rebuild_scp_quorum_set(&config, &discovery, &local_peer_id);
+                            if consensus.reconfigure_quorum(new_qs) {
+                                info!(
+                                    threshold = consensus.quorum_set().threshold,
+                                    members = consensus.quorum_set().members.len(),
+                                    "Consensus quorum reconfigured after peer disconnect"
+                                );
+                            }
 
                             // Re-evaluate minting eligibility
                             if minting_enabled {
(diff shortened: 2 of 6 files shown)


rjwalters:
✅ **Approved!**

Thorough review of this sensitive consensus change. The dynamic quorum reconfiguration correctly fixes the latched-solo bug from #401 without altering SCP protocol semantics.

## Correctness

- **`ScpNode::set_quorum_set`** (`node_impl.rs`) mirrors the existing `reset_slot_index` pattern exactly — same `Slot::new` construction, clears `externalized_slots` — but preserves the current slot index and installs the new membership/threshold. No slot regression.
- **Threshold policy** uses the recommended BFT formula `n − floor((n−1)/3)`. `rebuild_scp_quorum_set` calls `effective_threshold(n - 1)` where `n = members.len()` (local + peers); `effective_threshold` internally re-adds self, so the math is consistent with `QuorumConfig` and the original `build_scp_quorum_set`. Explicit mode is clamped with `.min(n).max(1)`, so churn can never demand an unreachable quorum or drop below 1.
- **Solo toggle** (`is_solo_mode` = 1-of-1) flips correctly in both directions; `apply_quorum_set` logs the enter/exit transitions.

## Safety (mid-round vs slot boundary)

- The deferral path is verified end-to-end: `advance_slot` sets `externalized = None` and clears `proposed_values` **before** the pending-apply block, so a deferred quorum set is applied at a genuine slot boundary. `reconfigure_mid_round_is_deferred_until_advance_slot` confirms the in-flight round keeps the old quorum and the change lands on the next `advance_slot`.
- Empty-set rejection and the order-independent no-op debounce (compares against active *and* pending set) are correct and tested.

## CI

- `Compile Benchmarks`, `Fuzz Build Guard (PR)`, `rustfmt`, and all benchmark jobs: **pass**.
- `cargo build --workspace (Linux)` is path-filtered (`**/Cargo.toml`, `crypto/**`, `web/packages/desktop/**`, workflow file). This PR touches only `.rs` files under `botho/` and `consensus/scp/`, so the job correctly does not trigger — its absence is expected, not a missing signal.
- `mergeStateStatus`: **CLEAN**. Skipped Fuzz jobs are event/path-gated.

## Test Execution

- `cargo test -p botho --lib consensus::service` — ✅ 6/6 new tests pass (solo→peered transition, disconnect/flap churn, empty-set reject, no-op debounce, mid-round defer, boots-solo).
- `cargo test -p bth-consensus-scp` — ✅ 98 existing SCP tests pass, no regressions.
- `cargo clippy -p botho -p bth-consensus-scp` — ✅ no errors; no new warnings in any touched file (`service.rs`, `node_impl.rs`, `node_trait.rs`, `scp_log.rs`). Remaining workspace warnings are pre-existing and unrelated.
- The solo→peered transition test uses `ConsensusService` directly with no SCP wire/deserialize path, so it does **not** depend on sibling #400, as the PR claims.

## Follow-up

Filed #405 to track one documented tradeoff: the immediate-apply path keys "slot boundary" off service-level local state (`proposed_values`/`externalized`), which does not capture inbound peer ballot state in multi-node operation, so an immediate `set_quorum_set` could abandon a partial slot. This is acceptable for this pre-mainnet latched-solo fix (SCP re-syncs; reconfigure is debounced/rare; full ballot exchange needs #400), and worth hardening once #400 lands.

Solid, well-tested, well-documented change. Approved.

PR #406, opened 2026-06-17
fix(consensus): make QuorumSetMember externally tagged so SCP msgs decode over the wire

### LINKED ISSUE #400: consensus: SCP Msg fails to deserialize over the wire (bincode deserialize_identifier) — multi-node ballots never exchanged

rjwalters:
## Summary

Real multi-node SCP could not exchange ballot messages: every received SCP message failed to deserialize with `Bincode does not support Deserializer::deserialize_identifier`, so peered minters stalled at height 0.

**Root cause:** `QuorumSetMember` (in `bth-consensus-scp-types`) carried `#[serde(tag = "type", content = "args")]` — an *adjacently-tagged* serde representation. Adjacently/internally-tagged enums deserialize by reading the tag *field name* first, which routes through `Deserializer::deserialize_identifier`. Non-self-describing binary codecs refuse to implement that method (bincode errors; postcard explicitly says "a feature that PostCard will never implement"). Since `QuorumSet` is embedded in every SCP `Msg`, the whole message failed to decode even though it serialized fine.

**Fix chosen (issue option 1 — minimal repr change):** remove the tag attribute so the enum uses serde's default externally-tagged, index-based representation. I verified empirically that this round-trips under both bincode and postcard, whereas the tagged form fails under *both*. Option 2 (switching the wire codec off bincode) would NOT have fixed it — the workspace's alternative codec (postcard, via `bth-util-serial`) hits the same `deserialize_identifier` wall. So the repr change is the actual fix, and it keeps the existing bincode wire codec with no new dependency.

## Changes

- `consensus/scp/types/src/quorum_set.rs`: remove `#[serde(tag = "type", content = "args")]` from `QuorumSetMember`; document why it must stay externally tagged.
- `consensus/scp/src/msg.rs`: add `scp_msg_round_trips_through_wire_codec_all_variants` (round-trips every `Topic` variant of `Msg<V>` through bincode with a realistic consensus-value payload and a quorum set containing a nested `InnerSet`) and `scp_msg_wire_encoding_is_deterministic` (wire bytes independent of value insertion order).
- `consensus/scp/Cargo.toml` + `Cargo.lock`: add `bincode` as a dev-dependency for the regression test.
- `consensus/scp/play/src/main.rs`: update the JSON doc example to the new externally-tagged shape (`{"Node": ...}`).

Block/transaction serialization is untouched: the `#[prost(...)]` protobuf tags (used for Digestible block hashing) are unchanged; only the serde representation of one enum changed. The encoding remains deterministic (`BTreeSet` ordering + index-based enum tags; verified by the determinism test).

| 1. Unit test round-trips EVERY `Msg<V>` variant; FAILS on main, PASSES after fix | done | `scp_msg_round_trips_through_wire_codec_all_variants` covers Nominate, NominatePrepare, Prepare, Commit, Externalize. Confirmed it FAILS on the pre-fix repr with `Bincode does not support Deserializer::deserialize_identifier` and PASSES after removing the tag attribute. |
| 2. `deserialize_identifier` path gone; encode→decode symmetric at call site and crate | done | `botho/src/consensus/service.rs` bincode encode/decode now round-trips (52 botho consensus tests pass); crate-level round-trip test passes. |
| 3. No change to block/tx serialization; determinism preserved | done | Only serde repr of `QuorumSetMember` changed; prost tags untouched. `scp_msg_wire_encoding_is_deterministic` asserts byte-identical output across 50 shuffles. |

## Test Plan

- `cargo test -p bth-consensus-scp --lib msg_tests` — 31 passed (incl. 2 new).
- Temporarily re-added the tag attribute: round-trip test FAILED with the exact `deserialize_identifier` error; restored fix and it passes.
- `cargo test -p bth-consensus-scp-types --lib` — 5 passed.
- `cargo test -p botho --lib consensus` — 52 passed (incl. #401 dynamic-quorum tests).
- `cargo check -p botho -p bth-consensus-scp-play` — clean.
- `cargo clippy`/`cargo fmt` on touched crates — no new warnings.

Closes #400

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1124,6 +1124,7 @@ dependencies = [
 name = "bth-consensus-scp"
 version = "7.1.0"
 dependencies = [
+ "bincode",
  "bth-common",
  "bth-consensus-scp-types",
  "bth-crypto-digestible",
--- a/consensus/scp/Cargo.toml
+++ b/consensus/scp/Cargo.toml
@@ -32,6 +32,7 @@ bth-consensus-scp-types = { workspace = true, features = ["test_utils"] }
 bth-util-logger-macros = { workspace = true }
 bth-util-test-helper = { workspace = true }
 
+bincode = { workspace = true }
 crossbeam-channel = { workspace = true }
 dashmap = { workspace = true }
 maplit = { workspace = true }
--- a/consensus/scp/play/src/main.rs
+++ b/consensus/scp/play/src/main.rs
@@ -30,8 +30,8 @@ pub struct Config {
     /// Quorum set.
     ///
     /// The quorum set is represented in JSON. For example:
-    /// {"threshold":1,"members":[{"type":"Node","args":"node2.test.botho.
-    /// com:8443"},{"type":"Node","args":"node3.test.botho.com:4843"}]}
+    /// {"threshold":1,"members":[{"Node":"node2.test.botho.
+    /// com:8443"},{"Node":"node3.test.botho.com:4843"}]}
     #[clap(long, value_parser = parse_quorum_set_from_json, env = "MC_QUORUM_SET")]
     pub quorum_set: Option<QuorumSet>,
 
--- a/consensus/scp/src/msg.rs
+++ b/consensus/scp/src/msg.rs
@@ -680,8 +680,155 @@ impl<V: Value, ID: GenericNodeId> fmt::Display for Msg<V, ID> {
 #[allow(clippy::unwrap_used, clippy::expect_used, clippy::panic)]
 mod msg_tests {
     use super::*;
-    use crate::test_utils::test_node_id;
+    use crate::{test_utils::test_node_id, QuorumSetMember};
     use rand::seq::SliceRandom;
+    use serde::{Deserialize, Serialize};
+
+    /// A realistic consensus value that mirrors the shape of the node's
+    /// `ConsensusValue` (a 32-byte tx hash, a minting flag, and a priority).
+    /// Defined locally so the codec regression test exercises the same
+    /// serialized layout the node sends over the wire without depending on the
+    /// `botho` crate (which depends on this one).
+    #[derive(
+        Clone, Copy, Debug, Deserialize, Digestible, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize,
+    )]
+    struct TestConsensusValue {
+        tx_hash: [u8; 32],
+        is_minting_tx: bool,
+        priority: u64,
+    }
+
+    impl TestConsensusValue {
+        fn new(seed: u8, minting: bool, priority: u64) -> Self {
+            Self {
+                tx_hash: [seed; 32],
+                is_minting_tx: minting,
+                priority,
+            }
+        }
+    }
+
+    /// Build a non-trivial quorum set that includes a nested `InnerSet` member.
+    ///
+    /// The nested `InnerSet` exercises the `QuorumSetMember` enum, which is the
+    /// type whose serde representation previously broke binary deserialization.
+    fn realistic_quorum_set() -> QuorumSet {
+        let inner = QuorumSet::new_with_node_ids(1, vec![test_node_id(3), test_node_id(4)]);
+        QuorumSet::new(
+            2,
+            vec![
+                QuorumSetMember::Node(test_node_id(1)),
+                QuorumSetMember::Node(test_node_id(2)),
+                QuorumSetMember::InnerSet(inner),
+            ],
+        )
+    }
+
+    /// One representative `Msg` for every SCP `Topic` variant, each carrying a
+    /// realistic consensus-value payload and a quorum set with a nested
+    /// `InnerSet`.
+    fn all_topic_messages() -> Vec<Msg<TestConsensusValue>> {
+        let v1 = TestConsensusValue::new(1, false, 100);
+        let v2 = TestConsensusValue::new(2, true, 200);
+        let qs = realistic_quorum_set();
+
+        let nominate = NominatePayload {
+            X: BTreeSet::from_iter([v1]),
+            Y: BTreeSet::from_iter([v2]),
+        };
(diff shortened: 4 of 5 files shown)


rjwalters:
✅ **Approved!**

Thorough review of this determinism-sensitive consensus wire-format change. The fix is correct, minimal, and well-tested.

## Fix correctness (issue #400)
Confirmed the root cause: `#[serde(tag = "type", content = "args")]` on `QuorumSetMember` forced deserialization through `Deserializer::deserialize_identifier`, which bincode (the SCP wire codec at `botho/src/consensus/service.rs:415`) refuses to support. Removing the attribute reverts to serde's default externally-tagged, index-based representation, which round-trips cleanly.

**Regression test verified empirically:**
- New repr: `scp_msg_round_trips_through_wire_codec_all_variants` PASSES for all 5 Topic variants (Nominate, NominatePrepare, Prepare, Commit, Externalize).
- I temporarily restored the tag attribute on a local copy and re-ran the test — it FAILED with the exact `Bincode does not support Deserializer::deserialize_identifier` error. The builder's "fails on old, passes on new" claim is confirmed.

## Determinism
The determinism test is meaningful. The wire-serialized `Msg` path contains no nondeterministic containers: `QuorumSet.members` is a `Vec` (construction order), `NominatePayload.{X,Y}` are `BTreeSet` (value-sorted), and the `HashSet` usages in `quorum_set.rs` are confined to computation methods (`nodes()`, validation), not serialized fields. `scp_msg_wire_encoding_is_deterministic` shuffles insertion order into the `BTreeSet`s across 50 iterations and asserts byte-identical output — correctly proving the encoding reflects sorted order, not insertion order.

## Scope (block/tx serialization unchanged)
Confirmed only the serde representation of one enum changed:
- `#[prost(message, tag = 1)]` / `tag = 2` on `QuorumSetMember` are **untouched** — Digestible block hashing is unaffected.
- No transaction/block serde or prost attributes touched.
- `bincode` is added under `[dev-dependencies]` (Cargo.toml line 35) — dev-only, confirmed.
- The scp-play JSON doc change (`{"type":"Node","args":...}` → `{"Node":...}`) is a CLI `--quorum-set` doc example in `bth-consensus-scp-play`, a standalone debug binary that is NOT a dependency of the botho node and persists nothing. Acceptable for a pre-mainnet diagnostic tool.

## Local verification
- `cargo test -p bth-consensus-scp --lib msg_tests` — 31 passed (incl. 2 new).
- `cargo test -p bth-consensus-scp-types --lib` — 5 passed.
- `cargo test -p botho --lib consensus` — 52 passed (exercises the real node-level bincode round-trip).
- `cargo clippy` on touched crates — no new warnings (the 4 pre-existing `useless vec!` warnings are in `slot.rs`/`logger`, untouched by this PR).

## CI
- cargo build --workspace (Linux) — pass (4m47s)
- Generate Coverage Report — pass
- rustfmt --check (pinned nightly) — pass
- mergeStateStatus: CLEAN

Solid fix. The added documentation in `quorum_set.rs` explaining *why* the enum must stay externally tagged is a nice guard against regression.

PR #408, opened 2026-06-17
fix(consensus): defer quorum reconfigure when SCP slot holds peer ballot/nominate state

### LINKED ISSUE #405: consensus: immediate quorum reconfigure can abandon in-flight SCP peer ballot state (multi-node)
## Background

Discovered while evaluating PR #404 (closes #401), which makes the SCP `QuorumSet` track live peer membership.

`ConsensusService::reconfigure_quorum` applies a new quorum set immediately when `proposed_values.is_empty() && externalized.is_none()` (the "slot boundary" guard). Those two fields track only the *local* node's proposal/externalization state. In multi-node operation, inbound peer ballot/nominate messages flow directly into the SCP node and are **not** reflected in the service-level guard.

As a result, an immediate `set_quorum_set` (which rebuilds `current_slot` fresh and clears `externalized_slots`) could discard partial SCP slot state that was accumulated from peer messages for an in-flight slot, even though the service-level guard considered the slot "idle."

## Why this is acceptable for #404

- SCP tolerates partial-slot loss and re-syncs via subsequent peer messages and catch-up.
- Reconfiguration only fires on actual membership change (debounced), which is rare.
- Full multi-node ballot exchange requires sibling #400, which is not yet landed; #404 deliberately scopes to the latched-solo wiring.

## Suggested hardening (once #400 lands)

- Consider deriving `at_slot_boundary` from the SCP node's own slot phase (e.g. whether `current_slot` has received any nominate/ballot messages) rather than only service-level local state.
- Or always defer reconfiguration to the next `advance_slot` in multi-node mode, accepting at most one slot of latency.

Found during code evaluation of PR #404.

---

### Status update (2026-06-16)

The original body deferred this hardening until sibling #400 landed. **Both prerequisites are now merged to `main`:**

- #401 (PR #404, MERGED) — SCP `QuorumSet` now tracks live peer membership via `ConsensusService::reconfigure_quorum`.
- #400 (PR #406, MERGED) — `QuorumSetMember` is externally tagged so SCP `Msg` decodes over the wire, enabling real multi-node ballot exchange.

The "once #400 lands" precondition is satisfied. This is now **actionable hardening**, not a deferred follow-up. The narrow latched-solo path that #404 scoped to is widening into a real multi-node path where the bug below becomes reachable.

### Problem Statement

`ConsensusService::reconfigure_quorum` (`botho/src/consensus/service.rs`) decides whether it can swap the quorum set immediately or must defer to the next slot boundary using:

```rust
let at_slot_boundary = self.proposed_values.is_empty() && self.externalized.is_none();
```

Both `proposed_values` and `externalized` are **service-level fields that only reflect the LOCAL node's activity** for the current slot — what *this* node proposed and whether *this* node has observed externalization. They are populated by `propose_pending_values` / `check_externalized`.

In multi-node operation, inbound peer `nominate`/`ballot` messages are fed directly into `self.scp_node` via `handle_message` and accumulate inside the SCP node's `current_slot` (its `X`/`Y`/`Z` nomination sets and ballot counter `B.N`). **None of that peer-driven slot state is mirrored into the service-level guard.** So a slot can be genuinely in flight — peers have voted/accepted nominations, a ballot is being prepared — while `proposed_values.is_empty() && externalized.is_none()` is still `true` because this node hasn't locally proposed or externalized yet.

When the guard wrongly reports `at_slot_boundary == true`, `apply_quorum_set` calls `self.scp_node.set_quorum_set(...)`. Per its own contract (`consensus/scp/src/node/node_trait.rs`), that method *"abandons any in-progress state for the current slot and any stored externalized slots"* and rebuilds `current_slot` fresh at the same index. The result: **partial SCP ballot/nominate state accumulated from peers is silently discarded mid-round.**

### Root Cause

The boundary check derives "is the slot idle?" from the wrong source of truth (local service fields) instead of from the SCP node's own slot phase. The SCP node already exposes this via `get_current_slot_metrics() -> SlotMetrics` (`consensus/scp/src/slot.rs`):

```rust
pub struct SlotMetrics {
    pub phase: Phase,                 // NominatePrepare | Prepare | Commit | Externalize
    pub num_voted_nominated: usize,   // |X|
    pub num_accepted_nominated: usize,// |Y|
    pub num_confirmed_nominated: usize,// |Z|
    pub cur_nomination_round: u32,
    pub bN: u32,                      // highest ballot counter
}
```

A slot that has received any peer nominate/ballot activity will show non-empty nomination sets and/or `bN > 0` and/or a phase past the initial state — even when local `proposed_values` is empty.

### Severity / Why it matters

- This is **determinism-relevant consensus code in a pre-mainnet chain.** Per project philosophy (no hard forks; validate empirically pre-mainnet), silent loss of agreed-upon protocol state is exactly the class of subtle correctness bug to eliminate before testnet hardening.
- It is **low-probability but not impossible**: reconfiguration only fires on a debounced membership change, and the race window is the gap between peers populating a slot and this node proposing/externalizing. But peer churn during an active round is precisely when this fires.
- SCP can re-sync via subsequent messages and catch-up, so this is a liveness/efficiency hazard (extra rounds, possible stalled slot) rather than a guaranteed safety break — but it can manifest as a spurious fork or a stuck slot under churn, which is hard to diagnose.

### Implementation Guidance

Two directions, from the issue. Recommended: **Option A** (precise) with **Option B** as a simpler fallback.

#### Option A — Derive `at_slot_boundary` from the SCP node's slot phase (recommended)

Augment the guard so it also consults the SCP node's current-slot metrics, treating the slot as in-flight if peers have populated it:

```rust
let metrics = self.scp_node.get_current_slot_metrics();
let scp_slot_active = metrics.num_voted_nominated > 0
    || metrics.num_accepted_nominated > 0
    || metrics.num_confirmed_nominated > 0
    || metrics.bN > 0
    || metrics.phase != Phase::NominatePrepare; // any phase past initial nominate
let at_slot_boundary =
    self.proposed_values.is_empty() && self.externalized.is_none() && !scp_slot_active;
```

- Keeps the immediate-apply fast path for genuinely idle slots (solo bootstrap, fresh slot).
- Note: `get_current_slot_metrics(&mut self)` takes `&mut self` on the trait — confirm `reconfigure_quorum` can call it (it already holds `&mut self`). `Phase` is in `bth_consensus_scp` (`consensus/scp/src/slot.rs`); import as needed.
- Verify the exact "initial / idle" phase value: the enum is `NominatePrepare | Prepare | Commit | Externalize` (nominate and prepare begin concurrently, so a brand-new slot is in `NominatePrepare` with empty X/Y/Z). The empty-nomination-sets + `bN == 0` check is the robust signal; phase is a secondary guard.

#### Option B — Always defer in multi-node mode (simpler, ≤1 slot latency)

When not in solo mode, always stash to `pending_quorum_set` and let `advance_slot` apply it. Accepts at most one slot of reconfiguration latency in exchange for never touching an in-flight SCP slot. Use `self.is_solo_mode()` to gate: solo keeps the immediate path, multi-node always defers.

```rust
let at_slot_boundary = self.is_solo_mode()
    && self.proposed_values.is_empty()
    && self.externalized.is_none();
```

This is the lowest-risk change and is easy to reason about for a pre-mainnet chain; the cost (one slot of latency on a rare membership change) is negligible. If unsure, prefer Option B for safety and revisit precision later.

### Constraints — do NOT drift SCP semantics

- Do **not** modify `set_quorum_set` semantics in `bth_consensus_scp` — it correctly documents that it abandons in-progress state and must only be called at a boundary. The fix belongs in the *caller's* boundary decision, not the SCP primitive.
- Do **not** add a code path that mutates an active SCP slot's nomination/ballot state. The only allowed behaviors are: apply at a true boundary, or defer to `advance_slot`.
- Preserve existing debounce/no-op behavior for unchanged membership and the empty-set rejection.

## Acceptance Criteria

- [ ] `reconfigure_quorum` no longer applies a new quorum set immediately when the SCP node's current slot has accumulated peer nominate/ballot state (non-empty X/Y/Z nomination sets, `bN > 0`, or a phase past initial nominate); it defers to the next `advance_slot` instead.
- [ ] A new unit test demonstrates that reconfiguration during an in-flight, **peer-populated** slot (not just the local solo-externalized case) does NOT discard the SCP slot's ballot/nominate state — i.e. `pending_quorum_set` is set, the active quorum set is unchanged, and `set_quorum_set` is not invoked until `advance_slot`.
- [ ] The immediate-apply fast path still works for a genuinely idle/fresh slot (no peer state, nothing locally proposed) — verified by an existing or new test.
- [ ] No SCP protocol-semantics drift: `bth_consensus_scp::node::Node::set_quorum_set` behavior and the `Phase`/`SlotMetrics` contracts are unchanged.
- [ ] The empty-quorum-set rejection and the same-membership debounce/no-op behavior are preserved.
- [ ] All existing consensus tests pass, including `reconfigure_mid_round_is_deferred_until_advance_slot` and the flapping/no-op tests in `service.rs`.

## Affected Files

- `botho/src/consensus/service.rs` — Update the `at_slot_boundary` computation in `reconfigure_quorum` (around line 338) to also consult the SCP node's current-slot state (Option A) or to always defer in multi-node mode (Option B). Add a new `#[cfg(test)]` unit test alongside `reconfigure_mid_round_is_deferred_until_advance_slot` exercising the peer-populated-slot case.
- `consensus/scp/src/node/node_trait.rs` — Reference only: `get_current_slot_metrics`, `set_quorum_set` (and its boundary contract) are defined here. No change expected unless a read-only `&self` metrics accessor is needed (currently `get_current_slot_metrics` takes `&mut self`).
- `consensus/scp/src/slot.rs` — Reference only: `SlotMetrics` and `Phase` definitions used by Option A. No change expected.
- `consensus/scp/src/node/node_impl.rs` — Reference only: `get_current_slot_metrics` / `set_quorum_set` implementations. No change expected.

## Test Plan

- [ ] Automated test (new): Construct a multi-node `ConsensusService`, feed an inbound peer SCP nominate/ballot message via `handle_message` so the SCP `current_slot` is populated, then call `reconfigure_quorum` with a changed membership. Assert it is deferred (`pending_quorum_set.is_some()`, active `quorum_set` unchanged, `scp_node.quorum_set()` unchanged), then `advance_slot` and assert the new set is applied. If directly injecting a peer message is impractical in the unit harness, assert via `get_current_slot_metrics()` that the boundary decision treats a non-empty slot as in-flight.
- [ ] Automated test (regression): A genuinely idle/fresh slot still takes the immediate-apply path (quorum set applied without deferral).
- [ ] Automated tests (existing, must still pass): `reconfigure_mid_round_is_deferred_until_advance_slot`, `reconfigure_with_same_membership_is_a_noop`, the empty-set rejection, and the flapping/churn test in `botho/src/consensus/service.rs`.
- [ ] Build/test: `cargo test -p botho consensus::service` (or the workspace consensus tests) and `cargo build --workspace`.
- [ ] Edge cases: solo->multi-node transition during an active slot; reconfiguration arriving exactly at `advance_slot`; phase exactly at `NominatePrepare` with empty nomination sets (must be treated as boundary, not in-flight).

rjwalters:
## Summary

Closes #405.

`ConsensusService::reconfigure_quorum` decided whether it could swap the quorum set immediately vs. defer to the next slot boundary using only **local** service fields:

```rust
let at_slot_boundary = self.proposed_values.is_empty() && self.externalized.is_none();
```

Those fields track only *this* node's proposal/externalization. In multi-node operation, inbound peer `nominate`/`ballot` messages accumulate inside the SCP node's `current_slot` (its X/Y/Z nomination sets and ballot counter `B.N`) and are never mirrored into that guard. So an "immediate" `apply_quorum_set` -> `set_quorum_set` could rebuild the slot fresh and **silently discard peer-agreed ballot/nominate state mid-round**, even though the local guard considered the slot idle.

## Fix (Option A — recommended)

Augment the boundary decision to also consult `scp_node.get_current_slot_metrics()`. The slot is treated as in-flight (and reconfiguration deferred to the next `advance_slot`) when it holds peer-driven state:

```rust
let metrics = self.scp_node.get_current_slot_metrics();
let scp_slot_active = metrics.num_voted_nominated > 0
    || metrics.num_accepted_nominated > 0
    || metrics.num_confirmed_nominated > 0
    || metrics.bN > 0
    || metrics.phase != ScpPhase::NominatePrepare;
let at_slot_boundary =
    self.proposed_values.is_empty() && self.externalized.is_none() && !scp_slot_active;
```

The immediate fast path is preserved for genuinely idle/fresh slots (solo bootstrap, brand-new slot in `NominatePrepare` with empty X/Y/Z and `bN == 0`).

**Why Option A over Option B:** Option A is the issue's recommended (precise) direction, keeps the immediate-apply fast path for idle multi-node slots, and is cleanly testable via `get_current_slot_metrics()`. It also proved straightforward to exercise with a real peer message, so the lower-precision Option B fallback was unnecessary.

## Constraints honored
- No change to `set_quorum_set`, `Phase`, or `SlotMetrics` contracts — the fix is confined to the caller's boundary decision.
- Empty-quorum-set rejection and the same-membership debounce/no-op are preserved.

## Tests
- **New** `reconfigure_during_peer_populated_slot_is_deferred`: feeds a real peer `NominatePrepare` message (carrying a structurally valid transfer tx so the SCP validity callback accepts the value). The SCP slot reaches `Y=1 / Z=1 / bN=1 / phase=Prepare` while the local guard fields are still idle. Asserts the reconfiguration is **deferred** (`pending_quorum_set` set, active service + SCP quorum sets unchanged, in-flight slot metrics untouched), then `advance_slot` applies it.
- **New** `reconfigure_on_idle_slot_applies_immediately`: a genuinely idle/fresh slot still takes the immediate-apply fast path.
- Existing `reconfigure_mid_round_is_deferred_until_advance_slot`, `reconfigure_with_same_membership_is_a_noop`, `empty_quorum_set_is_rejected`, and the flapping/churn test all still pass.

### Results
- `cargo test -p botho --lib consensus` — 54 passed, 0 failed
- `cargo test -p bth-consensus-scp` — 100 passed, 0 failed (2 ignored)
- `cargo build --workspace` — clean

### DIFF
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -1128,8 +1128,18 @@ mod tests {
         // The peer-driven slot state must survive the deferral untouched.
         let after = svc.scp_node.get_current_slot_metrics();
         assert_eq!(
-            (after.num_voted_nominated, after.num_accepted_nominated, after.bN, after.phase),
-            (metrics.num_voted_nominated, metrics.num_accepted_nominated, metrics.bN, metrics.phase),
+            (
+                after.num_voted_nominated,
+                after.num_accepted_nominated,
+                after.bN,
+                after.phase
+            ),
+            (
+                metrics.num_voted_nominated,
+                metrics.num_accepted_nominated,
+                metrics.bN,
+                metrics.phase
+            ),
             "deferral must not disturb the in-flight SCP slot state"
         );
 
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -6,7 +6,8 @@ use super::{validation::TransactionValidator, value::ConsensusValue};
 use crate::ledger::ChainState;
 use bth_common::NodeID;
 use bth_consensus_scp::{
-    create_null_logger, msg::Msg as ScpMsg, node::Node, QuorumSet, ScpNode, SlotIndex,
+    create_null_logger, msg::Msg as ScpMsg, node::Node, slot::Phase as ScpPhase, QuorumSet,
+    ScpNode, SlotIndex,
 };
 use serde::{Deserialize, Serialize};
 use std::{
@@ -333,9 +334,29 @@ impl ConsensusService {
         }
 
         // Determine whether the current slot is at a safe boundary to swap the
-        // quorum set without stranding an in-flight round. A slot is safe if we
-        // have not proposed anything into it and it has not externalized.
-        let at_slot_boundary = self.proposed_values.is_empty() && self.externalized.is_none();
+        // quorum set without stranding an in-flight round.
+        //
+        // The local service fields (`proposed_values`, `externalized`) only
+        // reflect *this* node's activity for the slot. In multi-node operation,
+        // inbound peer nominate/ballot messages accumulate directly inside the
+        // SCP node's `current_slot` and are NOT mirrored into those fields. So
+        // we must also consult the SCP node's own slot phase: if peers have
+        // populated the slot (non-empty nomination sets, a ballot counter past
+        // zero, or a phase past the initial NominatePrepare), an "immediate"
+        // `set_quorum_set` would silently discard that agreed-upon protocol
+        // state. Treat any such slot as in-flight and defer instead.
+        //
+        // A brand-new slot is in `NominatePrepare` with empty X/Y/Z and
+        // `bN == 0`; the empty-nomination-sets + `bN == 0` signal is the robust
+        // boundary indicator, with the phase check as a secondary guard.
+        let metrics = self.scp_node.get_current_slot_metrics();
+        let scp_slot_active = metrics.num_voted_nominated > 0
+            || metrics.num_accepted_nominated > 0
+            || metrics.num_confirmed_nominated > 0
+            || metrics.bN > 0
+            || metrics.phase != ScpPhase::NominatePrepare;
+        let at_slot_boundary =
+            self.proposed_values.is_empty() && self.externalized.is_none() && !scp_slot_active;
 
         if at_slot_boundary {
             self.apply_quorum_set(new_quorum_set);
@@ -800,8 +821,16 @@ impl fmt::Debug for ConsensusService {
 #[allow(clippy::unwrap_used)]
 mod tests {
     use super::*;
-    use bth_consensus_scp::{QuorumSetMember, ScpNode};
+    use crate::transaction::{
+        ClsagRingInput, RingMember, Transaction, TxOutput, MIN_RING_SIZE, MIN_TX_FEE,
+    };
+    use bth_consensus_scp::{
+        ballot::Ballot,
+        msg::{Msg, NominatePayload, PreparePayload, Topic},
+        QuorumSetMember, ScpNode,
+    };
     use bth_consensus_scp_types::test_utils::test_node_id;
+    use bth_transaction_types::ClusterTagVector;
 
     /// A NodeID for each numeric id; node 1 is "us".
     fn node(n: u32) -> NodeID {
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Thorough review of this determinism-relevant consensus hardening (closes #405). The fix is correct, minimal, and the load-bearing test is genuine.

## Correctness of the augmented boundary guard
The guard now consults `scp_node.get_current_slot_metrics()` and treats the slot as in-flight when `num_voted_nominated > 0 || num_accepted_nominated > 0 || num_confirmed_nominated > 0 || bN > 0 || phase != NominatePrepare`. This closes the original under-defer bug: peer nominate/ballot state accumulated inside `current_slot` is now visible to the boundary decision, so an immediate `set_quorum_set` can no longer silently discard it.

**No over-defer risk.** I traced `advance_slot` (service.rs:767-802): when a reconfiguration is deferred, `advance_slot` applies the stashed `pending_quorum_set` unconditionally (modulo a re-check that it still differs from the active set) — it does NOT re-evaluate `scp_slot_active`. So deferral is bounded to at most one slot; there is no code path that re-defers forever and stalls reconfiguration. This matches the issue's intended "≤1 slot latency."

## SCP contracts genuinely untouched
`git diff --stat HEAD~1` confirms only `botho/src/consensus/service.rs` changed. `set_quorum_set`, `Phase`, and `SlotMetrics` in `bth_consensus_scp` are unmodified — the fix lives entirely in the caller's decision, as required. No protocol-semantics drift.

## The load-bearing test is NOT a tautology
`reconfigure_during_peer_populated_slot_is_deferred` feeds a real peer `NominatePrepare` carrying a structurally valid transfer tx (cached so the SCP validity callback accepts the value). I verified the SCP slot genuinely reaches the peer-populated state **before** the deferral assertion by temporarily forcing the internal sanity assert to fail and capturing the metrics:

```
peer message must leave the SCP slot in an in-flight state: X=0 Y=1 Z=1 bN=1 phase=Prepare
```

This confirms the builder's claim exactly (Y=1, Z=1, bN=1, phase past `NominatePrepare`). Had the value been dropped by the validity callback, the slot would stay empty and the test's own pre-assertion would fail — so the test cannot pass vacuously. It then asserts deferral (`pending_quorum_set` set, active service + SCP quorum sets unchanged, slot metrics survive untouched), then `advance_slot` applies it. `reconfigure_on_idle_slot_applies_immediately` covers the fast path.

## Preserved behavior
Empty-set rejection (`reconfigure_quorum` early return on empty members) and same-membership debounce (effective-set comparison) are unchanged and still pass.

## Test Execution
- `cargo test -p botho --lib consensus` — 54 passed, 0 failed
- `cargo test -p bth-consensus-scp` — 100 passed, 0 failed (2 ignored)
- `cargo +nightly-2025-12-03 fmt --all --check` — clean
- New peer-populated + idle-fast-path tests — both pass

## Fixed during evaluation
The head commit (`a001f40`) failed the pinned-nightly `rustfmt --check` CI job: the multi-line `assert_eq!` tuple in `reconfigure_during_peer_populated_slot_is_deferred` was not formatted to the repo baseline. I applied `cargo +nightly-2025-12-03 fmt` (formatting-only, zero behavior change), committed as `0ea9822`, and pushed. CI is now green on the new head.

**CI on `0ea9822`:** all checks pass (rustfmt, Compile Benchmarks, Fuzz Build Guard, all benchmark suites). `mergeStateStatus` is CLEAN.

PR #410, opened 2026-06-17
fix(consensus): propagate minting txs so multi-node SCP can validate & externalize (#409)

### LINKED ISSUE #409: bug(consensus): real multi-node SCP never externalizes a block (stuck in NominatePrepare) — blocks #396 mine-after-join
## Summary

Real multi-node SCP consensus (two+ `botho run` processes peering over libp2p) **never externalizes a block** — it stays in the `NominatePrepare` phase indefinitely and the chain is stuck at height 0. This blocks the "participate + mine an accepted block" half of #396 (the fresh-joiner-mines test), and contradicts the #396 unblock comment's claim that #400/#401/#405 wired that path end-to-end.

The JOIN + CATCH-UP half works (a solo minter + a follower that syncs 0→N, proven by `join-consensus-live-node.test.ts` / #398). The MINT-an-accepted-block half does not, because no multi-node block is ever produced.

## Reproduction (hermetic, localhost)

Two real nodes, `recommended` quorum (`min_peers = 1`), `BOTHO_SLOT_DURATION_SECS=1`, on loopback:

- **Case 1 — both mint** (2-of-2): peer up, both report "Quorum satisfied: 2-of-2", both `mintingActive`, both stuck at `chainHeight: 0`. SCP runs `Nominate Round(2)`, `Round(3)`, ... forever; never externalizes.
- **Case 2 — A mints, B is validator (`mint=false`)**: same 2-of-2 quorum; only A proposes a single value (`Proposing values to SCP slot=1 count=1`); B participates; still stuck at `chainHeight: 0`, never externalizes.

Both cases verified directly with `cargo build --release --bin botho` at `main` (12fc1ba).

## Evidence

- SCP messages DO flow and decode after #406 (no `deserialize_identifier` errors): with `RUST_LOG=...bth_consensus_scp=debug`, `Received SCP message from network slot=1` and `Broadcast SCP message slot=1` both appear. So #406 is genuinely fixing the wire-decode.
- Dynamic reconfigure (#404/#408) works: `Applying quorum set reconfiguration slot=1 threshold=2 members=2 was_solo=true` / `Exited solo mode: peers present, switching to SCP consensus path`.
- BUT the SCP slot never leaves `phase=NominatePrepare`. No `Slot externalized!` log ever fires on either node; `chainHeight` stays 0.

## Likely root cause(s) to investigate

1. The SCP nomination → ballot → commit → externalize state machine is not being driven to completion in the real run-loop (`botho/src/commands/run.rs` consensus tick + `botho/src/consensus/service.rs`). Even with a single proposer (Case 2) and quorum messages flowing, the protocol never transitions out of `NominatePrepare`.
2. Minting-tx instability: in `propose_pending_values` (`service.rs`), each minting node proposes its own best PoW minting tx, whose `priority` changes every tick as new solutions arrive. In multi-node mode two minters nominate different, ever-changing values, so nomination can never stabilize — but Case 2 shows externalization fails even with one stable-ish proposer, so this is at most a contributing factor, not the whole story.

## Impact

- `#396`'s acceptance criteria ("3rd node mines a block accepted by the whole network; tip advances on all nodes; coinbase reward on-chain") are unreachable until this is fixed.
- This is a Critical pre-public-testnet item: a network of independent minters cannot agree on a single chain.

## Acceptance

- A network of 2+ real `botho run` minters externalizes blocks and advances the chain to height N on a shared tip (identical hashes) over real libp2p.
- Then #396's full mine-after-join test (already written, attached to #396 as a breadcrumb) passes.

---

## Curator Enhancement (2026-06-16)

Grounded against `origin/main` (`botho/src/commands/run.rs`, `botho/src/consensus/service.rs`, `consensus/scp/src/node/node_impl.rs`, `consensus/scp/src/slot.rs`). The reproduction and the two leads in the original body hold up against the code. Below is the code-level root-cause analysis that sharpens them.

### Root cause lead #1 (PRIME SUSPECT): the host run-loop never feeds the node its own SCP messages when peers are present

SCP requires a node to count **its own** votes toward quorum thresholds. In Botho's wiring the consensus service emits each outgoing message as a `ConsensusEvent::BroadcastMessage` and the run-loop broadcasts it — but it only loops the message back into the local node in **solo** mode:

`botho/src/commands/run.rs:1115-1128`
```rust
ConsensusEvent::BroadcastMessage(msg) => {
    if let Err(e) = NetworkDiscovery::broadcast_scp(&mut swarm, &msg) { warn!(...); }
    // For solo mining: loop our own message back to ourselves
    if discovery.peer_count() == 0 {            // <-- ONLY when zero peers
        if let Err(e) = consensus.handle_message(msg) { ... }
    }
}
```

With ≥1 peer, the local node's own nominate/ballot messages are broadcast to peers but never re-applied locally. Whether this stalls depends on how the SCP slot accounts for self: in `consensus/scp/src/slot.rs`, `handle_messages` explicitly **drops messages where `msg.sender_id == self.node_id`** (slot.rs:365-368) and never inserts a self entry into the `M` map (the only `self.M.insert` is at slot.rs:411, inside the peer-message loop). The accepted/confirmed-nominated predicates and the prepare/commit quorum checks must therefore treat the local node's own latest state implicitly. The investigation must determine empirically whether the real run-loop satisfies that contract:

- If SCP's quorum math implicitly includes self (the usual design), then the gap is elsewhere (see lead #1b / #2).
- If the implementation expects the host to re-deliver self-messages (as the solo-mode loopback hints), then the `peer_count() == 0` guard is the bug: it must loop self-messages back **in all modes** so the node sees its own nominate/prepare/commit and can cross quorum thresholds. Removing the guard (or always re-delivering) is the candidate fix to test first.

This is the single most likely cause and the cheapest to test: instrument or unit-test whether a 2-node in-process node pair externalizes when each node's own emitted messages are (a) never re-applied vs (b) always re-applied.

### Root cause lead #1b: SCP timer callbacks ARE driven, but verify they actually advance nomination

The original body's "SCP timers never fire" hypothesis is partially answered by the code: the run-loop **does** drive timeouts. `consensus_tick` fires every 500ms (`run.rs:513`) and calls `consensus.tick()` (`run.rs:1057`), which calls `self.scp_node.process_timeouts()` (`service.rs:444-451`). `process_timeouts` advances nomination rounds and ballot timers (`slot.rs:224-308`): each fired nomination round increments `nominate_round`, recomputes `find_max_priority_peer`, and re-runs `do_nominate_phase` (slot.rs:230-250). The 500ms tick is faster than the 1000ms `base_round_interval` (slot.rs:473), so rounds advance.

So timers fire — which matches the observed symptom: the logs show `Nominate Round(2)`, `Round(3)`, ... advancing forever. The problem is **not** that nomination rounds don't fire; it is that nomination never **converges to a confirmed value** (`Z` never becomes non-empty), so `do_nominate_phase` never sets `B = Ballot::new(1, ...)` (slot.rs:656-663) and the ballot protocol never starts. The fact that rounds keep incrementing while `Z` stays empty is exactly the signature of "the node isn't seeing enough nominate votes to cross the accept/confirm threshold" — which points straight back to lead #1 (self-vote not counted) or to lead #2 (the value set churns so no single value ever accumulates a quorum of nominate votes).

Investigation step: add a debug log (or test assertion) for `Z.len()`, `X.len()`, `Y.len()`, and the `M` map size each round. If `M` only ever contains the *peer* (never self) and `Z` stays empty even though both nodes nominate the same value, the self-accounting contract (lead #1) is violated.

### Root cause lead #2 (contributing, secondary): minting-tx value churn prevents nomination convergence

Confirmed plausible against `service.rs`. Each minting node proposes its current best PoW minting tx; as the miner finds new solutions, `submit_minting_tx` keeps inserting new `ConsensusValue`s with different `tx_hash`/`priority` (`service.rs:385-401`), and `propose_pending_values` re-selects `max_by_key(priority)` each cycle (`service.rs:493-498`). Two minters nominate different, ever-changing values, so no single value can accumulate a confirmed-nominated quorum before it's superseded.

HOWEVER, the original body's Case 2 (single proposer A, B is `mint=false` validator) **still fails to externalize**. A single stable proposer removes the inter-node value disagreement, so value churn alone cannot explain Case 2. This makes lead #2 at most a contributing factor on top of lead #1, **not** the root cause. Fix lead #1 first, then re-test Case 1 to see whether value churn still prevents convergence; if so, stabilize proposal (e.g. freeze the proposed minting value for the duration of a slot / nomination, only re-proposing on slot advance) as a follow-up.

### Why solo + join/catch-up work but multi-node mint does not

- Solo (1-of-1) bypasses SCP entirely: `propose_pending_values` directly externalizes (`service.rs:533-560`) and `advance_slot` manually bumps the slot (`service.rs:673-679`). So solo mining never exercises the nominate→ballot→commit→externalize path.
- Join/catch-up (#398) uses the chain-sync state machine (`run.rs:1151-1192`), not SCP. It applies already-externalized blocks from a peer; it does not produce consensus.
- Multi-node minting is the **only** path that exercises real SCP nomination→externalization, and it is the untested path that this issue targets.

### Recommended investigation/fix order

1. Write the in-process multi-node SCP test first (see Test Plan) so the bug is reproduced deterministically in CI without libp2p.
2. Determine the self-message contract: does the SCP node's quorum math include self, or must the host re-deliver self-messages? This is the decisive question.
3. If host must re-deliver: remove/relax the `discovery.peer_count() == 0` guard at `run.rs:1123` so self-messages are applied in all modes. Re-run the in-process test and Case 2.
4. Re-test Case 1 (two minters). If still stuck, stabilize the proposed value across a nomination window (lead #2).
5. Validate end-to-end with two real `botho run` minters over libp2p; then unblock #396.

### Out of scope / guardrails

- Do NOT change SCP protocol semantics (the nomination/ballot/commit predicates in `slot.rs`). The fix should be in the **host wiring** (`run.rs`/`service.rs`) and/or proposal stability — not in the SCP state machine math, unless a genuine bug in the predicate quorum accounting is proven, in which case scope it as a separate, carefully-reviewed change.
- Determinism must be preserved: the `combine_fn` tie-break by `tx_hash` (`service.rs:229-233`) and value ordering must not change.
- Keep the change small and focused; the run-loop is large but the suspect surface is ~10 lines.

## Acceptance Criteria

- [ ] A network of 2+ real `botho run` minters over real libp2p externalizes blocks and advances the chain to height N (N ≥ 3) on a **shared tip** — all nodes report identical tip hashes at each height. `Slot externalized!` is logged on the proposing node and blocks propagate to peers.
- [ ] Case 1 (both nodes `--mint`, 2-of-2 `recommended` quorum) reaches height ≥ 3 with identical hashes.
- [ ] Case 2 (one `--mint`, one validator with `mint=false`) reaches height ≥ 3 with identical hashes.
- [ ] The SCP slot demonstrably leaves `phase=NominatePrepare`, enters Prepare→Commit→Externalize, and `Z` (confirmed-nominated) becomes non-empty during a normal slot.
- [ ] A regression test (see Test Plan) reproduces the original stall on the pre-fix code and passes on the fixed code, runnable in CI.
- [ ] No SCP protocol-semantics drift: the predicate/quorum math in `consensus/scp/src/slot.rs` is unchanged (or, if a genuine predicate bug is found, the change is isolated, justified, and covered by a dedicated SCP unit test). Existing `cargo test -p bth-consensus-scp` continues to pass.
- [ ] Determinism preserved: identical externalized value sets given identical inputs (combine_fn tie-break unchanged).
- [ ] #396's mine-after-join test (its attached breadcrumb) passes after this fix.

## Affected Files

- `botho/src/commands/run.rs` — Consensus tick wiring (`consensus_tick` at ~:513, `ConsensusEvent` handling at ~:1056-1133). PRIME suspect: the `discovery.peer_count() == 0` guard at ~:1123 that prevents self-message loopback when peers are present.
- `botho/src/consensus/service.rs` — `tick`/`process_timeouts` wiring (~:443-464), `propose_pending_values` (~:481-578) including the minting-value-churn path (lead #2), `handle_message` (~:413-432), `advance_slot` (~:660-682). Add proposal stabilization here if lead #2 is confirmed.
- `consensus/scp/src/slot.rs` — Reference for the nominate→ballot→commit→externalize state machine and self-accounting contract (`handle_messages` self-filter ~:365-368, `do_nominate_phase` ~:603-665, `process_timeouts` ~:224-308). Read-only unless a proven predicate bug requires a scoped change.
- `consensus/scp/src/node/node_impl.rs` — `handle_messages`/`process_timeouts`/`externalize`/`reset_slot_index` (read-only reference; ~:96-313).
- New/updated test file (location TBD by implementer) — in-process multi-node SCP regression test asserting externalization. Candidate: a new test module under `botho/src/consensus/` or a `consensus/scp/` integration test that wires N `ConsensusService`/`Node` instances and shuttles their emitted messages between them, driving `tick()` until externalization.

## Test Plan

- [ ] Automated (regression, CI): In-process multi-node SCP test. Construct 2 (and 3) `ConsensusService` (or raw SCP `Node`) instances sharing a quorum set; submit a stable value to the proposer(s); pump each node's emitted `BroadcastMessage`s into the other node(s) — and into itself if the contract requires it — while calling `tick()` on a loop with a simulated clock or short real sleeps. Assert every node externalizes the same value set within a bounded number of ticks. This test MUST fail on the pre-fix code (stuck, never externalizes) and pass on the fixed code.
- [ ] Automated (SCP unit, no drift): `cargo test -p bth-consensus-scp` passes unchanged, proving SCP protocol semantics were not altered.
- [ ] Manual / e2e (real libp2p): Build release (`cargo build --release --bin botho`), run two `botho run --mint` nodes with `recommended` quorum (`min_peers = 1`), `BOTHO_SLOT_DURATION_SECS=1`, on loopback. With `RUST_LOG=...bth_consensus_scp=debug`, confirm `Slot externalized!` fires on both nodes and `chainHeight` advances past 0 to a shared tip. Repeat for Case 2 (A `--mint`, B validator).
- [ ] Edge cases: (a) value churn — both nodes actively mint and propose changing minting txs (Case 1); confirm convergence still occurs. (b) 3-node network reaches a shared tip. (c) a node restarted mid-run rejoins and resumes externalizing (overlap with #396).

## Dependencies

This issue has no blocking dependencies; the prerequisite wire-decode (#406) and dynamic reconfigure (#404, #408) fixes are already merged on `main`. This issue **blocks** #396 (the mine-after-join half cannot pass until multi-node externalization works).

rjwalters:
Closes #409

## Decisive finding: SCP counts self implicitly — the `peer_count() == 0` guard is NOT the bug

The Curator's prime suspect was the `run.rs:1123` `if discovery.peer_count() == 0` self-message loopback guard. I verified empirically that this is **not** the root cause:

- `QuorumSet::findQuorum` (`consensus/scp/src/quorum_set_ext.rs:95-108`) seeds `nodes_so_far` with the **local node** and counts it toward the threshold **without** requiring a message from self in `M`. So SCP's quorum/blocking-set math includes the local node's own state implicitly.
- The existing in-process multi-node SCP harness (`consensus/scp/tests/mock_network/mod.rs`) asserts `!peers.contains(self)` and delivers each node's emitted messages to **peers only, never to itself** — and the mesh/cyclic/metamesh tests pass. That is the contract: the host must **not** re-deliver self-messages.

Therefore self-votes were already counted, and removing that guard would be wrong. The bug is in host wiring, exactly as the issue's guardrails anticipated.

## Actual root cause

SCP messages carry only a value's `tx_hash`. Botho's `validity_fn` (`service.rs`) validates a `ConsensusValue` by looking up the referenced tx bytes in a per-node cache (`tx_cache`). A minting node populated its **own** cache on propose, but the minting-tx bytes were **never broadcast to peers** (unlike transfer txs). So when node A nominated its minting value:

1. Node B received A's SCP nominate (carrying A's minting `tx_hash`).
2. B's `tx_cache` had no entry → `validity_fn` returned `"Transaction not in cache"`.
3. `Slot::handle_messages` (`slot.rs:399-404`) **dropped A's message** (never inserted into `self.M`).
4. No quorum of nominate votes → `Z` (confirmed-nominated) stayed empty → `do_nominate_phase` never set `B = Ballot::new(1, ...)` → balloting never began → nothing externalized.

This explains both Case 1 (both mint) and Case 2 (one mint, one validator). The mock-network tests never reproduced it because they use a trivial `|_| Ok(())` validity fn.

## The fix (host wiring; SCP predicate math unchanged)

- **Propagate minting txs**: new `botho/minting-txs/1.0.0` gossip topic. The proposing minter broadcasts its minting-tx bytes (`NetworkDiscovery::broadcast_minting_tx`); on receipt a peer validates against its chain state and registers the bytes into the consensus cache (`ConsensusService::register_minting_tx`), so the local SCP node can validate and accept the peer's nominate/ballot messages.
- **Redundant-connection handling** (`discovery.rs`): `ConnectionClosed` now only reports `PeerDisconnected` when `num_established == 0`, so closing one of two concurrent connections to the same peer doesn't collapse the SCP quorum below threshold.
- **Bootstrap reconnect** (`run.rs`): periodically re-dial bootstrap peers when `peer_count == 0`, so a node that loses its only connection at startup can rejoin.
- Declared `bincode` as an scp dev-dependency (used by `src/msg.rs` wire-encoding tests) so the new test target resolves regardless of feature unification.

## Regression test (fails pre-fix, passes post-fix)

`consensus/scp/tests/test_value_cache_validity.rs` — in-process 2-node SCP with a cache-backed `validity_fn` mirroring Botho's contract, shuttling messages **peer-only** (proving self is counted implicitly):

- `stuck_when_peer_value_uncached`: reproduces the stall — when the peer cannot validate the proposed value, no node externalizes.
- `externalizes_when_peer_value_cached`: once the value is registered in every node's cache (the #409 fix), both nodes externalize the same value within a bounded number of ticks, asserting identical externalized sets (shared tip + determinism).

## Determinism / scope

- No SCP protocol-semantics drift: the nominate/ballot/commit predicates and `combine_fn` tie-break are untouched. `cargo test -p bth-consensus-scp` is green (100 unit tests + new regression test).
- The #404/#408 dynamic quorum-reconfigure wiring is preserved.

## Verification status — IMPORTANT

The consensus root cause is fixed and proven by the in-process regression test. While validating end-to-end over **real libp2p** with two `botho run` processes on localhost, I hit a **separate, pre-existing transport-layer bug**: two fresh nodes complete the noise + yamux upgrade but the connection is immediately closed by the remote (`multistream-select: "Stream closed. Confirmation from remote for optimistic protocol negotiation still pending."`), repeating on every (re)dial. This prevents a stable peer connection between two local nodes and therefore blocks empirical multi-node externalization — independent of consensus. It affects all two-node-on-one-host scenarios, not just minting, and predates this change. It warrants its own networking issue; chasing it here would exceed the consensus scope the issue defined.

The redundant-connection and reconnect changes here are correct, low-risk steps toward multi-node liveness, but they do not by themselves resolve that transport-stability bug.

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -215,6 +215,13 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
         config.network.dns_seeds.enabled
     );
 
+    // Keep a copy of the bootstrap multiaddrs so the main loop can re-dial them
+    // if the node ends up with no peers. The initial dial happens during the
+    // pre-loop discovery window; if that connection is lost (e.g. a transient
+    // drop right after startup) there is otherwise no path back to the network
+    // for a node whose only peers are bootstrap nodes (issue #409).
+    let reconnect_bootstrap_peers = bootstrap_peers.clone();
+
     // Start network discovery
     let mut discovery =
         NetworkDiscovery::new(config.network.gossip_port(network_type), bootstrap_peers);
@@ -521,6 +528,10 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     let mut sync_tick = tokio::time::interval(Duration::from_secs(2));
     let mut minting_check_interval = tokio::time::interval(Duration::from_millis(100));
     let mut faucet_balance_interval = tokio::time::interval(Duration::from_secs(10));
+    // Periodically re-dial bootstrap peers when we have no connections, so a
+    // node that lost its only connection right after startup can rejoin
+    // (issue #409). Cheap and a no-op once peers are connected.
+    let mut reconnect_interval = tokio::time::interval(Duration::from_secs(5));
 
     // Track pending compact blocks awaiting missing transactions
     // Key: block hash, Value: (compact block, missing indices)
@@ -598,6 +609,36 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                 if let Ok(mempool) = node.shared_mempool().read() {
                                     ws_broadcaster.mempool_update(mempool.len(), mempool.total_fees());
                                     metrics_updater.set_mempool_size(mempool.len());
+                                }
+                            }
+                        }
+                        NetworkEvent::NewMintingTx(minting_tx) => {
+                            // A peer proposed this minting tx for consensus.
+                            // Validate it against our chain state, then register
+                            // it in the consensus tx cache so the local SCP node
+                            // can validate (and therefore accept) the peer's
+                            // nominate/ballot messages referencing it (issue #409).
+                            let tx_hash = minting_tx.hash();
+                            debug!("Received minting tx {} from network", hex::encode(&tx_hash[0..8]));
+
+                            let chain_state = node
+                                .shared_ledger()
+                                .read()
+                                .ok()
+                                .and_then(|ledger| ledger.get_chain_state().ok());
+
+                            if let Some(chain_state) = chain_state {
+                                let temp_state = Arc::new(RwLock::new(chain_state));
+                                let validator = TransactionValidator::new(temp_state);
+                                match validator.validate_minting_tx(&minting_tx) {
+                                    Ok(()) => {
+                                        let tx_bytes = bincode::serialize(&minting_tx)
+                                            .expect("Failed to serialize minting tx");
+                                        consensus.register_minting_tx(tx_hash, tx_bytes);
+                                    }
+                                    Err(e) => {
+                                        debug!(error = %e, "Rejecting invalid peer minting tx");
+                                    }
                                 }
                             }
                         }
@@ -1175,6 +1216,23 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                 );
             }
 
+            // Reconnect tick: if we have no peers, re-dial configured bootstrap
+            // peers. The initial dial happens before the main loop; a node that
+            // lost that connection during startup would otherwise be stranded
+            // with no way back to the network (issue #409).
+            _ = reconnect_interval.tick() => {
+                if discovery.peer_count() == 0 && !reconnect_bootstrap_peers.is_empty() {
+                    for peer_addr in &reconnect_bootstrap_peers {
+                        if let Ok(addr) = peer_addr.parse::<libp2p::Multiaddr>() {
+                            debug!("Reconnect: re-dialing bootstrap peer {}", addr);
+                            if let Err(e) = swarm.dial(addr) {
+                                debug!("Reconnect dial failed: {}", e);
+                            }
+                        }
+                    }
+                }
+            }
+
             // Chain sync (IBD / catch-up) tick. Drives the sync state machine,
             // emitting status/block requests as needed so a node behind the
             // network tip backfills the missing blocks from a peer.
@@ -1296,6 +1354,15 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                 let tx_hash = minting_tx.hash();
 
                 consensus.submit_minting_tx(tx_hash, minted_tx.pow_priority, tx_bytes);
+
+                // Broadcast the minting tx so peers can validate the
+                // corresponding SCP consensus value when its hash appears in a
+                // nominate/ballot message. Without this, peers reject our SCP
+                // messages ("Transaction not in cache") and multi-node
+                // nomination never reaches quorum (issue #409).
+                if let Err(e) = NetworkDiscovery::broadcast_minting_tx(&mut swarm, minting_tx) {
+                    debug!("Failed to broadcast minting tx: {}", e);
+                }
 
                 // Also check for pending transfer transactions in mempool
                 // and submit them to consensus
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -507,6 +507,24 @@ impl ConsensusService {
         info!(?value, "Minting transaction submitted for consensus");
     }
 
+    /// Register a minting transaction received from a peer into the validation
+    /// cache, without adding it to our own pending/proposed set.
+    ///
+    /// SCP messages only carry a value's `tx_hash`. When a peer nominates its
+    /// minting tx, every node in the quorum must be able to validate that value
+    /// or the SCP slot rejects the peer's message (validity_fn: "Transaction
+    /// not in cache") and nomination never reaches quorum. Feeding the raw tx
+    /// bytes into the cache lets the local SCP node accept the peer's nominate
+    /// votes so balloting can begin. See issue #409.
+    pub fn register_minting_tx(&mut self, tx_hash: [u8; 32], tx_data: Vec<u8>) {
+        if let Ok(mut state) = self.shared_state.write() {
+            state.tx_cache.entry(tx_hash).or_insert(TxCacheEntry {
+                data: tx_data,
+                is_minting_tx: true,
+            });
+        }
+    }
+
     /// Handle an incoming SCP message from gossip
     #[instrument(
         name = "consensus.handle_message",
--- a/botho/src/network/discovery.rs
+++ b/botho/src/network/discovery.rs
@@ -40,7 +40,7 @@ use bth_transaction_types::{
 };
 
 use crate::{
-    block::Block,
+    block::{Block, MintingTx},
     consensus::ScpMessage,
     network::{
         compact_block::{BlockTxn, CompactBlock, GetBlockTxn},
@@ -77,6 +77,18 @@ const TRANSACTIONS_TOPIC: &str = "botho/transactions/1.0.0";
 
 /// Topic for SCP consensus messages
 const SCP_TOPIC: &str = "botho/scp/1.0.0";
+
+/// Topic for minting-transaction announcements.
+///
+/// Minting transactions referenced by an SCP `ConsensusValue` are not part of
+/// the mempool transaction flow (`TRANSACTIONS_TOPIC` carries user
+/// `Transaction`s), so a node that only learns a peer's minting `tx_hash` from
+/// an SCP nominate/ballot message has no way to validate it. Without the raw
+/// bytes the SCP slot rejects the peer's message (validity_fn: "Transaction not
+/// in cache") and nomination can never reach quorum. This topic propagates the
+/// minting-tx bytes so every consensus participant can validate a proposed
+/// minting value. See issue #409.
+const MINTING_TXS_TOPIC: &str = "botho/minting-txs/1.0.0";
 
 /// Topic for compact block announcements
 const COMPACT_BLOCKS_TOPIC: &str = "botho/compact-blocks/1.0.0";
@@ -253,6 +265,10 @@ pub enum NetworkEvent {
     NewBlock(Block),
     /// A new transaction was received from a peer
     NewTransaction(Transaction),
+    /// A minting transaction proposed for consensus was received from a peer.
+    /// Registered into the consensus tx cache so the local SCP node can
+    /// validate the corresponding `ConsensusValue` (see issue #409).
+    NewMintingTx(MintingTx),
     /// An SCP consensus message was received
     ScpMessage(ScpMessage),
     /// A compact block was received (for bandwidth-efficient relay)
@@ -532,6 +548,13 @@ impl NetworkDiscovery {
         let scp_topic = IdentTopic::new(SCP_TOPIC);
         swarm.behaviour_mut().gossipsub.subscribe(&scp_topic)?;
 
+        // Subscribe to minting-transactions topic (issue #409)
+        let minting_txs_topic = IdentTopic::new(MINTING_TXS_TOPIC);
+        swarm
+            .behaviour_mut()
+            .gossipsub
+            .subscribe(&minting_txs_topic)?;
+
         // Subscribe to compact blocks topic
         let compact_blocks_topic = IdentTopic::new(COMPACT_BLOCKS_TOPIC);
         swarm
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved!**

Rigorous review of this pre-mainnet consensus + networking fix. Verdict: **APPROVE**.

## (a) Fix correctness — CONFIRMED

The root-cause chain holds end-to-end:
- `validity_fn` (`service.rs:196-215`) validates a `ConsensusValue` by `state.tx_cache.get(&value.tx_hash)` → then `validate_from_bytes(&entry.data, entry.is_minting_tx)`. A missing entry yields `"Transaction not in cache"`, which `Slot::handle_messages` treats as invalid and drops the peer's message.
- `register_minting_tx` (`service.rs:510-524`) writes that exact `tx_cache` map keyed by `tx_hash` with `is_minting_tx: true` and the serialized bytes — precisely what `validity_fn` reads.
- The new `botho/minting-txs/1.0.0` gossip topic broadcasts minting-tx bytes on propose (`run.rs`), and peers validate + register on receipt. So a peer that has received the gossiped tx will validate the subsequent SCP nominate/ballot referencing it, the quorum forms, balloting begins, externalize proceeds. Correct.

## (b) #404/#408 intactness — CONFIRMED, no revert

`git diff origin/main...HEAD -- service.rs` = **18 insertions(+), 0 deletions**, purely the additive `register_minting_tx`. The #404/#408 dynamic-quorum-reconfigure logic is fully present and **unmodified**: `reconfigure_quorum` (320), `apply_quorum_set` (378), `pending_quorum_set` (169/285/370/813), and the #408 `get_current_slot_metrics` in-flight boundary guard (352-359). run.rs and discovery.rs have no consensus-logic deletions (discovery.rs's only `-` lines are a benign import-line merge and the `ConnectionClosed` pattern expansion). The transient stale-file copy the builder flagged did NOT leak into the final diff.

## (c) Gossip-path security — SOUND

- Size cap **before** deserialize (`MAX_TRANSACTION_SIZE`) — DoS guard.
- **Validate-before-register**: `validate_minting_tx` runs against current chain state before `register_minting_tx`; only valid minting txs enter the cache.
- `tx_cache.entry(tx_hash).or_insert(...)` — a peer **cannot overwrite** an existing entry (no cache poisoning of an already-present hash).
- **Defense in depth**: even a registered entry is re-deserialized + re-validated by `validity_fn` before SCP accepts the value.
- No panic on malformed bytes: both `validate_from_bytes` deserialize paths use `bincode::deserialize(...).map_err(...)`.
- Cache is bounded by slot-advance eviction (`advance_slot` removes externalized hashes). Flooding requires forging distinct valid (PoW-bound) minting txs — same surface as the existing mempool/minting cache paths, acceptable pre-mainnet.

## (d) Empirical-verification gap — ACCEPTABLE for merge

The in-process regression test is genuine and passes both directions:
- `stuck_when_peer_value_uncached` reproduces the stall (no node externalizes when the peer can't validate).
- `externalizes_when_peer_value_cached` shows both nodes externalize the **same** value once cached, using **peer-only** delivery (proving SCP counts self implicitly — the `peer_count()==0` guard was correctly ruled out as a red herring).

The real-libp2p e2e path is blocked by a **separate, pre-existing transport bug** (multistream-select "optimistic protocol negotiation still pending"), outside this PR's consensus scope. I filed it as **#411**. A correct, well-tested consensus fix that can't be e2e-verified due to an unrelated infra bug is mergeable; the limitation is clearly acknowledged in the PR body.

## Test Execution

- `cargo test -p bth-consensus-scp --features test_utils` — ✅ **100 unit + 2 new regression tests pass**, plus mesh/cyclic/metamesh suites green. No SCP protocol-semantics drift.
- CI on head `385fb95`: ✅ `cargo build --workspace (Linux)`, `rustfmt`, all benchmark + fuzz-build-guard checks **pass**. Merge state **CLEAN**.

Excellent diagnosis and a clean, additive, well-documented fix. Approving and relabeling to `loom:pr`.

Issue #414, opened 2026-06-17
Multi-node SCP receives peer messages but never externalizes: nomination/ballot stalls in NominatePrepare (downstream of #413)


rjwalters:
## Summary

With the #413 gossip rate-limiter fix in place (PR for #413), two real \`botho run\` minters on loopback now **exchange SCP messages** and **leave solo mode**, but SCP **never externalizes a block** — both nodes stay at height 0/1 with the slot stuck in \`NominatePrepare\`. This is the *next* root cause after #413, in the SCP nomination/ballot path itself, not the gossip layer.

## What #413 fixed (confirmed)

Two minters (\`recommended\`, \`min_peers=1\`, \`BOTHO_SLOT_DURATION_SECS=1\`, throwaway dirs, loopback bootstrap):

- \`Peer connected\` + \`Applying quorum set reconfiguration ... was_solo=true\` + \`Exited solo mode: peers present, switching to SCP consensus path\` fire on both nodes.
- \`flagged for disconnection\`: **0** (was 98x). \`Rate limited message from peer\`: **0**.
- Inbound peer SCP messages are now delivered to the consensus loop and handled:
  \`consensus.handle_message{slot_index=N peer_id=1100... msg_size=286}: scp.handle_messages{phase=NominatePrepare msg_count=1}\` fired **24x on node A, 14x on node B**.
- \`consensus.propose_values{solo_mode=false}: Proposing values to SCP\` fires repeatedly (correct non-solo path).

So the gossip rate limiter is no longer the blocker — peer SCP messages reach SCP.

## The remaining bug

Despite both nodes proposing values and handling each other's SCP messages:

- The slot never leaves \`phase=NominatePrepare\`. No \`Slot externalized!\`, no \`Solo mode: directly externalizing\` (correctly, since solo_mode=false), height stays 0->1.
- Each node keeps re-proposing its own minting tx every tick; \`pending_count\` climbs (11 -> 21 -> 126 ...) because nothing is ever externalized to drain it.
- The two nodes nominate **different** minting txs (each its own coinbase, different \`tx_hash\` + \`priority\`), and nomination never converges to a single accepted value, so the ballot protocol never starts PREPARE in earnest.

[...]

PR #415, opened 2026-06-17
fix(gossip): slot-rate-aware rate limits + violation decay so SCP/minting messages aren't dropped (#413)

### LINKED ISSUE #413: Multi-node SCP never externalizes a shared block: gossipsub per-peer rate limiter drops consensus/minting messages, nodes stay solo and fork
## Summary

With **all prior blockers merged** (#406 SCP wire-decode, #410 minting-tx propagation, #404/#408 dynamic quorum reconfigure, #412 sync-codec transport), two real `botho run` minters peering over loopback **still never reach consensus on a shared chain**. Each node externalizes its own minting tx every slot in **solo mode** and builds a **divergent fork**. This blocks #396 (the "fresh node joins, participates, mines an accepted block" flavor-A test) and any real multi-node operation.

This is a **different** failure than #409 claimed fixed. #410 added minting-tx propagation, but that is moot because the consensus messages themselves are being dropped at the gossip layer before they ever reach the consensus loop.

## Root cause (precise)

The gossipsub inbound message handler runs **every** message — including consensus (SCP) and transaction (minting-tx) topics — through a per-peer rate limiter, and **silently drops** any message that exceeds the limit:

`botho/src/network/discovery.rs:890-918`
```rust
if let Some(peer) = message.source {
    let msg_type = Self::topic_to_message_type(topic);
    match self.rate_limiter.record_message_typed(&peer, msg_type) {
        RateLimitResult::Allowed => { /* continue */ }
        RateLimitResult::RateLimited { .. } => { return None; }   // dropped
        RateLimitResult::Disconnect => {
            warn!(%peer, "Peer exceeded rate limit threshold, flagged for disconnection");
            return None;                                          // dropped
        }
    }
}
```

The default per-type limits (`gossip/src/config.rs:144-157`) are far below the message rate two active minters produce:

| message type        | default limit      |
|---------------------|--------------------|
| `consensus_per_minute` | **50/min** (< 1/s) |
| `transactions_per_minute` | **100/min** |
| `blocks_per_minute` | 10/min |

and `PeerRateLimitConfig` defaults (`gossip/src/config.rs:159-166`): `max_messages_per_second: 10`, `burst_limit: 50`, `disconnect_threshold: 3`.

A single SCP slot involves multiple nominate/ballot broadcasts, and minting txs are gossiped on the `transactions` topic. With `BOTHO_SLOT_DURATION_SECS=1` two minters easily exceed 50 consensus msgs/min and 100 tx/min within seconds, trip `disconnect_threshold = 3`, and from then on **every** consensus/minting message from the peer is dropped. Consensus-critical traffic is treated identically to flood traffic and discarded.

## Empirical repro (release binary, localhost)

Two minters, `recommended` + `min_peers = 1`, `BOTHO_SLOT_DURATION_SECS=1`, `RUST_LOG=info,botho::consensus=debug`. Both report `Quorum satisfied: 2-of-2`, `peerCount: 1`, `scpPeerCount: 1`, and climb to height ~35 — but on **two different chains**:

```
h=10 DIVERGE A=f5f073aa... B=29f977a0...
h=20 DIVERGE A=89571716... B=8708f400...
h=30 DIVERGE A=ecb0b42f... B=0e617aa2...
```

Per-node log evidence (node A):
- `Slot externalized!` fires 88x — but always under `solo_mode=true`:
  `consensus.propose_values{solo_mode=true}: Solo mode: directly externalizing values`
  followed by `Advanced to next slot (solo mode)`.
- `solo_mode=false` never appears; `Exited solo mode` never logged.
- **`Received SCP message from network`: 0** — zero inbound SCP messages reach the consensus loop.
- Peer minting txs registered: 0.
- **`Peer connected` / `Peer disconnected` run-loop events: 0 each** — so `reconfigure_quorum()` (the thing that lifts solo mode) is never invoked.
- `Peer exceeded rate limit threshold, flagged for disconnection`: **98x** for the single peer.

So although the discovery layer's peer counter shows 1 peer (status RPC `peerCount: 1`), the consensus run loop never gets a `PeerDiscovered` event and never gets a single SCP message — both are eaten by the rate limiter / connection churn. `is_solo_mode()` (`botho/src/consensus/service.rs:592`: `threshold == 1 && members.len() == 1`) therefore stays `true` forever.

## Why this matters

This is the active blocker for the whole "run your own node and participate in consensus" story (#396) and for any multi-minter testnet. Two honest minters cannot agree on a single chain on real processes today.

## Suggested fix directions (for the implementer to evaluate)

1. **Exempt consensus-critical topics from the drop path.** SCP (`Consensus`) and minting-tx (`Transaction`) messages from already-connected, quorum-member peers should not be silently dropped by the anti-flood limiter. At minimum, raise `consensus_per_minute` / `transactions_per_minute` dramatically (the SCP cadence is bounded by slot duration × #peers, not by a fixed small constant) and/or scale them with `BOTHO_SLOT_DURATION_SECS`.
2. **Do not silently `return None`.** Dropping a consensus message with only a WARN means the protocol stalls invisibly. Consider back-pressure or counting only genuinely-excess traffic.
3. **Decouple the `disconnect_threshold` from consensus topics** so a busy honest minter is never flagged for disconnection on the basis of normal SCP/minting volume.
4. Separately investigate why `PeerDiscovered` never reaches the consensus run loop in this 2-node scenario even though the discovery peer counter increments — the quorum reconfigure that lifts solo mode depends on that event firing.

## Acceptance

Two local `botho run` minters (`recommended`, `min_peers=1`, `BOTHO_SLOT_DURATION_SECS=1`) must:
- exchange SCP messages (`Received SCP message from network` > 0 on both),
- leave solo mode (`Exited solo mode` / `solo_mode=false`),
- converge on **identical block hashes** at multiple heights, and
- have `mine-after-join-live-node.test.ts` (the committed #396 breadcrumb on `feature/issue-396`) pass end-to-end.

## References
- Blocks #396 (flavor-A mine-after-join test; test committed on branch `feature/issue-396`).
- Supersedes the "fixed" claim in #409/#410 — minting-tx propagation is moot when SCP messages are dropped at the gossip layer.
- Files: `botho/src/network/discovery.rs:880-918`, `gossip/src/config.rs:120-166`, `botho/src/consensus/service.rs:320-395,592`.

rjwalters:
Closes #413

## Problem

The gossipsub per-peer rate limiter silently dropped consensus (SCP) and minting (Transaction) messages and *permanently* flagged honest peers for disconnection. As a result, multi-node nodes never received peer SCP messages, never left solo mode, and forked (observed `flagged for disconnection: 98x` for a single peer, `Received SCP message from network: 0`).

Three coupled root causes (per the Curator's grounding), all addressed via the recommended **Option 3 / hybrid** — generous slot-rate-aware ceiling for anti-flood + exempt consensus/minting from auto-disconnect + violation decay, while keeping a real flood ceiling:

### 1. Fixed per-type caps were far below honest cadence
`MessageTypeLimits::for_slot_duration(slot_secs, peers)` and `PeerRateLimitConfig::for_slot_duration(...)` derive the consensus/transaction caps from `msgs_per_slot * slots_per_min * peers * headroom`. A faster slot or larger quorum raises the ceiling instead of dropping honest traffic. `run.rs` plumbs the **effective** slot duration (min block time under dynamic timing, or `BOTHO_SLOT_DURATION_SECS` when fixed) and the gossip connection ceiling into the limiter.

### 2. Global gates tripped before per-type caps
The global `max_messages_per_second` (was 10) and `burst_limit` (was 50/5s) gates are checked first, so raising only per-type caps would still drop messages. They are now derived from the same slot-rate model and sized above a **full slot's instantaneous message cluster across peers** (not just the time-averaged rate), with headroom. Honest bursts pass; a true flood still trips.

### 3. Violations were sticky (no live cooldown)
`reset_violations()` was never called on a live path, so once a peer hit `disconnect_threshold` it stayed flagged forever. Added `PeerRateState::decay_violations()` invoked on every recorded message (decays one violation per 30s of good behavior), and made `GossipMessageType::Consensus`/`Transaction` **never** push a peer to the disconnect path (`is_disconnect_eligible()`), so an honest-but-busy minter is throttled at most, never banned. Non-consensus types still disconnect a genuine abuser.

## Anti-flood preserved
A sustained abusive rate is still `RateLimited` for consensus/minting (over-cap messages dropped, peer not banned), and still escalates to `Disconnect` for disconnect-eligible types (e.g. Block). Covered by regression tests.

## Tests (`cargo test -p bth-gossip` green, 69 pass)
- `test_honest_consensus_rate_allowed_at_fast_slots` / `_at_default_slots` — honest rates pass per-type AND global gates.
- `test_consensus_never_auto_disconnects` — a flood on consensus is throttled but never disconnects/flags.
- `test_flood_on_disconnect_eligible_type_still_disconnects` — a flood on Block still flags + disconnects (anti-flood intact).
- `test_violation_decay_clears_transient_burst` — transient burst decays back to 0.
- `test_slot_aware_global_gates_scale` — faster slots raise the global gates above the old fixed 10/s, 50/5s.

## Empirical two-minter verification (release binary, loopback, `recommended`/`min_peers=1`, `BOTHO_SLOT_DURATION_SECS=1`)
- `flagged for disconnection`: **0** (was 98x). `Rate limited message from peer`: **0**.
- `Exited solo mode` fires on both; `consensus.propose_values{solo_mode=false}` is the active path.
- Inbound peer SCP messages now reach SCP: `consensus.handle_message{peer_id=1100... }: scp.handle_messages{phase=NominatePrepare}` fired **24x on A, 14x on B** (was 0 — the rate limiter no longer drops them).

The gossip rate limiter is no longer the blocker. A **downstream** SCP nomination/ballot convergence stall remains (the two nodes receive each other's SCP messages and exit solo mode but still don't externalize a shared block — they nominate distinct per-slot minting txs and never converge). Per the task instructions I did not force it; filed as **#414** with exact logs.

## Files
- `gossip/src/config.rs` — slot-rate-aware `MessageTypeLimits`/`PeerRateLimitConfig` constructors + global-gate derivation.
- `gossip/src/rate_limit.rs` — `is_disconnect_eligible()`, `decay_violations()` live cooldown, consensus/minting exempt from disconnect, regression tests.
- `botho/src/commands/run.rs` — plumb effective slot duration + peer ceiling into the limiter.

## Determinism / security notes
- No SCP wire format or protocol changes; only the gossip-layer admission control. SCP message bytes are unchanged.
- Anti-flood/DoS protection is **raised and refined, not removed**: the limiter still throttles consensus/minting traffic and still disconnects genuine abusers on non-consensus topics.
- Limits are derived deterministically from slot duration + a fixed peer ceiling (no wall-clock/env nondeterminism in the values themselves); violation decay uses monotonic `Instant`.

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -4,6 +4,7 @@ use anyhow::{Context, Result};
 use bth_common::{NodeID, ResponderId};
 use bth_consensus_scp::QuorumSet;
 use bth_crypto_keys::Ed25519Public;
+use bth_gossip::{GossipConfig, PeerRateLimitConfig};
 use futures::StreamExt;
 use std::{
     net::SocketAddr,
@@ -222,9 +223,31 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // for a node whose only peers are bootstrap nodes (issue #409).
     let reconnect_bootstrap_peers = bootstrap_peers.clone();
 
-    // Start network discovery
-    let mut discovery =
-        NetworkDiscovery::new(config.network.gossip_port(network_type), bootstrap_peers);
+    // Start network discovery.
+    //
+    // Size the gossipsub per-peer rate limiter from the *effective* slot
+    // cadence and the peer ceiling so honest multi-node SCP/minting traffic is
+    // never silently dropped (issue #413). The fastest cadence the node may run
+    // at determines the highest honest message rate: under dynamic timing the
+    // slot duration can drop to `MIN_BLOCK_TIME_SECS`, and under the
+    // `BOTHO_SLOT_DURATION_SECS` test override it is whatever was configured.
+    // We use the smaller of the two as the rate-limit basis (a smaller slot ->
+    // higher rate -> higher cap), and the gossip layer's connection ceiling as
+    // the peer bound (each connected peer may broadcast SCP/minting traffic).
+    let consensus_cfg = consensus_config_from_env();
+    let effective_slot_secs = if consensus_cfg.dynamic_timing {
+        ConsensusConfig::MIN_BLOCK_TIME_SECS
+    } else {
+        consensus_cfg.slot_duration.as_secs().max(1)
+    };
+    let rate_limit_peers = GossipConfig::default().max_connections.max(1) as u32;
+    let rate_limit_config =
+        PeerRateLimitConfig::for_slot_duration(effective_slot_secs, rate_limit_peers);
+    let mut discovery = NetworkDiscovery::with_rate_limit_config(
+        config.network.gossip_port(network_type),
+        bootstrap_peers,
+        rate_limit_config,
+    );
 
     let mut swarm = discovery.start().await?;
 
--- a/gossip/src/config.rs
+++ b/gossip/src/config.rs
@@ -143,28 +143,139 @@ pub struct MessageTypeLimits {
 
 impl Default for MessageTypeLimits {
     fn default() -> Self {
+        // These caps must comfortably exceed the message rate that honest
+        // multi-node SCP consensus produces, while remaining far below a
+        // genuine flood. A single SCP slot is not one message: nomination
+        // rounds plus the ballot protocol (PREPARE / CONFIRM / EXTERNALIZE),
+        // each potentially rebroadcast on timeout, plus a per-slot minting tx.
+        // A conservative bound is ~10-30 consensus messages per slot per peer
+        // during active rounds. At the *default* slot duration (20s) that is
+        // tiny, but the same binary is run with `BOTHO_SLOT_DURATION_SECS=1`
+        // in tests, where it is ~600-1800 consensus msgs/min/peer. The old
+        // fixed caps (50/min consensus, 100/min tx) silently dropped honest
+        // SCP/minting traffic, kept nodes in solo mode, and caused forks
+        // (issue #413). We size the defaults for the fast-slot, multi-peer
+        // case and rely on `for_slot_duration` to scale further when the
+        // effective slot duration / quorum size is known.
+        Self::for_slot_duration(DEFAULT_SLOT_DURATION_SECS, DEFAULT_QUORUM_PEERS)
+    }
+}
+
+/// Default slot duration (seconds) assumed when sizing rate limits without an
+/// explicit value. Matches `ConsensusConfig::default().slot_duration`.
+const DEFAULT_SLOT_DURATION_SECS: u64 = 20;
+
+/// Default quorum size assumed when sizing rate limits without an explicit
+/// value. Small testnets run 2-of-2; we leave generous headroom.
+const DEFAULT_QUORUM_PEERS: u32 = 8;
+
+/// Estimated number of consensus (SCP) messages a single peer broadcasts per
+/// slot during active nomination + ballot rounds (with rebroadcast headroom).
+const CONSENSUS_MSGS_PER_SLOT: u32 = 30;
+
+/// Estimated number of transaction (incl. minting tx) messages a single peer
+/// broadcasts per slot.
+const TX_MSGS_PER_SLOT: u32 = 8;
+
+/// Multiplicative safety headroom applied to derived rate limits so transient
+/// bursts (e.g. several rebroadcasts colliding) never trip honest traffic.
+const RATE_LIMIT_HEADROOM: u32 = 4;
+
+impl MessageTypeLimits {
+    /// Derive per-type per-minute caps from the effective slot duration and the
+    /// quorum size. The consensus/transaction caps scale with `peers` and with
+    /// `60 / slot_secs`, so a faster slot or a larger quorum raises the ceiling
+    /// instead of silently dropping honest consensus traffic.
+    ///
+    /// The result is still a *flood ceiling*: a genuinely abusive peer sending
+    /// orders of magnitude more than the honest cadence is still caught.
+    pub fn for_slot_duration(slot_secs: u64, peers: u32) -> Self {
+        let slot_secs = slot_secs.max(1);
+        // Slots per minute at this cadence: a 1s slot yields 60, a 20s slot
+        // yields 3. Floored at 1 so very long slots still get a sane cap.
+        let slots_per_min = (60 / slot_secs).max(1) as u32;
+        let peers = peers.max(1);
+
+        // consensus_per_minute = msgs_per_slot * slots_per_min * peers * headroom
+        let consensus_per_minute = CONSENSUS_MSGS_PER_SLOT
+            .saturating_mul(slots_per_min)
+            .saturating_mul(peers)
+            .saturating_mul(RATE_LIMIT_HEADROOM);
+        let transactions_per_minute = TX_MSGS_PER_SLOT
+            .saturating_mul(slots_per_min)
+            .saturating_mul(peers)
+            .saturating_mul(RATE_LIMIT_HEADROOM);
+
         Self {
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Rigorous review of the gossipsub rate-limiter fix for #413. This correctly and safely fixes the message-dropping bug without disabling anti-flood protection. Verdict: **APPROVE**.

## Verification
- `cargo test -p bth-gossip` in the PR worktree (head `89591f7`): **69 passed, 0 failed**, including all 7 new/updated tests (`test_consensus_never_auto_disconnects`, `test_flood_on_disconnect_eligible_type_still_disconnects`, `test_violation_decay_clears_transient_burst`, `test_honest_consensus_rate_allowed_at_fast_slots`/`_at_default_slots`, `test_slot_aware_global_gates_scale`, `test_message_type_rate_limits`).
- `cargo clippy -p bth-gossip --all-targets` and `cargo build -p botho`: clean (only pre-existing unrelated warnings).
- CI on head commit: **all checks pass**, merge state **CLEAN**.

## Assessment

**(a) Limit derivation + global-gate fix — correct.** Caps derive from `msgs_per_slot × slots_per_min × peers × headroom`. Critically, the core #413 failure (global `max_messages_per_second`/`burst_limit` trip *before* per-type caps, so raising only per-type caps wouldn't help) is genuinely addressed: the global per-second gate is `max(per_slot_instantaneous_cluster, averaged_per_second, 50)`, sized above the worst-case cross-peer instantaneous cluster rather than the time-averaged rate. At 1s slots this scales the gate into the thousands/s vs. the old fixed 10/s; at default 20s slots it is ~1216/s vs. 10. All three gates (burst → per-second → per-type, in `PeerRateState::record_message_typed` order) are raised together. `run.rs` plumbs the effective slot duration (fixed override secs, or `MIN_BLOCK_TIME_SECS=3` under dynamic timing) and the `max_connections=50` peer ceiling in.

**(b) Anti-flood DoS protection genuinely preserved — confirmed.** The disconnect-exemption (`is_disconnect_eligible()` false for Consensus/Transaction) removes only auto-*disconnect*, not throttling: over-cap consensus/minting messages are still `RateLimited` and dropped on the live ingress path (`discovery.rs:848` → `return None`, never deserialized/processed). No unbounded-work/memory vector: rejected messages return before being pushed into the per-type/global buffers, and the buffers retain only within their window/minute, so per-peer state is bounded by the caps regardless of flood volume. A genuine abuser on a disconnect-eligible topic (e.g. Block) still flags + disconnects. The exemption is a sound tradeoff — banning an honest-but-busy minter (the #413 fork cause) is strictly worse than tolerating a connected-but-throttled peer whose excess traffic is dropped.

**(c) Violation-decay soundness — sound.** Uses monotonic `Instant`; one violation decays per 30s of continuous good behavior. A fresh violation sets `last_warning = now`, so an abuser cannot bank decay credit while actively violating; alternating flood/quiet forces ~30s of silence per violation, which throttles effective abuse far below the threshold. For consensus/minting the decay is moot (never banned), so it cannot be gamed there at all.

**(d) Scope / determinism — clean.** Changes are confined to gossip-layer admission control (`config.rs`, `rate_limit.rs`) plus the `run.rs` plumbing. No SCP wire-format or protocol changes; SCP message bytes are unchanged. Limits are derived deterministically from slot duration + a fixed peer ceiling; decay uses monotonic time. No wall-clock/env nondeterminism in the computed values.

PR correctly scopes itself to the rate-limiter bug; the downstream SCP nomination-convergence stall (#414) is appropriately out of scope and not required here. PR description uses `Closes #413`.

Excellent, well-tested, security-conscious fix.

PR #416, opened 2026-06-17
fix(consensus): make multi-node SCP externalize on a shared tip (#414)

### LINKED ISSUE #414: Multi-node SCP receives peer messages but never externalizes: nomination/ballot stalls in NominatePrepare (downstream of #413)

rjwalters:
## Summary

Two real `botho run` minters on loopback exchanged SCP messages and left solo mode but **never externalized a block** — both stuck at height 0 with the slot spinning in `NominatePrepare`. This is the headline blocker for real multi-node consensus.

Empirical investigation (two real minters with `RUST_LOG=...=debug`) showed the defect was **NOT in the SCP nomination/ballot state machine** (the curator's leading hypothesis). The SCP slot logic converges correctly — the deterministic combiner and federated voting are sound, and all 100 existing SCP tests pass. The blocker was **three bugs in the botho node layer**:

### 1. `peer_id_to_node_id()` produced colliding public keys (`botho/src/commands/run.rs`)
`NodeID` `Eq`/`Hash`/`Ord` compare **only the Ed25519 public key**. The old mapping copied the first 32 bytes of `peer_id.to_bytes()`, whose leading bytes are a fixed multihash prefix shared by every Ed25519 peer, so distinct peers collapsed onto the **same (often invalid) public key**. The quorum set then had duplicate members → `QuorumSet::is_valid()` was false → `Msg::validate` rejected **every** outgoing SCP message with `Invalid quorum set`. No SCP messages were ever broadcast, so nothing externalized.

**Fix:** hash the full PeerId with a domain separator into a 32-byte seed and derive a real Ed25519 keypair — always a valid curve point, deterministic, and injective across peers.

### 2. `check_externalized()` off-by-one (`botho/src/consensus/service.rs`)
On externalize the SCP `Node` auto-advances `current_slot` to `N+1` and files slot `N` in `externalized_slots`. The code queried `get_externalized_values(current_slot_index())` = slot `N+1`, which never has values, so `Slot externalized!` never fired and the chain stayed at height 0 even though SCP reached the Externalize phase and advanced slots.

**Fix:** read the just-externalized slot (`current-1`), surface slots strictly in order one-per-call (`last_externalized_slot` guard) so none are skipped, gated to the non-solo path.

### 3. Stale minting txs never pruned in the multi-node path (`botho/src/consensus/service.rs`)
A minting tx is bound to a specific height/prev-block-hash. After externalize, queued minting txs build on the old tip and never validate, piling up in `pending_values` and stalling the next slot. The multi-node externalize path now prunes pending minting txs, mirroring the solo path.

## End-to-end verification (the headline)

Two real minters (`recommended`, `min_peers=1`, `BOTHO_SLOT_DURATION_SECS=1`, loopback) now externalize in **non-solo mode** and reach a **shared chain at increasing heights**:

```
node A: Slot externalized! slot=1 / slot=2 / slot=3 / slot=4   (solo_mode=false)
node B: Slot externalized! slot=1 / slot=2 / slot=3 / slot=4   (solo_mode=false)
A: height 4 tip 1ada00535fe1f89a3a0a6c3e7e1911d55cb4f729181b7a883879b21f54e38a30
B: height 4 tip 1ada00535fe1f89a3a0a6c3e7e1911d55cb4f729181b7a883879b21f54e38a30
```

Identical tip hashes at increasing heights = a shared chain, not a fork. Before the fix: both nodes pinned at `chainHeight: 0` for the full run with repeated `Invalid quorum set` rejections.

## Tests
- `consensus/scp/tests/test_mesh_networks.rs`: two nodes nominating DISTINCT values with a deterministic pick-one combiner must agree on one shared block (no fork); identical-value edge case. Guards the SCP agreement property.
- `botho` `commands::run::tests`: `peer_id_to_node_id` determinism + injectivity, and a 2-of-2 quorum from distinct peers is valid. These **FAIL on the old mapping** (key collision / invalid quorum) and **PASS after the fix** — verified by temporarily restoring the old impl.

## Safety / determinism
- No SCP state-machine change: `B` is never silently mutated mid-ballot; `consensus/scp/src/slot.rs` is untouched.
- `cargo test -p bth-consensus-scp` green: all 100 unit tests + cyclic/mesh/metamesh/value-cache integration tests pass.
- `cargo test -p botho --lib`: 960 pass.
- Determinism preserved: key derivation and the combiner remain deterministic and order-independent.

## Out of scope / follow-up
Producing each new height's coinbase is gated by PoW mining throughput (a fresh block takes time to mine at the genesis difficulty), which is independent of this consensus fix. The pre-existing `clippy::print_stdout` errors in `botho/src/network/transport/fingerprint.rs` are unrelated and untouched.

Closes #414

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 401
-# Branch: feature/issue-401
+# Issue: 414
+# Branch: feature/issue-414
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -1606,26 +1606,65 @@ fn rebuild_scp_quorum_set(
     QuorumSet::new(threshold, members)
 }
 
-/// Convert a libp2p PeerId to an SCP NodeID
+/// Convert a libp2p PeerId to an SCP NodeID.
+///
+/// SCP identifies, hashes and orders nodes solely by their Ed25519 *public
+/// key* (see `bth_common::NodeID`'s `Eq`/`Hash`/`Ord` impls, which ignore the
+/// `responder_id`). The quorum set is therefore considered invalid — and every
+/// outgoing SCP message is rejected by `Msg::validate` — if two distinct peers
+/// map to the same public key. So the mapping from `PeerId` to public key MUST
+/// be both *deterministic* (every node derives the same key for a given peer)
+/// and *injective* (distinct peers get distinct, valid keys).
+///
+/// The previous implementation copied the first 32 bytes of
+/// `peer_id.to_bytes()` straight into the key. For libp2p Ed25519 peers,
+/// `to_bytes()` is an identity multihash whose first six bytes (`00 24 08 01 12
+/// 20`) are a fixed prefix shared by *every* Ed25519 peer, and the trailing key
+/// bytes are truncated. In practice this collapsed multiple peers onto the same
+/// (often invalid) public key, producing an "Invalid quorum set" rejection that
+/// silently prevented any block from being externalized in multi-node consensus
+/// (issue #414).
+///
+/// We now hash the *full* PeerId bytes with a domain separator to obtain a
+/// well-distributed 32-byte seed, then derive a real Ed25519 keypair from it.
+/// Any 32-byte seed is a valid Ed25519 private key (the scalar is clamped
+/// internally), so the resulting public key is always a valid curve point, and
+/// distinct PeerIds yield distinct keys with overwhelming probability.
+///
+/// NOTE: this remains a deterministic *stand-in* for the peer's real signing
+/// key — it is not the key the peer actually signs SCP messages with. It exists
+/// only so every node builds an identical, valid quorum set from the same set
+/// of connected PeerIds. Exchanging and verifying real per-peer signing keys is
+/// tracked separately; this fix is limited to making the quorum-set membership
+/// well-formed so SCP can externalize.
 fn peer_id_to_node_id(peer_id: &libp2p::PeerId) -> NodeID {
-    // Use the PeerId's string representation as the responder ID
-    // This provides a deterministic mapping from PeerId to NodeID
+    use sha2::{Digest, Sha256};
+
+    // Use the PeerId's string representation as the responder ID. This is purely
+    // informational for SCP (NodeID equality ignores it) but keeps logs
+    // readable.
     let peer_str = peer_id.to_string();
     let responder_id =
         ResponderId::from_str(&format!("{}:8443", &peer_str[..12.min(peer_str.len())]))
             .unwrap_or_else(|_| ResponderId::from_str("peer:8443").unwrap());
 
-    // Derive a deterministic Ed25519 public key from the PeerId bytes
-    // This is a placeholder - in production, peers should exchange actual keys
-    let peer_bytes = peer_id.to_bytes();
-    let mut key_bytes = [0u8; 32];
-    let copy_len = peer_bytes.len().min(32);
-    key_bytes[..copy_len].copy_from_slice(&peer_bytes[..copy_len]);
+    // Derive a deterministic, unique, valid Ed25519 public key from the *full*
+    // PeerId bytes. Hashing avoids the fixed-multihash-prefix collision and the
+    // truncation of the old implementation.
+    let mut hasher = Sha256::new();
+    hasher.update(b"botho-scp-node-id-v1");
+    hasher.update(peer_id.to_bytes());
+    let seed = hasher.finalize();
+
+    // Any 32-byte value is a valid Ed25519 private key; deriving the public key
+    // cannot fail for a 32-byte input, but fall back defensively just in case.
+    let public_key = bth_crypto_keys::Ed25519Private::try_from(&seed[..])
+        .map(|private| Ed25519Public::from(&private))
+        .unwrap_or_default();
 
     NodeID {
         responder_id,
-        public_key: Ed25519Public::try_from(&key_bytes[..])
-            .unwrap_or_else(|_| Ed25519Public::default()),
+        public_key,
     }
 }
 
@@ -1799,5 +1838,90 @@ mod tests {
         let mid = LOW + (HIGH - LOW) / 2;
         assert!(!should_pause_for_balance(mid, HIGH, 0));
         assert!(!should_resume_from_balance_pause(mid, LOW, 0));
+    }
+
+    // ---- Issue #414: PeerId -> NodeID mapping must be deterministic + injective
+    // ----
+
+    /// Build a real libp2p Ed25519 PeerId from a 32-byte secret seed.
+    fn ed25519_peer_id(seed: [u8; 32]) -> libp2p::PeerId {
+        let secret = libp2p::identity::ed25519::SecretKey::try_from_bytes(seed.to_vec())
+            .expect("valid ed25519 secret");
+        let keypair =
+            libp2p::identity::Keypair::from(libp2p::identity::ed25519::Keypair::from(secret));
+        keypair.public().to_peer_id()
+    }
+
+    #[test]
+    fn peer_id_to_node_id_is_deterministic() {
+        // Regression test for issue #414: the mapping must be deterministic so
+        // every node derives the SAME quorum-set membership for a given peer.
+        let pid = ed25519_peer_id([7u8; 32]);
+        let a = peer_id_to_node_id(&pid);
+        let b = peer_id_to_node_id(&pid);
+        assert_eq!(a, b, "same PeerId must map to the same NodeID");
+        assert_eq!(a.public_key, b.public_key);
+    }
+
+    #[test]
+    fn peer_id_to_node_id_is_injective_for_distinct_peers() {
+        // Regression test for issue #414: distinct peers MUST map to distinct
+        // public keys. NodeID equality is by public key only, so a collision
+        // here makes the quorum set invalid (`Msg::validate` rejects every
+        // outgoing SCP message) and multi-node consensus can never externalize.
+        //
+        // The previous implementation copied the first 32 bytes of
+        // `peer_id.to_bytes()`, whose leading bytes are a fixed multihash prefix
+        // shared by all Ed25519 peers, so distinct peers collided.
+        let n = 16;
+        let mut keys = std::collections::HashSet::new();
+        let mut node_ids = std::collections::HashSet::new();
+        for i in 0..n {
+            let mut seed = [0u8; 32];
+            seed[0] = i as u8 + 1;
+            seed[31] = 0xAB;
+            let pid = ed25519_peer_id(seed);
+            let node_id = peer_id_to_node_id(&pid);
+            // Distinct public keys.
+            assert!(
+                keys.insert(node_id.public_key.clone()),
+                "duplicate public key derived for distinct PeerId {pid}"
+            );
+            // Distinct NodeIDs (which compare by public key).
+            assert!(node_ids.insert(node_id));
+        }
+        assert_eq!(keys.len(), n);
+    }
+
+    #[test]
+    fn two_peers_yield_a_valid_quorum_set() {
+        // Regression test for issue #414: a 2-of-2 quorum built from two distinct
+        // peers must be VALID. The bug produced two members with identical
+        // public keys, so `QuorumSet::is_valid()` returned false (duplicate
+        // member) and the externalize message was never broadcast, leaving both
+        // minters stuck at height 0.
+        use bth_consensus_scp_types::QuorumSetMember;
+
+        let local = peer_id_to_node_id(&ed25519_peer_id([1u8; 32]));
+        let peer = peer_id_to_node_id(&ed25519_peer_id([2u8; 32]));
+
+        assert_ne!(
+            local.public_key, peer.public_key,
+            "two distinct peers must not share a public key"
+        );
+
+        let qs = QuorumSet::new(
+            2,
+            vec![QuorumSetMember::Node(local), QuorumSetMember::Node(peer)],
+        );
+        assert!(
+            qs.is_valid(),
+            "2-of-2 quorum from distinct peers must be valid: {qs:?}"
+        );
+        assert_eq!(
+            qs.nodes().len(),
+            2,
+            "quorum must contain two distinct nodes"
+        );
     }
 }
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -164,6 +164,12 @@ pub struct ConsensusService {
     /// Current slot's externalized values (if any)
     externalized: Option<Vec<ConsensusValue>>,
 
+    /// Highest SCP slot index already surfaced as a `SlotExternalized` event.
+    /// Used in the multi-node path so an externalized slot is only emitted once
+    /// (the SCP node auto-advances on externalize, so the just-externalized
+    /// slot stays readable for several ticks).
+    last_externalized_slot: Option<SlotIndex>,
+
     /// A quorum set reconfiguration that arrived mid-round and was deferred to
     /// the next slot boundary to avoid stranding an in-flight slot.
     pending_quorum_set: Option<QuorumSet>,
@@ -282,6 +288,7 @@ impl ConsensusService {
             events: VecDeque::new(),
             last_slot_attempt: Instant::now(),
             externalized: None,
+            last_externalized_slot: None,
             pending_quorum_set: None,
         }
     }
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!** This is the capstone multi-node consensus fix and it holds up under maximum-rigor review. Verdict: **APPROVE**.

## Scope / safety
- **`consensus/scp/src/` is entirely untouched.** Verified against the merge-base (`cf67c71`): the only files changed are `botho/src/commands/run.rs`, `botho/src/consensus/service.rs`, `consensus/scp/tests/test_mesh_networks.rs`, and `.loom-managed`. No `slot.rs` / `node_impl.rs` / `msg.rs` / `ballot.rs` change. The fix is purely node-layer, as claimed.

## 1. `peer_id_to_node_id()` — the critical fix
- **Old impl genuinely collided.** `NodeID`'s `Eq`/`Hash`/`Ord` (`common/src/node_id.rs:58-82`) compare **only** `public_key`, ignoring `responder_id`. The old code copied the first 32 bytes of `peer_id.to_bytes()`, whose leading `00 24 08 01 12 20` multihash prefix is shared by every Ed25519 peer → collisions/invalid keys → duplicate quorum members → `QuorumSet::is_valid()` false → `Msg::validate` rejects every outgoing message. **Empirically confirmed**: I temporarily restored the old impl and `peer_id_to_node_id_is_injective_for_distinct_peers` failed with "duplicate public key derived for distinct PeerId" and `two_peers_yield_a_valid_quorum_set` failed with "two distinct peers must not share a public key".
- **Determinism**: `peer_id_to_node_id` is a pure function of `peer_id.to_bytes()` + the fixed domain separator `b"botho-scp-node-id-v1"`. No node-local salt, randomness, or clock. Same PeerId → same SHA-256 seed → same Ed25519 keypair on every node and run. Confirmed by `peer_id_to_node_id_is_deterministic`.
- **Self-consistency**: a node's own NodeID (`run.rs:420`) and the membership entries built by peers (`run.rs:1427`/`1434`) all flow through the same function applied to the same exchanged PeerId, so node A's self-NodeID == the NodeID B derives for A. A node will recognize messages about itself in the quorum.
- **Validity**: any 32-byte SHA-256 output is a valid `Ed25519Private` (dalek `SecretKey::try_from` only fails on length ≠ 32), and `Ed25519Public::from(&private)` always yields a real curve point. The `.unwrap_or_default()` is dead-but-defensive.

## 2. `check_externalized()` off-by-one
- Verified against `node_impl.rs`: `externalize()` swaps in `slot_index+1` as `current_slot` and files slot N into `externalized_slots`, so post-externalize `current_slot_index()` == N+1 and `get_externalized_values(N+1)` is `None`. The old `current_slot_index()` query never fired. New `current.checked_sub(1)` reads N correctly.
- The `last_externalized_slot` (`last+1`) guard surfaces slots strictly in order, one per apply-cycle (gated by `externalized.is_some()`, cleared in `advance_slot`), with no skip and no double-report. Since `check_externalized()` runs after every `handle_message` and `tick`, `current` advances by at most one between calls, so the in-order guarantee holds.
- Correctly gated to non-solo: returns early under `is_solo_mode()`; the solo externalize path (`propose_pending_values`, `advance_slot`) is unchanged.

## 3. Stale minting-tx prune
- The multi-node prune `pending_values.retain(|v| !v.is_minting_tx)` (service.rs:790) exactly mirrors the solo path (service.rs:684). Correct and necessary — minting txs are bound to height/prev-hash and go stale on externalize.

## 4-5. Tests
- `cargo test -p bth-consensus-scp`: **100 unit tests pass**; integration suite (`--features test_utils --test test_mesh_networks`): **9 pass** including the new `two_node_distinct_values_converge` (load-bearing SCP agreement: two distinct-value minters must externalize the SAME block — no fork) and `two_node_identical_values_converge`.
- `cargo test -p botho --lib`: **960 pass**.
- Test genuineness verified empirically: the three new run.rs regression tests **fail on the old impl, pass on the new** (the determinism test passes on both, correctly — determinism was never the bug).

## 6. Empirical shared-tip evidence
Both minters logging `Slot externalized! slot=1..4 (solo_mode=false)` and converging on the identical tip `1ada00535fe1f89a...` at height 4 is the right signal: identical block hash at the same increasing height proves a shared chain (a fork would diverge). `solo_mode=false` confirms the multi-node SCP path. The deterministic `two_node_distinct_values_converge` test guards this property in CI.

## 7. CI
Head `36c9ac2`. All checks settled green: Compile Benchmarks, Crypto/Node/Transaction Benchmarks, Benchmark Summary, Fuzz Build Guard, Fuzz Summary, rustfmt — all SUCCESS; conditional Fuzz jobs SKIPPED (not failures). **mergeStateStatus: CLEAN.**

Excellent root-cause work overturning the original SCP-ballot hypothesis with empirical evidence. The "stand-in key" caveat (peers don't yet exchange real signing keys) is correctly documented and scoped out. Ready to merge.

Issue #417, opened 2026-06-17
consensus: multi-minter SCP forks + stalls (2-of-2 externalizes divergent blocks; stale-minting jam; joiner SCP-slot desync)


rjwalters:
## Summary

While completing the #396 mine-after-join real-process test I empirically reproduced **three distinct, related defects in the multi-minter consensus path**, the most serious of which is an actual **SCP safety fork in a unanimous (2-of-2) quorum**. These block #396 and must be fixed before any multi-minter testnet. Discovered with real release `botho run` processes peering over libp2p loopback (`BOTHO_SLOT_DURATION_SECS=1`, `recommended` quorum, `min_peers=1`). This extends the (closed) #397, which only described *solo-mode* peer forks and the (now-fixed) wire-decode bug; the fork below happens with `solo_mode=false` and "Quorum satisfied: 2-of-2".

## Finding 1 (CRITICAL — safety) — a 2-of-2 quorum forks and never heals

Two minters A+B in a `recommended` 2-of-2 quorum were repeatedly observed externalizing **different blocks at the same heights** and never reconciling:

```
h=22 A=84102dd2fac4 B=326a371952bf
h=23 A=3307e2630088 B=32594141a59f
h=24 A=3122266cab72 B=b44f7edc71c4
```

After this they deadlock: each rejects the other's next minting tx with
`Minting tx has wrong prev_block_hash expected=<own tip> got=<peer tip>` and
the chain stops. A correct unanimous quorum CANNOT externalize divergent values
at the same slot — this is a genuine SCP safety violation (a fork), not just a
liveness stall. It is intermittent (some runs reach height 28+ cleanly; others
fork by height ~22), i.e. a race in the multi-node nominate/ballot path. This is
the hard blocker for #396 and directly violates Botho's no-hard-forks policy.

## Finding 2 (liveness) — stale-minting-tx jam

In a multi-minter net each node's PoW thread races to mint the next block. When
a peer's block externalizes first, the locally-mined minting tx is bound to the
now-stale tip (`prev_block_hash`/height no longer match) and is rejected by every
validator. In a unanimous quorum the slot can then never externalize (no node
can externalize a value any validator rejects), and the stale value stays latched
in the SCP slot's nomination/ballot state. Reproduced: a 2-node net jams within
a few blocks (e.g. stuck at height 1-4) with `pending_count` growing unbounded as
it re-proposes the doomed value.

## Finding 3 (liveness) — joiner's SCP slot is never advanced after block-sync

[...]

Issue #396, opened 2026-06-16
test: fresh node joins a running local chain, participates in consensus, and mines an accepted block


rjwalters:
## Summary (integration test — flavor A, hermetic local)

Demonstrate, via a real-process local network, that **a fresh node can start on localhost, join an already-running chain, participate in SCP consensus, and mine a block the network accepts** — the full "run your own node and participate" capability, end to end.

This is distinct from `e2e_consensus_integration` (which starts all nodes at genesis together — no node ever JOINS an existing chain). The new, untested property here: a node joining **mid-chain** (peers already at height N) gets admitted into the quorum, catches up via the #376 sync, and has its **mined block accepted** by the existing nodes.

## Approach — real botho processes (build on the node-harness)
Use real `botho run` processes (as `web/packages/wasm-signer/test/node-harness.ts` and the #373 betanet-smoke already do), NOT the in-process `TestNetwork` sim (which uses DashMap message-passing and does not exercise the real libp2p bootstrap + #376 catch-up + run-loop consensus). The point is to test the REAL join/sync/consensus/mint paths.

## Scope
1. **Start a local network of 2 botho nodes** (throwaway configs/data dirs, testnet) that peer with each other and form a quorum. Have them mine to a height N > ring size (so the chain has real history + decoy outputs). One or both mint; they agree via SCP.
2. **Start a 3rd fresh node** (empty ledger) that bootstraps to the running pair (`bootstrap_peers` → the local nodes' gossip ports).
3. **Assert it JOINS + catches up**: the 3rd node connects (peerCount ≥ 1), and syncs from height 0 to the network tip N via the real catch-up path (#376) — not stuck at genesis.
4. **Assert it PARTICIPATES + MINES**: enable minting on the 3rd node (it needs a wallet/reward address); it produces a block that the EXISTING nodes accept (the network tip advances to a block minted by the 3rd node, and all nodes converge on it). This is the core claim — a newly-joined node's block is accepted by consensus.
5. **Assert it earned coins**: the 3rd node's minting reward (coinbase to its address) is on-chain and detectable (its wallet balance / a coinbase output owned by it).
6. Bounded with timeouts so it can't hang; clean teardown of all node processes + temp dirs.

## Key things to nail (investigate during build)
[...]

PR #420, opened 2026-06-18
fix(consensus): tip-agnostic SCP validity to stop the multi-minter fork (#417 Finding 1)

### LINKED ISSUE #419: Proposal: fix the multi-minter SCP safety fork (#417 Finding 1) via tip-agnostic consensus-value validity
## Architect proposal for #417 — root cause of the 2-of-2 SCP safety fork (Finding 1) + fix direction for all three findings

Design-level, read-only analysis against `origin/main` (#410/#412/#415/#416 merged). No source was modified.

---

## TL;DR

- **Finding 1 is NOT a flaw in the SCP federated-voting core.** The MobileCoin SCP slot/node logic (`consensus/scp/src/slot.rs`, `node/node_impl.rs`) is safe: `externalize` only fires on a genuine `Topic::Externalize` (confirm-commit over a quorum), and `check_externalized` reads only truly-externalized slots from the SCP store. The 100 SCP tests use `trivial_validity_fn` (accepts everything) and so cannot exercise the failure.
- **The fork is a botho node-layer defect: a *state-dependent* `validity_fn` combined with SCP's "drop messages carrying an invalid value" rule splits the two minters into two disjoint federated-voting instances that each ballot their own value to externalize.** It is a localized bug, not a protocol redesign — but the correct fix touches a design rule (what makes a minting value "valid" inside SCP), so it needs an approved design decision before a builder implements it.
- **Findings 2 and 3 are downstream/liveness and are moot until Finding 1 is closed.** Their prototyped fixes are assessed below; the same-index slot-restart jam-recovery is itself a fork risk and should NOT be the chosen Finding-2 fix.

---

### The divergence mechanism (code-grounded)

1. Each minter independently solves PoW and submits its **own distinct** coinbase. `run.rs:~1318-1379` drains the local miner, validates against the local tip, then `consensus.submit_minting_tx(tx_hash, pow_priority, bytes)`. Two nodes produce different `tx_hash` AND different `priority` (`botho/src/consensus/value.rs:48-54` — `ConsensusValue` carries `priority` in its `Eq/Ord`), so A's value and B's value are two **distinct** `ConsensusValue`s competing for the same slot.

2. The SCP `validity_fn` is **state-dependent**: it looks the value up in `tx_cache` and runs `validate_minting_tx`, which requires `tx.prev_block_hash == state.tip_hash` and `tx.block_height == state.height + 1` (`botho/src/consensus/validation.rs:133-152`). A value is only valid while it matches the validator's *current local tip*.

3. A peer's value enters the local cache **only if it validates at the instant the gossip arrives** (`run.rs` `NetworkEvent::NewMintingTx`, ~656-664: `validate_minting_tx` then `register_minting_tx`; on error it logs "Rejecting invalid peer minting tx" and drops it). There is no re-validation or re-registration later.

4. SCP **drops any peer message that carries a value the local node cannot validate**: `consensus/scp/src/slot.rs:398-405` — in `handle_messages`, for each value in the message, `if self.validate(&value).is_err() { continue 'msg_loop; }`. The message is never stored in `self.M`, so it never contributes to nomination/ballot federated voting. A value dropped here is **never re-examined** even if local chain state later changes.

5. **The race:** at a 1s slot with two near-simultaneous PoW solutions, A and B can each cross from `confirm-nominate` into balloting on their own singleton `Z` before the peer's value is accepted locally. The combiner (`service.rs:217-249`) deterministically truncates to the single best minting tx — but it runs over each node's **own** `Z`, which contains only that node's value when the peer's value was dropped. Each node then runs the entire prepare/commit/externalize ballot sequence on its own value. Because each node's `M` map lacks the peer's contradicting ballots (they were dropped as "invalid value"), each node sees a *quorum-of-itself* confirm-commit and externalizes its own block. Result: divergent externalize at the same slot — the observed fork.

In short: SCP's safety theorem assumes a value that is valid for one honest node is valid for all of them (validity is a pure function of the value). Botho violates that assumption by making validity a function of mutable local chain state, and SCP's "silently drop invalid-value messages" rule then **partitions the quorum into two single-node voting instances that each complete independently.** That is the fork, and it is intermittent precisely because it depends on the gossip/PoW timing race in steps 3-5.

### Why the recent fixes don't close it
- #414/#416 `check_externalized` reading `current-1` is **correct** and is not the source — it only surfaces slots the SCP node genuinely externalized.
- #410 `register_minting_tx` is necessary but **insufficient**: it only helps when the peer's tx happens to validate at gossip-arrival time; under the race it is dropped, and dropped values are never retried.
- The block builder (`run.rs:build_block_from_externalized`) is deterministic from the agreed `values`, so two equal value-sets would yield equal blocks. The divergence is upstream, in *which* value set each node externalizes.

### Confidence / what runtime instrumentation would confirm it
High confidence from static analysis. To confirm empirically, add (temporary) debug logs at: (a) `slot.rs:401` when a value is dropped as invalid — log slot, sender, value hash; (b) `run.rs` NewMintingTx reject branch — log the WrongPrevBlockHash/WrongBlockHeight reason; (c) the contents of `Z` and `self.B.X` at the moment the slot enters `Commit`. A fork run should show each node dropping the peer's minting value (a) before it externalizes its own singleton `Z` (c).

---

## Bug vs. design-flaw verdict

**Localized bug in the botho consensus layer, with a required design decision about the validity model.** The SCP core needs no change. But "fix it" forces a choice about how a minting value is made valid-for-all-honest-nodes. A builder must not pick that silently. Options:

- **Option A — Deterministic single-proposer-per-slot (leader election).** Derive the slot's eligible minter deterministically (e.g. lowest `NodeID`, or VRF/PoW-rank over the connected quorum for slot N). Only the leader's coinbase is a valid consensus value for that slot; others are not proposed. Removes the *competing-distinct-values* condition entirely, so there is only one value to agree on. Tradeoff: needs a liveness fallback if the leader is absent/slow (rotate after timeout), and changes Botho's "anyone mints" model toward leader-per-slot. Strongest safety story.
- **Option B — Make minting-value validity tip-agnostic inside SCP, and never drop a peer value for tip mismatch.** Validate a minting consensus value on its *intrinsic* properties (well-formed coinbase, PoW meets the difficulty for `block_height`, height/reward self-consistent) but DROP the `prev_block_hash == local tip` and `height == local+1` checks from the SCP `validity_fn` (keep them only at block-apply time). Then both A's and B's values are valid for both nodes, their ballot messages are no longer dropped, federated voting converges, and the deterministic combiner picks one. Tradeoff: must guarantee the combiner's pick is reproducible across nodes (it is: `priority` then `tx_hash`, `service.rs:229-233`) and that the *agreed* value's `prev_block_hash` is the common tip both nodes share for that slot — which it is, because both built on the same height-N tip.
- **Option C — Re-drive SCP when local state changes / retry dropped values.** Keep tip-dependent validity but re-register and re-feed peer values after a tip change. Fragile (must re-inject into `M`, which SCP does not expose) and does not remove the partition window; not recommended as the primary fix.

**Recommended: Option B as the core safety fix (smallest, provably-converging, keeps "anyone mints"), optionally layered with Option A's deterministic combiner tie-break already present.** Option A is a larger model change to keep as a fallback if B's combiner determinism proves insufficient.

---

## Recommended fix direction, ordered

**Finding 1 (safety gate — do first):** Make the SCP `validity_fn` for minting values **a pure function of the value** (intrinsic + PoW-vs-difficulty-for-stated-height), removing the local-tip equality checks from the in-SCP validation path (`validation.rs:133-152` is split: tip checks stay at block-apply in `add_block_from_network`, intrinsic checks move into the SCP validity path). And: **never silently drop a peer's ballot for a tip mismatch** — once validity is tip-agnostic this drop no longer happens. Then both minters' values survive into `M`, federated voting converges, and the deterministic combiner (priority, then tx_hash) selects one shared value → one block → no fork. The losing minter's value is naturally discarded by the combiner.

**Finding 2 (liveness — after 1):** With Finding 1 fixed, the stale-minting jam largely dissolves because the agreed value is whatever the quorum converged on, and the existing post-externalize pruning (`service.rs` `check_externalized` `pending_values.retain(|v| !v.is_minting_tx)`) already drops stale local minting txs. Keep that pruning; do NOT adopt the prototyped same-index `restart_current_slot` jam-recovery as the Finding-2 fix — restarting a slot at the same index after it may have *accepted commit* can drop committed ballot state and is itself a fork risk (the issue body flags this). If a liveness backstop is still needed, prefer a timeout that advances only via legitimate SCP (e.g. nominate-round timer escalation already in `slot.rs`), never a state-discarding restart.

**Finding 3 (liveness — after 1):** The joiner SCP-slot desync is a genuine, independent gap and its prototyped fix is sound: after #376 block-sync applies blocks to height H, call a `sync_scp_slot_to_chain(H)` that does `reset_slot_index(H+1)` to match `ConsensusService::new`'s `initial_slot = initial_height + 1` (`service.rs:253`). `reset_slot_index` (`node_impl.rs:299-313`) is safe here because the joiner has **no in-flight ballot state** for the live slot (its messages were all being discarded as "future slots", `node_impl.rs:196-206`) — it is abandoning a stale genesis slot, not committed state. Gate it so it only runs when the SCP slot is strictly behind `H+1` and the node holds no peer ballot/nominate state for the current slot (mirror the #408 deferral guard). This is safe to ship once Finding 1 lands; on its own it is moot while the net forks.

---

## Safety argument (no-hard-forks consistent)

The recommended Finding-1 fix **restores the precondition of SCP's safety theorem**: validity becomes a pure function of the value, so "valid for one honest node ⇒ valid for all honest nodes." With that, no honest node drops another honest node's ballot, the single shared federated-voting instance runs as designed, and SCP's agreement property guarantees at most one value can be confirmed-committed per slot across any quorum — i.e. no fork — **without weakening liveness**, because the combiner deterministically reduces the candidate set to one and the losing value is dropped rather than jamming the slot. We do not touch the SCP core, do not add a state-discarding slot restart (which would be the actual fork hazard), and do not introduce a consensus-incompatible change requiring a fork: the change is to *what bytes count as a valid consensus value*, applied uniformly, before any mainnet. The block-apply path keeps the full tip-equality checks, so an actually-stale block can never be appended.

---

## Test strategy (deterministic regression guard, no flaky real-process race)

The fork must be reproducible **in-process** so the guard is deterministic:

1. **New SCP-layer test with a *stateful* validity_fn** (the current `trivial_validity_fn` masks the bug). In `consensus/scp/tests/`, build a 2-of-2 mesh where the `validity_fn` accepts a value only if it equals the node's "current tip token", and have the two nodes propose *distinct* values for the same slot (simulating two coinbases). Assert: both nodes externalize the **same** value set for the slot (no divergence). This test FAILS on today's drop-on-invalid behavior and PASSES once validity is tip-agnostic — a precise regression guard at the layer where the bug lives.
2. **Botho-layer unit test** in `botho/src/consensus` driving two `ConsensusService` instances (2-of-2 quorum, distinct minting `ConsensusValue`s with different priority) by hand-passing each other's `BroadcastMessage` events, with the real `validity_fn`/`combine_fn`. Assert both emit `SlotExternalized` with identical `values`, and that `build_from_externalized` yields identical block hashes. Drive the race by feeding the peer's nominate BEFORE registering its minting tx, to force the historical drop path.
3. **Re-enable the #396 real-process harness** (`web/packages/wasm-signer/test/mine-after-join-live-node.test.ts`, drop `&& false`) as an end-to-end soak (many runs) — but only as acceptance, not the primary guard, since it is inherently timing-dependent.

---

## References (file:line, origin/main)
- `botho/src/consensus/value.rs:18-54` — `ConsensusValue` carries `priority` in Eq/Ord (distinct per minter).
- `botho/src/consensus/validation.rs:123-152` — tip-equality checks that make validity state-dependent.
- `consensus/scp/src/slot.rs:398-405` — peer messages carrying an "invalid" value are silently dropped from `self.M`.
- `consensus/scp/src/slot.rs:611-664` — `do_nominate_phase` / combine(Z) bootstraps the ballot from the local (possibly singleton) `Z`.
- `consensus/scp/src/node/node_impl.rs:96-128, 230-241` — `externalize` only fires on a true `Topic::Externalize`; core is safe.
- `consensus/scp/src/node/node_impl.rs:196-206` — future-slot messages discarded (Finding 3).
- `consensus/scp/src/node/node_impl.rs:299-313` — `reset_slot_index` (Finding 3 fix primitive).
- `botho/src/consensus/service.rs:526-533` — `register_minting_tx` only caches on validate-at-arrival (race window).
- `botho/src/consensus/service.rs:217-249` — deterministic combiner (priority, then tx_hash).
- `botho/src/commands/run.rs:~638-664` — NewMintingTx gossip: validate-then-register, drop on mismatch.
- `botho/src/commands/run.rs:~1318-1379` — local miner submits its own distinct coinbase.
- `botho/src/commands/run.rs:1062-1113, ~1471` — externalize → deterministic block build → advance_slot.
- `consensus/scp/src/test_utils.rs:27` — `trivial_validity_fn` (why 100 SCP tests miss this).

---
*Filed by the Loom Architect. Awaiting human approval (`loom:architect`).*

rjwalters:
## Summary

Fixes the **CRITICAL SCP safety fork** (#417 Finding 1): a unanimous 2-of-2
quorum externalizing divergent blocks at the same height. The root cause was a
**state-dependent (tip-dependent) consensus `validity_fn`**, which violates
SCP's agreement (no-fork) theorem (validity must be a pure function of the
value). Under the fast-slot PoW race, two minters each dropped the peer's
competing coinbase as "invalid against my tip", the quorum partitioned into two
single-node voting instances, and each externalized its own block — a fork.

Closes #419
Refs #417 (Findings 1–3), unblocks #396.

## Checks moved out of SCP validity vs kept intrinsic

The SCP `validity_fn` now uses `validate_minting_tx_intrinsic` (tip-agnostic).

**Kept INTRINSIC (in SCP validity):**
- structural well-formedness (deserialize as `MintingTx`)
- PoW solution meets the difficulty **stated in the tx itself** (`verify_pow`
  hashes only the tx's own fields against `tx.difficulty`)
- timestamp not absurdly far in the future (wall-clock bound, not tip-relative)

**Moved to BLOCK-APPLY only (`LedgerStore::add_block`, already enforced there):**
- `prev_block_hash == tip`
- `height == tip + 1`
- `difficulty == chain difficulty`
- `reward == emission(height, total_mined)`
- `timestamp >= parent timestamp` (monotonicity)

A genuinely stale/fraudulent block can never be appended, because `add_block`
re-checks all of the above unconditionally. The gossip-ingest path
(`NewMintingTx`) now registers a peer coinbase on **intrinsic** validity only,
so its bytes are cached and SCP no longer drops the peer's ballots for a tip
mismatch (the partition mechanism).

## Findings 2 & 3

- **Finding 2 (liveness):** pin the first-proposed coinbase per slot in
  multi-node mode so the local miner's continuous fresh coinbases don't churn
  the SCP nomination set. Kept the existing post-externalize minting-tx
  pruning. Did **not** add the rejected same-index slot-restart.
- **Finding 3 (liveness):** `ConsensusService::sync_scp_slot_to_chain`, called
  from the block-sync path, fast-forwards a joiner's SCP slot to `height + 1`
  (matching `ConsensusService::new`). Guarded to only advance (never rewind) and
  to refuse if the current slot holds in-flight ballot state.

## Regression tests (the load-bearing safety guard)

The existing 100 SCP tests use a trivial validity_fn and miss this.

- `consensus/scp/tests/test_tip_dependent_validity_fork.rs` — SCP-layer 2-of-2
  with a stateful/tip-like validity_fn and two distinct values. **Confirmed: the
  no-fork test FAILS pre-fix** (the nodes stall/diverge under tip-dependent
  validity) and PASSES with tip-agnostic validity.
- `botho` service test `two_minters_converge_on_single_value_no_fork` — drives
  two `ConsensusService` instances with the **real** validity/combine fns and
  distinct coinbases, asserts one shared externalize. **Confirmed: FAILS pre-fix**
  (consensus stalls) and passes with the fix.
- Finding-3 tests: `sync_scp_slot_fast_forwards_after_block_sync`,
  `sync_scp_slot_is_noop_in_solo_mode`.

## Verification

- `cargo test -p bth-consensus-scp --features test_utils`: 100 lib + all
  integration tests green (incl. 2 new fork tests).
- `cargo test -p botho --lib`: 963 green (incl. 3 new tests).
- `cargo fmt --check` + `cargo clippy` clean on changed crates;
  `pnpm -C web typecheck` green.
- **Real 2-node `botho run` over loopback (RUST_LOG=info):** mined to height 35
  with **IDENTICAL block hashes at every height including the old fork zone
  (22/23/24)**, via real SCP `Slot externalized!` in non-solo mode. The fork
  that previously appeared by ~height 22 is **gone**.
- **Real 3-node:** the joiner connects, syncs 0→24, and fast-forwards its SCP
  slot ("Fast-forwarding SCP slot ... Finding 3"); the 3-of-3 net advanced and
  all three converged on identical tips.

## Known limitation (out of scope, pre-existing)

The full `#396` 3-of-3 "joiner mines and all converge several blocks past N"
end-state is **not yet** reliably green, blocked by a **pre-existing SCP-slot /
block-height DRIFT**: the established minters' SCP `current_slot_index` runs
ahead of their ledger height (a stale/duplicate-height minting value can be
externalized and rejected at apply without rewinding the slot), so the joiner's
`height + 1` target no longer matches the network's drifted slot and they
discard each other's messages. Re-aligning the SCP slot index with block height
is a deeper protocol change tracked separately. Per the issue's STOP guardrail
I did not force a misleading green: the **#396 soak harness is left gated** with
an updated rationale, while the **safety fork (the hard requirement) is fixed
and proven**.

## Files touched

- `botho/src/consensus/validation.rs` — intrinsic (tip-agnostic) validation
- `botho/src/consensus/service.rs` — tip-agnostic SCP validity_fn, coinbase
  pinning, `sync_scp_slot_to_chain`, 3 new tests
- `botho/src/commands/run.rs` — intrinsic gossip-ingest gate + block-sync slot
  fast-forward
- `consensus/scp/tests/test_tip_dependent_validity_fork.rs` (new) +
  `consensus/scp/Cargo.toml` ([[test]] entry)
- `web/packages/wasm-signer/test/mine-after-join-live-node.test.ts` — updated
  gating rationale

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 414
-# Branch: feature/issue-414
+# Issue: 419
+# Branch: feature/issue-419
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,0 @@
-# Loom-managed worktree marker
-# Created by .loom/scripts/worktree.sh
-# Issue: 419
-# Branch: feature/issue-419
-# Removing this file makes Loom treat the worktree as user-owned and refuse
-# to clean it up automatically.
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -637,31 +637,37 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                         }
                         NetworkEvent::NewMintingTx(minting_tx) => {
                             // A peer proposed this minting tx for consensus.
-                            // Validate it against our chain state, then register
-                            // it in the consensus tx cache so the local SCP node
-                            // can validate (and therefore accept) the peer's
-                            // nominate/ballot messages referencing it (issue #409).
+                            // Register it in the consensus tx cache so the local
+                            // SCP node can validate (and therefore accept) the
+                            // peer's nominate/ballot messages referencing it
+                            // (issue #409).
+                            //
+                            // SAFETY (issue #419 / #417 Finding 1): we gate
+                            // registration on INTRINSIC (tip-agnostic) validity
+                            // ONLY — well-formedness + PoW vs the tx's stated
+                            // difficulty. We must NOT require the peer tx to
+                            // match our local tip here: under the fast-slot PoW
+                            // race the peer's coinbase legitimately builds on the
+                            // same height-N tip but may not equal whatever our
+                            // local tip momentarily is, and dropping it would
+                            // mean the SCP validity_fn later reports "not in
+                            // cache" and SCP silently drops the peer's ballots —
+                            // the exact mechanism that partitions the quorum and
+                            // forks. The tip-relative checks are enforced at
+                            // block-apply, so a stale block can never be
+                            // appended. The intrinsic PoW check still rejects
+                            // junk/DoS values.
                             let tx_hash = minting_tx.hash();
                             debug!("Received minting tx {} from network", hex::encode(&tx_hash[0..8]));
 
-                            let chain_state = node
-                                .shared_ledger()
-                                .read()
-                                .ok()
-                                .and_then(|ledger| ledger.get_chain_state().ok());
-
-                            if let Some(chain_state) = chain_state {
-                                let temp_state = Arc::new(RwLock::new(chain_state));
-                                let validator = TransactionValidator::new(temp_state);
-                                match validator.validate_minting_tx(&minting_tx) {
-                                    Ok(()) => {
-                                        let tx_bytes = bincode::serialize(&minting_tx)
-                                            .expect("Failed to serialize minting tx");
-                                        consensus.register_minting_tx(tx_hash, tx_bytes);
-                                    }
-                                    Err(e) => {
-                                        debug!(error = %e, "Rejecting invalid peer minting tx");
-                                    }
+                            match TransactionValidator::validate_minting_tx_intrinsic(&minting_tx) {
+                                Ok(()) => {
+                                    let tx_bytes = bincode::serialize(&minting_tx)
+                                        .expect("Failed to serialize minting tx");
+                                    consensus.register_minting_tx(tx_hash, tx_bytes);
+                                }
+                                Err(e) => {
+                                    debug!(error = %e, "Rejecting intrinsically-invalid peer minting tx");
                                 }
                             }
                         }
@@ -855,7 +861,20 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                                 if let Ok(state) = ledger.get_chain_state() {
                                                     metrics_updater.set_block_height(state.height);
                                                     sync_manager.on_blocks_added(state.height);
+                                                    let synced_height = state.height;
                                                     consensus.update_chain_state(state);
+                                                    // Issue #419 / #417 Finding 3: after catching
+                                                    // up via block-sync, advance the SCP slot to
+                                                    // `height + 1` so this node stops discarding
+                                                    // the live network's slot messages as "future
+                                                    // slots" and can actually participate.
+                                                    if consensus.sync_scp_slot_to_chain(synced_height) {
+                                                        info!(
+                                                            slot = consensus.current_slot(),
+                                                            height = synced_height,
+                                                            "Advanced SCP slot to synced chain height"
+                                                        );
+                                                    }
                                                 }
                                             }
                                         }
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -197,7 +197,22 @@ impl ConsensusService {
         let chain_state_ref = Arc::new(RwLock::new(initial_chain_state));
         let validator = TransactionValidator::new(chain_state_ref.clone());
 
-        // Validation callback - validates transactions using shared state
+        // Validation callback — the SCP consensus validity function.
+        //
+        // SAFETY (issue #419 / #417 Finding 1): this MUST be a pure function of
+        // the value, i.e. TIP-AGNOSTIC. SCP's no-fork agreement theorem assumes
+        // "valid for one honest node ⇒ valid for all honest nodes". If validity
+        // depended on the local chain tip, SCP's "silently drop messages
+        // carrying an un-validatable value" rule would partition two competing
+        // minters into two single-node voting instances that each externalize
+        // their own block — the multi-minter fork. So we validate only the
+        // INTRINSIC properties of the minting tx (well-formedness, PoW vs the
+        // tx's stated difficulty, non-future timestamp). The tip-relative
+        // checks (prev_block_hash == tip, height == tip+1, chain difficulty,
+        // emission reward, parent-timestamp monotonicity) are enforced
+        // unconditionally at block-apply (`LedgerStore::add_block`), so a stale
+        // or fraudulent block can never be appended even though its minting tx
+        // is an acceptable consensus *value*.
         let validation_state = shared_state.clone();
         let validity_fn: Arc<dyn Fn(&ConsensusValue) -> Result<(), String> + Send + Sync> =
             Arc::new(move |value: &ConsensusValue| {
@@ -210,13 +225,15 @@ impl ConsensusService {
                     format!("Transaction not in cache: {:?}", &value.tx_hash[0..8])
                 })?;
 
-                // Create a temporary validator for this check
+                // Create a temporary validator for the (tip-agnostic) check.
+                // The chain_state is only used by the transfer-tx structural
+                // path; the minting-tx path is fully tip-agnostic.
                 let temp_state = Arc::new(RwLock::new(state.chain_state.clone()));
                 let temp_validator = TransactionValidator::new(temp_state);
 
-                // Validate based on transaction type
+                // Validate INTRINSIC (tip-agnostic) properties only.
                 temp_validator
-                    .validate_from_bytes(&entry.data, entry.is_minting_tx)
+                    .validate_from_bytes_intrinsic(&entry.data, entry.is_minting_tx)
                     .map_err(|e| e.to_string())
             });
 
@@ -621,15 +638,38 @@ impl ConsensusService {
         // This prevents minting txs from crowding out user transactions.
         let mut to_propose: BTreeSet<ConsensusValue> = BTreeSet::new();
 
-        // Find the best minting tx (highest priority)
-        let best_minting_tx = self
-            .pending_values
-            .iter()
-            .filter(|v| v.is_minting_tx)
-            .max_by_key(|v| v.priority)
-            .cloned();
+        // LIVENESS (issue #419 / #417 Finding 2): in multi-node mode, once we
+        // have already proposed a minting tx for the CURRENT slot, keep
+        // proposing that SAME value rather than swapping to a newly-mined,
+        // higher-priority one. The local PoW miner produces a fresh minting tx
+        // (distinct hash) several times a second; if every node kept replacing
+        // its proposed coinbase mid-slot, each node's SCP nomination set would
+        // churn with different values and the quorum could never reach a shared
+        // confirmed-nominate, jamming the slot in a unanimous quorum. Pinning
+        // the first-proposed coinbase per slot stabilizes the candidate set so
+        // federated voting + the deterministic combiner converge. The losing
+        // coinbases are simply not proposed; this does not affect safety
+        // (validity is tip-agnostic and block-apply still enforces every
+        // tip-relative rule). After externalize, all pending minting txs are
+        // pruned (see `check_externalized`), so the next slot starts fresh.
+        let already_proposed_minting = self.proposed_values.iter().any(|v| v.is_minting_tx);
+        let minting_tx = if already_proposed_minting {
+            // Re-propose the exact minting value we already committed to this
+            // slot, if it is still known.
+            self.proposed_values
+                .iter()
+                .find(|v| v.is_minting_tx)
+                .cloned()
+        } else {
+            // Find the best (highest priority) minting tx to propose.
+            self.pending_values
+                .iter()
+                .filter(|v| v.is_minting_tx)
+                .max_by_key(|v| v.priority)
+                .cloned()
+        };
 
-        if let Some(minting_tx) = best_minting_tx {
+        if let Some(minting_tx) = minting_tx {
             to_propose.insert(minting_tx);
         }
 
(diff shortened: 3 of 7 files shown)


rjwalters:
✅ **Approved!** — reviewed with maximum rigor as the most safety-critical PR of the #417/#419 effort. This correctly and *safely* fixes the multi-minter SCP fork.

## (1) SAFETY — every moved tip-check is enforced at apply-time on EVERY append path ✅ (the critical check)

Verified `LedgerStore::add_block` (`botho/src/ledger/store.rs:383`) enforces **unconditionally**, on every block:
- `height == tip + 1` (L388), `prev_block_hash == tip` (L397)
- `header.difficulty == chain difficulty` (L438) + PoW vs that verified difficulty (L446)
- `reward == calculate_block_reward(...)` (L455)
- timestamp monotonicity vs parent (L478) + future bound (L488)
- **header ↔ minting-tx consistency** (L408–431, L473) — closes the "declare one difficulty in the header, another in the tx" mismatch attack.

Both append paths route through it:
- **externalize → build-block:** `run.rs:1186` `add_block_from_network` → `add_block`. `build_from_externalized` (`block_builder.rs:139`) copies header fields straight from the minting tx, so a stale externalized value is rejected at apply — **wasted slot, never a fraudulent append**.
- **#376 block-sync:** `run.rs:832` `add_block_from_network` → `add_block`.

A node can never append a block with wrong prev_hash/height/difficulty/reward/timestamp even though SCP now accepts the value. **No tip-relative check was lost.**

## (2) Tip-agnostic validity preserves no-fork ✅

`validity_fn` (`service.rs:217`) now calls `validate_from_bytes_intrinsic` (well-formedness, PoW vs the tx's *stated* difficulty, non-future timestamp) — a pure function of the value. No honest node drops a peer's ballot for a tip reason, so federated voting converges. The combiner (`service.rs:246`) is deterministic across nodes: sort by `priority` desc (`priority = u64::MAX - pow_hash[0..8]`, pure fn of the tx) then `tx_hash` asc, truncate to one minting tx. Same input set ⇒ same chosen Z everywhere.

## (3) Regression tests genuinely fail-pre / pass-post ✅ (empirically verified)

`consensus/scp/tests/test_tip_dependent_validity_fork.rs::no_fork_with_tip_agnostic_validity` is the load-bearing guard. I temporarily reverted it to the pre-fix tip-dependent `validity_fn` (same two competing values) and ran it: it **FAILED** (`node A never externalized` — the partition/stall). Restored, it **PASSES**. The botho-layer `two_minters_converge_on_single_value_no_fork` (`service.rs:1504`) drives two real `ConsensusService` instances with the real validity/combine fns and `.expect()`s externalization, so it too fails on a stall. Both are real guards, not no-ops.

## (4) No SCP regression ✅

`cargo test -p bth-consensus-scp --features test_utils`: **100 lib + all integration green** (incl. 2 new fork tests). `cargo test -p botho --lib`: **963 passed, 0 failed**.

## (5) Adversarial analysis of the intrinsic checks ✅ (safe; one liveness note)

PoW is validated against the tx's *own stated* difficulty in SCP validity. An attacker can therefore get a low-stated-difficulty coinbase cached/externalized — but `add_block` rejects any block whose difficulty ≠ chain difficulty (L438/L420), so the worst case is a **wasted slot, not a safety break**. Non-blocking liveness note: the combiner ranks by PoW-hash priority (not difficulty), so a low-difficulty tx with a lucky hash could win a slot then be rejected at apply; gossip-ingest could additionally require `stated_difficulty >= chain difficulty` as a DoS filter. Out of scope for this safety fix.

## (6) Finding 3 fast-forward safety ✅

`sync_scp_slot_to_chain` (`service.rs:951`): solo-mode short-circuit; only advances when `current < target` (forward-only, never rewinds); refuses if the slot holds any in-flight nominate/ballot/commit state (`num_voted/accepted/confirmed_nominated`, `bN > 0`, `phase != NominatePrepare`). Correct.

## (7) Gating #396 + the separate drift bug ✅ acceptable

The safety fork is fixed and tested; #396's full 3-of-3 soak is gated (`&& false`) on a **separate pre-existing SCP-slot/block-height drift** liveness bug, which is correctly out of scope for #420. I filed **#421** to track that drift bug (no follow-up existed). Judging #420 on whether it correctly + safely fixes the fork: it does.

## Minor (non-blocking) — please clean up before/at merge

- The head commit adds **`.loom-managed`** to the repo — a local Loom worktree marker that should not be tracked. It's an inert comment file (no code/build/safety impact), so I'm not blocking on it, but it should be removed (and ideally gitignored). Flagging so it doesn't slip into `main`.

## Test Execution
- SCP suite (full, `--features test_utils`): ✅ 100 lib + all integration green
- `cargo test -p botho --lib`: ✅ 963 passed
- New SCP fork tests: ✅ pass; **fail-pre verified** by temporary tip-dependent revert
- Finding 3 + botho fork tests: ✅ pass
- CI on head `eeeb06f`: ✅ all checks pass (`cargo build --workspace (Linux)`, Compile/Crypto/Node/Transaction Benchmarks, Fuzz Build Guard, rustfmt)

PR #422, opened 2026-06-18
fix(consensus): keep SCP slot aligned with block height (height filter + gated rewind)

### LINKED ISSUE #421: consensus: SCP current_slot_index drifts ahead of block height (blocks #396 3-of-3 soak)
## Summary

Discovered/confirmed during the Judge review of PR #420 (the #417 Finding 1 tip-agnostic-validity safety fix). PR #420 correctly fixes the multi-minter SCP **safety fork** and is being approved on that basis. This issue tracks the **separate, pre-existing liveness bug** that PR #420 explicitly leaves out of scope.

## The bug

On an established minter, the SCP `current_slot_index` can run **ahead** of the node's ledger block height. Mechanism (as documented in PR #420's "Known limitation" and the `mine-after-join-live-node.test.ts` gating rationale):

- A stale/duplicate-height minting value can be externalized by SCP and then **rejected at block-apply** (`LedgerStore::add_block` enforces height/prev-hash/difficulty/reward/timestamp unconditionally — this is the safety property the fork fix relies on).
- SCP auto-advances its slot on externalize, but the ledger height does **not** advance when apply rejects the block. So the SCP slot index drifts forward of the block height.
- A joiner that catches up via block-sync fast-forwards its SCP slot to `height + 1` (PR #420 Finding 3 `sync_scp_slot_to_chain`), but the established nodes are now on a **higher, drifted** slot, so the two sides discard each other's messages as wrong-slot and cannot converge.

## Impact

- **Severity:** liveness, not safety. The fork fix in #420 holds (no divergent block can ever be appended); this is about the network failing to *make progress*/converge in the joiner-mines scenario.
- **Blocks:** the #396 3-of-3 end-to-end soak ("fresh node joins, participates, mines an accepted block"). That test is currently gated (`&& false`) in `web/packages/wasm-signer/test/mine-after-join-live-node.test.ts` with a rationale pointing here.

## Suggested direction

Re-align the SCP slot index with the ledger block height so they cannot diverge:
- Option A: when an externalized value is rejected at apply, **do not** leave the SCP slot advanced past the ledger height — rewind/resync the slot to `height + 1` (re-using the guarded `sync_scp_slot_to_chain` forward-only path will not help here because it only moves forward; a controlled re-alignment is needed).
- Option B: gate slot auto-advance on successful block-apply rather than on externalize.

This is a deeper protocol change and should be designed by the Architect.

## Acceptance

- Established minters' SCP slot index stays aligned with block height under stale/duplicate-height externalize+reject.
- A joiner that block-syncs to height N can rejoin consensus and the 3-of-3 net converges and advances past N.
- Re-enable the #396 soak (drop `&& false`) and it passes.

Discovered while evaluating PR #420. Refs #417 (Finding 3 area), #396, #419.

rjwalters:
## Summary

Fixes the issue #421 liveness bug: the SCP `current_slot_index` could drift
ahead of the ledger block height, wedging a freshly-synced joiner against the
established minters (each side discarded the other's messages as wrong-slot).

The fix is **proposal-side**, not validity-side — PR #420's tip-agnostic
`validity_fn` is left completely untouched, so the no-fork safety guarantee is
preserved. Implements the architect's recommendation: a **primary** proposal
height filter plus the maintainer-chosen **gated-rewind backstop**.

Unblocks the SCP-slot-drift half of #396 (the remaining #396 blocker is a
separate, pre-existing catch-up-sync gap — see "Real-process validation").

## Changes

- **PRIMARY — proposal-side height filter** (`botho/src/consensus/service.rs`,
  `propose_pending_values`): only nominate a minting value whose coinbase builds
  the immediate next block (`block_height == tip_height + 1` AND
  `prev_block_hash == tip_hash`, read from the cached `MintingTx` + local chain
  state via the new `minting_value_matches_tip`). Honest proposers never
  nominate an already-taken/future-height coinbase — the doomed value whose
  externalize-then-reject created the drift — so `slot ≡ height` holds by
  construction. The `validity_fn` is **not** changed, so peer values still enter
  federated voting and #420's no-fork property is untouched.
- **BACKSTOP — gated backward slot re-alignment**:
  - `consensus/scp/src/node/node_trait.rs` + `node_impl.rs` + `scp_log.rs`: new
    `ScpNode::realign_slot_to(index)`, a backward-capable counterpart to the
    forward-only `reset_slot_index` (which keeps its `debug_assert!(>)`). It
    refuses forward moves and documents that the caller must guarantee the
    target was never finalized and the slot is idle.
  - `botho/src/consensus/service.rs`: `realign_scp_slot_to_chain(height)`, gated
    by (1) slot strictly ahead, (2) target strictly above the **authoritative
    local ledger height** (never re-open a finalized/committed index → no fork),
    (3) current slot idle (reuses the `sync_scp_slot_to_chain` activity guard).
  - `botho/src/commands/run.rs`: on a rejected `add_block_from_network`, invoke
    `realign_scp_slot_to_chain` so a slipped drift self-heals instead of wedging.
- **Regression tests** (`botho/src/consensus/service.rs`, reusing the existing
  `intrinsic_valid_minting_tx` / `submit_and_register` / `run_two_services`
  scaffolding): drift characterization; joiner non-convergence; the height
  filter skipping doomed coinbases (with a deterministic fail-pre guard); the
  backstop self-heal + its no-reopen refusal.
- **#396 soak** (`web/.../mine-after-join-live-node.test.ts`): gating comment
  rewritten to record the drift fix AND the newly-found, separate catch-up-sync
  blocker; the test stays gated until that is resolved (see below).

| Established minters' SCP slot stays aligned under stale/duplicate externalize+reject | ✅ | In-process tests + real 2-node run: A externalized slots 1..N one-per-slot, no "Failed to add", no drift; backstop test self-heals an injected drift |
| Joiner block-syncs to N, rejoins, 3-of-3 converges past N | ⚠️ partial | SCP slot fast-forward + 2-node convergence verified; full 3-of-3 advance is blocked by a SEPARATE pre-existing catch-up-sync gap (reproduces on clean #420) |
| Re-enable #396 soak and it passes | ❌ kept gated | Re-enabling would fail on the catch-up-sync gap above, not the drift; gating comment updated to point at it |
| No #420 no-fork regression | ✅ | `validity_fn` untouched; all 100 `bth-consensus-scp` tests pass incl. the #420 fork tests; real 2-node A+B produce identical tip hashes at every height |

## Real-process validation (release `botho run`, loopback, 1s slots)

- **2-node A+B**: converge via real SCP, **identical block hashes at every
  height**, zero solo-mode externalizes — no fork. Confirmed identical on a
  clean #420 (e0226ff) build, so this PR introduces no regression.
- **3-node joiner**: A+B stay converged (identical tips) but the fresh joiner C
  never catches up 0→N — it only receives the live compact block at the tip and
  cannot apply it across the gap ("Expected height 1, got 9"). This is a
  **pre-existing catch-up block-sync gap**: it reproduces **identically on a
  clean #420 binary**, so it is not caused by the drift fix. The `slot == height
  + 1` fast-forward only helps a joiner that has already block-synced; here the
  block-sync itself never runs. Tracked separately; the #396 soak stays gated on
  it.

Per the issue's STOP guardrail, a correct, well-tested partial fix is shipped
rather than forcing an end-to-end pass that depends on an unrelated blocker.

## Test Plan

- `cargo test -p bth-consensus-scp` → 100 passed (incl. #420 fork tests).
- `cargo test -p botho --lib` → 968 passed (incl. 5 new #421 tests).
- Fail-pre/pass-post verified: with the height filter removed,
  `proposal_height_filter_proposes_nothing_when_all_doomed` FAILS; with it,
  PASSES.
- `cargo build`, `cargo fmt`, `cargo clippy` clean on the changed crates.

Closes #421
Refs #420, #396, #417

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -1185,6 +1185,36 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                     // Add to ledger
                                     if let Err(e) = node.add_block_from_network(&block) {
                                         warn!("Failed to add consensus block: {}", e);
+
+                                        // BACKSTOP (issue #421): a doomed/superseded-height
+                                        // coinbase can be externalized by SCP (the validity_fn
+                                        // is tip-agnostic for #420's no-fork safety) and then
+                                        // REJECTED here at block-apply. SCP already auto-advanced
+                                        // its slot on externalize, but the ledger height did not
+                                        // move, so the SCP slot now drifts ahead of the block
+                                        // height and the net can no longer converge with a
+                                        // freshly-synced joiner. Re-align the SCP slot back down
+                                        // to ledger_height + 1. This is gated to never re-open an
+                                        // already-finalized index (see
+                                        // ConsensusService::realign_scp_slot_to_chain), so #420's
+                                        // no-fork guarantee is preserved. With the proposal-side
+                                        // height filter in place this should essentially never
+                                        // fire; it self-heals a slipped race rather than wedging.
+                                        let ledger_height = node
+                                            .shared_ledger()
+                                            .read()
+                                            .ok()
+                                            .and_then(|ledger| ledger.get_chain_state().ok())
+                                            .map(|state| state.height);
+                                        if let Some(h) = ledger_height {
+                                            if consensus.realign_scp_slot_to_chain(h) {
+                                                warn!(
+                                                    ledger_height = h,
+                                                    "Re-aligned drifted SCP slot to ledger height \
+                                                     after rejected block-apply (issue #421)"
+                                                );
+                                            }
+                                        }
                                     } else {
                                         // Record for dynamic timing
                                         consensus.record_block(block.header.timestamp, block.transactions.len());
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -317,6 +317,58 @@ impl ConsensusService {
         }
     }
 
+    /// The (height, tip_hash) of the local chain tip as last observed via
+    /// [`update_chain_state`]. Used by the proposal-side height filter and the
+    /// gated-rewind backstop. Returns `None` only if the shared-state lock is
+    /// poisoned.
+    fn chain_tip(&self) -> Option<(u64, [u8; 32])> {
+        self.shared_state
+            .read()
+            .ok()
+            .map(|state| (state.chain_state.height, state.chain_state.tip_hash))
+    }
+
+    /// Decide whether a pending minting value is safe to *nominate* given the
+    /// local chain tip: its coinbase must build the immediate next block, i.e.
+    /// `block_height == tip_height + 1` AND `prev_block_hash == tip_hash`.
+    ///
+    /// LIVENESS (issue #421): with the tip-agnostic SCP `validity_fn` (#420),
+    /// a coinbase at an already-taken or future height is a perfectly valid SCP
+    /// *value* and can be nominated, externalized, then REJECTED at block-apply
+    /// (`LedgerStore::add_block` enforces height/prev-hash unconditionally —
+    /// that is the no-fork safety property and MUST stay). SCP auto-advances
+    /// its slot on externalize but the ledger height does not move on a
+    /// rejected apply, so the SCP slot drifts ahead of the block height and the
+    /// net can no longer converge with a freshly-synced joiner.
+    ///
+    /// This filter is PROPOSAL-side only: it changes only which value *this*
+    /// node nominates. It does NOT touch the `validity_fn`, so a peer's value
+    /// is never dropped as invalid and #420's no-fork guarantee is preserved.
+    /// Honest proposers therefore only ever nominate the next-height coinbase,
+    /// so the externalized value applies cleanly and `slot ≡ height` is kept by
+    /// construction.
+    fn minting_value_matches_tip(&self, value: &ConsensusValue) -> bool {
+        let Some((tip_height, tip_hash)) = self.chain_tip() else {
+            // Lock poisoned: cannot verify, so do not nominate.
+            return false;
+        };
+
+        let Ok(state) = self.shared_state.read() else {
+            return false;
+        };
+        let Some(entry) = state.tx_cache.get(&value.tx_hash) else {
+            return false;
+        };
+        if !entry.is_minting_tx {
+            return false;
+        }
+
+        match bincode::deserialize::<crate::block::MintingTx>(&entry.data) {
+            Ok(mtx) => mtx.block_height == tip_height + 1 && mtx.prev_block_hash == tip_hash,
+            Err(_) => false,
+        }
+    }
+
     /// Get the current quorum set (for status/RPC/test assertions).
     pub fn quorum_set(&self) -> &QuorumSet {
         &self.quorum_set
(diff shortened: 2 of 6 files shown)


rjwalters:
✅ **Approved!** — high-rigor review of safety-sensitive consensus code (Closes #421).

## Critical safety checks

**1. #420 no-fork NOT regressed [CRITICAL] — PASS.**
The `validity_fn` is genuinely untouched. The only `validity_fn` references in the diff are doc comments plus one `self.validity_fn.clone()` inside the new `realign_slot_to` (it reuses the *existing* fn when rebuilding a `Slot` — it does not define a new validity policy). The height filter (`minting_value_matches_tip`) is PROPOSAL-side only: applied in `propose_pending_values` on both the new-best (`pending_values`) and re-propose-pinned (`proposed_values`) branches, and there is exactly one `scp_node.propose_values(...)` call site (service.rs:795). Transfer values are unaffected (correct — they don't drive height). The #420 fork regression suite passes: `no_fork_with_tip_agnostic_validity` and `fork_with_tip_dependent_validity` both green (`cargo test -p bth-consensus-scp --features test_utils --test test_tip_dependent_validity_fork`).

**2. Gated-rewind backstop cannot re-open a committed index [CRITICAL] — PASS.**
`realign_scp_slot_to_chain` Gate 2 reads the authoritative `ledger_height` from `self.chain_tip()` (chain state) — NOT the caller-supplied `chain_height` — and refuses when `target_slot <= ledger_height`. The re-seat target is therefore always strictly above the finalized frontier (`ledger_height + 1`), so no index that produced a committed block can be re-opened. The caller in run.rs reads ledger height from the authoritative ledger store. Adversarially: even if a doomed index in `(ledger_height, current)` were re-balloted, block-apply remains the unconditional final arbiter (the #420 tip-consistency check still rejects anything not building on the real tip), so no divergent *finalized* block can result. The backstop is a pure liveness mechanism; safety stays enforced downstream.

**3. `realign_slot_to` primitive safety — PASS.**
Backward-capable but refuses forward moves (`slot_index > current` → returns `false`), keeping `reset_slot_index` (forward-only `debug_assert`) as the single auditable advance path. The no-reopen invariant is explicitly delegated to the caller and documented in the trait. Trait + node_impl + LoggingScpNode wrapper are consistent.

**4. Tests fail-pre / pass-post — VERIFIED.**
By temporary revert (removing the `minting_value_matches_tip` filter from both proposal branches), `proposal_height_filter_proposes_nothing_when_all_doomed` FAILS with the exact assertion the builder documented as the deterministic fail-pre guard; restored cleanly afterward. TEST 4 (`backstop_realigns_drift_and_refuses_to_reopen_finalized_index`) asserts both the self-heal and the no-reopen refusal (stale-height target onto a finalized index is refused, slot unchanged).

**5. No regression — PASS.** `cargo test -p bth-consensus-scp` → 100 passed; `cargo test -p botho --lib` → 968 passed.

## Other checks

- **#396 gating — correct.** The 3-of-3 soak stays gated (`&& false`) on a SEPARATE, pre-existing catch-up-block-sync gap (fresh joiner never pulls 0→N; reproduces on clean #420). This is out of scope for #421, which correctly fixes the drift on its own merits. Filed follow-up **#423** to track the catch-up-sync gap.
- **Clean diff:** 6 files, no `.loom-managed`/marker/stray files.
- **CI:** all checks pass (head `e130f2e`); merge state CLEAN. Note: this PR was branched from e0226ff (pre-#402), so the `cargo build --workspace (Linux)` signal added in #402 is not in this run's rollup; the maintainer's rebase-on-merge will pick it up, and the affected crates build/test cleanly locally.

Excellent, carefully-scoped fix with thorough documentation of the safety invariants. Approving.

Issue #424, opened 2026-06-18
Two-minter A+B pair deterministically stalls at height 9 (never externalizes height 10) over loopback


rjwalters:
## Summary

Uncovered while validating the #423 catch-up fix by re-enabling the #396 soak
(`web/packages/wasm-signer/test/mine-after-join-live-node.test.ts`). With the
#423 catch-up trigger fixed, the soak now runs past its previous blocker and
reaches Step 1, where it surfaces a SEPARATE, distinct defect:

**A two-minter pair (A + B, explicit 2-of-2 quorum, `min_peers = 1`, each
minting) deterministically stalls at height 9 — it never externalizes height
10.** The shared chain is stuck at 9, so the test fails at Step 1 with:

```
A+B only mined a shared height of 9/24
```

This happens BEFORE the third node C ever joins, so it is unrelated to the
catch-up sync trigger fixed in #423 (A+B mine in lockstep from genesis and never
invoke catch-up).

## Repro

```
cd web
BOTHO_E2E_NODE=1 BOTHO_BIN=$(pwd)/../target/release/botho \
  pnpm exec vitest run packages/wasm-signer/test/mine-after-join-live-node.test.ts
```
(temporarily drop the `&& false` in the `const enabled = ...` gate). Observed on
two consecutive runs; deterministic.

## Symptom (node logs, RUST_LOG=info)

Both A and B repeatedly log, without ever externalizing:

```
INFO Submitting minting tx to consensus height=10 priority=...
INFO Minting transaction submitted for consensus value=ConsensusValue { tx_hash: "...", is_minting_tx: true, priority: ... }
```

i.e. each node keeps proposing its own height-10 coinbase, but the 2-of-2 quorum
never converges on a single height-10 value, so height 10 is never externalized
and the tip stays at 9.

## Likely area

Multi-minter SCP liveness: competing same-height coinbase nominations from two
equal minters failing to converge to a single externalized value. This is in the
same family as #414 / #416 / #420 / #422 (multi-node SCP convergence), but those
fixed FORK / slot-drift behavior; here the failure mode is a LIVENESS stall (no
progress past height 9), not a fork. Note the prior #396 gate comment claimed
"2-node A+B convergence (no fork, identical tips at every height) ... verified" —
that no longer holds in this configuration; A+B do not progress past 9.

## Impact

- Blocks the #396 3-of-3 mine-after-join soak (which is otherwise unblocked on
  the catch-up side by #423).
[...]

PR #425, opened 2026-06-18
fix(sync): trigger catch-up download for any gap > 1, not only > 10 (#423)

### LINKED ISSUE #423: Fresh joiner catch-up block-sync never fires in local-loopback config (0→N gap)
## Summary

Discovered while validating #421 (SCP slot/height drift) on real `botho run` processes over loopback. A fresh joiner C connects to a running A+B pair (peerCount >= 1) but never block-syncs 0 → N: it only receives the live compact block at the current tip (e.g. height 9), which it cannot apply across the 0..9 gap:

```
Failed to add reconstructed block: Expected height 1, got 9
```

C stays at height 0, so the 3-of-3 net cannot advance. The `slot == height + 1` SCP fast-forward only helps a joiner that has ALREADY block-synced — here the historical block-sync (catch-up download) itself never runs.

## Not a #421/#420 regression

This reproduces IDENTICALLY on a clean #420 build (binary at e0226ff), so it is a SEPARATE, pre-existing blocker — independent of the SCP-slot-drift fix shipped in #421/#422. The 2-node A+B convergence (no fork, identical tips at every height) and the joiner SCP fast-forward are verified; the missing piece is the catch-up download.

## Impact

- Blocks the #396 3-of-3 mine-after-join soak (`web/packages/wasm-signer/test/mine-after-join-live-node.test.ts`), which remains gated (`&& false`) pending this fix.
- A fresh node joining an already-advanced chain cannot catch up in the local-loopback configuration.

---

## Curator Enhancement (2026-06-17)

Investigated the #376 catch-up path on `origin/main` (includes #420 + #422). The plumbing is fully wired — the bug is in the **trigger condition**, not in the request/response transport. There are two distinct, compounding defects; the first fully explains the height-9 repro.

### Root cause 1 (PRIMARY — explains the height-9 repro): the catch-up trigger is gated behind `SYNC_BEHIND_THRESHOLD`, which suppresses the *initial* 0→N download for any tip within 10 of genesis

The catch-up trigger fires only when a peer is **more than `SYNC_BEHIND_THRESHOLD` (= 10)** blocks ahead. See the constant at `botho/src/network/sync.rs:49` and the three gate sites that all use `status.height > self.local_height + SYNC_BEHIND_THRESHOLD`:

- `ChainSyncManager::on_status` — `botho/src/network/sync.rs:436-447`
- `ChainSyncManager::tick`, Discovery arm — `botho/src/network/sync.rs:612-625`
- `ChainSyncManager::tick`, Synced arm — `botho/src/network/sync.rs:651-660`

For the repro, the fresh joiner C is at `local_height = 0` and the network tip is height 9. The gate evaluates:

```
status.height (9) > local_height (0) + SYNC_BEHIND_THRESHOLD (10)   ->   9 > 10   ->   FALSE
```

So in the Discovery arm (`sync.rs:612`), C falls into the `else` branch and transitions straight to `SyncState::Synced` (`sync.rs:622`) **without ever entering `Downloading`**. The historical download never runs. C then receives only the gossiped compact tip (height 9), which fails the contiguous-height check (`Expected height 1, got 9`), and stays at height 0 forever.

The threshold exists to suppress thrash when a node is only a *few blocks behind near the tip* and the gap will be closed by normal gossip of contiguous blocks. But that reasoning does **not** hold for a node behind by a gap that gossip cannot bridge: a gossiped compact/full block only applies if it is exactly `local_height + 1` (contiguous). Any gap > 1 — including the entire 0→9 initial download — must go through the sync state machine. Gating the *initial* catch-up behind a 10-block "near-tip lag" threshold is the defect.

**Fix direction:** the trigger must distinguish "near-tip lag that gossip will close" from "a gap gossip cannot bridge." Concretely, a node should enter `Downloading` whenever `status.height > local_height + 1` (i.e. any peer ahead by more than one block — gossip can only deliver the next contiguous block). The `SYNC_BEHIND_THRESHOLD` may still be appropriate as a *re-sync* hysteresis for an already-synced node lagging slightly, but it must not gate the *initial* download out of Discovery. Recommended approach: introduce a separate, tighter trigger for the initial/behind case (e.g. `INITIAL_SYNC_GAP = 1`) used by the Discovery arm and the compact-block fallback (root cause 2), keeping `SYNC_BEHIND_THRESHOLD` only as Synced-state hysteresis. Builder should pick the cleanest factoring; the load-bearing requirement is that **a node behind by any gap > 1 must initiate sync.**

### Root cause 2 (LATENT — masked today by RC1, but must also be fixed): the failed compact-block apply never pokes the sync manager

When C receives the gossiped tip it can't apply, the compact-block handler logs and drops it with no sync trigger:

- `botho/src/commands/run.rs:929-931` — `if let Err(e) = node.add_block_from_network(&block) { warn!("Failed to add reconstructed block: {}", e); }` — the error arm does nothing else. This is the exact log line in the repro.

There is no fallback that says "I just received a gossiped block at height N >> my height; I'm behind → request status / enter catch-up." The only real-path sources of `on_status` are (a) the Discovery `RequestStatus` round-trip and (b) the Synced-state `STATUS_REFRESH_INTERVAL` re-poll (`sync.rs:53`, 30s). With RC1 fixed, the Discovery round-trip alone is sufficient for the join case, so RC2 is not strictly required to fix the repro — but it is a real robustness gap: a node that is gossiped a far-ahead tip should treat that as a height advertisement and (re-)enter catch-up promptly rather than waiting up to 30s for the next status refresh. Recommended: in the `add_block_from_network` error arm (and/or the early "already have / too far ahead" check at `run.rs:905-911`), when the gossiped height exceeds `local_height + 1`, call `sync_manager.on_status(peer, height, tip_hash)` (or an equivalent "observed higher tip" hint) so the manager re-evaluates the gate immediately.

### Why the `chain_sync_catchup_integration` test passes but the real joiner does not

`botho/tests/chain_sync_catchup_integration.rs::test_fresh_node_syncs_existing_chain_to_tip` uses `target_height = RING_SIZE + 75 = 95` (`chain_sync_catchup_integration.rs:262`). With local height 0 and peer at 95, the gate `95 > 0 + 10` is **TRUE**, so the test never exercises the small-gap (`N <= 10`) regime that the repro hits. The test height was chosen to be "well past ring size," which incidentally is also well past the threshold, so the threshold bug is invisible to it.

Additionally, `run_catchup` (`chain_sync_catchup_integration.rs:161-245`) contains a fallback the **real node does not have**: when `tick()` returns `None` while still behind, the test manually injects `sync_manager.on_status(...)` (`chain_sync_catchup_integration.rs:198`) to re-arm the manager. This silently papers over root cause 2 — the real run loop has no equivalent "I'm behind but got no action, so feed myself a fresh status" path. A faithful regression test must NOT inject status out of band on the behind-but-idle path; it should rely only on the triggers the production loop actually has.

### Real-path plumbing (verified correct — not the bug)

For the builder's confidence, the rest of the path is sound and should not be touched:
- Peer table is populated for both dialer and listener on `ConnectionEstablished` — `botho/src/network/discovery.rs:1299-1315`.
- `get_connected_peers` reads the peer table and feeds `sync_manager.tick(&connected)` every 2s — `botho/src/commands/run.rs:60-61`, `551`, `1321-1322`.
- `RequestStatus` / `RequestBlocks` actions are dispatched as real `SyncRequest` messages — `botho/src/commands/run.rs:1324-1342`.
- `GetStatus` is served from the ledger — `botho/src/commands/run.rs:790-796`; `GetBlocks` serves up to 100 blocks/batch — `run.rs:797-808`.
- `SyncResponse::Status` feeds `sync_manager.on_status` — `run.rs:883-886`; `SyncResponse::Blocks` feeds `on_blocks` and applies blocks, then `sync_scp_slot_to_chain` fast-forwards SCP — `run.rs:818-882`.

## Implementation Guidance

1. Change the **initial / Discovery** catch-up trigger so a fresh or behind node enters `Downloading` whenever a peer is ahead by more than one block (gossip can only ever deliver the next contiguous block). Do not gate the initial download behind the 10-block near-tip threshold. (`botho/src/network/sync.rs` — `on_status` ~436, Discovery arm of `tick` ~612.)
2. Decide the role of `SYNC_BEHIND_THRESHOLD`: keep it as hysteresis only for the **Synced → re-Downloading** transition (so a synced node doesn't thrash on 1-2 block lag closed by gossip), or remove it in favor of the gap-1 rule with a small debounce. Whatever the choice, ensure the **Synced arm** (`sync.rs:651`) still re-enters catch-up for any real gap.
3. Add a compact-block fallback (root cause 2): in `botho/src/commands/run.rs` (~905-911 early check and/or ~929-931 error arm), when a gossiped block's height exceeds `local_height + 1`, hand the observed `(peer, height, tip_hash)` to the sync manager so it re-evaluates immediately instead of waiting for the 30s status refresh.
4. Re-tune / add tests (see Test Plan) so the small-gap initial-join regime is covered and the integration test no longer masks RC2 via its out-of-band `on_status` injection.

## Affected Files

- `botho/src/network/sync.rs` - Fix the catch-up trigger condition (`SYNC_BEHIND_THRESHOLD` gating in `on_status` ~436, `tick` Discovery arm ~612, Synced arm ~651). Introduce a gap-1 initial-sync trigger; restrict `SYNC_BEHIND_THRESHOLD` to Synced-state hysteresis (or remove with debounce).
- `botho/src/commands/run.rs` - Add a "behind on gossiped tip" fallback in the compact-block handler (~905-911 / ~929-931) that pokes the sync manager with the observed higher tip so initial/lagging catch-up fires promptly.
- `botho/tests/chain_sync_catchup_integration.rs` - Add a small-gap (e.g. tip = 9, well under the old threshold) fresh-join regression test; tighten `run_catchup` so it does not inject `on_status` out of band on the behind-but-idle path (so the test exercises the real trigger, not the test-only fallback). Keep existing tests green.
- `web/packages/wasm-signer/test/mine-after-join-live-node.test.ts` - Re-enable the #396 soak by dropping `&& false` from `const enabled = wasmBuilt && RUN_LOCAL_NODE && false` (line ~135) once the joiner catch-up works.

## Acceptance Criteria

- [ ] A fresh joiner at height 0 connecting to peers at height N (including small N, e.g. N = 9) triggers the historical catch-up download and syncs 0→N in the local-loopback configuration — not just the in-process integration test.
- [ ] The catch-up trigger is robust to gap size: a node behind by any gap > 1 (not only gaps > 10) initiates sync from a peer. The 10-block threshold no longer suppresses the initial download.
- [ ] A node only ~1 block behind near the tip still does NOT thrash into a redundant download (the legitimate purpose of the old threshold is preserved, as hysteresis on the Synced→Downloading transition).
- [ ] Receiving a gossiped tip block that cannot be applied across a gap promptly (re)enters catch-up rather than relying solely on the 30s status refresh.
- [ ] The #396 soak (`mine-after-join-live-node.test.ts`, with `&& false` removed) passes: the 3-of-3 net converges, C catches up to A/B's tip and mines an accepted block on matching block hashes.
- [ ] No regression to existing `chain_sync_catchup_integration` tests.

## Test Plan

- [ ] Manual verification: run 3 `botho run` processes over loopback (A+B mine to height ~9, then start C). Confirm C logs entering `Downloading`, requests `GetBlocks [1..N]`, applies historical blocks, reaches A/B's tip, and the net advances past N.
- [ ] Automated (unit, trigger condition): add a `ChainSyncManager` unit test asserting that `on_status(peer, 9, ..)` from `local_height = 0` transitions to `Downloading { target_height: 9 }` (currently it goes to `Synced`). Add a boundary test for gap = 1 (no download) vs gap = 2 (download).
- [ ] Automated (integration): add a small-gap variant of `test_fresh_node_syncs_existing_chain_to_tip` (tip = 9) in `chain_sync_catchup_integration.rs`, and remove the out-of-band `on_status` injection from `run_catchup` so the test exercises the real trigger. Confirm both small-gap and large-gap fresh-join cases converge.
- [ ] Automated (soak / #396): drop `&& false` in `mine-after-join-live-node.test.ts` and confirm the 3-of-3 mine-after-join soak passes.
- [ ] Edge cases: tip exactly at the old threshold boundary (N = 10, N = 11); a joiner that connects, syncs to N, then the chain advances to N+1 (the Synced→re-Downloading hysteresis path); a joiner that only ever receives a gossiped tip and no status response.

## Related

- Blocks #396 (3-of-3 mine-after-join soak — actively pursued by the maintainer).
- #376 (catch-up sync implementation — the path with the defective trigger).
- #412 (`SyncCodec`/`SyncRequest`/`SyncResponse` transport — verified correct).
- Refs #420, #421, #422 (confirmed NOT the cause; reproduces on a clean #420 build).

rjwalters:
## What

Fixes the fresh-joiner catch-up block-sync that never fired in the local-loopback
config (0 -> N gap). A fresh node at height 0 against a tip within 10 of genesis
(e.g. 9) used to stall at 0 forever because the catch-up trigger was gated behind
`SYNC_BEHIND_THRESHOLD = 10`.

Closes #423. Unblocks the catch-up half of #396 (see "#396 status" below).

## Root causes (verified against the Curator's analysis)

**RC1 (primary):** the catch-up trigger fired only when a peer was more than
`SYNC_BEHIND_THRESHOLD` (10) blocks ahead. A fresh joiner at height 0 vs tip 9
evaluated `9 > 0 + 10` = FALSE, so the Discovery arm jumped straight to `Synced`
and never entered `Downloading`. Gossip only delivers the next contiguous block,
so any gap >= 2 must go through the sync state machine.

**RC2 (latent):** a gossiped compact block more than one block ahead failed the
contiguous-height check (`Expected height 1, got 9`) and the handler just logged
the error — it never poked the sync manager.

## Fix

- Introduce `SYNC_INITIAL_GAP = 1`. During initial discovery / recovery
  (`on_status` when not yet Synced, and the Discovery arm of `tick`), enter
  `Downloading` whenever a peer is ahead by more than one block (gap >= 2). A
  1-block lag is left to gossip so it does not thrash.
- Keep `SYNC_BEHIND_THRESHOLD` strictly as hysteresis for the already-`Synced`
  -> `Downloading` transition (`on_status` when Synced, and the Synced arm of
  `tick`), preserving the legitimate anti-thrash purpose of the old threshold.
- RC2: add `ChainSyncManager::note_gossiped_tip(connected, height, tip_hash)`
  and call it from the compact-block handler in `run.rs`. When a gossiped tip is
  more than one block ahead, it (re)enters catch-up immediately (gap-1 rule, even
  from Synced) instead of waiting up to 30s for the next status refresh.

## Trigger condition summary

| State | Trigger gap | Rationale |
|-------|-------------|-----------|
| Discovery / Failed (initial join) | `> local + 1` (gap >= 2) | gossip can't bridge gaps >= 2; must sync |
| Synced (re-poll drift) | `> local + 10` | hysteresis; gossip closes small near-tip lags |
| Gossiped far-ahead tip (any state) | `> local + 1` (gap >= 2) | direct evidence of a real gap |

## Regression test (fail-pre / pass-post)

`test_fresh_node_syncs_small_gap_chain_discovery_only` builds a chain to tip = 9
(under the old threshold) and drives catch-up using ONLY the Discovery status
round-trip — no out-of-band `on_status` injection. Verified it FAILS pre-fix
(node loops, never leaves height 0 -> panics on non-convergence) and PASSES
post-fix. Also:
- `run_catchup` no longer injects `on_status` out of band on the behind-but-idle
  path; it now models the real production trigger (`note_gossiped_tip`), so the
  integration test stops masking RC2.
- Added unit/boundary tests for gaps 0/1/2/9 and the gossiped-tip fallback.
- Updated pre-existing threshold-encoded assertions to the gap-1 semantics.

## Real-process verification (release binary, loopback)

- `join-consensus-live-node.test.ts` (enabled, green): A solo-mines to **N=24**,
  fresh **B catches up 0 -> 26** onto A's exact chain (identical hashes at
  [1, 12, 23, 24]).
- A throwaway small-N run (not committed): A solo-mines to **N=8**, **B catches
  up 0 -> 8** — the exact pre-fix repro regime (`8 > 0 + 10` = FALSE), now works.

## Tests / build

- `cargo test -p botho`: all green (970 lib + all integration). The only failures
  in the full run are two PRE-EXISTING doctests in `network/privacy/` (`RelayCapacity::measure` does not exist) — untouched by this PR.
- `cargo build --release --bin botho`, `cargo fmt`, lib clippy: clean (no new
  warnings/errors; the pre-existing `println!` clippy lints in
  `transport/fingerprint.rs` are unrelated).
- `pnpm -C web typecheck`: green.

## #396 status (capstone)

The catch-up half of #396 is fixed and proven over real processes. Re-enabling
the 3-of-3 soak surfaced a SEPARATE, distinct blocker: the two-minter A+B pair
**deterministically stalls at height 9** (never externalizes height 10), before
C ever joins — a multi-minter SCP liveness gap independent of the catch-up
trigger. Per the STOP guardrail I did not force a flaky pass: the #423 fix is
complete and independently verified, and the new blocker is filed as **#424**.
The `mine-after-join-live-node.test.ts` gate comment is updated to point at #424
and the soak stays gated until two minters can advance past height 9.

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -901,13 +901,41 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                 "Received compact block"
                             );
 
-                            // Check if we already have this block
+                            // Check if we already have this block, and detect
+                            // whether this gossiped tip is ahead of us by a gap
+                            // gossip cannot bridge.
+                            let mut local_height_for_gossip: Option<u64> = None;
                             if let Ok(ledger) = node.shared_ledger().read() {
                                 if let Ok(state) = ledger.get_chain_state() {
                                     if state.height >= height {
                                         debug!("Already have block {}, ignoring compact block", height);
                                         continue;
                                     }
+                                    local_height_for_gossip = Some(state.height);
+                                }
+                            }
+
+                            // RC2 fallback (issue #423): a gossiped block more
+                            // than one block ahead of us (height > local + 1)
+                            // can never be applied contiguously — it must be
+                            // backfilled via catch-up. Keep the sync manager's
+                            // local height current and poke it with the observed
+                            // tip so it (re)enters Downloading immediately rather
+                            // than waiting up to 30s for the next status refresh.
+                            // The reconstruction/apply below will fail with
+                            // "Expected height L+1, got H"; this is the path that
+                            // turns that failure into a real catch-up trigger.
+                            if let Some(local_h) = local_height_for_gossip {
+                                if height > local_h + 1 {
+                                    sync_manager.set_local_height(local_h);
+                                    let connected = get_connected_peers(&discovery);
+                                    info!(
+                                        height = height,
+                                        local_height = local_h,
+                                        peers = connected.len(),
+                                        "Gossiped tip is ahead by a gap gossip can't bridge; triggering catch-up"
+                                    );
+                                    sync_manager.note_gossiped_tip(&connected, height, block_hash);
                                 }
                             }
 
@@ -929,7 +957,13 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
 
                                     // Add to ledger
                                     if let Err(e) = node.add_block_from_network(&block) {
-                                        warn!("Failed to add reconstructed block: {}", e);
+                                        // A non-contiguous gossiped block (height
+                                        // > local + 1) lands here ("Expected
+                                        // height L+1, got H"). The catch-up
+                                        // trigger was already poked above
+                                        // (issue #423 RC2), so the sync state
+                                        // machine will backfill the gap.
+                                        warn!("Failed to add reconstructed block: {} (catch-up handles non-contiguous gaps)", e);
                                     } else {
                                         // Record for dynamic timing
                                         consensus.record_block(block.header.timestamp, block.transactions.len());
--- a/botho/src/network/sync.rs
+++ b/botho/src/network/sync.rs
@@ -45,8 +45,28 @@ pub const BLOCKS_PER_REQUEST: u32 = 100;
 /// Request timeout duration
 pub const REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 
-/// Blocks behind threshold before re-syncing
+/// Blocks behind threshold before a *Synced* node re-enters catch-up.
+///
+/// This is hysteresis for the `Synced -> Downloading` transition only: when a
+/// node is already caught up and falls a *few* blocks behind near the tip, the
+/// gap is normally closed by gossip of contiguous blocks, so we don't want to
+/// thrash into a redundant historical download for every 1-2 block lag. It is
+/// deliberately NOT used to gate the *initial* catch-up download — see
+/// [`SYNC_INITIAL_GAP`].
 pub const SYNC_BEHIND_THRESHOLD: u64 = 10;
+
+/// Minimum gap that must trigger a historical catch-up download, regardless of
+/// the near-tip hysteresis [`SYNC_BEHIND_THRESHOLD`].
+///
+/// Gossip can only ever deliver the *next contiguous* block (`local_height +
+/// 1`); any larger gap must go through the sync state machine. So a node behind
+/// by more than one block (i.e. `peer_height > local_height + 1`, a gap of >=
+/// 2) must enter `Downloading`. A 1-block lag (`peer_height == local_height +
+/// 1`) is left to gossip and does NOT trigger a download, avoiding thrash.
+///
+/// This is what makes a fresh joiner at height 0 against a tip of, say, 9 enter
+/// catch-up: `9 > 0 + 1` is true even though `9 > 0 + 10` is false.
+pub const SYNC_INITIAL_GAP: u64 = 1;
 
 /// How often a synced node re-polls peers for their chain status.
 ///
@@ -430,21 +450,102 @@ impl ChainSyncManager {
             },
         );
 
-        // If we're not already downloading and a peer is far enough ahead,
-        // (re)enter catch-up. This covers both the initial join (Discovery) and
-        // a synced node that just learned, via a status refresh, that a peer
-        // advanced past us.
+        // If we're not already downloading and a peer is ahead by a gap that
+        // gossip cannot bridge, (re)enter catch-up. The required gap depends on
+        // our current state:
+        //
+        // - Initial join (Discovery) or recovery (Failed): use the gap-1 rule
+        //   (`SYNC_INITIAL_GAP`, gap >= 2). Gossip only ever delivers the next
+        //   contiguous block, so ANY gap >= 2 — including the entire 0->N initial
+        //   download for a small N — must go through the sync state machine. This is
+        //   the #423 fix: a fresh joiner at height 0 against a small tip (e.g. 9)
+        //   enters Downloading instead of jumping to Synced.
+        //
+        // - Already Synced (learned via a status refresh that a peer advanced): use the
+        //   hysteresis threshold (`SYNC_BEHIND_THRESHOLD`). An already-caught-up node
+        //   that lags a few blocks near the tip normally has that gap closed by gossip,
+        //   so we avoid thrashing into a redundant historical download for every small
+        //   near-tip lag.
+        //
+        // Either way, a 1-block lag is left to gossip and never triggers a
+        // download.
         if !matches!(self.state, SyncState::Downloading { .. }) {
+            let trigger_gap = if matches!(self.state, SyncState::Synced) {
+                SYNC_BEHIND_THRESHOLD
+            } else {
+                SYNC_INITIAL_GAP
+            };
             if let Some((best_peer, status)) = self.best_peer() {
-                if status.height > self.local_height + SYNC_BEHIND_THRESHOLD {
+                if status.height > self.local_height + trigger_gap {
                     self.state = SyncState::Downloading {
                         peer: best_peer,
                         target_height: status.height,
                     };
                     self.download_height = self.local_height;
                 } else if matches!(self.state, SyncState::Discovery) {
-                    // We're close enough during initial discovery: synced.
+                    // Within one block of the tip during initial discovery:
+                    // gossip will close the gap. Mark synced.
                     self.state = SyncState::Synced;
+                }
+            }
+        }
+    }
+
+    /// React to a gossiped tip block we cannot apply because it is ahead of us
+    /// by a gap gossip cannot bridge.
+    ///
+    /// Gossip only delivers the next contiguous block (`local_height + 1`).
+    /// When a node receives a gossiped compact/full block at a height
+    /// beyond that, it is behind by a gap that only the catch-up state
+    /// machine can close. The run loop's only other sources of peer height
+    /// are the Discovery `RequestStatus` round-trip and the 30s
+    /// `STATUS_REFRESH_INTERVAL` re-poll; without this hint, a node that is
+    /// gossiped a far-ahead tip while already `Synced` would wait up to 30s
+    /// before re-entering catch-up.
+    ///
+    /// Since gossip does not tell us which peer relayed the block, we record
+    /// the observed height against the currently connected peers (at least
+    /// one of them is at or beyond this height, having relayed it) and
+    /// re-evaluate the catch-up gate immediately. This is a best-effort
+    /// hint; the authoritative height is still confirmed by the `GetBlocks`
+    /// response during download.
+    pub fn note_gossiped_tip(
+        &mut self,
+        connected_peers: &[PeerId],
+        height: u64,
+        tip_hash: [u8; 32],
+    ) {
+        // Only act if this is genuinely ahead of us by a gap gossip can't bridge
+        // (gap >= 2). A 1-block lag is left to gossip.
+        if height <= self.local_height + SYNC_INITIAL_GAP {
+            return;
+        }
+
+        // Record the observed tip against the connected peers as a hint.
+        for peer in connected_peers {
+            self.peer_statuses.insert(
+                *peer,
+                PeerStatus {
+                    height,
+                    tip_hash,
+                    last_updated: Instant::now(),
+                },
+            );
+        }
+
+        // Receiving a gossiped block we cannot apply is direct evidence of a
+        // real gap (gap >= 2), so trigger catch-up with the gap-1 rule even from
+        // the Synced state — do NOT defer to the near-tip hysteresis threshold,
+        // which exists only to suppress thrash on lags gossip *can* close. If we
+        // are already Downloading we leave the existing target alone.
+        if !matches!(self.state, SyncState::Downloading { .. }) {
+            if let Some((best_peer, status)) = self.best_peer() {
+                if status.height > self.local_height + SYNC_INITIAL_GAP {
+                    self.state = SyncState::Downloading {
+                        peer: best_peer,
+                        target_height: status.height,
+                    };
+                    self.download_height = self.local_height;
                 }
             }
         }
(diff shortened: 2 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed thoroughly as networking/sync-liveness code. CI green, merge state CLEAN, diff limited to the 5 intended files (no `.loom-managed`/stray files — that artifact is untracked and never committed). Verdict: approve.

## (1) Trigger logic + no-thrash — correct
All three gate sites are consistent and state-aware:
- `on_status` (sync.rs:472-491): `trigger_gap = SYNC_BEHIND_THRESHOLD` only when already `Synced`, else `SYNC_INITIAL_GAP` (covers Discovery AND Failed). The `else if Discovery` arm now only marks Synced for a true gap ≤ 1 — a genuinely-behind joiner (gap ≥ 2) enters Downloading instead of being sent straight to Synced.
- Discovery arm of `tick` (sync.rs:714-734): `status.height > local + SYNC_INITIAL_GAP`.
- Synced arm of `tick` (sync.rs:757-776): keeps `SYNC_BEHIND_THRESHOLD` hysteresis.

Boundary conditions verified by `test_on_status_gap_triggers_download_boundary`: gap 0 → Synced, gap 1 → Synced (gossip closes it, no thrash), gap 2 → Downloading, gap 9 → Downloading. Failed → Discovery transition (tick:791-795) then re-evaluates under the gap-1 rule. Correct.

## (2) Regression test genuineness — confirmed
`test_fresh_node_syncs_small_gap_chain_discovery_only` (tip=9) drives catch-up via ONLY the Discovery status round-trip — `run_catchup_discovery_only` has no out-of-band `on_status` injection and no gossiped-tip hint; if the production triggers are insufficient it stalls and panics. I reproduced the builder's claim: temporarily setting `SYNC_INITIAL_GAP = 10` makes the test FAIL ("discovery-only catch-up did not converge" — node stuck at height 0); restoring to 1 passes. The old `test_fresh_node_syncs_existing_chain_to_tip` blind spot (tip=95 + out-of-band injection) is closed: small tip + no injection.

## (3) No regression — confirmed
- `chain_sync_catchup_integration`: 8/8 ✅
- `network_integration`: 18/18 ✅
- `network::sync` lib: 51/51 ✅

The two failing `network/privacy/capacity.rs` doctests (`RelayCapacity::measure`) are genuinely pre-existing: that file is **byte-identical to main** on this branch (`git diff origin/main...HEAD -- capacity.rs` empty), untouched by this PR. Not introduced here.

## (4) `note_gossiped_tip` safety — acceptable
- Gap-guarded (`height <= local + SYNC_INITIAL_GAP` → early return); only acts on gap ≥ 2.
- Uses connected-peer state: records against `get_connected_peers(&discovery)` only, and re-evaluates via `best_peer()` which filters banned peers by reputation.
- A bogus high tip is **not worse than the existing `on_status` path** — a malicious peer could already report a fake height via status and trigger the same download. Downloaded blocks are validated on apply (`on_blocks`/`add_block`), so no invalid block can be injected; worst case is a wasted GetBlocks attempt that times out (30s) and returns to Discovery/Failed.
- `run.rs:928-940` correctly calls `set_local_height(local_h)` before `note_gossiped_tip` so the internal `best_peer` comparison uses a current local height. The all-connected-peers hint write is a documented best-effort that self-corrects on the next real status/GetBlocks round-trip. Acceptable for liveness code.

## (5) #423 complete on its own merits; #396 gating on #424 correct
#423's fix is independently verified (real-process `join-consensus-live-node.test.ts` 0→26, plus the small-N regression test). #396's 3-of-3 soak stays gated because of a SEPARATE blocker #424 (A+B two-minter stall at height 9), reached only after #423 unblocked Step 1 — correctly not required for this PR. The `mine-after-join-live-node.test.ts` comment update accurately re-points the gate at #424.

## (6) Is #424 a #421/#422 regression? — worth flagging to maintainer
I did NOT block on this (it's out of scope for #425), but: #424 (A+B deterministic stall at height 9, never externalizes 10) is suspicious given PR #420 demonstrated a clean 2-of-2 run to height 35. The PR description claims it reproduces on a clean #420 build, but the symptom — both minters repeatedly "Submitting minting tx to consensus height=10" while the 2-of-2 quorum never converges on a single height-10 coinbase — is exactly what a too-strict proposal-side height filter (`block_height == tip+1 && prev_block_hash == tip_hash`) from #421/#422 would produce in a race: valid competing coinbases get filtered so `propose_pending_values` proposes nothing → stall. **Recommend the maintainer bisect #424 against the #421/#422 merge points** before assuming it's pre-existing.

Issue #427, opened 2026-06-18
consensus(design): SCP cluster health — minimum node count, n-of-n threshold below 4, competing-coinbase model & solo-latch startup race


rjwalters:
## Summary

Investigation (code analysis + SCP/Stellar literature + empirical 2/3/4-node loopback runs) into whether botho's multi-minter consensus troubles (#424 stall, #421 drift, #396 soak) are fundamental to small SCP clusters. Conclusion: **partly yes** — botho runs SCP in a non-standard, degenerate-at-small-N configuration. This issue captures the findings and proposes design directions. Awaiting maintainer decision (no code change yet).

## Finding 1 — botho's quorum is degenerate below 4 nodes

`effective_threshold` (`botho/src/config.rs`) under `recommended` mode = `n - floor((n-1)/3)`:
- n=2 → **2-of-2** (unanimity, 0 fault tolerance)
- n=3 → **3-of-3** (unanimity, 0 fault tolerance)
- n=4 → **3-of-4** (first config tolerating f=1)

So **3 nodes is *also* degenerate** (n-of-n) — increasing 2→3 does NOT buy fault tolerance. BFT needs 3f+1 ⇒ **≥4 nodes** to tolerate one fault. Stellar's own docs: "a quorum with fewer than 4 entities will tolerate zero node failures … run at least three validators for high availability" ([Stellar configuring guide](https://developers.stellar.org/docs/validators/admin-guide/configuring)).

## Finding 2 — MobileCoin uses SCP differently than botho

MobileCoin (the SCP we forked) runs **10 validators across 7 orgs** with **threshold supermajority** quorum sets, never n-of-n, never 2 nodes ([Crawling the MobileCoin Quorum System, arXiv 2111.12364](https://arxiv.org/abs/2111.12364); [MobileCoin quorum config docs](https://mobilecoin.gitbook.io/mobilecoin-consensus-fog/configuring-your-node/configuring-your-node-to-connect-with-trusted-peers/defining-a-consensus-quorum-set)). Critically, MobileCoin validators agree on a **composable** value — the *set of pending user txs* (values union, so nomination converges naturally) — and there is **no mining / no per-node reward block**.

[...]

PR #429, opened 2026-06-18
fix(consensus): gate minting on quorum participation (fix solo-latch fork)

### LINKED ISSUE #428: consensus: gate minting on quorum participation (fix solo-latch fork; preserve 2-node tests)
## Goal

Make a small multi-node minter cluster (incl. the 2-node #396 A+B phase) converge reliably **without adding nodes**, by blocking minting in any round where there aren't enough connected participants to form the configured quorum. Implements direction #2 of #427 (the solo-latch startup race) and should unblock #396.

## Root cause (from #427 empirical investigation)

A node configured to expect peers (`min_peers >= 1`) boots, builds a transient **solo (1-of-1) quorum**, and if it mints a block in that window *before* peers connect, it produces a divergent solo chain → the two nodes fork (reproduced deterministically with a staggered 2-node start; #424). With simultaneous start they happen to converge, but the window is a latent fork hazard. `is_solo_mode()` (`botho/src/consensus/service.rs`) is `threshold==1 && members==1`; #401/#404 dynamically reconfigure to N-of-N once peers connect, but minting during the pre-quorum solo window is unguarded.

## The fix

**Block minting/propose in any round where the node does not have enough connected participants for its configured quorum.**

- Gate condition: do NOT mint/propose this round if `connected_peer_count < min_peers` (the configured expectation). Equivalently: a node that expects peers must not externalize a block while effectively solo.
- A node configured `min_peers == 0` (genuine single-node dev/genesis network) **continues to mint solo** — no regression. The gate only applies when the operator declared they expect peers.
- When peers connect and the quorum forms (connected >= min_peers; dynamic reconfig to N-of-N per #401/#404), minting resumes via the normal SCP path.
- If peers later DISCONNECT below `min_peers`, minting pauses (block) rather than falling back to solo-minting — a quorum that loses a member must halt (SCP safety-over-liveness), not fork.

This is a **liveness/proposer gate at the botho node layer** — it does NOT change SCP protocol semantics, validity, or the no-fork guarantee from #420. Peering/discovery is independent of minting, so nodes still discover each other and form the quorum (no deadlock): they wait for *connections*, not for *blocks*.

## Investigate / implement

- `botho/src/commands/run.rs`: the minting trigger, the existing `min_peers` / "Waiting for peers" startup logic (there is already some startup gating — make the gate apply CONTINUOUSLY to minting, not just at boot), and the dynamic quorum reconfig (#401/#404) on PeerDiscovered/PeerDisconnected.
- `botho/src/consensus/service.rs`: `is_solo_mode()`, `propose_pending_values`, the solo direct-externalize path. Add the participation gate so a node expecting peers does not propose/externalize while under-quorum.
- Plumb `min_peers` (and live connected-peer count) to wherever the mint/propose decision is made.

## Acceptance criteria

1. A node with `min_peers >= 1` does NOT mint/externalize while it has fewer than `min_peers` connected peers (verified by log/behavior: no solo block produced pre-quorum).
2. A node with `min_peers == 0` still mints solo (no regression for single-node dev/genesis).
3. **Staggered-start 2-node A+B no longer forks**: start A, let it idle (no solo blocks), start B; once peered they converge on ONE chain with identical hashes. (This was the deterministic fork repro.)
4. Simultaneous-start 2/3/4-node minter clusters still converge cleanly (no regression).
5. The #396 soak (`web/packages/wasm-signer/test/mine-after-join-live-node.test.ts`) is re-enabled (drop `&& false`) and the A+B phase reaches its target height reliably; ideally the full 3-node soak passes (C joins, catches up via #423, mines an accepted block). If #421 drift still blocks the *full* soak, get as far as the participation gate allows and report precisely what remains.
6. `cargo test -p botho` + `cargo test -p bth-consensus-scp` green (no regression; #420 fork tests still pass). Build/fmt/clippy clean; `pnpm -C web typecheck` green.

## Verification (empirical, mandatory)

Build `cargo build --release --bin botho`. Run: (a) a STAGGERED 2-node start (A alone for ~15s, then B) and confirm A produces NO blocks while alone and the pair converges on one chain after peering (no fork); (b) a simultaneous 2-node run (no regression); (c) the #396 soak. Capture evidence (no pre-quorum blocks, identical hashes post-peering, heights advancing). Clean up processes + temp dirs. Do NOT commit `.loom-managed`.

## References
- Parent design issue: #427 (SCP cluster health).
- Unblocks #396; related #421 (drift, throughput), #424 (closed), #401/#404 (dynamic reconfig), #420 (no-fork validity).

rjwalters:
## Summary

Implements **#427 direction #2** (the solo-latch startup race) and **unblocks #396**: blocks minting/propose in any round where a node that *expects* peers does not yet have enough connected participants to form its configured quorum, so it never produces a divergent pre-quorum solo block.

Closes #428

## The gate (where + condition)

Added a participation gate to `ConsensusService` (`botho/src/consensus/service.rs`):
- Two fields: `min_peers` (the operator's declared expectation) and a live `connected_peers` count, with setters `set_min_peers` / `set_connected_peers`.
- `should_propose_this_round()`: returns `false` when `min_peers >= 1 && connected_peers < min_peers`; `true` when `min_peers == 0` (genuine solo) or once `connected_peers >= min_peers`.
- `propose_pending_values()` consults the gate up front and withholds **both** the solo direct-externalize path **and** the normal SCP propose while under-quorum. Withheld values stay queued (`pending_values`) and are proposed once the quorum forms.

Wired from the node run loop (`botho/src/commands/run.rs`):
- `consensus.set_min_peers(min_peers_to_wait)` + initial `set_connected_peers(...)` at startup.
- `consensus.set_connected_peers(new_peer_count)` refreshed on every `PeerDiscovered` / `PeerDisconnected` event.

Discovery/peering is independent of minting, so the node still forms its quorum while gated — **no deadlock** (it waits for *connections*, not *blocks*). This is a node-layer liveness/proposer gate only; it does **not** change SCP protocol semantics, validity, or the #420 no-fork guarantee.

Behavior:
- `min_peers >= 1` + connected < min_peers → **no block** (pre-quorum window silent).
- `min_peers == 0` → mints solo (no regression for dev/genesis).
- peers drop below `min_peers` → minting **pauses** (halt), does not fall back to solo (no fork).

## Regression test

`participation_gate_blocks_solo_block_until_peers_connected` in `service.rs` exercises the real gate: a `min_peers==0` node externalizes solo; a `min_peers>=1` node with 0 connected peers does **not** externalize and keeps its value pending; the gate opens once `connected >= min_peers` and re-closes if peers drop.

## Empirical verification (real loopback `botho run`, release binary)

**Staggered 2-node (the deterministic fork repro):**
- A (`min_peers=1`) launched alone: stayed at **height 0 for the full 15s** while alone (`peers=0`), logs show `participation gate min_peers=1, connected=0` and `Quorum not satisfied`. No pre-quorum solo block.
- B launched + peered: both converged on the **identical tip hash** (`597e452e…2919`) at heights 50/51. The previously-deterministic fork is **gone**.

**Simultaneous 2-node (no regression):** converged on identical tip (`8be2f0cd…7010`) at height 17.

**#396 soak (`BOTHO_E2E_NODE=1`, 2x):**
- A+B (2-of-2) reliably reach shared target **N=24 with identical tips**; C joins (`peerCount=2`) and **catches up 0→24 onto A's exact chain**. All three sit on one tip at N.
- The FINAL step (network advancing **past N** with C's freshly-mined block accepted by all three) stalls on the pre-existing **#421 SCP-slot/height drift** (re-opened by the #422 revert), which is independent of the proposer gate.
- Per #428's STOP guardrail the soak is **re-gated with an accurate comment** rather than forced green; it stays gated on the full 3-node end state until #421 is re-fixed without regressing two-minter liveness.

## Tests / build

- `cargo test -p botho` — green (the 2 failing `network/privacy` *doctests* are pre-existing on the base commit, unrelated).
- `cargo test -p bth-consensus-scp` — 100 passed; `no_fork_with_tip_agnostic_validity` (+ `fork_with_tip_dependent_validity`) #420 tests pass.
- Build + `cargo fmt` clean; clippy warnings are all pre-existing unused-import noise.
- `pnpm -C web typecheck` — green.

## Files touched
- `botho/src/consensus/service.rs` — participation gate + regression test
- `botho/src/commands/run.rs` — wire `min_peers` + live connected count into the gate
- `web/packages/wasm-signer/test/mine-after-join-live-node.test.ts` — #396 soak status comment + re-gate note

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -477,9 +477,24 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
         chain_state,
     );
 
+    // Participation gate (issue #428): tell the consensus service how many
+    // connected peers this node expects before it may produce a block, and the
+    // current live count. `min_peers_to_wait` is the configured expectation
+    // (Explicit mode: at least one trusted peer; Recommended mode: the
+    // configured `min_peers`). When it is 0 the node is a genuine single-node
+    // network and mints solo (no gate). When it is >= 1 the consensus service
+    // withholds minting/propose until at least that many peers are connected,
+    // so the node never externalizes a divergent solo block during the
+    // pre-quorum startup window. The live count is refreshed on every
+    // PeerDiscovered / PeerDisconnected event below.
+    consensus.set_min_peers(min_peers_to_wait);
+    consensus.set_connected_peers(discovery.peer_count());
+
     info!(
-        "Consensus service initialized at slot {}",
-        consensus.current_slot()
+        "Consensus service initialized at slot {} (participation gate min_peers={}, connected={})",
+        consensus.current_slot(),
+        min_peers_to_wait,
+        discovery.peer_count()
     );
 
     // Chain-sync state machine: drives initial block download (IBD) / catch-up.
@@ -695,6 +710,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             // Broadcast peer event to WebSocket clients
                             ws_broadcaster.peer_connected(new_peer_count, &peer_id.to_string());
 
+                            // Refresh the participation gate's live peer count
+                            // (issue #428) so the consensus service can resume
+                            // proposing once enough peers are connected.
+                            consensus.set_connected_peers(new_peer_count);
+
                             // Reconfigure the consensus quorum to include the
                             // newly connected peer. This is what lifts a node
                             // out of latched solo mode without a restart.
@@ -747,6 +767,13 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
 
                             // Broadcast peer event to WebSocket clients
                             ws_broadcaster.peer_disconnected(new_peer_count, &peer_id.to_string());
+
+                            // Refresh the participation gate's live peer count
+                            // (issue #428). If this drops below the configured
+                            // expectation the consensus service pauses proposing
+                            // rather than falling back to solo minting — a
+                            // quorum that loses a member halts, it does not fork.
+                            consensus.set_connected_peers(new_peer_count);
 
                             // Shrink the consensus quorum to drop the departed
                             // peer so we don't deadlock waiting on a member that
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -173,6 +173,22 @@ pub struct ConsensusService {
     /// A quorum set reconfiguration that arrived mid-round and was deferred to
     /// the next slot boundary to avoid stranding an in-flight slot.
     pending_quorum_set: Option<QuorumSet>,
+
+    /// Participation gate (issue #428): the number of connected peers the
+    /// operator declared they expect before this node may produce blocks
+    /// (the configured `min_peers`). `0` means the node is a genuine
+    /// single-node dev/genesis network and mints solo forever — no gate.
+    /// `>= 1` means the node expects peers and MUST NOT mint/propose a block
+    /// while it has fewer than this many connected peers, so it never produces
+    /// a divergent solo block during the pre-quorum startup window (or after a
+    /// quorum member disconnects). See [`Self::should_propose_this_round`].
+    min_peers: usize,
+
+    /// Live count of currently connected peers, kept up to date by the node
+    /// run loop on PeerDiscovered/PeerDisconnected (see
+    /// [`Self::set_connected_peers`]). Compared against `min_peers` by the
+    /// participation gate.
+    connected_peers: usize,
 }
 
 impl ConsensusService {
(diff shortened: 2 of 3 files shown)


rjwalters:
## APPROVED — participation gate is correct, narrowly scoped, and does not touch SCP protocol/validity

Verified at head `9ba4c3601f5ab0e816525ad0049180dfdd41f072`. CI: all checks **pass** (benchmarks + fuzz build guard + rustfmt green; fuzz jobs intentionally `skipping`). `mergeStateStatus`: **CLEAN**.

### 1. Gate logic correct + covers BOTH mint paths ✅
`should_propose_this_round()` is exactly `block iff min_peers >= 1 && connected_peers < min_peers`: early-returns `true` for `min_peers == 0` (genuine solo/genesis — no regression), else `connected_peers >= min_peers`. The gate is a single early-return at the **top** of `propose_pending_values()` (before value selection), so it sits ahead of **both** the solo direct-externalize path (`if self.is_solo_mode()`) **and** the normal `scp_node.propose_values()` path. A partial gate is structurally impossible — both minting paths live below it in the same function.

### 2. No deadlock + values queued ✅
Discovery/peering and quorum reconfig (`rebuild_scp_quorum_set`/`reconfigure_quorum`) are independent of the gate; a gated node still connects and forms its quorum (waits for *connections*, not *blocks*). On withhold, the function returns before any `pending_values.remove(...)`, so values stay queued and are proposed on the next tick once `connected_peers >= min_peers`. Test asserts `pending_values` non-empty while gated.

### 3. Disconnect-halt correct ✅
`PeerDisconnected` refreshes `set_connected_peers(new_peer_count)`, re-closing the gate if peers drop below `min_peers` — the intended SCP safety-over-liveness halt (a quorum that loses a member pauses, it does not fall back to solo and fork). Matches #428's stated intent.

### 4. connected_peers / min_peers consistency — no off-by-one ✅
Both fed from `discovery.peer_count()` (= `self.peers.len()`, peers excluding self). `min_peers_to_wait` is the same peer-count metric already used by the pre-existing startup wait loop (`discovery.peer_count() < min_peers_to_wait`): Explicit mode → 1, Recommended mode → configured `min_peers`. Comparison is apples-to-apples; no over/under-block.

### 5. No SCP/validity regression ✅
Diff is +143/-0 in service.rs confined to two new fields, two setters, the gate predicate, the early-return, and the test. `is_solo_mode`, `combine_fn`, and the #420 tip-agnostic validity logic are untouched. `cargo test -p bth-consensus-scp`: 100 passed. `--test test_tip_dependent_validity_fork --features test_utils`: **`no_fork_with_tip_agnostic_validity` + `fork_with_tip_dependent_validity` both pass.**

### 6. Regression test genuine (not tautological) ✅
`participation_gate_blocks_solo_block_until_peers_connected` calls the **real** `propose_pending_values()` and checks the real `externalized` field: `min_peers==0` solo node externalizes; the SAME 1-of-1 solo quorum with `set_min_peers(1)` + 0 peers does **not** externalize and keeps its value pending; gate opens at `connected>=min_peers`; re-closes on drop. This is the #424 fork repro at the unit level. Passes locally.

### 7. #396 soak gating accurate + acceptable ✅
Soak stays gated (`&& false` → `describe.skip`). Added comment accurately separates what the gate fixed (A+B reach shared target N with identical tips; C catches up 0→N onto A's chain) from the residual pre-existing **#421** SCP-slot/height drift (re-opened by the #422 revert in base `942bdd0`) that stalls only the final past-N 3-of-3 step. Correctly NOT forced green — requiring full #396 conflates two issues; #428's own acceptance is proven separately.

### 8. Clean diff ✅
Exactly 3 files vs main (run.rs +31, service.rs +143, test.ts +21). No `.loom-managed`/stray files. The 2 `network/privacy` doctest failures are pre-existing on base (PR touches no network/privacy files) — unrelated.

Node-layer proposer gate only; SCP protocol/validity untouched as intended. Shipping.

PR #426, opened 2026-06-18
revert: #422 SCP slot-alignment (regressed two-minter liveness #424)

rjwalters:
## Revert PR #422 — fixes #424

PR #422 ("keep SCP slot aligned with block height: height filter + gated rewind", commit `5824f6f`) **regressed two-minter liveness** and is reverted here.

### Why
A **bisect** (real `botho run` minters, loopback) confirmed #424 is a regression introduced by #422:
- **At #420 (`e0226ff`, tip-agnostic validity):** clean — A+B advance past height 9 to 35 with identical tips. **4/4 runs.**
- **At #422 (`5824f6f`):** A+B deterministically WEDGE (~2/5 runs), stalling at height ~6-9, never externalizing the next height. Both minters endlessly resubmit their own next-height coinbase; the 2-of-2 quorum never converges.

The wedge is the **whole #422 mechanism**, not one line: the proposal-side height filter (`minting_value_matches_tip`) drops both minters' competing-but-valid coinbases in a race → `propose_pending_values` proposes nothing; and the gated-rewind backstop (`realign_scp_slot_to_chain`) thrashes the slot via a drift→reject→realign→future-slot-discard cycle. A throwaway experiment loosening only the filter left the realign-backstop wedge in place (stalled at 10 instead of 9), so a surgical tweak does not close #424.

### What this restores
Reverting #422 returns to #420's clean behavior. Verified empirically on this branch: **4/4 two-minter runs advance past height 9 (to 18) with identical block hashes at every common height, in genuine non-solo SCP** (no fork, no stall).

### Trade-off (re-opens #421)
This **re-opens #421** (the SCP slot/height drift #422 was solving). That drift only affects the **3-of-3 joiner step** of the #396 soak (a freshly-synced joiner vs drifted established minters), not two-minter liveness. #421 must be re-approached **without** the externalize-then-reject + downward-realign loop that wedges the 2-minter race. The #396 soak stays gated (`&& false`) with an updated rationale.

### Safety / scope
- #420's tip-agnostic `validity_fn` (no-fork guarantee) is **intact** — the #420 fork regression tests (`no_fork_with_tip_agnostic_validity`, `fork_with_tip_dependent_validity`) still pass.
- #423's catch-up trigger (`SYNC_INITIAL_GAP`) is **intact**.
- Only #422's code is removed (`realign_slot_to`, `realign_scp_slot_to_chain`, `minting_value_matches_tip` — 582 deletions across 6 files).
[...]

PR #430, opened 2026-06-19
fix(consensus): non-wedging SCP slot/height drift fix — hybrid A1 + C (#421)

### LINKED ISSUE #421: consensus: SCP current_slot_index drifts ahead of block height (blocks #396 3-of-3 soak)

rjwalters:
## Summary

Implements the maintainer-approved **hybrid A1 + C** design for the SCP slot / ledger-height drift that gated the #396 3-of-3 soak. This is the non-wedging re-approach after PR #422 was reverted (#426) for regressing 2-minter liveness (#424). It deliberately avoids both of #422's wedge mechanisms.

Closes #421.

Built on latest `origin/main` (3fb656a — has #420 tip-agnostic validity, #423 catch-up, #429 participation gate; does NOT include the reverted #422).

This is NOT the reverted #422 approach: no proposal-time coinbase filtering, no backward SCP slot re-seat.

## A1 — tolerant / idempotent duplicate-height apply (apply path only)

`botho/src/commands/run.rs` `SlotExternalized` handler. When `add_block_from_network` rejects an externalized coinbase whose height is **already filled** in the ledger (`block.height() <= ledger tip height` — a duplicate / lost-the-race coinbase), classify it as a **benign skip** instead of a hard `warn!` error. The winning coinbase already landed at that height, so the externalized duplicate is provably redundant.

Discrimination (the safety-critical part): we read the ledger tip height after the failure and only treat the rejection as benign when `block.height() <= ledger_height` — i.e. the height is **genuinely already in the ledger**. A block for a NEW height that fails for any other reason (bad PoW / difficulty / reward, etc.) is still surfaced as a real failure. No validation masking; no real winning block dropped. All #420 no-fork `add_block` checks stay unconditional.

## C — forward-only, idle-gated anchoring to the network's live SCP slot

`botho/src/consensus/service.rs` new `anchor_scp_slot_to_peer`, wired into `handle_message`. When an inbound peer SCP message is for a slot **ahead** of ours (which `node_impl::handle_messages` would otherwise silently discard as a "future slot") and our current slot is **idle**, fast-forward FORWARD to the peer's slot via `reset_slot_index`. This lets a freshly-synced / fallen-behind node converge with established nodes **even if they drifted ahead**.

- **Forward-only:** only moves to a strictly higher index, so `reset_slot_index`'s `debug_assert!(slot_index > current)` stays **satisfied** (never relaxed). Never re-seats backward, never re-opens an externalized index — so #422's backstop re-externalize fork risk is absent.
- **Idle-gated:** refuses to anchor while the local slot holds in-flight nominate/ballot state (same activity guard as `sync_scp_slot_to_chain`), so it never discards accepted-commit state.
- **Proposal-untouched:** does not filter or withhold any coinbase. Two established minters sharing a tip are on the SAME slot, so C never triggers for them — only a genuinely behind node fast-forwards toward the leaders.

## Hard constraints honored

- No proposal-time filtering of valid competing coinbases (`propose_pending_values` / `validate_minting_tx_intrinsic` untouched).
- No backward SCP slot re-seat (`reset_slot_index`'s assert unchanged and satisfied).
- #420 no-fork preserved (every `add_block` check stays unconditional).
- 2-minter liveness preserved (the #424 regression) — guarded in-process by T5.
- #429 participation gate untouched.

## Tests (T1–T5, `botho/src/consensus/service.rs`)

- **T1** characterizes the externalize-then-reject drift mechanism.
- **T2** reproduces joiner non-convergence with a drifted leader (pre-C behavior).
- **T3 (key guard)** Option C anchors a behind, idle node FORWARD to the leader's live slot; asserts forward-only (never rewinds) and the idle-gate refusal. **Fails pre-fix** (verified by disabling the `handle_message` hook → T3 FAILED, all others still passed).
- **T4** the A1 benign-skip discrimination (`height <= ledger_height` benign; higher height not masked).
- **T5 (anti-#424, REQUIRED)** two minters each racing a distinct competing coinbase make sustained progress past height 9 with no wedge. Green with this change; would fail under #422's proposal filter.

## Verification (empirical)

- `cargo test -p bth-consensus-scp`: **100 passed**; #420 fork tests `no_fork_with_tip_agnostic_validity` / `fork_with_tip_dependent_validity` both **pass**.
- `cargo test -p botho`: lib **971 passed**, all integration suites green.
- `cargo fmt --all --check` clean; `cargo clippy -p botho --lib` no new warnings; `pnpm typecheck` green.
- **2-minter loopback (release binary, the #424 regression check):**
  - **Simultaneous** A+B (min_peers=1): reached height 25 (past target 24, past the #424 wedge point ~9) with **identical tips, no wedge** → PASS.
  - **Staggered** A+B: reached height ~26-28, **no wedge** (no stall/plateau).
- **3-node #396 soak (real loopback, BOTHO_E2E_NODE=1):** steps 1–3 PASS — A+B reach N=24 with identical tips; C joins, connects (3-of-3), catches up 0→24 onto A's exact tip. **Option C proven to work:** C anchored FORWARD from slot 1 to the leaders' drifted live slot 29 ("Anchoring SCP slot forward ... to_slot=29"), whereas baseline (3fb656a) leaves C stuck at slot 1 forever discarding the leaders' messages.

## What the full soak still needs (out of scope for #421, see #427)

The soak's FINAL end-state (C mines a block accepted past N) is still blocked by an **independent** concern, NOT the drift: at n=3 the recommended quorum is **3-of-3 unanimous (n-of-n)**, and the 2→3 reconfiguration at C's join is **deferred** on the busy established minters (round in flight). The join-boundary slot then runs under **mismatched memberships** (C on 3-of-3, the minters still on 2-of-2 until the deferred reconfig applies — which itself needs that slot to externalize → deadlock). This is the SCP cluster-health / n-of-n threshold + join-boundary reconfig coordination concern tracked by **#427**.

Baseline (3fb656a) stalls at the SAME height N with C never even participating, so this PR strictly improves convergence (the drift symptom is closed: all three align on the same slot index) without regressing 2-minter liveness.

Per #421's STOP guardrail I did **not** force the soak green and did **not** reintroduce any #422-style filter/backward-rewind. The `mine-after-join-live-node.test.ts` soak stays gated (`&& false`) with an accurate comment pointing at #427; re-enable it when #427 lands.

## Files touched
- `botho/src/commands/run.rs` — A1 benign-skip classification at the apply boundary.
- `botho/src/consensus/service.rs` — Option C `anchor_scp_slot_to_peer` + `handle_message` wiring; T1–T5 tests.
- `web/packages/wasm-signer/test/mine-after-join-live-node.test.ts` — accurate #421-status / #427-gating comment.

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -1245,7 +1245,56 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
 
                                     // Add to ledger
                                     if let Err(e) = node.add_block_from_network(&block) {
-                                        warn!("Failed to add consensus block: {}", e);
+                                        // Issue #421 (Option A1): tolerant /
+                                        // idempotent duplicate-height apply.
+                                        //
+                                        // SCP auto-advances its slot on
+                                        // externalize, but a duplicate /
+                                        // lost-the-race coinbase for a height the
+                                        // ledger has ALREADY filled is rejected
+                                        // here by `add_block`'s unconditional
+                                        // height/prev-hash checks (the #420
+                                        // no-fork property — KEEP IT). That
+                                        // rejection is EXPECTED and benign: the
+                                        // winning coinbase already landed at that
+                                        // height, so the externalized duplicate is
+                                        // provably redundant. Treat it as a benign
+                                        // SKIP (not a hard error) so we don't spam
+                                        // misleading error logs for the common
+                                        // 2-minter race where the loser's
+                                        // externalize is redundant.
+                                        //
+                                        // SAFETY: we ONLY classify it as benign
+                                        // after VERIFYING the height is genuinely
+                                        // already in the ledger (block height <=
+                                        // current tip height). A block for a NEW
+                                        // height that fails for any other reason
+                                        // (genuinely invalid: bad PoW, wrong
+                                        // difficulty/reward, etc.) is still a hard
+                                        // failure — we must NOT mask real
+                                        // validation failures, and we must NOT drop
+                                        // a real winning block.
+                                        let ledger_height = node
+                                            .shared_ledger()
+                                            .read()
+                                            .ok()
+                                            .and_then(|l| l.get_chain_state().ok())
+                                            .map(|s| s.height);
+                                        let height_already_filled = matches!(
+                                            ledger_height,
+                                            Some(h) if block.height() <= h
+                                        );
+                                        if height_already_filled {
+                                            debug!(
+                                                height = block.height(),
+                                                ledger_height = ledger_height.unwrap_or_default(),
+                                                "Skipping duplicate-height consensus block \
+                                                 (height already filled — benign, issue #421 A1): {}",
+                                                e
+                                            );
+                                        } else {
+                                            warn!("Failed to add consensus block: {}", e);
+                                        }
                                     } else {
                                         // Record for dynamic timing
                                         consensus.record_block(block.header.timestamp, block.transactions.len());
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -639,6 +639,19 @@ impl ConsensusService {
         Span::current().record("msg_type", msg_type);
         trace!(msg_type, "Processing SCP message");
 
+        // Issue #421 (Option C): forward-only anchoring to the network's live
+        // SCP slot. If this message is for a slot AHEAD of ours, the SCP node
+        // would otherwise silently DISCARD it as a "future slot" (see
+        // node_impl::handle_messages). When we are behind the network — e.g. a
+        // freshly-synced joiner, or a node that fell behind while the leaders
+        // drifted ahead via the #421 apply-boundary drift — and our current
+        // slot is idle, fast-forward FORWARD to the peer's slot so we re-enter
+        // federated voting at the network's actual slot instead of stalling.
+        // This is forward-only and idle-gated (see `anchor_scp_slot_to_peer`),
+        // so it never re-seats backward, never re-opens an externalized index,
+        // and never triggers for two established minters sharing a tip.
+        self.anchor_scp_slot_to_peer(scp_msg.slot_index);
+
         // Handle the message
         if let Some(response) = self.scp_node.handle_message(&scp_msg)? {
             self.queue_broadcast(response);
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!** — rigorous review of the hybrid A1 + C non-wedging drift fix (#421). Head `4376b2d`. CI: all required checks pass, merge state CLEAN.

This is the #1 priority I was asked to verify: **it does NOT reintroduce #422's 2-minter liveness regression.** Confirmed on every axis below.

### 1. NO 2-minter liveness regression (THE critical check) — PASS
- **No proposal filter anywhere.** `grep` for `minting_value_matches_tip` / `matches_tip` / `value_matches` returns nothing. `git diff 3fb656a..HEAD` touches neither `propose_pending_values` nor `validate_minting_tx_intrinsic` — proposal selection is entirely untouched.
- **No backward slot re-seat.** `reset_slot_index`'s `debug_assert!(slot_index > self.current_slot_index())` (consensus/scp/src/node/node_impl.rs:330) is **unchanged and unrelaxed**. Both `reset_slot_index` callers in service.rs guard `peer_slot > current` / `current >= target` first, so the assert always holds. C is strictly forward-only.
- **T5 is genuine and green.** Two real services in a 2-of-2, each racing a DISTINCT valid coinbase per height, sustained progress past height 9 (the #424 wedge point) with fork-safety + single-coinbase assertions. Passes. I also confirmed T5 stays green with the C hook disabled — i.e. T5 guards liveness independently of C.

### 2. A1 drops no real block / masks no validation — PASS
A1 only changes the **log level** of an already-failed `add_block` — it never re-applies or skips a successful apply. Benign-skip triggers ONLY when `block.height() <= ledger tip`; any failure at a fresh height (`> tip`) still `warn!`s. Adversarially: a valid block at `tip+1` rejected by `add_block` has `height > tip` → NOT masked. It also fails *safe* — if the ledger read/`get_chain_state` errors, `ledger_height` is `None`, `matches!` is false, and it falls through to `warn!`. #420's no-fork checks in `LedgerStore::add_block` are untouched (ledger crate not in the diff). T4 asserts the discrimination.

### 3. C forward-only + idle-gated + not finality-breaking — PASS (with tracked follow-up)
Forward-only and idle-gated; the idle guard mirrors `sync_scp_slot_to_chain` exactly (num_voted/accepted/confirmed_nominated, bN, phase). It cannot discard in-flight consensus or re-open an externalized slot. T3 verifies forward anchor + no-backward-rewind + idle-gate refusal, and **T3 FAILS with the hook disabled while all others stay green** — C is load-bearing, not a tautology (I reproduced this).
- One hardening note (non-blocking): `anchor_scp_slot_to_peer` acts on the *claimed* `slot_index` of any gossiped message before sender/quorum validation, so a peer advertising a bogus high slot could fast-forward an idle node and grief its liveness. This is safety-preserving (no fork, idle-gated, forward-only), is on an already-unauthenticated pre-mainnet gossip path, and does NOT reintroduce the #422/#424 failure mode — so it does not block. Tracked in **#431**.

### 4. No #420 regression — PASS
`cargo test -p bth-consensus-scp` → 100 passed; the tip-dependent fork integration tests (`--features test_utils --test test_tip_dependent_validity_fork`) → `no_fork_with_tip_agnostic_validity` and `fork_with_tip_dependent_validity` both pass.

### 5. T1–T5 genuine + fail-pre/pass-post — PASS
All five pass (`cargo test -p botho --lib`). Spot-checked: they drive real services/SCP nodes, not tautologies. Verified the claimed load-bearing relationship myself: disabling the `handle_message` C hook makes **T3 FAIL** while **T5 stays PASS** — exactly as the builder reported.

### 6. #421 drift fixed + remaining blocker independent (#427) + gating acceptable — PASS
The drift mechanism (T1) and its convergence symptom are addressed by C (forward anchor 1→leader's live slot), with end-to-end loopback evidence in the PR. The remaining soak blocker is genuinely independent: n-of-n (3-of-3) quorum + deferred 2→3 reconfig at C's join → mismatched-membership join-boundary deadlock (#427), not the slot/height drift. Baseline stalls at the same height with C never participating, so this strictly improves convergence. Keeping the soak gated (`&& false`, comment points at #427) is correct — not requiring full #396 green to approve.

### 7. Clean diff — PASS
Exactly 3 expected files (run.rs, service.rs, the soak test comment+gate). No `.loom-managed`/stray files.

### Local verification run
- `cargo test -p bth-consensus-scp`: 100 passed
- tip-dependent fork integration tests: both pass
- `cargo test -p botho --lib`: 971 passed; T1–T5 all pass
- C-hook-disabled experiment: T3 FAILED, T5 still passed (C load-bearing, T5 independent)
- `cargo fmt --all --check`: clean
- CI: all required checks pass, merge state CLEAN

Excellent, disciplined work — it honors every #421 STOP guardrail and avoids both of #422's wedge mechanisms. Created #431 to track the Option C peer-advertised-slot hardening.

PR #434, opened 2026-06-19
feat(consensus): add quorum.fault_model (crash 2f+1 default / bft 3f+1) (#432)

rjwalters:
Closes #432

Implements the maintainer-approved #427 Finding-1 decision and the n=3
join-boundary half of #396: default small homogeneous clusters to
**crash-fault tolerance (2f+1 simple majority)** so a lagging/crashed node no
longer acts like a fault that stalls liveness, while keeping a 3f+1 Byzantine
posture available as `bft`.

## What changed

**Config (`botho/src/config.rs`)**
- New `FaultModel` enum (`crash` = DEFAULT | `bft`), added as
  `quorum.fault_model` next to `quorum.mode`. Serde `rename_all = "lowercase"`;
  an invalid value errors clearly at config load.
- `QuorumConfig::effective_threshold` (Recommended mode) now branches on the
  fault model:

  | n | crash `floor(n/2)+1` | bft `n-floor((n-1)/3)` |
  |---|----------------------|------------------------|
  | 1 | 1 | 1 |
  | 2 | 2 | 2 |
  | 3 | **2 (2-of-3)** | 3 |
  | 4 | 3 | 3 |
  | 5 | 3 | 4 |
  | 6 | 4 | 5 |

  Explicit mode is unchanged (operator-set exact threshold).

**Threshold reuse (`botho/src/commands/run.rs`)**
- `rebuild_scp_quorum_set` already delegates to `effective_threshold`, so it is
  fault-model-aware with no duplicated formula. Docstring updated.

**Tests (`config.rs`)**
- Unit tables for BOTH models, n=1..6
  (`test_quorum_effective_threshold_crash` / `_bft`).
- `fault_model` parse tests: crash, bft, default-when-absent, invalid-errors.
- Updated `can_reach_quorum` recommended-mode tests for the new 2-of-3 default.

**Docs**
- `docs/operations/configuration.md`: full crash/bft posture + threshold tables;
  notes crash = HA/liveness for trusted clusters (default), bft = Byzantine and
  needs >=4 nodes for genuine BFT.
- `docs/minting.md`: fault-model summary + cross-link.

## Hard constraints honored
- **#420 no-fork preserved**: any two 2f+1 majority subsets intersect; block-apply
  tip checks are untouched. #420 `no_fork_with_tip_agnostic_validity` /
  `fork_with_tip_dependent_validity` both pass.
- **n=2 unchanged**: crash n=2 == bft n=2 == 2-of-2, so 2-minter behavior is not
[...]

PR #435, opened 2026-06-19
fix(consensus): eliminate n=2 startup solo-mode fork (#433)

### LINKED ISSUE #433: consensus: intermittent two-minter (n=2) solo-mode fork — A and B both mine divergent solo chains despite 'Quorum satisfied: 2-of-2'
## Summary

A pre-existing, intermittent **two-minter (n=2) liveness/safety defect**: two
established minters A and B that have peered and report `Quorum satisfied:
2-of-2` sometimes BOTH fall into SCP **solo mode** and mine DIVERGENT solo
chains that never reconcile.

This is independent of `quorum.fault_model` (#432): at n=2 the crash and bft
thresholds are both 2-of-2, so this reproduces under either model. It is also
independent of any third joiner node — it reproduces in a plain A+B-only run.

## Evidence (real loopback `botho run`, release binary, BOTHO_E2E_NODE=1)

In an A+B-only 2-minter run (no node C):
- Both nodes log `Starting minting - Quorum satisfied: 2-of-2` and
  `Peers: 1 | Minting: active | Quorum satisfied: 2-of-2`.
- Yet both also log `Advanced to next slot (solo mode)` and
  `Solo mode: directly externalizing values slot=… count=1` for every slot.
- Result: A and B fork — in a failing run they disagreed as low as **height 20**
  and ran to DIFFERENT tips at heights 111 vs 115 (distinct hashes the whole way).
- In passing runs A+B agree at all historical heights and the tip (so this is
  intermittent — a startup race, roughly 1 in 2–3 runs at slot=1s, ringSize=20,
  fast-lottery overrides).

Reproduced repeatedly with two `recommended`/`min_peers=1` minters peering over
127.0.0.1 loopback (the same harness as
`web/packages/wasm-signer/test/mine-after-join-live-node.test.ts`).

## Suspected cause

A mismatch between the **minting / `can_reach_quorum`** view (which reports
2-of-2 satisfied once a peer connects) and the SCP **`is_solo_mode`** /
`rebuild_scp_quorum_set` view at the **1-of-1 → 2-of-2 transition**: a node that
boots solo (1-of-1) appears to keep treating itself as solo and directly
externalizing values, instead of running federated SCP with the connected peer.
Because both nodes do this, neither defers to the other and they fork.

Likely areas:
- `is_solo_mode` determination vs the dynamically rebuilt quorum set
  (`botho/src/commands/run.rs` `rebuild_scp_quorum_set`, and the SCP service's
  solo-mode gate in `botho/src/consensus/service.rs`).
- The order/timing of `reconfigure_quorum` relative to the first peer connect
  and the first proposed slot.

## Impact

- Blocks re-enabling the #396 capstone soak
  (`mine-after-join-live-node.test.ts`), which currently stays gated ONLY
  because of this n=2 fork (the n=3 join-boundary blocker is fixed by #432's
  2-of-3 crash quorum — the 3-node capstone itself passes when A+B converge).
- More importantly it is a real two-minter safety/liveness concern: two honest
  minters can produce a persistent fork.

## Acceptance

1. Two `recommended`/`min_peers=1` minters peering over loopback NEVER both run
   in solo mode once peered; they run federated SCP and converge on identical
   tips (and identical historical block hashes) reliably across many runs.
2. Add a deterministic in-process/regression test for the 1-of-1 → 2-of-2
   transition that asserts the node leaves solo mode and does not directly
   externalize once a peer is in its quorum set.
3. With this fixed, re-enable the #396 soak (drop the `&& false` in
   `mine-after-join-live-node.test.ts`) and confirm it passes reliably.

## References
- Found while implementing #432 (quorum.fault_model). The fault_model change is
  complete and unit-tested; this is the remaining, separate blocker for #396.
- Related: #420 (no-fork apply checks), #424 (2-minter liveness guard), #429
  (participation gate), #430 (drift fix), #427 (SCP cluster health).

rjwalters:
Closes #433. Unblocks the n=2 phase of the #396 capstone soak.

## Confirmed root cause (code refs on current main)

The intermittent two-minter fork is a **startup race at the 1-of-1 -> 2-of-2 transition**, but the stranding mechanism is NOT the #408 deferral — it's the pre-consensus **wait-for-peers loop swallowing the `PeerDiscovered` event**:

1. `botho/src/commands/run.rs:269` — the startup "wait for peers" loop polls the swarm and CONSUMES `PeerDiscovered` events (just logs them). The peer is inserted into `discovery.peers` but the main event loop never sees that event.
2. `botho/src/commands/run.rs:456` — the consensus service is then seeded via `build_scp_quorum_set(&quorum, ...)`, which uses the **static config** and yields a **1-of-1 solo quorum** for a `recommended`/`min_peers=1` node.
3. `botho/src/commands/run.rs:491` — `set_connected_peers(1)` opens the participation gate (#429): `should_propose_this_round()` now returns `true` (`connected >= min_peers`).
4. Result: **gate open + quorum still 1-of-1**. The next `consensus.tick()` -> `propose_pending_values()` (`service.rs:807`) takes the **solo direct-externalize** path (`is_solo_mode()` true), mining a divergent solo chain. No further `PeerDiscovered` ever arrives (peer already connected), so the quorum is never reconfigured out of 1-of-1. Both nodes do this -> fork.

This matches the issue's refined hypothesis (solo mint wins the race against the reconfig) but pins the actual stranding to the wait-loop event consumption rather than the #408 active-slot deferral.

## The fix (localized)

- **Primary** (`run.rs`): seed the INITIAL SCP quorum from the peers ALREADY CONNECTED at startup (`rebuild_scp_quorum_set`) instead of the static config, so a node that has already peered boots into federated SCP (e.g. 2-of-2), never solo. Lone node (no peers) still yields 1-of-1 -> solo, no regression.
- **Defense-in-depth** (`service.rs`): `is_transitional_solo()` guard withholds the solo direct-externalize whenever `min_peers >= 1 && connected_peers >= 1 && is_solo_mode()` (peer present but still solo = transitional, must wait for reconfig). Values stay queued.

`min_peers == 0` still solo-mints. No proposal-time filtering of valid competing coinbases, no backward slot re-seat (no #422 wedge). #420/#429/#430/#434 untouched.

## Regression test (fail-pre / pass-post)

`consensus::service::tests::transitional_solo_does_not_directly_externalize`: a node with a peer connected but a still-1-of-1 quorum must NOT directly externalize, and must run federated SCP once reconfigured. Verified it FAILS without the guard (`expected externalized.is_none()`) and PASSES with it.

## Empirical evidence (real loopback `botho run`, release binary)

A faithful A+B 2-minter harness (recommended, min_peers=1, 1s slots, mutual bootstrap, fast-lottery — same config as the #396 test):
- **Pre-fix**: forked in **2 of 8** runs (`Solo mode: directly externalizing` on BOTH A and B) — reproduces the issue's ~1-in-2-3 rate.
- **Post-fix**: **10/10 runs clean** — federated SCP (`Slot externalized!`), ZERO solo-mode externalizes, identical tips + identical historical hashes 1..N.

## #396 capstone status

Re-ran the 3-node capstone 6x. Steps 1-4(a) pass on EVERY run: A+B converge (no n=2 fork), C catches up onto A's exact chain, and **all three converge past N onto one identical tip** with ZERO solo mode. The FINAL step — C winning its OWN coinbase — is still intermittent (~4/6 pass), gated by a SEPARATE pre-existing **n=3 SCP-slot drift** join-boundary concern (#421/#427-family): C's proposals land on an earlier slot and get discarded as "future slots", so C rarely gets its own block externalized. Per the STOP guardrail this PR does NOT force #396 green or mask that flake — #396 stays gated (`&& false`) with an accurate updated comment documenting exactly what's fixed (n=2 fork) vs what remains (n=3 C-coinbase). The #433 n=2 fix itself is complete and proven.

## Tests / build

- `cargo test -p bth-consensus-scp`: 100 passed (incl #420 fork tests)
- `cargo test -p botho --lib`: 976 passed (consensus module: 62 passed)
- `cargo fmt --all --check`: clean
- `cargo clippy -p botho --lib`: no errors, no new warnings (pre-existing `println!`/missing-docs warnings unchanged)
- `pnpm -C web typecheck`: green

## Files touched

- `botho/src/commands/run.rs` — seed initial quorum from connected peers
- `botho/src/consensus/service.rs` — `is_transitional_solo()` guard + regression test
- `web/packages/wasm-signer/test/mine-after-join-live-node.test.ts` — updated #433 status comment (soak stays gated on the separate n=3 C-coinbase flake)

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -452,8 +452,31 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     let local_peer_id = *discovery.local_peer_id();
     let node_id = peer_id_to_node_id(&local_peer_id);
 
-    // Build SCP quorum set from connected peers (or just ourselves for solo mining)
-    let scp_quorum_set = build_scp_quorum_set(&quorum, &local_peer_id);
+    // Build the initial SCP quorum set.
+    //
+    // Issue #433: peers that connect during the pre-consensus "wait for peers"
+    // loop above have their `PeerDiscovered` events CONSUMED by that loop — the
+    // main event loop never sees them and therefore never calls
+    // `reconfigure_quorum` for them. If we seed the consensus service from the
+    // *static config* (`build_scp_quorum_set`, which yields a 1-of-1 solo set
+    // for a `recommended`/`min_peers=1` node), the node starts in solo mode even
+    // though a peer is already connected. The participation gate
+    // (`set_connected_peers` below) then opens (connected >= min_peers) while the
+    // quorum is still 1-of-1, so the very next consensus tick takes the solo
+    // direct-externalize path and the node mines a divergent solo chain forever
+    // (no further `PeerDiscovered` ever arrives, so the quorum is never
+    // reconfigured out of 1-of-1). Two such nodes fork.
+    //
+    // Fix: seed the initial quorum from the peers ALREADY CONNECTED at startup
+    // (the same `rebuild_scp_quorum_set` used on every later peer event), so a
+    // node that has already peered starts in federated SCP (e.g. 2-of-2), never
+    // solo. A genuine lone node (no peers connected) still yields a 1-of-1 solo
+    // quorum and mints solo — no regression for `min_peers == 0`.
+    let scp_quorum_set = if discovery.peer_count() > 0 {
+        rebuild_scp_quorum_set(&config, &discovery, &local_peer_id)
+    } else {
+        build_scp_quorum_set(&quorum, &local_peer_id)
+    };
 
     // Capture the starting height before chain_state is moved into the
     // consensus service; the sync state machine needs it to know how far
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -700,6 +700,26 @@ impl ConsensusService {
         self.quorum_set.threshold == 1 && self.quorum_set.members.len() == 1
     }
 
+    /// Issue #433: are we in a *transitional* solo state — a peer is connected
+    /// but the SCP quorum has not yet been reconfigured out of 1-of-1?
+    ///
+    /// A node that expects peers (`min_peers >= 1`) and currently has at least
+    /// one connected (`connected_peers >= 1`) but is still holding a solo
+    /// (1-of-1) quorum set is mid-transition: it must run federated SCP with
+    /// the connected peer, not directly solo-externalize. Taking the solo
+    /// path here produces a divergent solo chain and forks once both peered
+    /// nodes do it. This is the exact race from #433 (a peer that connected
+    /// during the pre-consensus startup window leaves the gate open while
+    /// the quorum is still 1-of-1). We block the solo direct-externalize
+    /// until the quorum reconfigures (the run loop calls
+    /// `reconfigure_quorum` on peer events; the startup path now seeds the
+    /// quorum from connected peers — see `commands::run`). A genuine lone
+    /// node (`min_peers == 0`) is never transitional: it has no peers to
+    /// wait for and mints solo.
+    fn is_transitional_solo(&self) -> bool {
+        self.min_peers >= 1 && self.connected_peers >= 1 && self.is_solo_mode()
+    }
+
     /// Propose pending values to SCP
     #[instrument(
         name = "consensus.propose_values",
@@ -728,6 +748,23 @@ impl ConsensusService {
                 min_peers = self.min_peers,
                 connected_peers = self.connected_peers,
                 "Withholding propose: not enough connected peers for quorum (issue #428)"
+            );
+            return;
+        }
+
+        // Transitional-solo guard (issue #433): a node that expects peers and
+        // has one connected, but whose SCP quorum is still 1-of-1, must NOT take
+        // the solo direct-externalize path below — that produces a divergent
+        // solo chain and forks once both peered nodes do it. Withhold the block
+        // until the quorum reconfigures out of solo (the run loop reconfigures
+        // on peer events; startup seeds the quorum from connected peers). Values
+        // stay queued in `pending_values`, so nothing is lost.
+        if self.is_transitional_solo() {
+            debug!(
+                min_peers = self.min_peers,
+                connected_peers = self.connected_peers,
+                "Withholding propose: peer connected but quorum still solo \
+                 (awaiting 1-of-1 -> N-of-N reconfig, issue #433)"
             );
             return;
         }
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!** Rigorous consensus/safety review complete — the n=2 startup solo-mode fork (#433) is closed.

## Verdict
The fix correctly eliminates the 1-of-1 → 2-of-2 transition fork via two complementary mechanisms, verified at the code and test level.

### (1) Fork race closed — no remaining window
- **Primary (run.rs:475-479)**: the initial SCP quorum is now seeded from peers already connected at startup. `discovery.peer_count() > 0` → `rebuild_scp_quorum_set` (federated, e.g. 2-of-2) instead of `build_scp_quorum_set` (static 1-of-1 solo). A node that already peered boots federated → `is_solo_mode()` is false → never reaches the solo direct-externalize path.
- **Defense-in-depth (service.rs:719-720, 762)**: `is_transitional_solo() = min_peers >= 1 && connected_peers >= 1 && is_solo_mode()`. This guard runs at `propose_pending_values:762`, **before** the solo direct-externalize path at `:848`. So even if a node transiently boots 1-of-1 with a peer connected (the original race), it returns early and never solo-externalizes. The only exit from solo is `reconfigure_quorum` flipping `is_solo_mode()` to false.
- Together: once `connected_peers >= 1` and `min_peers >= 1`, a node can NEVER take the solo direct-externalize path. No window remains.

### (2) min_peers==0 not regressed
`is_transitional_solo()` short-circuits on `min_peers >= 1`, so a genuine single-node network (`min_peers == 0`) is never transitional and mints solo. Startup seeding with no peers still yields 1-of-1 via `build_scp_quorum_set`. Covered by `participation_gate_blocks_solo_block_until_peers_connected` (asserts min_peers==0 always externalizes solo) — passing.

### (3) No deadlock + later-join reconfig intact
- Seeding from connected peers IS the deadlock fix: a peered node boots federated and does not wait for a reconfig that never comes (the bug was that the wait-loop consumed `PeerDiscovered` so no reconfig fired).
- The later-join path (run.rs:719-745) is untouched: `PeerDiscovered` in the main loop still updates `set_connected_peers` + `rebuild_scp_quorum_set` + `reconfigure_quorum`. A peer connecting after consensus start still reconfigures normally. The startup seed uses the identical `rebuild_scp_quorum_set`, so the two paths produce consistent quorums.

### (4) No cascade regression (#420/#429/#430/#434)
- `cargo test -p bth-consensus-scp`: **100 passed** (incl #420 no-fork tests).
- `cargo test -p botho --lib consensus::service::tests`: **18 passed**, incl `two_minters_converge_on_single_value_no_fork`, `t5_two_minters_make_sustained_progress_no_wedge`, the #429 gate test, the #430 drift tests (t1/t2/t3), and the #434/#432 fault-model path.
- The #429 gate (`should_propose_this_round`) and #430 `reconfigure_quorum` slot-boundary logic are unchanged.

### (5) Regression test is genuine (not a tautology)
Verified empirically: with the `is_transitional_solo()` guard neutralized, `transitional_solo_does_not_directly_externalize` **FAILS** at `service.rs:1488` ("transitional-solo node must NOT directly externalize"). With the guard, it passes. It exercises the real transition: starts from a genuine 1-of-1 `solo_service()`, asserts the solo path is withheld pre-reconfig AND that values stay queued, then `reconfigure_quorum`s to 2-of-2 and asserts the node ballots via federated SCP (`proposed_values` non-empty, `externalized` still none) rather than direct-externalizing.

### (6) Empirical confidence
Builder reports pre-fix 2/8 forks, post-fix 10/10 clean. For a ~25% intermittent bug, 10 clean runs by luck if unfixed is ≈ 0.75^10 ≈ 5.6%. Combined with the identified root cause + a regression test proven to fail pre-fix, this is convincing evidence the fork is closed.

### (7) #433 fixed; #396 correctly stays gated
The n=2 fork is fixed on its own merits (above). The residual #396 flake is the FINAL step (C winning its own coinbase), failing ~1/3 due to a SEPARATE pre-existing n=3 SCP-slot drift ("future slots" discards at the join boundary, #421/#427-family) — convergence (the n=2 fork's domain) now always succeeds. The soak remains gated (`&& false`, line 252) per the STOP guardrail; the `.test.ts` change is comment-only (documents what's fixed vs. what remains). Not requiring #396 green to approve.

### (8) Clean diff + CI
- Only 3 files touched: `run.rs` (+25/-2), `service.rs` (+102, guard + test), `mine-after-join-live-node.test.ts` (comment-only). No stray files (the `.loom-managed` artifact in the worktree is untracked and not part of the PR).
- `cargo fmt --all --check`: clean.
- All CI checks **pass** (Benchmarks, Fuzz Build Guard, rustfmt). Skipping fuzz jobs are conditional and not triggered by this diff. Merge state **CLEAN**.

head_sha: 32487a24be3db480b2a6a328d72e66bf97c395b7

PR #440, opened 2026-06-20
feat(network): persist libp2p node key for stable peer IDs across restarts (#439)

### LINKED ISSUE #439: consensus/network: persist libp2p node key — peer IDs are ephemeral per restart, breaking DNS-seed discovery
## Summary
botho nodes do not persist their libp2p network identity key, so a **new peer ID is generated on every process start**. This makes **DNS-seed discovery non-durable** (the `seeds.testnet.botho.io` TXT records pin peer IDs, which go stale on any node restart) and also resets peer-level identity state (reputation/bans) across restarts.

## Evidence (live testnet, 2026-06-19/20)
- The DNS-seed code path WORKS: with `[network.dns_seeds] enabled = true` + empty `bootstrap_peers`, a node queries `seeds.testnet.botho.io`, parses `PEER_ID@ADDRESS:PORT` into multiaddrs, and dials them ("Discovered 3 seeds via DNS", "Using 3 bootstrap peers (DNS: true)", "Dialing bootstrap peer: /dns4/seed.botho.io/tcp/17100/p2p/12D3Koo…").
- BUT a DNS-only node failed to actually connect (`peerCount=0`): it dialed the **pinned peer IDs from the TXT records, which no longer matched** the live nodes' current peer IDs → libp2p peer-ID mismatch → connections rejected.
- The nodes' peer IDs differ from the IDs collected at an earlier boot, and there is **no persistent key file** in `~/.botho/<network>/` (only `config.toml` + `ledger/`). A single startup even logs multiple distinct `12D3Koo…` IDs.
- The static bootstrap (`/ip4/IP/tcp/17100`, no `/p2p/<id>`) is immune (no pinned ID), which is why the 3-node fleet peers fine via static config.

## Impact
- **DNS-seed discovery (the whole `dns_seeds.rs` feature) is effectively unusable** for a real deployment: any node restart (deploy, crash, reboot) rotates its peer ID and invalidates its published TXT record, so DNS-only joiners get peer-ID-mismatch failures. Cloudflare TXT records for `seeds.testnet.botho.io` were published and verified resolving, but cannot stay valid.
- Peer reputation / ban state keyed by peer ID does not persist across restarts (a node "becomes a new peer" each boot).
- Pinned-peer-ID multiaddrs anywhere (DNS seeds, hardcoded seeds with `/p2p/<id>`) are fragile.

## Fix
Persist the libp2p `Keypair` to disk and load it on startup, so the peer ID is stable across restarts. Suggested: store an ed25519 key at `~/.botho/<network>/node_key` (0600), generate on first run, load thereafter. Where the swarm keypair is created in `botho/src/network/` (and the run/node startup), read-or-create the persisted key instead of generating an ephemeral one. Add a config/CLI override if useful (`network.node_key_path`).

## Acceptance
- A node's peer ID is identical across restarts (same key file).
- After publishing a node's stable peer ID to `seeds.testnet.botho.io`, a fresh node with `dns_seeds.enabled = true` and empty `bootstrap_peers` discovers and CONNECTS via DNS (peerCount ≥ 1) — and still connects after the seed nodes restart.
- Investigate/clean up the "multiple peer IDs logged per startup" observation (ensure one canonical identity).

## Context
Found while wiring Cloudflare DNS-seed records for the redeployed testnet (#323). Cloudflare API access + TXT records + the DNS discovery code path are all set up/verified; this key-persistence gap is the remaining blocker for functional DNS-seed discovery. Related: #323, `botho/src/network/dns_seeds.rs`, `botho/src/network/seeds.rs`, `botho/src/config.rs` (`bootstrap_peers_async`), `botho/src/commands/run.rs`.

rjwalters:
Closes #439

## Problem
botho generated a fresh libp2p `Keypair` on every process start, so the node's peer ID changed on every restart. This broke DNS-seed discovery (the `seeds.testnet.botho.io` TXT records pin peer IDs, which went stale on any restart) and reset peer reputation/ban state each boot. Confirmed live: a DNS-only node dialed stale pinned IDs -> peer-ID mismatch -> 0 peers.

## The "multiple peer IDs per startup" finding
`NetworkDiscovery` created **two** distinct ephemeral keypairs per startup:
1. In the constructor (`identity::Keypair::generate_ed25519()`), used only to derive the logged `local_peer_id`.
2. A **second** one via `SwarmBuilder::with_new_identity()` for the actual swarm.

`start()` then overwrote `local_peer_id` with the swarm's id. So a single startup logged a peer ID (from keypair #1) that did not even match the swarm's real identity (keypair #2) — the "multiple distinct 12D3Koo… per startup" observation in the issue. This is now consolidated to ONE canonical keypair.

## Fix
- **`botho/src/network/node_key.rs`** (new): `load_or_create_keypair(path)` — reads the key if present, else generates an ed25519 key and persists it. Stored at `<data_dir>/node_key` (next to `ledger/` + `config.toml`), `0600` perms, parent dirs created as needed. The private key is never logged (only the derived peer ID).
  - **Serialization**: libp2p protobuf keypair encoding (`Keypair::to_protobuf_encoding` / `from_protobuf_encoding`) — self-describing (records key type), lossless round-trip. libp2p 0.56.
- **`discovery.rs`**: new `NetworkDiscovery::with_keypair(keypair, …)`; `start()` now uses `SwarmBuilder::with_existing_identity(self.keypair.clone())`. The logged `local_peer_id` and the swarm's peer ID are now the same key and stable across restarts. Existing `new`/`with_rate_limit_config` constructors keep generating ephemeral keys (preserves test behavior).
- **`run.rs`**: load-or-create the key on startup; path defaults to `<data_dir>/node_key`, overridable via `network.node_key_path`.
- **`config.rs`**: optional `node_key_path: Option<PathBuf>` field + `node_key_path_from_config()` helper.

## Hard constraints honored
- Each node still gets a UNIQUE key (generate-on-first-run per data dir; no hardcoded/shared key).
- No-key node generates + persists one; existing-key node loads it.
- Network-identity only — consensus stack untouched.
- Key file `0600`; private key never logged.

## Tests (`node_key.rs`)
- `generates_on_first_call_and_persists_file`
- `same_file_yields_same_peer_id_across_calls` (restart -> identical peer ID)
- `distinct_files_yield_distinct_peer_ids`
- `protobuf_serialization_round_trips`
- `persisted_key_has_0600_permissions`

## Results
- `cargo test -p botho`: all pass (one stress test initially failed due to a disk-full ENOSPC in the env; passes after freeing space — unrelated to this change, it touches no networking).
- `cargo test -p bth-consensus-scp`: 100 passed.
- build / fmt clean; clippy clean on new code (the 8 pre-existing `println!` clippy errors in `fingerprint.rs` tests exist on `main`).
- No web files touched.

## Deployment note
After merge, deploy this and restart the seed nodes to give them STABLE peer IDs. Then refresh the `seeds.testnet.botho.io` TXT records ONCE with the new stable IDs, after which DNS-seed discovery becomes durable across restarts.

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -243,7 +243,21 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     let rate_limit_peers = GossipConfig::default().max_connections.max(1) as u32;
     let rate_limit_config =
         PeerRateLimitConfig::for_slot_duration(effective_slot_secs, rate_limit_peers);
-    let mut discovery = NetworkDiscovery::with_rate_limit_config(
+
+    // Load (or, on first run, generate + persist) the node's identity keypair so
+    // its peer ID is STABLE across restarts (issue #439). An unstable peer ID
+    // breaks DNS-seed discovery (TXT records pin peer IDs) and resets peer
+    // reputation/ban state on every boot. The key lives at <data_dir>/node_key
+    // by default, or at the configured `network.node_key_path` override.
+    let node_key_path = config
+        .network
+        .node_key_path
+        .clone()
+        .unwrap_or_else(|| crate::config::node_key_path_from_config(config_path));
+    let node_keypair = crate::network::load_or_create_keypair(&node_key_path)?;
+
+    let mut discovery = NetworkDiscovery::with_keypair(
+        node_keypair,
         config.network.gossip_port(network_type),
         bootstrap_peers,
         rate_limit_config,
--- a/botho/src/config.rs
+++ b/botho/src/config.rs
@@ -79,6 +79,13 @@ pub struct NetworkConfig {
     #[serde(default)]
     pub dns_seeds: DnsSeedConfig,
 
+    /// Optional override for the persistent libp2p node identity key path
+    /// (issue #439). If unset, the key lives at `<data_dir>/node_key` alongside
+    /// the ledger and config. The key is generated on first run and loaded
+    /// thereafter so the node's peer ID is stable across restarts.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub node_key_path: Option<PathBuf>,
+
     /// Quorum configuration
     #[serde(default)]
     pub quorum: QuorumConfig,
@@ -580,6 +587,8 @@ impl Default for NetworkConfig {
             cors_origins: default_cors_origins(),
             bootstrap_peers: Vec::new(), // Uses DNS discovery or network-specific defaults
             dns_seeds: DnsSeedConfig::default(),
+            node_key_path: None, // Defaults to <data_dir>/node_key
+
             quorum: QuorumConfig::default(),
             api_keys: Vec::new(),
             max_connections_per_ip: default_max_connections_per_ip(),
@@ -693,6 +702,16 @@ pub fn ledger_db_path(network: Network) -> PathBuf {
 /// Get the ledger database path from config file path
 pub fn ledger_db_path_from_config(config_path: &Path) -> PathBuf {
     config_path.parent().unwrap_or(config_path).join("ledger")
+}
+
+/// Get the persistent libp2p node-key path from the config file path (issue
+/// #439).
+///
+/// Defaults to `<data_dir>/node_key` (the data dir is the directory containing
+/// `config.toml`, the same dir that holds `ledger/`). This is the file that
+/// stores the node's identity keypair so its peer ID is stable across restarts.
+pub fn node_key_path_from_config(config_path: &Path) -> PathBuf {
+    config_path.parent().unwrap_or(config_path).join("node_key")
 }
 
 /// Get the wallet database path for a network
--- a/botho/src/network/discovery.rs
+++ b/botho/src/network/discovery.rs
@@ -331,6 +331,13 @@ pub struct BothoBehaviour {
 
 /// Network discovery and gossip service
 pub struct NetworkDiscovery {
+    /// Persistent libp2p identity keypair (issue #439).
+    ///
+    /// This is the single, canonical identity for the node: it is used both to
+    /// derive `local_peer_id` and to build the swarm in [`start`], so the
+    /// logged peer ID always matches the swarm's actual peer ID. Persisting it
+    /// to disk keeps the peer ID stable across restarts.
+    keypair: identity::Keypair,
     /// Local peer ID
     local_peer_id: PeerId,
     /// Gossip port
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed carefully as network-identity code. The fix is correct, well-tested, and scoped strictly to network identity (consensus stack untouched).

## Critical: unique key per node — PASS
`load_or_create_keypair` generates a random ed25519 key via `Keypair::generate_ed25519()` **only** when the file is absent, then persists it; on subsequent runs it loads. There is **no hardcoded/shared key constant**. The path defaults to `<data_dir>/node_key` (per-network, per-data-dir), so distinct nodes get distinct keys. Verified by `distinct_files_yield_distinct_peer_ids`. No quorum-collapse risk.

## Persistence correctness — PASS
- Same file → same peer ID across loads (`same_file_yields_same_peer_id_across_calls`).
- Lossless round-trip via libp2p 0.56 `to_protobuf_encoding`/`from_protobuf_encoding` (`protobuf_serialization_round_trips`).
- File written `0600` (`mode(0o600)` + verified by `persisted_key_has_0600_permissions`), parent dirs via `create_dir_all`, `sync_all()` for durability.
- Private key **never logged** — only `PeerId::from(keypair.public())` is emitted at INFO.

## Two-keypair consolidation — FIXED
`with_keypair` derives `local_peer_id` from the single canonical keypair, and `start()` now builds the swarm via `SwarmBuilder::with_existing_identity(self.keypair.clone())` (was `with_new_identity`, which minted a throwaway second key). The `debug_assert_eq!(self.local_peer_id, *swarm.local_peer_id())` guards against future drift. Logged ID now matches the real swarm identity.

## No consensus/network regression — PASS
- `cargo test -p botho --lib`: **985 passed, 0 failed**
- `cargo test -p bth-consensus-scp`: **100 passed, 0 failed** (consensus untouched)
- `cargo build -p botho` clean; clippy shows only the pre-existing `event_tx`/`subscribe_all` warnings on main (not introduced here).

## Tests genuine — PASS
All 5 node_key tests exercise real behavior (generate-on-first-call, stable-reload, distinct-dirs, protobuf round-trip, 0600 perms). The ENOSPC stress-test blip noted by the builder is environmental (disk full) and touches no networking — confirmed not a real failure.

## Config compat — PASS
`node_key_path: Option<PathBuf>` with `#[serde(default, skip_serializing_if = "Option::is_none")]` — existing `config.toml` files without the field deserialize to `None` and fall back to `<data_dir>/node_key`.

## CI / diff — CLEAN
All required checks pass (Benchmarks, Fuzz Build Guard, rustfmt; fuzz jobs `skipping` is path-conditional, not a failure). `mergeStateStatus: CLEAN`. No `.loom-managed`/stray files in the diff.

Head: bad8d5b9eff92a4c55232686f3bb6ec5f52498e9

Issue #441, opened 2026-06-20
epic: Botho product architecture — mobile app, managed mining rigs, and operator-curated trust


rjwalters:
> **Type:** Architecture EPIC (vision-level design proposal) — `loom:architect`.
> **Status:** Awaiting human approval. NOT ready-to-build. No source is modified by this issue.
> This decomposes the maintainer's product vision into phases; each phase becomes its own implementation issue *after* approval. Grounded in `origin/main` as of `509f2a2`.

## North star (maintainer-stated vision)

The **Botho mobile app is the primary entry point** — people pay each other from their phones. The app offers two modes:

1. **Paid managed mining rig** (~$50/month) — the app provisions a Botho mining node for the user on AWS (datacenter near them) that mines on their behalf; the app is a mobile wallet pointed at that node.
2. **Connect to a trusted node** (free) — to just transact, point the app's wallet at a friend's / chosen node's RPC.

**"Bring Botho to your village":** the unit of adoption is a community around a node operator. One *village host* pays for a cloud mining node; they and all their friends thin-client into it — hub-and-spoke (one mining node, many thin wallets). Anyone may still run their own consensus nodes; the managed rig is a convenience product (Botho-as-a-service), not the only path.

---

## 1. Target architecture

```
                         ┌──────────────────────────────────────────┐
   Mobile App (primary)  │  React Native + UniFFI  (mobile/app)      │
   non-custodial wallet  │  keys on device (iOS Keychain/biometric)  │
                         └───────────────┬──────────────┬───────────┘
                                         │ JSON-RPC      │ JSON-RPC
                       ┌─────────────────▼───┐      ┌────▼─────────────────┐
            MODE 1     │  Managed mining rig │      │  Trusted node (free) │  MODE 2
          ($50/mo)     │  EC2, provisioned   │      │  friend / village    │
                       │  by Botho-as-a-svc  │      │  host's node         │
                       └─────────┬───────────┘      └─────────┬────────────┘
                                 │  libp2p gossip + SCP        │
                                 └───────────┬─────────────────┘
                                             ▼
                    ┌───────────────────────────────────────────────┐
                    │  Consensus / validator set                     │
                    │  operator-curated quorum slices (FBA)          │
                    │  optimistic-trust default + reactive shun       │
                    └───────────────────────────────────────────────┘

   Roles:
   • Thin clients  — most users; wallet only; NOT in consensus (small Sybil surface)
[...]

PR #443, opened 2026-06-20
feat(pow): replace SHA-256 PoW with RandomX (ASIC-resistant CPU mining)

### LINKED ISSUE #442: consensus: wire RandomX into PoW (replace SHA-256) — realize ASIC-resistant CPU mining
## Goal
Wire the **already-declared `randomx-rs` dependency** (`Cargo.toml:140`) into the proof-of-work, replacing the current single **SHA-256** PoW. This realizes the intended design: **ASIC/GPU-resistant, CPU-egalitarian mining** with a **fair latency edge for well-connected node operators** — the basis for the managed-rig / "village" economic model (epic #441). Pre-mainnet consensus change.

## Current state (the gap)
- `randomx-rs = "1.3"` is declared but **never imported/used** in any `.rs` (phantom dependency).
- The live PoW is **single SHA-256**: `botho/src/block.rs:78` `pow_hash()` = `Sha256(nonce ‖ prev_block_hash ‖ minter_view_key ‖ minter_spend_key)`; `is_valid_pow()` checks `u64(hash[0..8]) < difficulty`. Same pattern in `MintingTx::pow_hash` (`block.rs:236`). Validation at `consensus/src/.../validation.rs:296` ("verify PoW last").
- Consequence (see #441 economics): SHA-256 is the most ASIC/GPU-friendly hash → one GPU/ASIC captures ~all coinbases → cheap CPU rigs earn ~$0; the egalitarian-mining premise collapses.

## Design (decide during build)
1. **Hash function:** replace `pow_hash`'s SHA-256 with RandomX (via `randomx-rs`). Keep the preimage = `nonce ‖ prev_block_hash ‖ minter_view_key ‖ minter_spend_key` as the RandomX *input* so per-block binding AND the node-operator latency edge are preserved (miner can't start block N+1 until it has N's hash).
2. **RandomX key/seed rotation:** RandomX hashing is parameterized by a *key* that seeds the cache/dataset. It must rotate deterministically from the chain (Monero rotates the seedhash every 2048 blocks). Choose botho's rule (e.g. seed = block hash at `height - (height mod K)` for some K; all nodes derive it identically). Document K and the rotation/lookahead handling at boundaries.
3. **Light vs fast mode (RAM) — EMPIRICALLY MEASURED (t4g.medium Graviton2, RandomX v2.0.1 JIT):** miners run **fast mode** (~2 GB dataset) = **~415 H/s @ 2 vCPU (~207 H/s/vCPU, linear)**; verifiers run **light mode** (~256 MB cache) = **18.5 ms/hash** (~54 verifies/s/core — ample for 1 PoW-verify per 5s block). IMPLICATION: **t4g.small (1.8 GB) CANNOT fast-mine** (needs ~2 GB dataset) but CAN verify → minimum *mining* tier is **t4g.medium** (4 GB; the efficiency sweet spot at 16.9 H/s per $/mo, raw ~$24.5/mo → ~$50 retail). t4g.small can still be a validator/thin-serving node. Use JIT (`randomx-rs` JIT flag) — interpreter is ~6× slower (73 vs 415 H/s). (Full economics + tier table in #441.)
4. **Difficulty recalibration:** RandomX hashrate is orders of magnitude below SHA-256, so `initial_difficulty`, the difficulty target, and bounds (`monetary.rs`) must be retuned for RandomX H/s at the 5s target block time. The `DifficultyController` adapts over time, but the *initial* values must be RandomX-appropriate or blocks stall/flood at genesis.
5. **Verification cost / DoS:** RandomX verify (light) is ~ms-scale vs SHA-256 ~µs. At 5s blocks (1 verify/block) this is fine; keep PoW verification *last* (after cheap structural checks — already the case) so invalid-PoW spam is cheap to reject. Thin clients don't verify PoW (they trust their node), so only full nodes carry the RandomX cost.
6. **VM lifecycle:** manage RandomX VM/cache/dataset init (expensive, ~seconds) — init once, reuse across hashes, re-init only on key rotation. Avoid per-hash allocation.

## Hard constraints
- Deterministic + identical across nodes (same seed-rotation rule, same RandomX flags affecting output). RandomX output must not depend on fast-vs-light mode (it doesn't — same result, different speed) — verify this holds with `randomx-rs` flags.
- Consensus-breaking → requires fresh genesis + coordinated testnet redeploy (we have the flow). Coordinate per the no-hard-forks/pre-mainnet stance.
- Preserve the prev_block_hash latency edge (don't drop prev_block_hash from the preimage).
- Don't regress the consensus stack (#420/#429/#430/#434/#435/#437/#440).

## Acceptance criteria
1. `pow_hash`/`is_valid_pow` (header + minting tx) use RandomX via `randomx-rs`; SHA-256 PoW removed.
2. Deterministic chain-derived RandomX seed with a documented rotation rule; all nodes agree.
3. Difficulty constants retuned so a CPU-mined testnet holds ~5s blocks from genesis.
4. Miners use fast mode, verifiers light mode; documented RAM/instance requirements.
5. Tests: PoW round-trip/determinism; a known-seed RandomX test vector; difficulty-at-genesis sanity; light-vs-fast produce identical hashes.
6. Empirically: a multi-node CPU-mined testnet advances at ~5s blocks; no GPU/ASIC advantage (by construction).

## References
- Economics + benchmark: #441 (open question (a)). Realizes the Mode-1 mining value prop.
- Code: `botho/src/block.rs` (pow_hash/is_valid_pow, ×2), `botho/src/monetary.rs` (difficulty), `consensus` validation, `Cargo.toml:140` (dep present).
- Prior art: Monero RandomX (seedhash rotation every 2048 blocks; light/fast mode split).

rjwalters:
Closes #442

Replaces the single **SHA-256** PoW with **RandomX** (Monero's CPU-egalitarian, ASIC/GPU-resistant PoW) via the already-declared `randomx-rs` dependency, realizing the Mode-1 mining value prop in epic #441. **Consensus-breaking — requires a fresh genesis + coordinated testnet redeploy** (expected pre-mainnet).

## How RandomX is wired in

New `botho::pow` module (`botho/src/pow.rs`):

- **Hash:** `pow_hash(seed_key, preimage)` runs RandomX over the **same preimage** as the legacy PoW: `nonce ‖ prev_block_hash ‖ minter_view_key ‖ minter_spend_key`. `prev_block_hash` is preserved → per-block binding + node-operator latency edge intact. Target check unchanged: `u64::from_be_bytes(hash[0..8]) < difficulty` (lower = better).
- **Seed rotation (K = 2048):** `seed_key = SHA256("BOTHO_RANDOMX_SEED_V1" ‖ epoch)`, `epoch = height / SEED_ROTATION_INTERVAL`. A **pure function of height** — no chain lookup, so the SCP tip-agnostic intrinsic validity function stays a pure function of the value, every node agrees trivially, and there are no lookahead/boundary hazards. Epoch 0 is the fixed genesis bootstrap key. ASIC resistance comes from RandomX's per-hash random program, so a deterministic height-rotated key is as ASIC-resistant as Monero's past-block-hash seed while being far safer for consensus.
- **Light vs fast:** verifiers use **light mode** (recommended flags incl. JIT, no full-mem, ~256 MB); miners use **fast mode** (recommended | FULL_MEM, ~2 GB dataset). RandomX output is **identical** across modes/flags — proven by a hard-coded known-answer vector and a light==fast test. On macOS/arm64 the JIT needs `FLAG_SECURE` (provided by `get_recommended_flags()`); bare `FLAG_JIT` faults.
- **VM lifecycle:** one process-cached light VM, rebuilt only on seed rotation; miners build the fast dataset once per seed epoch and reuse it across all nonces/blocks (never per-hash).

`block.rs`: `BlockHeader::pow_hash` / `MintingTx::pow_hash` / `is_valid_pow` / `verify_pow` / `pow_priority` now use RandomX; both `pow_hash` sites stay mutually consistent (same preimage + same per-height seed, header `height` == tx `block_height` enforced at apply). SHA-256 PoW removed (SHA-256 still used for non-PoW hashing).

`minter.rs`: mint loop hashes via RandomX fast mode, rebuilding the dataset only on seed-epoch change.

## Difficulty recalibration

`INITIAL_DIFFICULTY` retuned from `0x00FF_FFFF_FFFF_FFFF` (~256 hashes/block) to **`9_000_000_000_000_000`** (9e15). Expected hashes/block = `2^64 / difficulty ≈ 2049`. At RandomX ~415 H/s on a 2-vCPU t4g.medium that's ≈ **4.9 s/block for a single node from genesis** (≈ 2.5 s for two nodes); the `EmissionController` adapts from there. All call sites flow through the one constant (`MAX_DIFFICULTY`, genesis chain state, ledger default). Empirically confirmed: a fast-mode VM at this difficulty found ~1 solution per ~2049 hashes (math holds).

## Tests

`botho::pow` unit tests (all green): seed-rotation epochs + boundary, genesis bootstrap seed, preimage layout, big-endian target value, verify determinism/round-trip, **known-answer RandomX test vector** (guards against future algorithm/flag drift — any change is consensus-breaking), and **light == fast identical output**.

Heavy fast-dataset tests (`#[ignore]`d, run manually): minter fast==light, real single-node mine. Integration harnesses pin the ledger difficulty to the trivial target (`u64::MAX`) so RandomX PoW solves in one hash, keeping suites fast.

## Verification

- `cargo build --release -p botho` ✅ (RandomX C++ builds via cmake/cc; `randomx-rs = "1.3"` resolves to 1.4.x and builds cleanly on macOS/arm64; cache 2 vCPU build worked).
- `cargo test -p botho --lib` ✅ 990 passed, 2 ignored.
- `cargo test -p bth-consensus-scp` ✅ 100 passed.
- Integration suites ✅ (ledger_consistency 26, chain_sync_catchup 8, compact_block 16, rpc 39, tx_lifecycle 13).
- Real fast-mode mine empirically produces a valid PoW below genesis difficulty that light-mode verifies identically.

## Notes / follow-ups

- RandomX **fast dataset build is single-threaded (~50 s)** per seed epoch in this code. That's amortized over 2048 blocks (~2.8 h) and only at epoch rotation, so it's acceptable, but multithreaded dataset init (`randomx-rs` exposes per-thread `start`) would speed node startup — worth a follow-up. The `#[ignore]`d real-mine test can exceed a 120 s debug-build timeout for this reason; it is intentionally manual.
- CI must have a C++ toolchain + cmake for `randomx-rs` (already required transitively; release build verified locally).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -828,6 +828,7 @@ dependencies = [
  "rand_chacha 0.3.1",
  "rand_core 0.6.4",
  "rand_distr",
+ "randomx-rs",
  "rcgen",
  "reqwest",
  "ring",
@@ -2032,6 +2033,15 @@ name = "clear_on_drop"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "38508a63f4979f0048febc9966fadbd48e5dab31fd0ec6a3f151bbf4a74f7423"
+dependencies = [
+ "cc",
+]
+
+[[package]]
+name = "cmake"
+version = "0.1.58"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c0f78a02292a74a88ac736019ab962ece0bc380e3f977bf72e376c5d78ff0678"
 dependencies = [
  "cc",
 ]
@@ -7289,6 +7299,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "513962919efc330f829edb2535844d1b912b0fbe2ca165d613e4e8788bb05a5a"
 dependencies = [
  "rand_core 0.9.3",
+]
+
+[[package]]
+name = "randomx-rs"
+version = "1.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d108ec2b3f83ea06f8290685fa4a20c8f0803b1de43c2301639d7a0592474554"
+dependencies = [
+ "bitflags 1.3.2",
+ "cmake",
+ "libc",
+ "thiserror 1.0.69",
 ]
 
 [[package]]
--- a/botho/Cargo.toml
+++ b/botho/Cargo.toml
@@ -78,6 +78,11 @@ rand_distr = "0.4"
 hex = "0.4"
 sha2 = "0.10"
 bs58 = "0.5"
+
+# RandomX proof-of-work (ASIC/GPU-resistant, CPU-egalitarian).
+# Builds the RandomX C++ library at compile time (needs cmake + a C++
+# toolchain). "1.3" resolves to the latest 1.x (currently 1.4.x).
+randomx-rs = "1.3"
 siphasher = "1.0"
 
 # System
--- a/botho/src/block.rs
+++ b/botho/src/block.rs
@@ -74,21 +74,36 @@ impl BlockHeader {
         hasher.finalize().into()
     }
 
-    /// Compute the PoW hash (what minters are trying to get below target)
+    /// Compute the PoW hash (what minters are trying to get below target).
+    ///
+    /// This is **RandomX** (ASIC/GPU-resistant CPU PoW) over the preimage
+    /// `nonce ‖ prev_block_hash ‖ minter_view_key ‖ minter_spend_key`, keyed by
+    /// the chain-deterministic seed for this block's height (see
+    /// [`crate::pow`]). Verification uses RandomX **light mode**; it produces
+    /// the identical hash a miner's fast mode produces.
+    ///
+    /// The seed key is a pure function of `self.height`, so this remains a pure
+    /// function of the header — no chain lookup is required and every node
+    /// agrees on the result.
     pub fn pow_hash(&self) -> [u8; 32] {
-        let mut hasher = Sha256::new();
-        hasher.update(self.nonce.to_le_bytes());
-        hasher.update(self.prev_block_hash);
-        hasher.update(self.minter_view_key);
-        hasher.update(self.minter_spend_key);
-        hasher.finalize().into()
+        let seed = crate::pow::seed_key_for_height(self.height);
+        let preimage = crate::pow::pow_preimage(
+            self.nonce,
+            &self.prev_block_hash,
+            &self.minter_view_key,
+            &self.minter_spend_key,
+        );
+        crate::pow::verify_pow_hash(&seed, &preimage)
     }
 
-    /// Check if PoW is valid (hash < difficulty target)
+    /// Check if PoW is valid (hash < difficulty target).
+    ///
+    /// Big-endian read of the first 8 bytes of the RandomX output, compared to
+    /// the difficulty target (lower = better PoW) — same convention as the
+    /// legacy SHA-256 PoW.
     pub fn is_valid_pow(&self) -> bool {
         let hash = self.pow_hash();
-        let hash_value = u64::from_be_bytes(hash[0..8].try_into().unwrap());
-        hash_value < self.difficulty
+        crate::pow::pow_value(&hash) < self.difficulty
     }
 
     /// Create header for genesis block (defaults to testnet for backward
@@ -232,21 +247,29 @@ impl MintingTx {
     }
 
     /// Compute the PoW hash.
-    /// Uses minter keys to match BlockHeader::pow_hash for block validation.
+    ///
+    /// RandomX over `nonce ‖ prev_block_hash ‖ minter_view_key ‖
+    /// minter_spend_key`, keyed by the chain-deterministic seed for
+    /// `self.block_height`. This MUST stay byte-for-byte consistent with
+    /// [`BlockHeader::pow_hash`] — they use the same preimage and the same
+    /// per-height seed (header `height` == minting tx `block_height`, enforced
+    /// at block-apply time), so a header and its minting tx always agree on the
+    /// PoW.
     pub fn pow_hash(&self) -> [u8; 32] {
-        let mut hasher = Sha256::new();
-        hasher.update(self.nonce.to_le_bytes());
-        hasher.update(self.prev_block_hash);
-        hasher.update(self.minter_view_key);
-        hasher.update(self.minter_spend_key);
-        hasher.finalize().into()
+        let seed = crate::pow::seed_key_for_height(self.block_height);
+        let preimage = crate::pow::pow_preimage(
+            self.nonce,
+            &self.prev_block_hash,
+            &self.minter_view_key,
+            &self.minter_spend_key,
+        );
+        crate::pow::verify_pow_hash(&seed, &preimage)
     }
 
     /// Verify the PoW is valid
     pub fn verify_pow(&self) -> bool {
         let hash = self.pow_hash();
-        let hash_value = u64::from_be_bytes(hash[0..8].try_into().unwrap());
-        hash_value < self.difficulty
+        crate::pow::pow_value(&hash) < self.difficulty
     }
 
     /// Get the PoW hash value as u64 (lower = better, used for priority in
@@ -254,7 +277,7 @@ impl MintingTx {
     pub fn pow_priority(&self) -> u64 {
         let hash = self.pow_hash();
         // Invert so that lower hash = higher priority
-        u64::MAX - u64::from_be_bytes(hash[0..8].try_into().unwrap())
+        u64::MAX - crate::pow::pow_value(&hash)
     }
 
     /// Compute the hash of this minting transaction (for consensus)
--- a/botho/src/lib.rs
+++ b/botho/src/lib.rs
@@ -17,6 +17,7 @@ pub mod mempool;
 pub mod monetary;
 pub mod network;
 pub mod node;
+pub mod pow;
 pub mod rpc;
 pub mod telemetry;
 pub mod transaction;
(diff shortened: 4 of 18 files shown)


rjwalters:
✅ **Approved!** — RandomX PoW replacement reviewed with maximum rigor (most safety-critical consensus change since the fork fix). All seven determinism/consensus checks pass, all CI is green, merge state is CLEAN.

## 1. Cross-platform / cross-node determinism (THE critical check) — PASS

This is the existential risk and it is empirically closed:

- **`get_recommended_flags()` does NOT vary hash output across machines.** On this arm64 (Apple Silicon) box, recommended flags resolve to `FLAG_HARD_AES | FLAG_JIT | FLAG_SECURE`. I ran the canonical `(epoch-0 seed, zero preimage)` through **both** the recommended-flag path **and** the portable `FLAG_DEFAULT` path (soft-AES, interpreter, no JIT, no secure W^X). Both produced the **identical** hash `04779c03247d2b9f45bd745529ae60cc02ddab82d5eb3c2fdcb8b1fcaaa7dfb4`, matching the pinned `KNOWN_ANSWER_HEX`. This is the strongest possible local proof that hardware-AES vs soft-AES, JIT vs interpreter, and Argon2 variant flags are performance-only and never affect RandomX output — so an x86 operator and the arm64 Graviton seed nodes will agree byte-for-byte. (Probe code was reverted; worktree left clean.)
- **Cross-platform confirmed on actual x86 Linux CI.** `Generate Coverage Report` (which runs the test suite, including `test_known_answer_vector` and `test_light_equals_fast`) passed on the x86 Linux runner. The KAT passing on x86 Linux AND arm64 macOS against the same pinned hash is definitive cross-arch agreement.
- **light == fast:** `test_light_equals_fast` asserts the verifier's light VM and the miner's `FLAG_FULL_MEM` fast VM produce identical output. Passes locally.
- **Known-answer vector is real** (not self-referential): hex-decoded constant `KNOWN_ANSWER_HEX` cross-checked against a freshly computed hash; would catch any future randomx-rs algorithm bump or flag-output drift. `randomx-rs = "1.3"` is caret-pinned but `Cargo.lock` pins `1.4.1`, and the KAT guards drift regardless.

## 2. Seed-rotation purity + #420 invariant — PASS

`seed_key_for_height(h) = SHA256("BOTHO_RANDOMX_SEED_V1" ‖ (h/2048).to_le_bytes())` is a **pure function of height** — no chain/tip lookup. Validity stays a pure function of the value, so the #417 tip-dependence fork class is not reintroduced. Genesis bootstrap handled: heights `0..2048` are all epoch 0 (`test_genesis_bootstrap_seed_is_epoch_zero`, `test_seed_rotation_epochs` cover boundaries). Every validator derives the seed from the committed `height` field alone.

## 3. Both pow_hash sites consistent — PASS

`BlockHeader::pow_hash` (uses `self.height`) and `MintingTx::pow_hash` (uses `self.block_height`) call the identical `pow_preimage` + `verify_pow_hash` path. Header/minting equality is enforced by construction in `block_builder.rs` (header.height = minting_tx.block_height) AND defensively at acceptance in `ledger/store.rs:408-431` (height, prev_block_hash, difficulty, minter keys all cross-checked). SHA-256 PoW fully removed from both sites; SHA-256 retained only for non-PoW hashing (tx_root, tx hash, seed derivation). Preimage layout preserved byte-for-byte from the legacy PoW.

## 4. Difficulty sanity — PASS

`INITIAL_DIFFICULTY = 9e15` → expected hashes/block = 2^64/9e15 ≈ 2049 → ≈4.9s/block @415 H/s single node (verified arithmetically). Won't stall genesis, won't flood. The `EmissionController` adjustment is emission-based and correctly handles the inverted target semantics (higher value = easier; comment at block.rs:1056-1057). The legacy block-time `TARGET_BLOCK_TIME=20` only feeds the deprecated `calculate_new_difficulty` path, not the active controller. `MAX_DIFFICULTY = INITIAL_DIFFICULTY` correctly caps the easy end.

## 5. CI builds randomx-rs — PASS

`cargo build --workspace (Linux)` passed in 5m8s on x86 Linux — CI has cmake + C++ toolchain and compiles the RandomX C++ lib cleanly. Verified locally too (cmake 4.3.4, arm64 build succeeded).

## 6. No consensus regression — PASS

`cargo test -p bth-consensus-scp` → 100 passed / 0 failed locally. PoW change is properly isolated from the SCP layer (#420/#429/#430/#434/#435/#437/#440 untouched). Test-harness change pinning `TRIVIAL_DIFFICULTY = u64::MAX` is legitimate: tests still run real RandomX hashes through `verify_pow()`, just with a trivial target so one hash solves — PoW is exercised, not stubbed.

## 7. VM wrapper soundness — PASS

`SendCache(LightCache)` with `unsafe impl Send` is correct: the VM is only ever touched while holding `LIGHT_VM`'s `Mutex`, so access is serialized to one thread at a time (RandomX VM is single-thread-safe but not `Sync`; the Mutex provides the required exclusivity, the newtype asserts `Send`). Re-init happens only on seed rotation (`needs_rebuild` compares `seed_key`), never per-hash. Miner side mirrors this: `FastHasher` rebuilt only on epoch change (`minter.rs:276-297`), dataset built once per epoch and reused — not catastrophically per-hash.

## Test Execution

- `cargo test -p botho --lib pow::` — ✅ 7/7 passed (KAT, light==fast, seed rotation, genesis bootstrap, preimage layout, big-endian value, determinism roundtrip)
- `cargo test -p bth-consensus-scp` — ✅ 100 passed, 0 failed
- FLAG_DEFAULT vs recommended-flags hash equality probe — ✅ identical output (reverted after)
- ⏭️ Heavy `#[ignore]`d fast-dataset / real-mine tests skipped (>2min, RAM-heavy) — intentionally manual per PR notes

## Notes (non-blocking, already tracked in PR)

- Fast dataset build is single-threaded (~50s/epoch); amortized over 2048 blocks (~2.8h) so acceptable. Multithreaded init is a reasonable follow-up.
- Consensus-breaking → fresh genesis + coordinated testnet redeploy expected (per PR / memory: testnet is ephemeral).

Outstanding work. The determinism story is airtight: a pure-height seed, a real cross-checked known-answer vector, and empirical proof that recommended-vs-default flags produce identical output across both arm64 and x86. Ready to merge.

PR #450, opened 2026-06-20
fix(consensus): deterministic + peer-cached transfer-tx path; chain no longer halts on a transfer (#449)

### LINKED ISSUE #449: consensus: chain halts when a regular transfer tx enters the mempool (non-deterministic value priority + no peer tx-cache registration)
## Summary
**CRITICAL: the multi-node chain halts when a regular (non-coinbase) transfer transaction enters the mempool.** Coinbase-only blocks externalize fine and the chain advances; the instant a real transfer tx (e.g. a faucet payout) is pending, the SCP slot never externalizes and the chain freezes. This blocks ALL payments (faucet payouts, sends) — the testnet can mine but cannot transact. Reproduced live on the healthy full-mesh 3-node testnet (frozen at h32, mempoolSize=1, mintingActive=true).

This is the transfer-tx parallel of the minting-convergence work (#409/#419): the minting path was made deterministic + cache-registered during the consensus cascade, but the **transfer-tx path never got the equivalent fixes.**

## Root cause — two compounding defects (both localized)

**Defect A — non-deterministic value identity (`priority = SystemTime::now()`).**
`ConsensusValue::from_transaction` (`botho/src/consensus/value.rs:33-45`) sets `priority = SystemTime::now().as_secs()`. `ConsensusValue` derives `Eq/Ord/Hash` over ALL fields incl. `priority` (`value.rs:18`), and SCP nomination agrees on exact value identity. So the same tx becomes a DIFFERENT `ConsensusValue` on each node (and between ticks) → nodes propose non-identical values → nomination can never converge on a shared confirmed set. (Minting txs are immune: priority = deterministic PoW hash, `value.rs:48-54`.)

**Defect B — no peer cache registration for transfer txs.**
The SCP validity_fn (`service.rs:217-238`) does a HARD `tx_cache` lookup first (`service.rs:224`: "Transaction not in cache" error) before structural checks. The `NewTransaction` peer handler (`run.rs:622-637`) routes a received tx to the mempool ONLY — it never seeds the consensus `tx_cache`. Contrast `NewMintingTx` (`run.rs:638-672`) which calls `consensus.register_minting_tx` (the #409 fix). So peers fail the validity lookup on any nominate carrying the transfer value → drop it → quorum can't form on a set containing it.

Either defect alone stalls the slot; together they guarantee it. Confirmed via logs: coinbase-only slots externalize in 1-4s; the instant a transfer enters (slot 33, value_count jumps 1→7→…→100), the slot NEVER externalizes and `add_block`/`Building block` for that height is never reached (dies in nomination, before externalize — NOT an apply bug).

## Fix (localized, buildable — mirror the minting path)
1. **Deterministic transfer value identity.** In `ConsensusValue::from_transaction` (`value.rs:33-45`), replace `SystemTime::now()` with a deterministic, cross-node-identical priority. Use **`tx.fee`** (matches the combine_fn's documented "sort regular txs by priority (fee)" at `service.rs:266-271`, so higher-fee txs order first within a block — sensible) — or `tx.created_at_height` if fee isn't readily available. Thread the needed field into `from_transaction` (callers `run.rs:1422`, `service.rs:498` have the tx). Must be identical on all nodes for the same tx.
2. **Register transfer txs in the consensus cache on receipt.** Add `ConsensusService::register_transfer_tx` mirroring `register_minting_tx` (`service.rs:543-550`), and call it from the `NewTransaction` handler (`run.rs:622-637`) so every node seeds its `tx_cache` (gate on intrinsic/structural validity like the minting path). This makes the validity_fn cache lookup succeed on peers. Also seed the cache on local submit if not already (the local node already does via `submit_transaction`).

No SCP/architecture change. (Minor design note: the `priority` field feeds intra-block tx ordering via the combine_fn — using `fee` is the intended ordering, so it's a feature not just a fix.)

## Test-coverage gap (why this slipped)
No test externalizes a block CONTAINING a real transfer tx through the production `ConsensusService` multi-node path:
- `two_minters_converge_on_single_value_no_fork` (`service.rs:1503-1552`) proposes minting txs ONLY.
- `submit_transaction` (transfer path) is exercised only in a SOLO-mode test (`service.rs:1150`), where SCP is bypassed.
- The 5-node `e2e_consensus_integration` uses a SEPARATE harness (`tests/common/network.rs`) that masks BOTH defects: validity_fn = `|_| Ok(())` (accepts all, never hits the cache gate, `network.rs:356`) and its own ConsensusValue sets transfer `priority = created_at_height` (deterministic, `network.rs:408-411`).

**Add a regression test:** two real `ConsensusService` instances each `submit_transaction` a transfer tx (+ peer `register_transfer_tx`); assert the slot externalizes a block CONTAINING the tx and both nodes agree. Must fail pre-fix, pass post-fix.

## Acceptance criteria
1. `ConsensusValue::from_transaction` uses deterministic priority (identical across nodes for the same tx).
2. `register_transfer_tx` exists + is called from the `NewTransaction` peer handler (validity-gated); peers' `tx_cache` is seeded so the validity_fn lookup succeeds.
3. Regression test: 2 real `ConsensusService` instances converge + externalize a block containing a transfer tx (fails pre-fix, passes post-fix).
4. No regression to minting convergence (#409/#419/#420 etc.) or other tests; `cargo test -p botho` + `bth-consensus-scp` green.
5. **Live verification after redeploy:** a faucet payout / a wallet→wallet send confirms (mempool drains, recipient balance rises) on the multi-node testnet — the demo's faucet→balance→send→recipient flow completes.

## References
Blocks the mobile demo (#446/#441) and is a fundamental payments-correctness bug. Parallels #409 (minting-tx propagation) and #419 (deterministic minting value). Code: `botho/src/consensus/value.rs`, `botho/src/consensus/service.rs`, `botho/src/commands/run.rs`. (Supersedes the closed #448, which was the same halt mis-attributed to topology.)

rjwalters:
Closes #449

## Problem
The multi-node chain HALTED the instant a regular (non-coinbase) transfer tx entered the mempool. Coinbase-only slots externalized fine; any pending transfer froze SCP nomination (reproduced live: frozen at h32, mempoolSize=1). This blocked ALL payments (faucet payouts, sends) — the testnet could mine but not transact. This is the transfer-tx parallel of the minting-convergence work (#409/#419), which the transfer path never received.

## Root cause — two compounding defects (both localized)

**Defect A — non-deterministic value identity.** `ConsensusValue::from_transaction` set `priority = SystemTime::now().as_secs()`. `ConsensusValue` derives `Eq/Ord/Hash` over ALL fields incl. `priority`, and SCP nomination agrees on exact value identity, so the same tx became a DIFFERENT value on each node (and each tick) and nomination could never converge. (Minting was immune: priority = deterministic PoW hash.)

**Defect B — no peer cache registration.** The SCP validity_fn does a HARD `tx_cache` lookup before structural checks. The `NewTransaction` peer handler routed received txs to the mempool ONLY, never seeding the consensus `tx_cache` (contrast `NewMintingTx` -> `register_minting_tx`). So peers failed validity ("Transaction not in cache") on any nominate carrying the transfer value and SCP silently dropped it.

## Fix (mirrors the minting path)
1. **`value.rs`** — `from_transaction(tx_hash, fee)` now uses the tx **`fee`** as a deterministic priority (identical on all nodes for the same tx; matches the combine_fn's documented "sort regular txs by fee" ordering). No more `SystemTime::now()`.
2. **`service.rs`** — added `register_transfer_tx(tx_hash, tx_bytes)` (mirrors `register_minting_tx`), and threaded `fee` through `submit_transaction(tx_hash, fee, tx_data)` (local submit already seeds the cache).
3. **`run.rs`** — the `NewTransaction` peer handler now calls `consensus.register_transfer_tx(...)` after the mempool accepts the tx (mempool acceptance is the structural/UTXO/signature validity gate), so every node seeds its cache. Local submit path threads `tx.fee`.
4. **`service.rs` combine_fn (latent bug surfaced by the first multi-value block)** — the combiner returned `[coinbase, ...regular]` in fee-selection order, but SCP wraps the output directly into a `Ballot` whose values MUST be strictly ascending per `ConsensusValue::Ord` (`Ballot::is_values_sorted`). A block containing BOTH a coinbase and a transfer therefore panicked SCP's outgoing-message validation non-deterministically (depending on the coinbase's PoW-derived hash). The combined output is now `sort()` + `dedup()` by `Ord`. Minting-only blocks (single value) were never affected, which is why this only surfaced with transfers. Block tx ordering is decided downstream by the block builder, so this is a pure ballot-invariant normalization (validity unchanged).

## Tests (fail pre-fix, pass post-fix — verified by temporarily reverting each defect)
- `value::tests::test_from_transaction_deterministic` — same `(tx_hash, fee)` -> same value.
- `service::tests::two_nodes_externalize_block_containing_transfer_tx` — **the load-bearing one**: two real `ConsensusService` instances each get a transfer tx via the production path (`submit_transaction` + peer `register_transfer_tx`) and externalize a block CONTAINING the transfer, both agreeing (exactly one transfer value + one coinbase). Pre-fix this panics in nomination ("ballot values not sorted" / value churn) and never externalizes the transfer.
- `service::tests::register_transfer_tx_enables_peer_nominate_acceptance` — isolates Defect B: a peer's transfer nominate is rejected ("not in cache") without `register_transfer_tx` and accepted with it.

## Results
- `cargo test -p botho` — lib **995 passed**, 0 failed (the new externalize test verified stable across repeated runs).
- `cargo test -p bth-consensus-scp` — **100 passed**, 0 failed.
- `cargo build --release -p botho` / `cargo fmt` clean.
- Pre-existing, UNRELATED doctest failures in `network/privacy/{capacity,selection}.rs` (`RelayCapacity::measure` doesn't exist) are present on clean `origin/main` (829ba5e) and untouched here.

## Live verification (post-merge)
Consensus-relevant: needs a fresh-genesis redeploy to fully take effect on the live testnet. The faucet->balance->send->recipient flow should be confirmed after merge + redeploy.

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -675,10 +675,28 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             debug!("Received transaction {} from network", hex::encode(&tx.hash()[0..8]));
                             let tx_hash = tx.hash();
                             let tx_fee = tx.fee;
+                            // Serialize before move so we can seed the consensus
+                            // cache once the mempool accepts the tx (issue #449).
+                            let tx_bytes = bincode::serialize(&tx)
+                                .expect("Failed to serialize transaction");
                             // Add to mempool for inclusion in next block
                             if let Err(e) = node.submit_transaction(tx) {
                                 debug!("Failed to add network transaction to mempool: {}", e);
                             } else {
+                                // Seed the consensus tx cache so the SCP
+                                // validity_fn lookup succeeds when a peer
+                                // nominates the value for this transfer tx.
+                                // Without this, peers fail validity ("Transaction
+                                // not in cache") and SCP silently drops the
+                                // nominate/ballot messages, so quorum can never
+                                // form on a set containing the transfer and the
+                                // chain halts (issue #449). This mirrors the
+                                // minting path's register_minting_tx (#409). The
+                                // mempool acceptance above is the validity gate:
+                                // it performs full structural + UTXO + signature
+                                // checks, so we only cache validated bytes.
+                                consensus.register_transfer_tx(tx_hash, tx_bytes);
+
                                 // Broadcast transaction and mempool update to WebSocket clients
                                 ws_broadcaster.new_transaction(&tx_hash, tx_fee, None);
                                 if let Ok(mempool) = node.shared_mempool().read() {
@@ -1557,6 +1575,7 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                 // and submit them to consensus
                 for tx in node.get_pending_transactions(10) {
                     let tx_hash = tx.hash();
+                    let tx_fee = tx.fee;
                     let tx_bytes = bincode::serialize(&tx)
                         .expect("Failed to serialize transaction");
 
@@ -1565,8 +1584,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                         debug!("Failed to broadcast transaction: {}", e);
                     }
 
-                    // Submit to consensus for ordering
-                    consensus.submit_transaction(tx_hash, tx_bytes);
+                    // Submit to consensus for ordering. The fee is threaded
+                    // through as the deterministic consensus-value priority
+                    // (issue #449) so the same tx maps to the same value on
+                    // every node and SCP nomination converges.
+                    consensus.submit_transaction(tx_hash, tx_fee, tx_bytes);
                 }
             }
 
--- a/botho/src/consensus/block_builder.rs
+++ b/botho/src/consensus/block_builder.rs
@@ -414,7 +414,7 @@ mod tests {
 
     #[test]
     fn test_build_from_externalized_no_minting_tx() {
-        let values = vec![ConsensusValue::from_transaction([1u8; 32])];
+        let values = vec![ConsensusValue::from_transaction([1u8; 32], 0)];
 
         let result = BlockBuilder::build_from_externalized(&values, |_| None, |_| None);
 
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -291,16 +291,33 @@ impl ConsensusService {
             // Only keep the best minting tx (one coinbase per block)
             minting_txs.truncate(1);
 
-            // Sort regular txs by priority (fee), then hash for determinism
+            // Sort regular txs by priority (fee), then hash for determinism.
+            // This is the SELECTION ordering (higher-fee txs are preferred); the
+            // resulting block tx order is decided later by the block builder.
             regular_txs.sort_by(|a, b| {
                 b.priority
                     .cmp(&a.priority)
                     .then_with(|| a.tx_hash.cmp(&b.tx_hash))
             });
 
-            // Combine: best minting tx first, then regular txs
+            // Combine the selected coinbase with the regular txs.
             let mut combined = minting_txs;
             combined.extend(regular_txs);
+
+            // BALLOT INVARIANT (issue #449): SCP wraps this output directly into
+            // a `Ballot` and `Msg::validate()` REQUIRES the ballot's values to be
+            // STRICTLY ascending per `ConsensusValue::Ord` (see
+            // `Ballot::is_values_sorted`). The fee/priority selection ordering
+            // above is NOT the `Ord` ordering (a coinbase's random PoW-derived
+            // tx_hash can sort before or after a transfer), so a block containing
+            // BOTH a coinbase and a transfer would violate the invariant and
+            // panic SCP's outgoing-message validation — exactly what halted any
+            // multi-value block. Re-sort by the canonical `Ord` and dedup so the
+            // ballot is always valid. (Minting-only blocks have a single value
+            // and were never affected, which is why this latent bug only
+            // surfaced once transfer txs reached the slot.)
+            combined.sort();
+            combined.dedup();
             Ok(combined)
         });
 
@@ -583,9 +600,13 @@ impl ConsensusService {
         self.scp_node.current_slot_index()
     }
 
-    /// Submit a transaction for consensus
-    pub fn submit_transaction(&mut self, tx_hash: [u8; 32], tx_data: Vec<u8>) {
-        let value = ConsensusValue::from_transaction(tx_hash);
+    /// Submit a transaction for consensus.
+    ///
+    /// `fee` is threaded through as the deterministic consensus-value priority
+    /// (issue #449): it must be identical on every node for the same tx so the
+    /// value identity converges in SCP nomination.
+    pub fn submit_transaction(&mut self, tx_hash: [u8; 32], fee: u64, tx_data: Vec<u8>) {
+        let value = ConsensusValue::from_transaction(tx_hash, fee);
 
         // Add to shared cache for validation callback
         if let Ok(mut state) = self.shared_state.write() {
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!** High-rigor consensus-correctness review complete. This fixes the chain-halt-on-transfer defect (#449) cleanly and the safety-adjacent combine_fn change is correct.

## Verdict on the three defects

**(A) Deterministic value identity — CORRECT.** `from_transaction(tx_hash, fee)` sets `priority = fee`. `fee` is a field of the transaction (deserialized from the same bytes on every node), not locally computed, so the resulting `ConsensusValue` is identical across nodes for the same tx. `ConsensusValue` derives `Ord/Eq/Hash` over `(tx_hash, is_minting_tx, priority)` — all value-intrinsic; no node-local field remains for a transfer. This is what lets nomination converge.
- Verified empirically: reverting A to `SystemTime::now()` makes `two_nodes_externalize_block_containing_transfer_tx` fail (`priority` came out as a wall-clock nanosecond value instead of the fee).

**(C) combine_fn ordering — CORRECT, deterministic, preserves coinbase/set (the subtle one).**
- `Ballot::is_values_sorted` (ballot.rs:50) requires *strictly* ascending (`w[0] < w[1]`). `sort()` (ascending by `Ord`) + `dedup()` (removes adjacent equals) yields exactly strictly-ascending order. Correct normalization.
- **Deterministic across nodes:** the combine_fn input is the SCP-agreed value SET (same on all nodes), and `sort()` uses the total `Ord` over purely value-intrinsic fields — same set ⇒ same sorted vector everywhere.
- **Cannot drop the coinbase or change the SET:** `dedup()` only removes consecutive *equal* elements; a coinbase (`is_minting_tx=true`) can only collide with an identical coinbase, never with a transfer (differs in `is_minting_tx`/`tx_hash`). No distinct value is lost.
- **Selection invariant preserved:** the "truncate minting to 1" + fee-sort selection runs *before* the final `sort()`, which only reorders for the ballot invariant. Membership is unchanged; block tx order is decided downstream by the block builder.
- Verified empirically: reverting the `sort()`+`dedup()` makes the load-bearing test FLAKE (failed 2/5 runs) with the exact predicted panic `B ballot values not sorted` — confirming the bug was real, non-deterministic (coinbase PoW hash sometimes sorts before/after the transfer), and that the fix is load-bearing. With the fix it passes 5/5.

**(B) register_transfer_tx validity-gated — CORRECT.** In `run.rs` the new `register_transfer_tx` call is placed in the `else` branch *after* `node.submit_transaction(tx)` succeeds (the full structural + UTXO + signature mempool gate), mirroring how the minting handler gates on intrinsic validity. Peers only cache validated bytes. The validity_fn's hard `tx_cache` lookup now succeeds for transfer values on peers; the existing `validate_from_bytes_intrinsic` structural check still applies, so no invalid tx can leak into a block.
- Verified: `register_transfer_tx_enables_peer_nominate_acceptance` fails the peer-nominate acceptance without register and passes with it.

## (4) No #420/#417 regression — CONFIRMED.
`validation.rs` is NOT touched by this PR (scope = run.rs/block_builder.rs/service.rs/value.rs). The transfer validity path's `state.height` read (`validate_transfer_tx`, the 100-block `MAX_TX_AGE` staleness window at validation.rs:335) is **pre-existing on `origin/main`** — both the `validate_from_bytes_intrinsic`→`validate_transfer_tx` routing and the `state.height` comparison already exist there unchanged. This PR does not introduce or worsen tip/state-dependence. It is a latent #417-class concern worth a separate follow-up (tip-dependent transfer staleness could theoretically diverge validity across nodes at a height boundary), but it is out of scope for #449 and not made worse here. Filing a tracking issue.

## (5) Tests genuine + fail-pre/pass-post — CONFIRMED.
Both new service tests use real `ConsensusService` instances driving the production path (`submit_transaction(hash, fee, bytes)` + `register_transfer_tx(hash, bytes)`), with `run_two_services` pumping real `BroadcastMessage`/`SlotExternalized` events through real `handle_message`/`tick` — no mocks. Fail-pre/pass-post independently verified for both A and C (above).

## (6) No regression / scope — CONFIRMED.
- `cargo test -p botho --lib`: **995 passed, 0 failed** (incl. both new tests + `two_minters_converge_on_single_value_no_fork` minting convergence).
- `cargo test -p bth-consensus-scp`: **100 passed, 0 failed.**
- `cargo fmt --check -p botho`: clean.
- Diff touches exactly the 4 stated files (+289/-24).

## (7) CI — green.
All checks pass or skip (benchmarks, fuzz guards, rustfmt all pass; no failing/pending). `mergeStateStatus: CLEAN`. head_sha `8fcfa44`.

Solid, well-documented fix with empirically-validated regression coverage. Ship it.

PR #454, opened 2026-06-20
fix(consensus): make transfer-tx SCP validity tip-agnostic; height-staleness at build+apply (#451)

### LINKED ISSUE #451: Transfer-tx SCP validity is tip-dependent (state.height staleness window) — potential #417-class fork risk
## Summary
Discovered during high-rigor review of PR #450 (#449). The SCP validity function for **transfer** txs is NOT fully tip-agnostic, unlike the minting path which was deliberately made tip-agnostic in #419/#417 (Finding 1).

`ConsensusService` validity_fn → `TransactionValidator::validate_from_bytes_intrinsic(.., is_minting_tx=false)` → `validate_transfer_tx`, which reads the local chain tip:

`botho/src/consensus/validation.rs:335`
```rust
const MAX_TX_AGE: u64 = 100;
if tx.created_at_height + MAX_TX_AGE < state.height {
    return Err(ValidationError::StaleTransaction);
}
```

`state.height` is the **local** chain height. For minting txs the intrinsic path deliberately drops tip-relative checks so a value valid for one honest node is valid for all (the no-fork agreement property). For transfers, this staleness check makes validity tip-dependent.

## Why it matters (theoretical)
SCP's no-fork agreement assumes "valid for one honest node ⇒ valid for all honest nodes". If two honest nodes are at heights straddling the `created_at_height + 100` boundary (e.g. one at height H, one at H+1 during catch-up/lag), one may accept the transfer value while the other silently drops it as `StaleTransaction`. That asymmetric validity is exactly the #417-class condition that, for minting, could partition nominators. Whether it can actually fork for transfers depends on quorum dynamics, but it violates the same invariant the minting fix established.

## Scope / status
- **Pre-existing** on `origin/main` — NOT introduced by PR #450 (which does not touch `validation.rs`). PR #450 was approved on the basis that it does not worsen this.
- Surfaced now because #450 is the first change to actually route transfer values through federated SCP voting in production.

## Suggested direction
Make the consensus validity gate for transfers tip-agnostic (mirror the minting `*_intrinsic` split): drop the `state.height` staleness comparison from the SCP validity path and enforce staleness unconditionally at block-apply (`LedgerStore::add_block`) instead, where it cannot diverge across nodes. Confirm no other tip-relative reads exist in the transfer consensus-value validity path.

## References
- PR #450, issue #449
- Minting precedent: #419 / #417 Finding 1 (tip-agnostic validity_fn)
- `botho/src/consensus/validation.rs:313-348` (`validate_transfer_tx`), `:488-502` (`validate_from_bytes_intrinsic`)
- `botho/src/consensus/service.rs:230-266` (validity_fn + its own SAFETY note that the transfer path uses chain_state)

Discovered during code review of PR #450.

rjwalters:
Closes #451

## Problem
The SCP validity gate for **transfer** txs (`validate_transfer_tx`, reached via the validity_fn → `validate_from_bytes_intrinsic`) read each node's **local current tip** (`state.height`) to enforce a 100-block staleness window:

```rust
if tx.created_at_height + MAX_TX_AGE < state.height { return Err(StaleTransaction); }
```

Two honest nodes straddling `created_at_height + 100` disagree on validity → asymmetric validity → **#417-class fork** (reachable since #449/#450 route transfers through federated SCP voting). The bug is not the use of block height (correct, deterministic metric — kept) but evaluating it against the **local current tip during nomination**.

## Fix — keep height-based staleness, move it from tip-dependent nomination to deterministic build/apply

1. **`validate_transfer_tx` is now purely intrinsic** — structure only, **no `chain_state`/tip read**. Validity is a pure function of the tx ⇒ "valid for one honest node iff valid for all" (the no-fork invariant the minting `*_intrinsic` split established in #419/#417). Verified: the transfer SCP-validity path (`validity_fn` → `validate_from_bytes_intrinsic` → `validate_transfer_tx`) has no remaining tip/state-dependent read; the `tx_cache` lookup is content-addressed.

2. **Block-build filter** (`BlockBuilder::build_from_externalized`): drop any transfer tx where `created_at_height + MAX_TX_AGE < block_height`, using the **block's own height** (deterministic across nodes). So an externalized block never contains a too-old tx and always applies cleanly — **no externalize-then-reject halt** (#449/#421 failure mode).

3. **Block-apply backstop** (`LedgerStore::add_block` via `first_stale_transfer_tx`): reject a block containing a transfer tx older than `MAX_TX_AGE` relative to `block.height()`. Deterministic defense-in-depth (mirrors #420 moving minting tip-equality checks to apply-time). Never fires for blocks honest builders produce.

`MAX_TX_AGE = 100` is now a single shared const (`consensus::MAX_TX_AGE`), used by both build and apply so they agree exactly.

The mempool's wall-clock `MAX_TX_AGE_SECS` is left unchanged (mempool memory hygiene, not the consensus rule).

## Why correct (honors "block height > wall-clock")
- Block height stays the staleness metric (no clock-skew/timezone/leap-second dependence).
- Fork risk came from evaluating staleness against the **local current tip** during nomination; evaluating against the **block's own height** at build/apply is deterministic → no fork.
- Building with stale txs filtered → externalized block is always apply-valid → no halt.

## Tests
- `test_transfer_tx_validity_is_tip_agnostic` / `test_transfer_tx_intrinsic_path_tip_agnostic` (Test A): validity_fn returns the SAME result at tip 10 and tip 111 for a tx created at height 10. **Fail-pre** (old code rejected it as `StaleTransaction` once local tip passes `created_at_height+100`); **pass-post** (tip-independent).
- `test_build_excludes_stale_transfer_tx` (Test B, build): block built at H=200 excludes a tx with `created_at_height = H-101`, keeps the boundary tx (`H-100`).
- `test_first_stale_transfer_tx_backstop` (Test B, apply): backstop flags the stale tx, accepts the boundary tx.

## Results
- `cargo test -p botho --lib`: 999 passed, 0 failed.
- `cargo test -p bth-consensus-scp`: 100 passed, 0 failed.
- `cargo build --release -p botho`: clean. `cargo fmt`/clippy clean on touched files.
- Pre-existing failures: 2 `network/privacy` doctests fail on a clean `origin/main` too — unrelated to this change.

## Files touched
- `botho/src/consensus/validation.rs` — `validate_transfer_tx` made tip-agnostic (drop `state.height` read) + Test A.
- `botho/src/consensus/block_builder.rs` — shared `pub const MAX_TX_AGE`, build-time staleness filter + Test B (build).
- `botho/src/consensus/mod.rs` — export `MAX_TX_AGE`.
- `botho/src/consensus/service.rs` — validity_fn doc/SAFETY note (now fully tip-agnostic).
- `botho/src/ledger/store.rs` — `first_stale_transfer_tx` apply backstop + Test B (apply).

## Deployment
This is a consensus-validity change (the SCP transfer-validity gate and block accept rules change). A **testnet redeploy is recommended** so all nodes run the tip-agnostic gate consistently.

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 444
-# Branch: feature/issue-444
+# Issue: 451
+# Branch: feature/issue-451
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/botho/src/consensus/block_builder.rs
+++ b/botho/src/consensus/block_builder.rs
@@ -21,6 +21,29 @@ use sha2::{Digest, Sha256};
 use std::collections::HashSet;
 use tracing::{debug, info, warn};
 
+/// Maximum age (in blocks) of a transfer transaction relative to the height of
+/// the block that includes it.
+///
+/// A transfer tx whose `created_at_height + MAX_TX_AGE < block_height` is
+/// considered stale and is excluded at block-build (and rejected at block-apply
+/// as a deterministic backstop).
+///
+/// This staleness rule used to live in the SCP transfer-validity function
+/// (`validate_transfer_tx`) and was evaluated against each node's *local
+/// current tip*. Because two honest nodes can be at different tips while
+/// nominating the same value, that made transfer-tx validity tip-dependent — a
+/// value valid for one honest node could be silently dropped as
+/// `StaleTransaction` by another, the #417-class asymmetric-validity fork
+/// condition (issue #451).
+///
+/// The fix keeps block-height as the staleness metric (NOT wall-clock — no
+/// timezone/leap-second/clock-skew dependence) but evaluates it against the
+/// height of the block being built/applied. Every honest node assembles and
+/// applies block N at the same height N, so the check is deterministic and
+/// cannot fork. Defining it once here (used by both build and apply) guarantees
+/// the two paths agree exactly.
+pub const MAX_TX_AGE: u64 = 100;
+
 /// Result of building a block from externalized values
 #[derive(Debug)]
 pub struct BuiltBlock {
@@ -78,6 +101,12 @@ impl BlockBuilder {
             "Building block from externalized minting tx"
         );
 
+        // The block's own height. Staleness is evaluated against THIS height
+        // (deterministic across all honest nodes), never against any node's
+        // local current tip — that tip-dependence was the #417-class fork risk
+        // removed from the SCP validity gate in issue #451.
+        let block_height = minting_tx.block_height;
+
         // Collect transfer transactions with key image deduplication
         // This is a defense-in-depth measure to prevent double-spend attacks where
         // two transactions with the same key image somehow both made it to consensus
@@ -88,6 +117,25 @@ impl BlockBuilder {
         for value in &transfer_values {
             match get_transfer_tx(&value.tx_hash) {
                 Some(tx) => {
+                    // Height-based staleness filter (issue #451). Exclude any
+                    // transfer tx that is too old relative to THIS block's
+                    // height. Filtering here (rather than rejecting after
+                    // externalize) guarantees the built block always applies
+                    // cleanly: a block we build never contains a stale tx, so
+                    // the apply-time backstop never fires for our own blocks
+                    // (no externalize-then-reject halt — the #449/#421 failure
+                    // mode).
+                    if tx.created_at_height + MAX_TX_AGE < block_height {
+                        warn!(
+                            tx_hash = hex::encode(&value.tx_hash[0..8]),
+                            created_at_height = tx.created_at_height,
+                            block_height,
+                            "Skipping stale transfer tx in block (created_at_height + \
+                             MAX_TX_AGE < block_height)"
+                        );
+                        continue;
+                    }
+
                     // Check for duplicate key images (defense in depth)
                     let mut has_duplicate = false;
                     for input in tx.inputs.clsag() {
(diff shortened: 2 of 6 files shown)


rjwalters:
## Judge verdict: APPROVE

Consensus-safety review of #451 (make SCP transfer-tx validity tip-agnostic — remove a latent #417-class fork condition). Reviewed at head `1863ae1`. CI: all checks SUCCESS (skipped fuzz jobs are conditional), `mergeStateStatus: CLEAN`. Locally reproduced `cargo test -p botho --lib` (999 passed, 0 failed) and `bth-consensus-scp` (100 passed, 0 failed).

### 1. Validity now tip-agnostic [crux] — CONFIRMED
The SCP validity_fn path is `validity_fn` → `validate_from_bytes_intrinsic(is_minting=false)` → `validate_transfer_tx`. `validate_transfer_tx` no longer reads `chain_state`/`state.height`; it checks only inputs-non-empty, outputs-non-empty, no-zero-amount-output — all pure functions of the tx. The validity_fn's only other input is `state.tx_cache.get(&value.tx_hash)`, which is content-addressed by hash (not tip-dependent). The `temp_state` clone in service.rs is retained solely because `TransactionValidator::new` requires a handle; `validate_from_bytes_intrinsic` never reads it. Grepped the full function and callees — no current-tip read remains in the transfer SCP path. This restores "valid for one honest node ⇒ valid for all" → fork condition removed.

### 2. Build + apply determinism + matching boundary — CONFIRMED
Both use the block's OWN height, never the local tip:
- Build (`build_from_externalized`): `tx.created_at_height + MAX_TX_AGE < block_height`, where `block_height = minting_tx.block_height`.
- Apply (`first_stale_transfer_tx` in `add_block`): `tx.created_at_height + MAX_TX_AGE < block.height()`, where `block.height() == header.height == minting_tx.block_height`.

Same shared const (`consensus::MAX_TX_AGE = 100`), same comparison, same value. Boundary verified: a tx at `created_at_height = H-100` gives `H < H` = false → KEPT by build, NOT flagged by apply; a tx at `H-101` gives `H-1 < H` = true → DROPPED by build, flagged by apply. Identical boundary — no tx the builder keeps can be rejected by apply.

### 3. No externalize-then-reject halt — CONFIRMED
The build filter runs inside the transfer-collection loop with `continue` (block_builder.rs:128-137), i.e. BEFORE the block is assembled (line 187). Stale txs never enter `block.transactions` or `transfer_tx_hashes`. So an externalized-but-stale transfer value is dropped from the built block rather than making the block fail apply. Since build and apply share the identical boundary, the apply backstop never fires for honest-built blocks → no #449/#421-class halt. Adversarial path (stale tx externalized AND included AND rejected) is closed: the only way a block contains a stale tx is a malformed/adversarial block, which the backstop correctly rejects (that block was never valid).

### 4. Staleness still enforced — CONFIRMED
Not deleted, just relocated out of the tip-dependent nomination gate: build filters stale transfers, apply rejects any block containing one. A too-old transfer cannot end up in the chain.

### 5. Block-height not wall-clock — CONFIRMED
The consensus staleness rule uses block height (maintainer's requirement). The mempool's wall-clock `MAX_TX_AGE_SECS` is untouched and remains mempool memory hygiene, not the consensus rule.

### 6. Tests genuine — CONFIRMED
- Test A (`test_transfer_tx_validity_is_tip_agnostic`, `test_transfer_tx_intrinsic_path_tip_agnostic`): tx created at height 10 validates identically at tip 10 and tip 111 (= 10+100+1, past the old boundary). Genuine fail-pre/pass-post — the old `state.height` check returned StaleTransaction at tip 111. Exercises both `validate_transfer_tx` and the real `validate_from_bytes_intrinsic` entry point.
- Test B build (`test_build_excludes_stale_transfer_tx`): block at H=200 drops `created_at_height = H-101`, keeps boundary `H-100`; asserts the stale hash is absent from `transfer_tx_hashes`.
- Test B apply (`test_first_stale_transfer_tx_backstop`): flags `H-101`, accepts boundary `H-100`. Boundary correct in all three.

### 7. Scope / no regression — CONFIRMED
Diff limited to validation.rs, block_builder.rs, mod.rs, service.rs, store.rs (+ the .loom-managed marker). #420/#417 minting tip-relative checks (validate_minting_tx, gossip path) intact; #449 transfer path intact; UTXO/double-spend/signature checks remain at mempool/apply. Full suites green.

### Note (non-blocking, out of scope)
`validate_quantum_private_tx` (validation.rs:399-402, `#[cfg(feature = "pq")]`) still has a `created_at_height + MAX_TX_AGE < state.height` tip read. It is NOT wired into the SCP validity_fn (the validity_fn only deserializes MintingTx/Transaction, never QuantumPrivateTransaction) and only its own tests call it, so it carries no fork risk today and is correctly out of #451 scope. Worth tracking as a follow-up if/when PQ txs ever route through federated voting.

Approving. Relabeling `loom:review-requested` → `loom:pr`.

PR #457, opened 2026-06-20
feat(web-wallet): 3-node SCP ingress picker + Cloudflare Pages PWA deploy config

rjwalters:
## Summary
Ships `web/packages/web-wallet` as a deployable mobile PWA modeled on
`../once-around/apps/web`. Adds a 3-node SCP ingress picker (the user chooses
which live testnet validator/faucet node relays their RPC), adds Cloudflare
Pages deploy config + scripts, and verifies the build, PWA assets, and the
faucet path against the live testnet. Reuses `@botho/wasm-signer` — keys never
leave the browser.

## Changes
- **3-node ingress picker** (`src/config/networks.ts`): model `INGRESS_NODES`
  (seed / seed2 / faucet), pin `FAUCET_ENDPOINT` to the faucet node regardless
  of ingress choice, add `fetchNodeHealth` (via `node_getStatus`) and ingress
  persistence in localStorage. Legacy custom-RPC + `NETWORKS` exports kept.
- **`src/contexts/network.tsx`**: `selectIngress` + on-demand per-node health
  polling (subscribed by the selector while mounted; the landing page makes no
  node calls).
- **`src/components/NetworkSelector.tsx`**: renders the 3 ingress nodes with
  live health dots (height/sync), a faucet badge, and keeps the Custom RPC
  option.
- **`src/contexts/wallet.tsx`**: routes the initial RPC adapter to the persisted
  ingress node.
- **Cloudflare Pages**: `wrangler.toml` (`pages_build_output_dir = "dist"`) +
  `deploy` / `deploy:all` / `build:wasm` scripts in package.json mirroring
  once-around (`wrangler pages deploy dist --project-name=botho-wallet`).

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Pick 1 of 3 SCP nodes; persists; routes RPC; node health shown | Done | `selectIngress` persists to localStorage; wallet adapter inits from selection; selector shows per-node `node_getStatus` health |
| Faucet pinned to faucet node | Done | `FAUCET_ENDPOINT` is fixed to the faucet node; FaucetButton uses `network.faucetEndpoint` which is always the faucet RPC |
| `build` (incl. wasm) + `typecheck` green | Done | `build:wasm` regenerated pkg; `tsc --noEmit` clean; `vite build` emits dist + PWA |
[...]

PR #463, opened 2026-06-20
fix(web-wallet): open wallet at root + real client-side tx history (no 0-BTH spam) (#459)

### LINKED ISSUE #459: web-wallet: open wallet at root (not landing) + real client-side tx history (fix 0-BTH-spam)
Two concrete bugs found using the live wallet at wallet.botho.io:

## Bug 1: wallet.botho.io opens the marketing homepage, not the wallet
`web/packages/web-wallet/src/App.tsx`: `/` → `<LandingPage/>`, wallet is at `/wallet`. On the `wallet.` subdomain users expect the wallet. Fix: make `/` open the wallet (e.g. redirect `/` → `/wallet`, or render WalletPage at `/` and move the landing to `/about`/`/home`). Keep the landing reachable. Decide the cleanest routing.

## Bug 2: transaction history shows ~102 "received 0 BTH" entries
`web/packages/adapters/src/remote.ts` `getTransactionHistory` is an unfinished STUB: it calls `chain_getOutputs` for a height range and maps EVERY output to `{type:'receive', amount: 0n}` ("Would need to decrypt"), with no owned-output filtering. So the wallet shows every chain output as a 0-BTH receive.

Fix: build transaction history CLIENT-SIDE from the user's OWNED outputs, like balance already is — scan outputs in wasm (the wallet already does this for balance via the spent-filter), decode the real amount (the node returns the transparent amount in the commitment via `getRawOutputs`/`chain_getOutputs`), and show only the user's actual receives (and sends, if derivable) with correct amounts + timestamps. Reuse the existing balance scan path (`web/packages/web-wallet/src/contexts/wallet.tsx` computeBalance / wasm-signer scanOwnedOutputs). Don't show non-owned outputs.

## Acceptance
1. wallet.botho.io opens the wallet directly; landing still reachable.
2. Transaction history shows only the user's owned receives with correct decoded amounts (no 0-BTH spam); ideally sends too. Verify against the live testnet (faucet a wallet → see one correct +10 BTH receive entry, not 100+ zeros).
3. typecheck + e2e green; redeploy to Cloudflare Pages (--branch=main).

rjwalters:
Closes #459

## Summary
Fixes two bugs found on the live wallet at wallet.botho.io.

### Bug 1 — root opens the wallet (not the marketing homepage)
`web/packages/web-wallet/src/App.tsx`: `/` now redirects to `/wallet` **only** when the app is served from a `wallet.*` hostname (i.e. wallet.botho.io). On every other host — botho.io, localhost, the Playwright preview — `/` keeps rendering the landing page. This satisfies the "wallet subdomain opens the wallet" requirement while keeping the existing smoke e2e (`a[href="/"]` → landing) green. The landing is always reachable at `/home` and `/about`, and the wallet header's back link points to `/home` on the wallet subdomain so the landing stays one click away.

### Bug 2 — real client-side transaction history (the main fix)
The adapter's `getTransactionHistory` was an unfinished stub: it pulled every chain output over a height range and mapped each to `{ type: 'receive', amount: 0n }` with **no owned-output filtering** (the adapter has no wallet keys), producing ~102 "received 0 BTH" rows.

History is now built **client-side from the user's OWNED outputs**, reusing the exact balance scan path:

- **`adapters/remote.ts`** — new `getRawOutputsWithMeta(start, end)` returns outputs with block height + tx hash, decoding the amount from the LE `amountCommitment` (same decode `getRawOutputs`/balance already use). The old `getTransactionHistory` is kept to satisfy the `NodeAdapter` interface but now returns `[]` (no more spam, ever).
- **`wasm-signer/send.ts`** — new `buildOwnedHistory(keys, rpc)`: fetch outputs → `scanOwnedOutputs` (the same wasm ownership primitive balance uses) keeps only the user's outputs with their **real decoded amounts** → `computeOwnedOutputKeyImages` + `chain_areKeyImagesSpent` flag spent ones. Each owned output = one `receive` (real amount + block height); spent owned outputs also emit a `spend` entry. Non-owned outputs are never returned.
- **`web-wallet/contexts/wallet.tsx`** — new `fetchHistory` derives keys from the mnemonic (non-custodial; keys stay in the browser) and maps entries to the `Transaction` shape, replacing every `adapter.getTransactionHistory` call. A locked wallet shows empty history rather than spam.

A freshly-faucet'd wallet now shows a single ~+10 BTH receive instead of 100+ zeros.

## Tests
- `pnpm --filter @botho/web-wallet typecheck` — green (also wasm-signer + adapters typecheck green).
- `pnpm --filter @botho/web-wallet build` — green (after `build:wasm`).
- New hermetic unit tests `packages/wasm-signer/test/history.test.ts` (4 tests, pass): only owned outputs returned, real amounts, no 0-BTH entries, spend entries for spent outputs, newest-first, empty when nothing owned.
- Existing hermetic `sign.test.ts` — pass.
- Playwright smoke e2e (local mock, system Chrome) — all routing tests pass, incl. "navigate back to landing from wallet". The one failing test ("Explorer › displays recent blocks") is a pre-existing flake (passes on retry / in isolation; references #334) unrelated to this change and untouched by it.
- Playwright `web-wallet` e2e (create + import, local mock) — 17/17 pass; wallet create/import/unlock flows intact.

## Verification notes
- The history-from-owned-outputs logic reuses the same `scanOwnedOutputs` + `chain_areKeyImagesSpent` + commitment-decode path that `spendableBalance` already uses, so receives are exactly the outputs counted in the balance.
- Did not run the live-testnet check end-to-end in this environment (live-node tests require a running node), but the logic is unit-tested and mirrors the proven balance path.

## Post-merge
After merge, redeploy to Cloudflare Pages with `--branch=main` (maintainer will handle the redeploy).

## Files touched
- `web/packages/web-wallet/src/App.tsx`
- `web/packages/web-wallet/src/pages/wallet.tsx`
- `web/packages/web-wallet/src/contexts/wallet.tsx`
- `web/packages/adapters/src/remote.ts`
- `web/packages/wasm-signer/src/send.ts`
- `web/packages/wasm-signer/src/index.ts`
- `web/packages/wasm-signer/test/history.test.ts` (new)

### DIFF
--- a/web/packages/adapters/src/remote.ts
+++ b/web/packages/adapters/src/remote.ts
@@ -297,38 +297,18 @@ export class RemoteNodeAdapter implements NodeAdapter {
     }
   }
 
-  async getTransactionHistory(_addresses: Address[], options?: TxHistoryOptions): Promise<Transaction[]> {
-    // Fetch outputs for the height range and filter client-side
-    const result = await this.call<Array<{
-      height: number
-      outputs: Array<{
-        txHash: string
-        outputIndex: number
-        targetKey: string
-        publicKey: string
-        amountCommitment: string
-      }>
-    }>>('chain_getOutputs', {
-      start_height: options?.startHeight || 0,
-      end_height: options?.endHeight || (options?.startHeight || 0) + 100,
-    })
-
-    // For now, return empty - client should process outputs locally
-    // Full implementation would scan outputs for matching addresses
-    return result.flatMap((block) =>
-      block.outputs.map((output) => ({
-        id: output.txHash,
-        type: 'receive' as const,
-        amount: BigInt(0), // Would need to decrypt
-        fee: BigInt(0),
-        privacyLevel: 'private' as const, // Ring signatures for privacy
-        cryptoType: 'clsag' as CryptoType, // Default to classical, actual type from RPC
-        status: 'confirmed' as const,
-        timestamp: Date.now(),
-        blockHeight: block.height,
-        confirmations: 0,
-      }))
-    )
+  /**
+   * @deprecated The adapter has no wallet keys, so it cannot tell which on-chain
+   * outputs the user owns or decode their amounts — its previous implementation
+   * mapped EVERY chain output to a `{ type: 'receive', amount: 0n }` entry,
+   * producing dozens of bogus "received 0 BTH" rows (#459). Transaction history
+   * is now built CLIENT-SIDE from the wallet's OWNED outputs in the wallet
+   * context (see `buildOwnedHistory`, which scans via the wasm signer exactly
+   * like balance does). This method is retained only so the `NodeAdapter`
+   * interface stays satisfied; it now returns an empty list rather than spam.
+   */
+  async getTransactionHistory(_addresses: Address[], _options?: TxHistoryOptions): Promise<Transaction[]> {
+    return []
   }
 
   /**
@@ -361,6 +341,51 @@ export class RemoteNodeAdapter implements NodeAdapter {
 
     return result.flatMap((block) =>
       block.outputs.map((output) => ({
+        targetKey: output.targetKey,
+        publicKey: output.publicKey,
+        amount: leHexToBigInt(output.amountCommitment),
+      })),
+    )
+  }
+
+  /**
+   * Like {@link getRawOutputs}, but preserves each output's block height (and
+   * its source transaction hash). The thin wallet uses this to build its
+   * transaction history CLIENT-SIDE: scan these for owned outputs (wasm), then
+   * stamp each owned receive with the block height it landed in (#459). Kept
+   * separate from {@link getRawOutputs} so the hot send/balance path stays a
+   * flat `{ targetKey, publicKey, amount }[]`.
+   */
+  async getRawOutputsWithMeta(
+    startHeight: number,
+    endHeight: number,
+  ): Promise<
+    Array<{
+      txHash: string
+      height: number
+      targetKey: string
+      publicKey: string
+      amount: bigint
+    }>
+  > {
+    const result = await this.call<Array<{
+      height: number
+      outputs: Array<{
+        txHash: string
+        outputIndex: number
+        targetKey: string
+        publicKey: string
+        amountCommitment: string
+      }>
+    }>>('chain_getOutputs', {
+      start_height: startHeight,
+      end_height: endHeight,
+    })
+
+    return result.flatMap((block) =>
+      block.outputs.map((output) => ({
+        txHash: output.txHash,
+        height: block.height,
         targetKey: output.targetKey,
         publicKey: output.publicKey,
         amount: leHexToBigInt(output.amountCommitment),
--- a/web/packages/wasm-signer/src/index.ts
+++ b/web/packages/wasm-signer/src/index.ts
@@ -237,10 +237,14 @@ export function setSigner(signer: WasmSigner): void {
 
 export {
   buildSendTransaction,
+  buildOwnedHistory,
   spendableBalance,
   spendableOwnedOutputs,
   type BuildSendParams,
   type BuildSendResult,
+  type ChainOutputWithMeta,
+  type HistoryEntry,
+  type HistoryRpc,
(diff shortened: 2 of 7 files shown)


rjwalters:
## Judge review: APPROVE

Reviewed at head `4a06bebb0fe60506fab0402d415df0ea89e4ad32`. Merge state CLEAN / MERGEABLE. Scope is exactly the 7 expected files, all under `web/` — no consensus/node changes, no stray files (the local `pnpm-workspace.yaml` allowBuilds noise is NOT in the PR; verified reverted).

### 1. Tx-history correctness (the main fix) — PASS
- `buildOwnedHistory` (wasm-signer/src/send.ts) reuses the SAME proven path as balance: `getOutputsWithMeta` -> `scanOwnedOutputs` (wasm ownership check) -> `computeOwnedOutputKeyImages` -> `chain_areKeyImagesSpent`. Only OWNED outputs survive the scan; non-owned outputs are never returned, so the 0-BTH-per-chain-output spam is gone.
- Amounts are REAL: decoded from `amountCommitment` LE bytes via `leHexToBigInt`, the identical decode used by the balance path (`getRawOutputs`/`getRawOutputsWithMeta` share the same helper). History and balance therefore agree.
- One `receive` per owned output (with true amount + landing height); spent owned outputs additionally emit a `spend` entry and flag the receive `spent` (covers on-chain `spent` OR mempool `pending`). Newest-first sort by blockHeight. Empty-when-nothing-owned handled (returns `[]` at two guards).
- The old adapter stub `getTransactionHistory` now returns `[]` (deprecated, kept only for interface conformance) — no more bogus rows even on the locked path.

### 2. Routing — PASS
- `isWalletHost()` uses `hostname.startsWith('wallet.')`: matches `wallet.botho.io`, does NOT match `botho.io`, `localhost`, or the Playwright preview. `/` -> `/wallet` only on the wallet subdomain; landing stays at `/` everywhere else.
- Landing always reachable via `/home` + `/about`; wallet header back-link is `/home` only on the wallet subdomain, else `/` — so the existing smoke selector `a[href="/"]` still matches on localhost and the back-to-landing test stays valid. SSR-safe (`typeof window` guards).

### 3. Non-custodial — PASS
`fetchHistory` derives keys from `mnemonicRef.current` (in-memory only) via `deriveKeypairs`, mirroring `fetchBalance` exactly. Spend/view privkeys are passed only into the wasm signer; only key images and locally-decoded amounts leave the browser. Keys are never sent to the node nor persisted. Locked wallet -> empty history (no spam, no fallback that would need keys).

### 4. No regression + tests genuine — PASS
- `pnpm typecheck`: green across all 7 packages.
- `pnpm vitest run packages/wasm-signer`: 8 passed (sign.test.ts 4/4 unchanged + new history.test.ts 4/4). The new tests genuinely assert owned-only filtering (1 owned among 100 foreign -> exactly 1 entry, no amount==0), real decoded amounts, spend-entry + receive-flagging, newest-first ordering, and empty-when-nothing-owned — via the documented `setSigner` test seam, no live node needed.
- All 5 `adapter.getTransactionHistory` call sites in `contexts/wallet.tsx` replaced with `fetchHistory`; create/import/unlock/send/faucet and the #457 ingress picker (`getIinitialNetwork` ingress routing) remain intact.
- e2e smoke: could not be executed in the review sandbox (browser process is killed at launch — `kill ESRCH`, environmental, fails before any navigation/assertion). The routing change is provably backward-compatible by inspection (localhost is not a `wallet.` host), and the explorer flake #334 is pre-existing/unrelated to this diff.

### 5. Scope — PASS
7 files, web-only (adapters/remote.ts, wasm-signer src/index.ts + src/send.ts + test/history.test.ts, web-wallet App.tsx + contexts/wallet.tsx + pages/wallet.tsx). No node/consensus code touched.

### CI note
No automated checks are reported on this PR because the repo's web-facing workflows (e2e-tests.yml) are `workflow_dispatch`-only and no PR workflow watches `web/` paths — a pre-existing repo config, not a gap in this PR. Local typecheck + unit tests substitute and are green.

Approving and relabeling `loom:pr`.

Issue #476, opened 2026-06-21
SECURITY/PRIVACY: encrypt address book at rest + document threat model + CSP/SRI hardening


rjwalters:
### (a) Address book stored in plaintext (privacy-at-rest)
`web/packages/core/src/storage/address-book.ts` writes contacts (names, addresses, notes, tx counts) to localStorage as plaintext JSON. This exposes the user's **transaction-counterparty graph and personal annotations** to anyone who can read localStorage (extension, shared device, XSS) — undermining Botho's privacy value prop at rest. Fix: encrypt the address book under the same password-derived vault key as the seed (only readable while unlocked).

### (b) Document the web-wallet threat model + add XSS/supply-chain hardening
localStorage is readable by ANY JavaScript running on the origin, so encryption-at-rest does not protect against XSS or a malicious dependency once the wallet is unlocked (the decrypted seed lives in memory). We should both document this honestly and reduce the attack surface:
- Add a strict **Content-Security-Policy** (no inline scripts, locked-down `connect-src` to the chosen node RPC + faucet, `default-src 'self'`) — via Cloudflare Pages `_headers` or meta. The PWA build should be CSP-compatible (no `unsafe-inline`).
- **Subresource Integrity** / pinned, audited dependencies; minimize third-party JS; review the wasm-signer supply chain.
- Write `web/packages/web-wallet/SECURITY.md` (or docs page) stating the threat model: non-custodial keys-in-browser, what is/ isn't protected (at-rest vs XSS vs malicious extension vs node-sees-your-queries), and guidance (dedicated browser profile, password, hardware/passkey roadmap).
- Note the **thin-client privacy tradeoff** already identified: querying a node leaks address/UTXO interest to that node (acceptable for a *trusted* node; document it).

Acceptance: address book encrypted at rest under the vault key (migrates on unlock); CSP headers shipped on wallet.botho.io and build verified CSP-clean; SECURITY.md threat-model doc added. Scope: web-wallet/core + Cloudflare Pages headers + docs.

This is partly design/policy — may warrant an architect pass on the KDF/vault-key approach shared with the seed and claim-link issues.

Issue #475, opened 2026-06-21
SECURITY: encrypt wallet seed at rest by default + stronger KDF/password policy


rjwalters:
## Severity: HIGH — wallet seed is stored in plaintext by default

Current behavior (`web/packages/web-wallet/src/pages/wallet.tsx` `CreateWalletView`/`ImportWalletView`, `core/src/wallet/index.ts` `saveWallet`):
- "Protect with password" is a checkbox that **defaults to OFF** (`usePassword = false`). If left off, `saveWallet(mnemonic, undefined)` writes the **24-word seed in plaintext** to localStorage (`STORAGE_KEY_ENCRYPTED='false'`).
- When on, the minimum password is **4 characters**.
- KDF is PBKDF2-SHA256 at **100,000 iterations** (OWASP 2023 guidance is ≥600k for PBKDF2-SHA256).

The crypto construction itself is sound (WebCrypto AES-256-GCM, random 16-byte salt + 12-byte IV per encrypt) — the problem is the defaults and policy.

### Fix
1. **Encrypt by default.** Make a password (or device passphrase / passkey-derived key) the default path; treat plaintext storage as an explicit, clearly-warned opt-out (or remove it entirely).
2. **Strengthen the password policy** (raise min length, basic strength meter) and **raise PBKDF2 iterations** to a modern value (≥600k SHA-256), or migrate to a memory-hard KDF (scrypt/argon2 via a vetted WASM lib). Keep a versioned KDF header so existing wallets can be re-wrapped on next unlock.
3. Consider a single "vault key" (password-derived) that encrypts ALL sensitive at-rest data (seed + claim-link secrets + address book) so one unlock protects everything.
- Acceptance: new wallets encrypted by default; old plaintext wallets migrate to encrypted on next unlock/save; configurable/raised iteration count; round-trip + migration unit tests; typecheck+build+vitest green. Scope: web-wallet/core only.

Relates to the claim-link-secret and address-book-encryption issues (shared vault key). Refs: `core/src/wallet/index.ts`, `pages/wallet.tsx`.

PR #481, opened 2026-06-21
feat(web-wallet): encrypt claim-link bearer secrets at rest under vault key

### LINKED ISSUE #474: SECURITY: claim-link bearer secrets stored in plaintext localStorage (drainable)
## Severity: HIGH/CRITICAL — bearer secrets stored in plaintext on device

`web/packages/core/src/storage/claim-links.ts` persists `ClaimLinkRecord`s to localStorage via `JSON.stringify` (line ~87). Each record contains the **ephemeral 12-word mnemonic** that owns the unclaimed funds (the code comment explicitly calls it "bearer-secret"). It is stored **unencrypted**, even when the user's own wallet seed IS encrypted with a password.

### Impact
Anyone able to read the origin's localStorage — a malicious/compromised browser extension, a shared or stolen device, or any XSS on wallet.botho.io — can extract every outstanding claim link's ephemeral mnemonic and **drain all unclaimed claim-link funds**. This is strictly worse than the seed exposure because it bypasses the wallet password entirely.

### Fix
- Encrypt claim-link records at rest under the same key material used for the wallet seed (the password-derived AES-GCM key), so outstanding-link secrets are only readable while the wallet is unlocked. Reuse the existing `encrypt`/`decrypt` (AES-256-GCM + PBKDF2) in `core/src/wallet/index.ts`.
- For wallets created WITHOUT a password (currently the default — see the encrypt-at-rest issue), at minimum gate this behind the broader "encrypt by default" change so there's always a key.
- Acceptance: outstanding claim-link mnemonics are not present in plaintext in localStorage; refund flow still works after unlock; round-trip unit tests; typecheck+build+vitest green. Scope: web-wallet/core only.

Refs: #465 (claim links), `core/src/storage/claim-links.ts`, `core/src/wallet/index.ts` encrypt/decrypt.

rjwalters:
Closes #474.

## Problem
`web/packages/core/src/storage/claim-links.ts` persisted each `ClaimLinkRecord` — including the **ephemeral 12-word mnemonic** that owns unclaimed claim-link funds (a bearer secret) — to localStorage as **plaintext JSON**, even when the wallet seed is encrypted. Anyone able to read the origin's localStorage (malicious extension, shared/stolen device, XSS) could drain all unclaimed claim-link funds, bypassing the wallet password. Fund-loss bug.

## Fix (reuses the #475 vault layer — no new crypto)
This is step 2 of the #475 → #474 → #476 security series and builds directly on the `VaultKey` foundation #475 landed.

1. **Encrypt at rest.** New `EncryptedClaimLinks` storage encrypts the entire record array as a single **versioned vault blob** via `VaultKey.encryptJSON` / `decryptJSON`. The blob carries the vault header (magic + version + KDF params); no ephemeral mnemonic is written in cleartext for a password-protected wallet.
2. **Migration.** On the first unlocked load, a legacy plaintext blob is detected with `isVaultBlob` and re-wrapped under the session vault key, overwriting the cleartext, while preserving refund ability.
3. **Locked / plaintext-wallet / no-key handling.** When `getVaultKey()` is null: `load()` degrades to empty (records unavailable until unlock) and leaves any legacy plaintext blob untouched so a later unlock can migrate it; `save()` throws `ClaimLinksLockedError` rather than silently persisting a bearer secret in cleartext.
4. **Wiring.** The wallet context constructs the claim-link store with `EncryptedClaimLinks(() => sessionVaultKey)` and keeps the module-scope key in sync with `vaultKeyRef` on create/import/unlock/reset, reloading (and migrating) claim links when the key becomes available.

## How the vault API was reused
- `VaultKey.encryptJSON` / `decryptJSON` for the record array (no AES/PBKDF2 reimplemented).
- `isVaultBlob` to distinguish encrypted blobs from legacy plaintext for migration detection.
- The session `VaultKey` from the wallet context (`getVaultKey()` / `vaultKeyRef`) is read lazily by the storage via a getter.

## Tests added (`claim-links.test.ts`)
- Encrypted round-trip: stored blob is a vault blob and contains no plaintext mnemonic/words/address.
- Legacy plaintext → encrypted migration on unlocked load.
- Locked/no-key: reads as empty, does not touch legacy plaintext, and refuses to write (throws `ClaimLinksLockedError`).
- Refund flow still works after encryption + reload (add → reload → setStatus refunded; bearer secret survives round-trip).

## Verification
- `pnpm -r typecheck` — green
- `pnpm --filter @botho/web-wallet build` — green
- `pnpm test:run` — 176 passed, 10 skipped (incl. 12 claim-link tests, 14 wallet-context tests)

## Scope
web-wallet + `@botho/core` only. No consensus/faucet/network/Rust changes. Reverted stray `.loom-managed` and `pnpm-workspace.yaml` mutations before pushing.

### DIFF
--- a/web/packages/core/src/storage/claim-links.test.ts
+++ b/web/packages/core/src/storage/claim-links.test.ts
@@ -2,9 +2,12 @@ import { describe, it, expect, beforeEach } from 'vitest'
 import {
   ClaimLinkStore,
   LocalStorageClaimLinks,
+  EncryptedClaimLinks,
+  ClaimLinksLockedError,
   CLAIM_LINK_EXPIRY_WINDOW_SECONDS,
   type ClaimLinkRecord,
 } from './claim-links'
+import { VaultKey, isVaultBlob } from '../wallet/vault'
 
 const localStorageMock = (() => {
   let store: Record<string, string> = {}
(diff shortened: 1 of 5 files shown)


rjwalters:
## Loom Judge Review — Changes Requested

Strong, well-scoped PR. The core crypto reuse is correct, the encrypted-at-rest storage is clean, and the test suite is thorough. I verified locally: `pnpm -r typecheck` green (7 pkgs), `pnpm --filter @botho/web-wallet build` green, `pnpm test:run` = **176 passed / 10 skipped**. Scope is clean (4 files, web-wallet + @botho/core only; no Rust/consensus/network; no .loom-managed / pnpm-workspace mutations vs merge-base). Merge state CLEAN.

I'm requesting one change because the PR introduces a **new fund-loss path** for plaintext (no-password) wallets — the exact bug class #474 set out to eliminate.

### Blocking: `sendViaLink` funds the ephemeral wallet BEFORE persisting the bearer secret, and persist throws for plaintext wallets

In `web/packages/web-wallet/src/contexts/wallet.tsx`, `sendViaLink` does:

1. guard on `mnemonicRef.current` — passes for ANY unlocked wallet, **including a legacy/plaintext no-password wallet**
2. `buildAndSubmitSend(...)` — funds the ephemeral address **on-chain**
3. `claimLinkStore.add(...)` → `save()` → **throws `ClaimLinksLockedError` when `sessionVaultKey === null`**

For a plaintext wallet, `sessionVaultKey` is always null (password is optional in `createWallet`/`importWallet`, and `applyVaultKey(null)` is published for no-password wallets), yet the wallet is unlocked. So step 2 sends real funds, then step 3 throws — and the ephemeral mnemonic (the ONLY key to those funds) is discarded from memory and never persisted. The funds become unrecoverable.

The UI itself does NOT crash — `SendLinkModal.handleCreate` catches and shows the error banner ("Claim-link store is locked: unlock the wallet…"), which is good. But the message is misleading for a plaintext wallet (it's not locked), and more importantly the money is already gone by the time the error shows.

This is point-5's UX edge, but worse than a silent broken button: it spends first, then fails.

### Required fix (pick one)

- **Preferred:** guard `sendViaLink` BEFORE funding — if `getVaultKey()`/`sessionVaultKey` is null, throw early (before `buildAndSubmitSend`) with a clear, actionable message that routes plaintext-wallet users to set a password / migrate (the #475 path). Disable/annotate the "Create Claim Link" button when no session key is available, so it's a conscious surfaced decision rather than a spend-then-fail.
- **Alternative:** if encrypt-by-default means plaintext wallets should be fully unable to create links, make that an explicit, pre-funding block with guidance — never reachable after `buildAndSubmitSend`.

Either way, no code path may fund an ephemeral address that it cannot then persist a recoverable secret for.

### Also please add (test gap)

`wallet.test.tsx` does not cover `sendViaLink` for a no-password wallet. Add a test asserting that creating a claim link on a plaintext wallet does NOT submit a funding tx that it can't persist (i.e. it throws/blocks before `buildAndSubmitSend`, or the secret is persisted). This is the regression that guards the fund-loss path.

### Verified GOOD (no change needed)

1. **No plaintext bearer secret** — confirmed. The stored blob is a vault blob (`isVaultBlob` true) and contains no mnemonic/words/address in cleartext; test asserts this. Reuses `VaultKey.encryptJSON`/`decryptJSON`/`isVaultBlob` — no reimplemented crypto.
2. **Migration / no fund loss on load** — legacy plaintext is detected via `isVaultBlob` (sound: hex magic `bv\0\1` cannot collide with a JSON `[...]` blob) and re-wrapped on first unlocked load; refund flow works after encryption AND after migration; best-effort migration keeps the working plaintext blob if re-wrap fails. Tests cover all of this.
3. **Locked/no-key read safety** — `load()` degrades to empty and leaves legacy plaintext untouched; `save()` throws instead of writing cleartext. Refund button (`OutstandingLinks`) handles errors gracefully. (The remaining issue is the WRITE path in `sendViaLink`, above.)
4. **Key-sync correctness** — `applyVaultKey` syncs `vaultKeyRef` + module-scope holder on create/import/unlock, reloads/migrates on key availability; `resetWallet` clears both refs and `setSessionVaultKey(null)`. No stale key surviving reset. Lazy getter avoids a load-before-key race.

Once the pre-funding guard + test land, I expect to approve. Nice work on the migration design and the no-cleartext-on-write principle — this just needs the same principle applied before the on-chain spend.

rjwalters:
## Doctor — Judge feedback addressed (blocking fund-loss path fixed)

Fixed the spend-then-fail fund-loss path in `sendViaLink` and added the requested regression test. Scope unchanged (web-wallet only), and all the parts the Judge approved (encrypted storage, migration, key-sync) are left intact.

### The fix
`web/packages/web-wallet/src/contexts/wallet.tsx` — `sendViaLink` now guards on the session vault key BEFORE any on-chain spend:

```ts
if (vaultKeyRef.current === null) {
  throw new Error(
    'Claim links require a password-protected wallet. Add a password to your wallet to send via link.',
  )
}
```

This runs *before* `buildAndSubmitSend`, so a plaintext / no-password wallet (where `sessionVaultKey` is always null and persisting the bearer secret would throw `ClaimLinksLockedError`) can never fund an ephemeral address it cannot then persist a recoverable secret for. The error is actionable and routes users to the #475 add-a-password path; `SendLinkModal` already catches and shows it in a banner — and now no spend has occurred when it does.

### Regression test
`web/packages/web-wallet/src/contexts/wallet.test.tsx` — new `describe('sendViaLink fund-loss guard (#474)')` with two tests:
1. **Plaintext wallet**: `sendViaLink` rejects with the actionable `/password-protected wallet/i` error, asserts `buildAndSubmitSend` was **never called** (mocked `../lib/claim-link-ops`), and no claim link is recorded — i.e. it throws/blocks before spending.
2. **Encrypted wallet (happy path)**: still funds (`buildAndSubmitSend` called once) and returns a valid claim link (`/claim` URL, correct amount + funding hash).

### Verification (all green)
- `pnpm install` — only the expected `ERR_PNPM_IGNORED_BUILDS` warning
- `pnpm -r typecheck` — green (7 pkgs)
- `pnpm --filter @botho/web-wallet build` — green
- `pnpm test:run` — **178 passed / 10 skipped** (up from 176; the 2 new tests), `wallet.test.tsx` now 16 tests

### Scope
Only `wallet.tsx` + `wallet.test.tsx` changed. No Rust/consensus/network. Reverted the stray `pnpm-workspace.yaml` `allowBuilds` mutation pnpm wrote during install before pushing.

Relabeled `loom:changes-requested` -> `loom:review-requested`.

rjwalters:
## Judge re-review (2nd/final pass) — APPROVED

The blocking fund-loss bug (spend-then-fail on a plaintext/no-key wallet) is genuinely closed and tested. Verdict: approve.

### 1. Guard runs before any on-chain spend, and uses a key source that can't drift
In `web/packages/web-wallet/src/contexts/wallet.tsx`, `sendViaLink` now checks `vaultKeyRef.current === null` and throws the actionable error **before** `createClaimLinkMnemonic` and **before** `buildAndSubmitSend` (the on-chain spend), which in turn precedes `claimLinkStore.add` (the persist that would throw `ClaimLinksLockedError`). Ordering confirmed on the branch: guard → ephemeral gen → `buildAndSubmitSend` → `claimLinkStore.add`. No path can spend then fail.

Key-source sync verified: `vaultKeyRef.current` (read by the guard) and the module-scope `sessionVaultKey` (read by `EncryptedClaimLinks` via its `() => sessionVaultKey` getter) are written together through the single `applyVaultKey` funnel on create/import/unlock, and both are nulled together on reset. There is no code path that sets one without the other, so the guard cannot be out of sync with the persist layer. `EncryptedClaimLinks.save` throwing on a null key is exactly the condition the guard pre-empts.

### 2. Regression tests assert the right invariants
`wallet.test.tsx` mocks `../lib/claim-link-ops` (the module `wallet.tsx` actually imports) so it can observe the spend:
- Plaintext wallet: `sendViaLink` rejects with `/password-protected wallet/i`, `buildAndSubmitSend` is asserted **never called**, and `claimLinks.length === 0`.
- Encrypted happy path: `buildAndSubmitSend` called once, returns a record whose `url` contains `/claim`, correct amount, and funding hash — with the real encrypted `claimLinkStore.add` running (genuine integration assertion).

### 3. No regressions to approved work; scope clean
The new commit (9c7cfa7) touched only `wallet.tsx` (+11) and `wallet.test.tsx` (+80); the previously-approved `claim-links.ts` encryption/migration/key-sync is unchanged. Diff vs merge-base is 5 files, web-wallet + @botho/core only. No `.loom-managed`, `pnpm-workspace.yaml`, lockfile, `package.json`, or Cargo mutations.

### 4. CI reproduced locally (green)
- `pnpm -r typecheck` — green (7 packages)
- `pnpm --filter @botho/web-wallet build` — green
- `pnpm test:run` — **178 passed / 10 skipped** (incl. 12 claim-link, 16 wallet-context tests)

mergeStateStatus: CLEAN.

PR #482, opened 2026-06-21
feat(web-wallet): encrypt address book at rest + CSP + threat-model doc (#476)

### LINKED ISSUE #476: SECURITY/PRIVACY: encrypt address book at rest + document threat model + CSP/SRI hardening

rjwalters:
Closes #476

Final step of the #475 → #474 → #476 web-wallet security/privacy series. Branched off latest main (HEAD dfc4ba0, after #475 vault and #474 claim-link encryption merged).

## (a) Encrypt the address book at rest (reuses the #475 vault — no new crypto)
`web/packages/core/src/storage/address-book.ts` previously persisted contacts (names, addresses, notes, tx counts) to localStorage as plaintext JSON, leaking the user's counterparty graph + annotations at rest.

- New `EncryptedAddressBook` storage impl, mirroring #474's `EncryptedClaimLinks`: the whole contact array is stored as **one versioned vault blob** via `VaultKey.encryptJSON`/`decryptJSON`, key read lazily via `() => sessionVaultKey`, wired through the same `applyVaultKey` funnel in `contexts/wallet.tsx`.
- Legacy plaintext contacts are transparently re-wrapped under the vault key on the first unlocked load (`isVaultBlob` migration). No plaintext contact graph is ever written for a password-protected wallet.
- **Plaintext / no-key handling** (the address book is less critical than bearer secrets, so it degrades gracefully instead of hard-failing like claim links):
  - `load()` with no key (locked, or a plaintext/no-password wallet) returns `[]` and leaves any legacy plaintext blob untouched (so a later unlock can migrate it).
  - `save()` with no key is a **NO-OP**: it does NOT throw and does NOT write the contact graph in plaintext. Contacts are simply not persisted until the wallet has a password.
  - This keeps `recordPayment` and the post-send address-book write from throwing or losing money when there is no vault key — no spend-then-fail, no uncaught crash.

## (b) Content-Security-Policy on wallet.botho.io
New `web/packages/web-wallet/public/_headers` (Cloudflare Pages). CSP shipped:

```
default-src 'self'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self' data:; connect-src 'self' https://seed.botho.io https://seed2.botho.io https://faucet.botho.io; worker-src 'self' blob:; manifest-src 'self'; object-src 'none'; base-uri 'self'; frame-ancestors 'none'; form-action 'self'; upgrade-insecure-requests
```

- **No `'unsafe-inline'` / `'unsafe-eval'` in `script-src`.** `'wasm-unsafe-eval'` is required only to instantiate the WebAssembly signer; the PWA SW registration runs from a bundled module (`vite-plugin-pwa injectRegister: null`), not an inline script. Verified the production build is CSP-clean: `dist/index.html` contains only an external `<script type="module" src=...>` — no inline scripts.
- `connect-src` is an allowlist of our origin plus exactly the seed/seed2/faucet RPC origins from `src/config/networks.ts`.
- Also ships `X-Content-Type-Options`, `Referrer-Policy: no-referrer`, `X-Frame-Options: DENY`, `Permissions-Policy`.

## (c) Threat-model doc
New `web/packages/web-wallet/SECURITY.md`: non-custodial keys-in-browser model; what IS protected at rest (#475 seed, #474 claim-link secrets, #476 address book — all under the password-derived vault) vs what is NOT (XSS / malicious extension reads memory once unlocked; a trusted node sees your address/UTXO queries — the thin-client privacy tradeoff); the plaintext/no-password caveats; CSP/SRI hardening; user guidance (set a password, dedicated browser profile).

## Verification
- `pnpm install` — only the expected `ERR_PNPM_IGNORED_BUILDS` warning.
- `pnpm -r typecheck` — green (all 7 projects).
- `pnpm --filter @botho/web-wallet build` — green; PWA SW generated; CSP-clean (no inline scripts).
- `pnpm test:run` — 184 passed / 10 skipped. Added tests: encrypted address-book round-trip (no plaintext contacts in blob), legacy plaintext migration, locked/no-key reads-empty without touching the blob, `save()` no-op-no-throw, recordPayment-style upsert with null key does not throw.

## Scope
web-wallet + @botho/core only. No consensus/faucet/network/Rust changes. Reverted stray `.loom-managed` / `pnpm-workspace.yaml` mutations before pushing.

### DIFF
--- a/web/packages/core/src/storage/address-book.test.ts
+++ b/web/packages/core/src/storage/address-book.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, beforeEach } from 'vitest'
-import { AddressBook } from './address-book'
-import type { Address, Timestamp } from '../types'
+import { AddressBook, EncryptedAddressBook } from './address-book'
+import type { Address, Contact, Timestamp } from '../types'
+import { VaultKey, isVaultBlob } from '../wallet/vault'
 
 const localStorageMock = (() => {
   let store: Record<string, string> = {}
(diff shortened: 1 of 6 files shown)


rjwalters:
## Judge review — Changes requested (one scoped CSP fix)

Reproduced locally on `feature/issue-476` (merge-base diff). Encryption, migration, tests, build, and docs are excellent; one shipped feature is broken by the CSP and needs to be addressed before merge.

### What's verified GOOD
- **Encryption + migration correctness**: `EncryptedAddressBook` stores the whole contact array as one versioned vault blob via `VaultKey.encryptJSON`/`decryptJSON` — no reimplemented crypto, reuses the #475 vault. `isVaultBlob` gates decrypt; legacy plaintext is re-wrapped (and the cleartext overwritten) on the first unlocked load. Tests assert the stored blob does NOT contain names/addresses/notes (`not.toContain` on all three), cover round-trip, migration, and re-availability.
- **No-key behavior**: `load()` → `[]` and leaves any legacy blob untouched (so a later unlock can migrate); `save()` with no key is a silent no-op that neither throws nor writes plaintext. Spend-path safety is explicitly tested (`recordPayment`-style upsert with null key does not throw) — no spend-then-fail, no cleartext contact graph. I accept the silent no-op for the legacy-plaintext minority (encryption is default since #475; no funds at risk; lower severity than #474). Suggest a follow-up to surface a one-line "set a password to save contacts" hint in the contacts UI — non-blocking.
- **Build is CSP-clean**: `dist/index.html` has only an external `<script type="module" src=...>` — no inline scripts, no inline handlers. SW registration is bundled (`injectRegister: null`); `wasm-unsafe-eval` is present for the signer. `_headers` is correctly copied into `dist/`.
- **No websockets anywhere** (`grep` for WebSocket/wss in web-wallet + core src = none) — all RPC is `fetch` POST, so no ws origin needs allowing. The three default RPC origins (seed/seed2/faucet.botho.io) match `connect-src` exactly. `botho.io/explorer` and `wallet.botho.io` references are navigation/string-building, not `fetch`, so they correctly don't need connect-src.
- **SECURITY.md** is accurate to the code and does not overclaim (correctly states the address book is unavailable/not-persisted without a password).
- **Scope clean**: only `core` + `web-wallet` (6 files); no Rust/consensus/network; no `.loom-managed`/`pnpm-workspace` mutations.
- **CI green locally**: `pnpm -r typecheck` ✅, `pnpm test:run` → 184 passed / 10 skipped ✅, `pnpm --filter @botho/web-wallet build` ✅.

### Blocking: the CSP breaks the shipped "Custom RPC" feature
`connect-src` is a static allowlist of only `'self'` + the three `*.botho.io` origins. But the wallet ships a first-class **Custom RPC** option (`src/components/NetworkSelector.tsx` → `setCustomEndpoint` → `createCustomNetwork`, persisted via `loadSelectedNetwork`/`networks.ts`) where the user enters an arbitrary `https://node.example.com/rpc`. Under this CSP the browser blocks BOTH:
1. the validation fetch (`fetchNodeHealth`) — so "Connect" always fails with the existing (misleading) "validation failed" error, and
2. all subsequent RPC to the custom origin.

A user pointing the wallet at their own/managed node — directly relevant to the "connect to a trusted node / managed rig" product direction — silently cannot connect. SECURITY.md even says "If you add a new node origin, update both connect-src here and networks.ts," which acknowledges the allowlist is exhaustive, yet the UI lets users add origins at runtime that can never be in a static header.

Pick one (small):
- Remove/hide the Custom RPC UI if it's intentionally out of scope for the hosted wallet, OR
- Relax `connect-src` to accommodate custom HTTPS endpoints (e.g. allow `https:` for connect-src, accepting the weaker exfil guarantee, and document the tradeoff in SECURITY.md), OR
- Ship the CSP as `Content-Security-Policy-Report-Only` first to avoid breaking the feature while you gather violations, then enforce.

Whatever you choose, please reconcile SECURITY.md + the Custom RPC UI so they agree.

Everything else is approve-ready — happy to flip to `loom:pr` on the CSP/custom-endpoint reconciliation.

rjwalters:
Addressed the blocking CSP / Custom RPC conflict.

**Fix:** relaxed ONLY `connect-src` to permit arbitrary HTTPS origins so runtime-added Custom RPC / self-hosted / managed-trusted nodes work; kept everything else strict.

Final CSP shipped (`public/_headers`, verified copied into `dist/_headers`):
```
default-src 'self'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self' data:; connect-src 'self' https:; worker-src 'self' blob:; manifest-src 'self'; object-src 'none'; base-uri 'self'; frame-ancestors 'none'; form-action 'self'; upgrade-insecure-requests
```
plus unchanged X-Content-Type-Options / Referrer-Policy / X-Frame-Options / Permissions-Policy headers.

- `http:` is NOT allowed (HTTPS-only); `upgrade-insecure-requests` kept.
- XSS-execution defenses unchanged: `script-src 'self' 'wasm-unsafe-eval'` (no `unsafe-inline`/`unsafe-eval`), `object-src 'none'`, `frame-ancestors 'none'`, `base-uri 'self'`, `form-action 'self'`.

**SECURITY.md:** updated the CSP block and the connect-src explanation to state it permits any HTTPS origin (so users can point at their own/trusted nodes, incl. runtime Custom RPC) and the exfil tradeoff — the primary anti-XSS defense is `script-src 'self'` + no inline/eval + dependency hygiene, not connect-src. Removed the now-inaccurate 'update connect-src when adding a node origin' instruction. The `_headers` comments were updated to match.

Encryption/migration, no-key graceful save() no-op, and all other approved parts are untouched. The non-blocking legacy-plaintext UI hint is intentionally left out of scope (separate follow-up).

**Verification:**
- `pnpm install` clean (only ERR_PNPM_IGNORED_BUILDS warning)
- `pnpm -r typecheck` green
- `pnpm --filter @botho/web-wallet build` green; `dist/_headers` carries the updated connect-src
- `pnpm test:run` green — 184 passed / 10 skipped (no test changes needed)
- Scope: web-wallet only (SECURITY.md + _headers); reverted a stray pnpm-workspace.yaml mutation before pushing.

Back to `loom:review-requested`.

rjwalters:
## Judge re-review: APPROVED

Final pass of the Doctor→Judge cycle on the CSP fix (commit `380935f`). The previously-approved encryption/migration/no-key behavior was not re-litigated — confirmed byte-identical to `596f542` (the fix touched only `SECURITY.md` and `public/_headers`).

### 1. connect-src fix genuinely unblocks Custom RPC — verified
- `connect-src 'self' https:` now permits any HTTPS origin. `createCustomNetwork(rpcEndpoint)` stores an arbitrary user-entered endpoint and `fetchNodeHealth` does a runtime `fetch(endpoint, { method: 'POST', ... })` against it. The old static allowlist could never include a runtime-added origin, so it blocked both the health-check and all subsequent RPC; `https:` resolves this.
- `http:` remains disallowed and `upgrade-insecure-requests` is kept; node RPC defaults and the Custom RPC placeholder are all HTTPS.
- Rest of CSP confirmed unchanged and still strict: `default-src 'self'`, `script-src 'self' 'wasm-unsafe-eval'` (no unsafe-inline/eval), `object-src 'none'`, `base-uri 'self'`, `frame-ancestors 'none'`, `form-action 'self'`, plus `X-Content-Type-Options: nosniff`, `Referrer-Policy: no-referrer`, `X-Frame-Options: DENY`, and `Permissions-Policy`.
- `dist/_headers` after `pnpm build:web` carries the updated policy verbatim; `dist/index.html` has no inline script bodies (build is CSP-clean).

### 2. SECURITY.md accurate, no overclaiming — verified
- Embedded CSP block matches the shipped header (`connect-src 'self' https:`).
- Correctly states connect-src permits any HTTPS origin by design (custom/trusted nodes added at runtime), documents the exfil tradeoff, and names `script-src 'self'` + no inline/eval as the primary anti-XSS defense rather than connect-src.
- The inaccurate "update connect-src when adding a node origin" instruction was removed.

### 3. Scope clean — verified
PR diff vs merge-base (`dfc4ba0`) touches only web-wallet + @botho/core: `address-book.ts`, `address-book.test.ts`, `storage/index.ts`, `wallet.tsx`, `SECURITY.md`, `_headers`. No `.loom-managed`, `pnpm-workspace.yaml`, or `networks.ts` mutations. No regressions to approved encryption/migration/no-key code.

### 4. CI green locally — reproduced
- `pnpm typecheck`: exit 0 (all packages)
- `pnpm build:web`: exit 0, PWA precache generated, dist/_headers updated
- `pnpm test:run`: exit 0 — **184 passed / 10 skipped** (matches Doctor report)

Merge state: CLEAN. No repo CI checks configured. The CSP now permits custom HTTPS nodes while keeping script-execution and framing defenses strict, and the doc is accurate. Approving.

PR #494, opened 2026-06-21
feat(web-wallet): set/change wallet password with full vault re-wrap (#489)

### LINKED ISSUE #489: web-wallet: add/change wallet password (encrypt a plaintext wallet; rotate password, re-wrap vault)
Let users SET a password on a plaintext (no-password) wallet and ROTATE an existing password. Today the only re-wrap path is automatic legacy migration; there is no user-initiated flow, so a plaintext-wallet user can never become encrypted (root cause behind #488 and the claim-link "requires a password-protected wallet" limitation), and an encrypted-wallet user can't change their password without re-importing from seed.

Build on the #475 vault + #474/#476 encrypted storage (all merged):
- `core/src/wallet/vault.ts` `VaultKey` (`fromPassword`, `encryptJSON`/`decryptJSON`), `MIN_PASSWORD_LENGTH`, `passwordStrength`.
- `core/src/wallet/index.ts` `saveWallet(mnemonic, password)`, `loadWallet(password)`.
- Encrypted storage: `EncryptedAddressBook` (#476), `EncryptedClaimLinks` (#474), wired through the `sessionVaultKey`/`vaultKeyRef` funnel + `applyVaultKey` in `contexts/wallet.tsx`.

Implement on the wallet context:
1. `setPassword(newPassword)` — for a plaintext wallet: derive a new VaultKey, re-save the seed ENCRYPTED (`saveWallet(mnemonic, newPassword)`), then re-wrap the address book and claim-link records under the new key (reuse the migration/re-wrap path — `applyVaultKey(newKey)`), and set the session key. After this the wallet is encrypted and contacts/claim-links work.
2. `changePassword(oldPassword, newPassword)` — for an encrypted wallet: verify old via `loadWallet(oldPassword)` (reject wrong old password), derive new key, re-encrypt the seed and re-wrap address book + claim links under the new key, update the session key. Old password must no longer decrypt.
3. Settings UI in `pages/wallet.tsx` (near the existing reveal-mnemonic/reset controls): "Set a password" (shown for plaintext wallets) / "Change password" (encrypted wallets), with the #475 password policy (min length + strength hint + confirm field).

Acceptance: plaintext→encrypted upgrade persists across reload and re-wraps all three data types (seed/contacts/claim-links) with NO plaintext left; change-password rotates and the old password is rejected; wrong-old-password rejected; unit tests for both flows incl. re-wrap correctness and old-key-invalidation. typecheck+build+vitest green. Scope: web-wallet + @botho/core only.
Refs: #475/#474/#476, #488.

rjwalters:
Closes #489

Let users SET a password on a plaintext wallet (upgrade to encrypted) and ROTATE an existing password, reusing the merged #475 vault + #474/#476 encrypted storage. No new crypto — only the existing `VaultKey` / `saveWallet` / `loadWallet` APIs.

## Context methods (`contexts/wallet.tsx`)
- `setPassword(newPassword: string) => Promise<void>` — plaintext → encrypted. Re-saves the seed via `saveWallet(mnemonic, newPassword)`, derives the session `VaultKey` (bound to the new seed blob's salt), then `applyVaultKey(newKey)` migrates the address book + claim links from legacy plaintext to encrypted under the new key (the same automatic plaintext→encrypted re-wrap used on unlock). Rejects if already encrypted / locked. Sets `isEncrypted=true`.
- `changePassword(oldPassword: string, newPassword: string) => Promise<void>` — rotate an encrypted wallet. Verifies the old password via `loadWallet(oldPassword)` (throws **"Incorrect current password"** on failure, before writing anything), re-encrypts the seed under the new password, then re-wraps the in-memory address book + claim links under the new key. The old password no longer decrypts any of the three data types.

## How the re-wrap reuses `applyVaultKey`
- **setPassword** reuses `applyVaultKey(newKey)` directly: its `load()` path already migrates legacy plaintext sibling blobs to encrypted under the freshly-applied key (this is the existing legacy-migration code).
- **changePassword** needs a different shape — the on-disk sibling blobs are still encrypted under the OLD key, so reloading would fail to decrypt and drop data. A sibling `rewrapUnderNewKey(newKey)` swaps the session key then re-persists the already-decrypted, in-memory store contents (new `AddressBook.rewrap()` / `ClaimLinkStore.rewrap()` primitives) so they're re-encrypted under the new key, overwriting the old-key blobs.

## UI (`pages/wallet.tsx`)
A Security card on the dashboard shows "Set a password" (plaintext) or "Change password" (encrypted), opening a modal that reuses the #475 `PasswordFields` policy (min length, strength hint, confirm), plus a current-password field for change.

## Files changed
- `web/packages/web-wallet/src/contexts/wallet.tsx` — `setPassword`, `changePassword`, `rewrapUnderNewKey`
- `web/packages/web-wallet/src/pages/wallet.tsx` — Security card + `PasswordSettingsModal`
- `web/packages/core/src/storage/address-book.ts` — `AddressBook.rewrap()`
- `web/packages/core/src/storage/claim-links.ts` — `ClaimLinkStore.rewrap()`
- `web/packages/web-wallet/src/contexts/wallet-password.test.tsx` — new tests

## Verification
- `pnpm -r typecheck` green
- `pnpm --filter @botho/web-wallet build` green
- `pnpm test:run` green — 209 passed / 10 skipped (live-node suites). New tests: (a) setPassword encrypts seed + re-wraps address book (no plaintext left) + session key set; (b) changePassword rotates, old password rejected by `loadWallet`, new key decrypts re-wrapped data, stale old key cannot; (c) wrong old password rejected with no rotation.

Scope: web-wallet + @botho/core only. No consensus/faucet/network/Rust changes.

### DIFF
--- a/web/packages/core/src/storage/address-book.ts
+++ b/web/packages/core/src/storage/address-book.ts
@@ -145,6 +145,21 @@ export class AddressBook {
     this.contacts = new Map(contacts.map(c => [c.id, c]))
   }
 
+  /**
+   * Re-persist the CURRENT in-memory contacts through the storage layer without
+   * re-reading from disk.
+   *
+   * This is the re-wrap primitive used when the wallet's password changes
+   * (#489): the encrypted storage reads the session {@link VaultKey} lazily, so
+   * once the context has swapped the session key to the NEW key, calling this
+   * re-encrypts the existing contact graph under the new key (overwriting the
+   * blob that was encrypted under the old key). The in-memory contacts must
+   * already be loaded (the wallet is unlocked) before this is called.
+   */
+  async rewrap(): Promise<void> {
+    await this.persist()
+  }
+
   private async persist(): Promise<void> {
     await this.storage.save(Array.from(this.contacts.values()))
   }
--- a/web/packages/core/src/storage/address-book.ts
+++ b/web/packages/core/src/storage/address-book.ts
@@ -2,6 +2,29 @@ import type { Address, Contact, Timestamp } from '../types'
 import { VaultKey, isVaultBlob } from '../wallet/vault'
 
 /**
+ * Thrown by {@link EncryptedAddressBook.loadStrict} when a PRESENT encrypted
+ * address-book blob fails to decrypt under the current key (wrong key /
+ * tampered / locked).
+ *
+ * The lenient {@link EncryptedAddressBook.load} returns `[]` on decrypt failure
+ * for graceful degradation; the strict path used by the password-rotation
+ * re-wrap (#489) surfaces the failure so the rotation never re-wraps an empty
+ * book over real (but un-decryptable) contacts.
+ */
+export class AddressBookDecryptError extends Error {
+  constructor(
+    message = 'Address book failed to decrypt under the current key',
+    options?: { cause?: unknown },
+  ) {
+    super(message)
+    this.name = 'AddressBookDecryptError'
+    if (options && 'cause' in options) {
+      ;(this as { cause?: unknown }).cause = options.cause
+    }
+  }
+}
+
+/**
  * Storage interface for address book persistence
  * Implementations can use localStorage, IndexedDB, or other storage
  */
@@ -76,6 +99,22 @@ export class EncryptedAddressBook implements AddressBookStorage {
   }
 
   async load(): Promise<Contact[]> {
+    return this.loadInternal(false)
+  }
+
+  /**
+   * Strict load for the password-rotation re-wrap path (#489).
+   *
+   * Unlike {@link load}, if a PRESENT encrypted (vault) blob fails to decrypt
+   * under the current session key, this THROWS instead of returning `[]`, so the
+   * rotation can distinguish a genuinely empty book from a decrypt FAILURE and
+   * never re-wrap an empty book over real (but un-decryptable) contacts.
+   */
+  async loadStrict(): Promise<Contact[]> {
+    return this.loadInternal(true)
+  }
+
+  private async loadInternal(strict: boolean): Promise<Contact[]> {
     const data = localStorage.getItem(this.key)
     if (!data) return []
 
@@ -84,11 +123,20 @@ export class EncryptedAddressBook implements AddressBookStorage {
     // Encrypted (vault) blob: only readable while unlocked. If locked, degrade
     // to "unavailable" rather than crashing.
     if (isVaultBlob(data)) {
-      if (!vaultKey) return []
+      if (!vaultKey) {
+        if (strict) {
+          throw new AddressBookDecryptError()
+        }
+        return []
+      }
       try {
         return await vaultKey.decryptJSON<Contact[]>(data)
-      } catch {
-        // Wrong key / tampered blob — degrade to empty rather than throw.
+      } catch (err) {
+        // Wrong key / tampered blob. In strict mode (rotation) surface this;
+        // otherwise degrade to empty rather than throw.
+        if (strict) {
+          throw new AddressBookDecryptError(undefined, { cause: err })
+        }
         return []
       }
     }
@@ -142,6 +190,25 @@ export class AddressBook {
 
   async load(): Promise<void> {
     const contacts = await this.storage.load()
+    this.contacts = new Map(contacts.map(c => [c.id, c]))
+  }
+
+  /**
+   * Like {@link load}, but if the underlying storage supports a strict load
+   * (e.g. {@link EncryptedAddressBook.loadStrict}) it THROWS when a present blob
+   * fails to decrypt instead of silently loading empty. Used by the
+   * password-rotation re-wrap (#489) so a decrypt failure aborts the rotation
+   * rather than clobbering the on-disk contacts with an empty book. Falls back
+   * to {@link load} for storages without strict support.
+   */
+  async loadStrict(): Promise<void> {
+    const strictCapable = this.storage as AddressBookStorage & {
+      loadStrict?: () => Promise<Contact[]>
+    }
+    const contacts =
+      typeof strictCapable.loadStrict === 'function'
+        ? await strictCapable.loadStrict()
+        : await this.storage.load()
     this.contacts = new Map(contacts.map(c => [c.id, c]))
   }
 
--- a/web/packages/core/src/storage/claim-links.ts
+++ b/web/packages/core/src/storage/claim-links.ts
@@ -113,6 +113,28 @@ export class ClaimLinksLockedError extends Error {
 }
 
 /**
+ * Thrown by {@link EncryptedClaimLinks.loadStrict} when a PRESENT encrypted
+ * claim-link blob fails to decrypt under the current key (wrong key / tampered).
+ *
+ * The lenient {@link EncryptedClaimLinks.load} returns `[]` on decrypt failure
+ * (so a locked app degrades gracefully); the strict path used by the
+ * password-rotation re-wrap (#489) must instead surface the failure so it never
+ * mistakes a decrypt FAILURE for an empty store and clobbers bearer secrets.
+ */
+export class ClaimLinksDecryptError extends Error {
+  constructor(
+    message = 'Claim-link store failed to decrypt under the current key',
+    options?: { cause?: unknown },
+  ) {
+    super(message)
+    this.name = 'ClaimLinksDecryptError'
+    if (options && 'cause' in options) {
+      ;(this as { cause?: unknown }).cause = options.cause
+    }
+  }
+}
+
+/**
  * localStorage-based claim-link storage encrypted under the session
  * {@link VaultKey} (#474).
  *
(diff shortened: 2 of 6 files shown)


rjwalters:
## Judge review — changes requested

Solid feature: scope is clean (web-wallet + @botho/core only, no Rust/consensus/.loom/lockfile changes), it reuses the #475 vault with no reimplemented crypto, and verification reproduces green locally — `pnpm -r typecheck` (7 pkgs), `pnpm --filter @botho/web-wallet build`, and `pnpm test:run` (209 passed / 10 skipped). The happy-path correctness is good:

- **setPassword** correctly encrypts the seed, derives the salt-bound session key, and reuses `applyVaultKey`'s legacy plaintext→encrypted migration to re-wrap contacts/claim links. Survives reload; no plaintext left.
- **changePassword** verifies the old password via `loadWallet` BEFORE writing anything (wrong-password test confirms no mutation), and the explicit `addressBook.load()` / `claimLinkStore.load()` DO run under the still-active OLD session key before the key swap, so in-memory data is real before `rewrapUnderNewKey` persists it under the new key. Old key is invalidated for all three data types.

I'm requesting changes on two related data/fund-loss concerns around the claim-link bearer secrets, plus a test-coverage gap.

### 1. (Blocking) The claim-link re-wrap path — the fund-loss-relevant one — is untested, and the test name misrepresents coverage
Test #2 is titled `changePassword rotates the password; the old password no longer decrypts and **a claim link survives re-wrap**`, but the test body never creates a claim link (`createClaimLink`) and never asserts a `ClaimLinkRecord` (with its `ephMnemonic` bearer secret) decrypts under the new key. It only exercises the address book. Please make this test actually populate a claim link AND a contact, change the password, and assert BOTH survive and decrypt under the new key (and that a stale old key cannot). As written, the path whose failure means loss of bearer secrets (= fund loss) has zero test coverage despite the title claiming it.

### 2. (Blocking) Error handling can silently strand/lose claim-link bearer secrets
Two spots swallow failures in a way that can lose data without surfacing any error:

- In `changePassword`, `try { await claimLinkStore.load() } catch { /* keep in-memory */ }` is misleading. `EncryptedClaimLinks.load()` returns `[]` (it does NOT throw) on a decrypt failure, and `ClaimLinkStore.load()` then assigns that empty map to memory. So a decrypt miss does not "keep in-memory" — it clobbers memory to empty, and the subsequent rewrap would overwrite the on-disk blob with empty data. In the normal unlocked flow with a valid session key this doesn't fire, but the defense the comment promises isn't real.

- In `rewrapUnderNewKey`, the seed is already re-saved under the NEW password (in `changePassword` step 2) BEFORE the store rewrap runs, and `claimLinkStore.rewrap()` failures are swallowed (`catch { /* best-effort */ }`) after the session key has already been swapped to the new key. If the claim-link rewrap throws, you get a half-rotated state: the seed is under the new password, but the claim-link blob on disk is still under the OLD key. On the next session that blob fails to decrypt under the new key → `load()` returns `[]` → bearer secrets silently gone, no error shown.

Suggested direction: bearer-secret loss should never be silent. Treat `claimLinkStore.rewrap()` as a hard failure (don't swallow it), and consider ordering so the seed re-save is the LAST irreversible step (or wrap the rotation so a store-rewrap failure surfaces to the user and leaves a recoverable state). The address book swallow is acceptable (non-bearer, privacy-only) but the claim-link store is not.

### Nits
- The pre-rewrap explicit `load()` calls assume the on-disk blob decrypts under the current session key; if you keep them, the empty-on-decrypt-failure behavior above should be guarded (e.g., skip persist when load yielded empty but a non-empty blob exists on disk) to avoid overwriting a non-empty old-key blob with empty data.

Happy to re-review once the claim-link path is genuinely tested and the rewrap error handling no longer loses bearer secrets silently. CI/build/typecheck all green, so this is purely about the claim-link safety + coverage.

rjwalters:
## Doctor — Judge feedback addressed (claim-link bearer-secret safety + tests)

All three blockers are fixed. Scope unchanged: web-wallet + @botho/core only (no Rust/consensus/.loom/lockfile changes; stray `pnpm-workspace.yaml` allowBuilds mutation reverted before push).

### 1. Half-rotated silent fund-loss — fixed (atomic + abortable + reordered)
`changePassword` now orders the steps so the seed re-save is the LAST irreversible step:
1. Verify old password (unchanged).
2. STRICT-load both stores under the still-active OLD key (`loadStrict`, see #2) — throws if a present blob fails to decrypt, so the in-memory data is provably the real decrypted contents.
3. Derive the NEW key independently via `VaultKey.fromPassword(newPassword)` (fresh salt; each blob is self-describing, so it decrypts the new sibling blobs directly and the new seed blob via salt-fallback).
4. ATOMICALLY re-wrap stores under the new key.
5. ONLY THEN `saveWallet(mnemonic, newPassword)`.

`rewrapUnderNewKey` is now transactional and never swallows the claim-link failure:
- Snapshots both on-disk blobs + the old session key.
- Re-wraps the BEARER-SECRET claim-link store FIRST; its failure is fatal (throws).
- On any re-wrap throw: restores the original blobs + OLD session key and re-throws, so the rotation aborts BEFORE the seed is re-saved → no seed-under-new-key / store-under-old-key state.
- On success returns a `rollback()` that `changePassword` invokes if the final `saveWallet` itself fails, restoring the old blobs+key so the inverse half-rotated state (stores-new / seed-old) is also prevented.

Net: any failure leaves the wallet consistently on the OLD password with all three data types readable; no swallowed errors on the bearer-secret path.

### 2. Decrypt-failure indistinguishable from empty — fixed
Added strict load variants alongside the existing lenient `load()`:
- `EncryptedClaimLinks.loadStrict()` / `EncryptedAddressBook.loadStrict()` THROW (`ClaimLinksDecryptError` / `AddressBookDecryptError`) on decrypt failure (or locked) of a PRESENT `isVaultBlob` blob, instead of returning `[]`. A genuinely empty/absent blob still returns `[]`.
- `ClaimLinkStore.loadStrict()` / `AddressBook.loadStrict()` delegate to the strict storage method (graceful fallback to `load()` for storages without it).
- The lenient `load()` ([]-on-locked) behavior is unchanged, so the normal app paths (mount, unlock, `applyVaultKey`) are unaffected.
The rotation uses `loadStrict()`, so a decrypt failure aborts instead of re-wrapping empty over real bearer secrets.

### 3. Test gap / misleading name — fixed
- The renamed test now persists a REAL `ClaimLinkRecord` (with its `ephMnemonic` bearer secret) encrypted under the OLD key, changes the password, and asserts the full record (including `ephMnemonic`, `ephAddress`, `amount`, `status`) decrypts under the NEW key, that no plaintext bearer secret remains, that a stale OLD key can no longer decrypt either the claim-link or address-book blob, and that the link is visible in context state. A contact is verified alongside.
- New abort-on-failure test: forces the final seed re-save to throw, then asserts the rotation rejected, the seed blob is unchanged (OLD password still decrypts; NEW rejected), both sibling blobs are restored to the OLD-key versions (contact + bearer secret still decrypt under the OLD key), and the live session key is the OLD key (wallet stays usable). Real stores + native vitest matchers only.

### Verification
- `cd web && pnpm install` — only the ERR_PNPM_IGNORED_BUILDS warning.
- `pnpm -r typecheck` — green (7 pkgs).
- `pnpm --filter @botho/web-wallet build` — green.
- `pnpm test:run` — green: **210 passed / 10 skipped**, including the rewritten claim-link-survives-rotation test and the new abort-on-failure test.

Back to `loom:review-requested`.

rjwalters:
## Judge re-review: APPROVED

Final Judge pass of the Doctor→Judge cycle on the password-rotation fund-safety fixes (commit `26119bb`). All three change requests are resolved; happy paths were not touched and not re-litigated.

### 1. Atomic / abortable rotation (closest scrutiny) — VERIFIED
`changePassword` now orders the single irreversible write (the seed re-save) LAST:
1. verify old password (read-only)
2. `claimLinkStore.loadStrict()` + `addressBook.loadStrict()` under the still-active OLD key (throws on decrypt failure → aborts before any write)
3. derive new key
4. `rewrapUnderNewKey(newKey)` — swaps session key, rewraps the **bearer-secret claim-link store FIRST (fatal, throws)**, then the address book; snapshots both on-disk blobs + old key, and on ANY failure `restore()`s them and rethrows
5. only then `saveWallet(mnemonic, newPassword)`; on failure `rollbackRewrap()` restores the old blobs + old session key

- **Fund-loss direction (seed-new while a bearer-secret blob is still old) is unreachable**: step 5 runs only after both rewraps succeeded; a rewrap throw restores and rethrows before the seed is ever re-saved → wallet stays consistently on the OLD password.
- **Inverse (stores-new / seed-old) is rolled back**: `saveWallet` only writes the `botho-wallet-*` keys (confirmed in `wallet/index.ts`), never the sibling blobs, so the rewrap and the seed save are independent; if the seed save throws, `rollbackRewrap()` restores the sibling blobs + old key byte-for-byte.
- **Rewrap failures are not swallowed**: claim-link rewrap is first and fatal; the catch restores + rethrows.
- **Snapshot/restore is real**: `restore()` captures the literal `localStorage` blobs (null → `removeItem`, else `setItem`) and the previous `VaultKey`.

### 2. Decrypt-failure detection — VERIFIED
New `loadStrict()` on `EncryptedClaimLinks` / `EncryptedAddressBook` (+ store wrappers) routes through `loadInternal(strict=true)` and THROWS `ClaimLinksDecryptError` / `AddressBookDecryptError` on decrypt failure (and locked-but-present blob), but returns `[]` for genuinely absent/empty. The lenient `load()` (`loadInternal(false)`, `[]`-on-locked/decrypt-fail) is UNCHANGED, so mount / `applyVaultKey` / unlock callers are unaffected. Rotation uses the strict path, so a decrypt failure can no longer masquerade as an empty store and clobber bearer secrets.

### 3. Tests — real and meaningful
- Survives-rotation test persists a REAL serialized `ClaimLinkRecord` (with `ephMnemonic` bearer secret) under the old key, rotates, and asserts the full record (id/ephMnemonic/ephAddress/amount/status) decrypts under the NEW key, no plaintext `ephMnemonic` remains, and a re-derived OLD key `.rejects.toThrow()`.
- Abort-on-failure test spies `localStorage.setItem` to throw only on the seed key (rewraps succeed, final seed save fails), then asserts seed blob byte-identical to old (`toBe`), old password still decrypts, new password rejected, both sibling blobs restored byte-identical to the old-key versions, and the live session key is the OLD key (decrypts the on-disk bearer secret). Assertions are on byte-identity, not trivially passing.

### 4. Scope — clean
Diff vs merge-base (`49057ac`) touches only web-wallet + @botho/core (6 files). No Rust/consensus, no `.loom-managed`/`pnpm-workspace`/`Cargo` mutations. Crypto reuses the existing `VaultKey` API — no reimplemented crypto.

### 5. CI reproduced locally (worktree, frozen lockfile)
- `pnpm typecheck`: 7 packages green
- `pnpm test:run`: 210 passed / 10 skipped
- `pnpm build`: green

Merge state CLEAN. Approving.

PR #495, opened 2026-06-21
feat(web-wallet): wallet lock — Lock button + inactivity auto-lock (#490)

### LINKED ISSUE #490: web-wallet: wallet lock — Lock button + auto-lock on inactivity (clear in-memory seed/vault key)
The wallet has no real lock. `isLocked` is only set true on initial load for an encrypted wallet (`contexts/wallet.tsx`); once unlocked, the decrypted seed (`mnemonicRef`) and vault key (`vaultKeyRef`/`sessionVaultKey`) stay in memory until a full page refresh. There is no Lock button and no inactivity auto-lock — flagged repeatedly during the security sweep.

Implement:
1. `lockWallet()` on the wallet context: clear the in-memory secrets (`mnemonicRef.current = null`, `applyVaultKey(null)`), set `isLocked: true`, and stop balance/history polling (already gated on `!isLocked`). Do NOT wipe localStorage — the wallet still exists, just locked; unlocking re-derives via `unlockWallet(password)` (exists). When locked, `WalletPage` already renders `UnlockWalletView`.
2. A **Lock** button in the wallet UI (near settings/reset).
3. **Auto-lock on inactivity**: a configurable idle timeout (default e.g. 15 min) that calls `lockWallet()`; reset the timer on user activity (pointer/keyboard/visibility). Persist the timeout preference in localStorage; allow an "off" option. Use a sensible default.
4. Plaintext (no-password) wallets have nothing to unlock with — for them, lock should be DISABLED or prompt the user to set a password first (ties into the add/change-password issue). Do not strand a plaintext wallet in a locked state it can't exit.

Acceptance: Lock button clears in-memory seed + vault key and shows the unlock screen; auto-lock fires after the timeout and is reset by activity; unlock restores access and data (encrypted contacts/claim-links reload via the vault key); plaintext-wallet handling is safe (no lockout). Unit tests: lockWallet clears secrets; auto-lock timer fires/resets; unlock round-trip. typecheck+build+vitest green. Scope: web-wallet (+ @botho/core only if needed).

rjwalters:
Closes #490

## Summary
Adds a real wallet lock to the web wallet. Until now `isLocked` was only set on initial load for an encrypted wallet; once unlocked the decrypted seed (`mnemonicRef`) and vault key (`vaultKeyRef`/`sessionVaultKey`) stayed in memory until a full page refresh, with no Lock button and no inactivity auto-lock.

## What changed
- **`lockWallet()` on the wallet context** (`contexts/wallet.tsx`): wipes in-memory secrets — `mnemonicRef.current = null` and `applyVaultKey(null)` (clears `vaultKeyRef` + module-scope `sessionVaultKey`) — and sets `isLocked: true`. Does NOT touch localStorage; the encrypted wallet still exists. Balance/history polling (already gated on `!isLocked`) stops. Unlock uses the existing `unlockWallet(password)`, which re-derives the key and reloads encrypted contacts/claim-links via `applyVaultKey`.
- **Plaintext safety**: `lockWallet()` is a NO-OP when there is no session vault key (plaintext wallet), so a plaintext wallet can never be stranded in a locked state it can't exit. Lock button + auto-lock are gated on `isEncrypted`.
- **Auto-lock on inactivity**: `autoLockMinutes` preference persisted in localStorage (default 15 min, `0` = Off/Never). An effect arms a `setTimeout` that calls `lockWallet()`, reset on `pointermove/pointerdown/keydown/click/scroll/wheel/touchstart` and `visibilitychange`. Listeners + timer are cleaned up on unmount, when locked, for plaintext wallets, and when off.
- **UI** (`pages/wallet.tsx`): a Lock button in the balance-card actions (disabled for plaintext) and a Settings sheet with the auto-lock timeout selector, Lock now, and Reset. Plaintext wallets are routed to "Set a password" first (reuses #489).

## How secrets are cleared
`lockWallet()` sets `mnemonicRef.current = null` (decrypted seed gone) and calls `applyVaultKey(null)` which sets `vaultKeyRef.current = null` and `setSessionVaultKey(null)` (vault key gone from both the ref and the module-scope holder the encrypted stores read). After lock, the encrypted address book / claim-link stores read as empty until unlock.

## Tests (native vitest matchers only)
New `wallet-lock.test.tsx` using real crypto (only adapter + wasm signer stubbed):
- lockWallet clears in-memory seed + vault key and sets isLocked
- encrypted contacts unreadable after lock, restored on unlock round-trip
- plaintext-wallet lock is a no-op (no lockout)
- auto-lock timer fires after timeout (fake timers), is reset by activity, is disabled for plaintext, and off (0) disables it

## Verification
- `pnpm -r typecheck` green
- `pnpm --filter @botho/web-wallet build` green
- `pnpm test:run` green (218 passed, 10 skipped)

Scope: web-wallet only. No consensus/faucet/network/Rust changes. Stray `.loom-managed` / `pnpm-workspace.yaml` mutations reverted before push.

### DIFF
--- a/web/packages/web-wallet/src/contexts/wallet-lock.test.tsx
+++ b/web/packages/web-wallet/src/contexts/wallet-lock.test.tsx
@@ -1,0 +1,285 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Tests for the wallet LOCK + auto-lock-on-inactivity feature (#490).
+ *
+ * Like wallet-password.test.tsx (and unlike wallet.test.tsx) this file does NOT
+ * mock @botho/core's AddressBook / ClaimLinkStore: the whole security point of
+ * locking is that the decrypted seed + session vault key are wiped from memory
+ * so the REAL encrypted address book / claim-link blobs become unreadable until
+ * an unlock re-derives the key. Only the network adapter and wasm signer are
+ * stubbed.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
+import { render, waitFor, act, cleanup } from '@testing-library/react'
+import { WalletProvider, useWallet } from './wallet'
+
+// Stub the wasm signer so the module import is hermetic (no wasm artifact on a
+// fresh checkout). None of these tests exercise real signing/scanning.
+vi.mock('@botho/wasm-signer', () => ({
+  spendableBalance: vi.fn().mockResolvedValue(0n),
+  buildOwnedHistory: vi.fn().mockResolvedValue([]),
+  buildSendTransaction: vi.fn().mockResolvedValue({ txHex: '0xstub' }),
+}))
+
+// Real-backing localStorage mock so the encrypted stores actually persist blobs
+// we can inspect across lock/unlock.
+const localStorageMock = (() => {
+  let store: Record<string, string> = {}
+  return {
+    getItem: (key: string) => store[key] ?? null,
+    setItem: (key: string, value: string) => { store[key] = value },
+    removeItem: (key: string) => { delete store[key] },
+    clear: () => { store = {} },
+    _dump: () => ({ ...store }),
+  }
+})()
+Object.defineProperty(globalThis, 'localStorage', { value: localStorageMock })
+
+// Stub the adapter. isConnected returns false so unlock/create do not attempt a
+// balance fetch (which would reach the stubbed wasm path); the lock-state
+// assertions do not depend on balances.
+vi.mock('@botho/adapters', () => ({
+  RemoteNodeAdapter: class MockRemoteNodeAdapter {
+    connect = vi.fn().mockResolvedValue(undefined)
+    disconnect = vi.fn()
+    isConnected = vi.fn().mockReturnValue(false)
+    getNodeInfo = vi.fn().mockReturnValue({ version: '1.0.0', network: 'testnet' })
+    getBalance = vi.fn().mockResolvedValue({ available: 0n, pending: 0n, total: 0n })
+    getTransactionHistory = vi.fn().mockResolvedValue([])
+    onNewBlock = vi.fn().mockReturnValue(() => {})
+    onTransaction = vi.fn().mockReturnValue(() => {})
+    onMempoolUpdate = vi.fn().mockReturnValue(() => {})
+    onPeerStatus = vi.fn().mockReturnValue(() => {})
+    getWsStatus = vi.fn().mockReturnValue('disconnected')
+    onWsStatusChange = vi.fn().mockReturnValue(() => {})
+  },
+}))
+
+const TEST_MNEMONIC_12 =
+  'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about'
+const A_CONTACT = 'tbotho://1/contactaddrabc'
+
+const STORAGE_AUTO_LOCK = 'botho-auto-lock-minutes'
+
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed PR #495 (wallet lock + inactivity auto-lock, closes #490). Verified all five focus areas against the worktree at `feature/issue-490` (head `8f00c10`).

### 1. Secrets actually cleared (the security point) — CONFIRMED
`lockWallet()` calls `applyVaultKey(null)`, which sets BOTH `vaultKeyRef.current = null` AND `setSessionVaultKey(null)` — the module-scope holder that `EncryptedAddressBook(() => sessionVaultKey)` and `EncryptedClaimLinks(() => sessionVaultKey)` read. It also nulls `mnemonicRef.current`. The synchronous wipes run before the first `await` in `applyVaultKey` (called via `void`), so the decrypted seed and vault key are gone from memory immediately, not eventually. The test `after lock, encrypted contacts are unreadable until unlock restores access` exercises this with the REAL (unmocked) `@botho/core` EncryptedAddressBook/VaultKey — only the adapter + wasm-signer are stubbed — and asserts contacts vanish (store reads empty with no key) while the on-disk blob is preserved, then reappear after unlock. The companion test also asserts `getVaultKey() === null` and that `exportWallet()` rejects with "password required" post-lock, proving the cleartext mnemonic is no longer held.

### 2. No plaintext-wallet lockout — CONFIRMED
`lockWallet` early-returns when `vaultKeyRef.current === null` (true for plaintext), so a plaintext wallet can never enter a locked state. The auto-lock effect early-returns on `!state.isEncrypted`. UI: header Lock button and the settings auto-lock selector are both `disabled={!isEncrypted}`, with a link routing to set-password (#489 'set' mode). Tests `lockWallet is a NO-OP for a plaintext wallet` and `does not auto-lock a plaintext wallet even with a timeout set` both pass.

### 3. Auto-lock correctness — CONFIRMED
Fake-timer tests genuinely advance timers: fires at the configured timeout (unlocked at 59s, locked at 61s, vault key null after); activity (`keydown`) resets the countdown so it does not fire early; off (0) disables entirely; plaintext never locks. Effect deps `[autoLockMinutes, state.isLocked, state.isEncrypted, state.hasWallet, lockWallet]` are complete; `lockWallet` is a stable useCallback. Timer + all listeners (pointer/key/click/scroll/wheel/touch + visibilitychange) are torn down in cleanup on unmount/lock/dep-change. On unlock-without-activity the effect re-runs and starts a fresh full-interval timer — no stale immediate fire.

### 4. Unlock round-trip — CONFIRMED
`unlockWallet(password)` re-derives the key via `loadWalletWithKey` and `applyVaultKey(stored.vaultKey)` reloads the encrypted stores. The unreadable-after-lock test confirms the 'Alice' contact and the seed are fully restored after unlock. No data loss.

### 5. Scope + CI — CONFIRMED
Diff vs merge-base touches exactly 3 files: `contexts/wallet.tsx`, `pages/wallet.tsx`, `contexts/wallet-lock.test.tsx`. No Rust/consensus, no .loom-managed, no pnpm-workspace mutations. Reproduced locally: typecheck clean (`tsc --noEmit`), `vitest run` 59 passed / 0 failed (8 new lock tests), `build` succeeds. mergeStateStatus is CLEAN.

Note: `gh pr checks` reports no checks on this branch (web-wallet changes may not trigger the configured workflows), but I reproduced typecheck/build/vitest green locally as the equivalent validation.

Solid, security-focused implementation. Approving.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

Issue #468, opened 2026-06-21
web-wallet: remove redundant per-tx 'Private' badges + PWA auto-reload on update (fix stale cache)


rjwalters:
Two wallet UX fixes from live use of wallet.botho.io:

## 1. Remove redundant "Private" badges from transaction list
Every Botho value transfer is private (CLSAG ring sigs + confidential amounts), so a per-transaction "Private" badge is pure noise. Remove it from the wallet's transaction display.
- `web/packages/features/src/wallet/components/transaction-row.tsx` renders `<PrivacyBadge>`; `transaction-list.tsx` has a show-privacy-badges prop. Stop showing the badge in the wallet tx list (default it off / remove from the row). Keep the `PrivacyBadge` component itself if used elsewhere (e.g. send-modal privacy selector / explorer), but the wallet transaction history should not show a per-row "Private" tag.
- Verify the wallet transaction list renders cleanly without the badge (amounts, direction, time, status — no "Private" chip per row).

## 2. PWA: auto-reload on update (fixes stale-cache after deploys)
`web/packages/web-wallet/vite.config.ts` uses `registerType: 'autoUpdate'`, but the generated registerSW is the bare `navigator.serviceWorker.register('/sw.js')` (no auto-reload), so after a deploy users keep seeing the OLD cached app until a manual hard-refresh. This caused wallet.botho.io to keep showing the landing page even after the root-redirect fix shipped.
- Make the PWA take over + reload on update so users get new versions automatically. Options: use vite-plugin-pwa's virtual `virtual:pwa-register` with `registerType:'autoUpdate'` (which reloads on new SW), OR add a `controllerchange` listener that calls `window.location.reload()` once when the new SW activates. The SW already has skipWaiting+clientsClaim. Avoid reload loops (guard with a flag).
- Goal: after a new deploy, an existing visitor gets the new app within one navigation/refresh automatically — no manual hard-refresh / incognito needed.

## Acceptance
1. Wallet transaction list shows NO per-row "Private" badge; layout clean. PrivacyBadge still available where genuinely used (don't break send-modal/explorer if they use it).
2. After a deploy, an existing PWA user auto-updates to the new version (no manual hard refresh) — verified by the SW update→reload path; no reload loop.
3. typecheck + build green; existing e2e green; redeploy to Cloudflare Pages (--branch=main) after merge (maintainer/automation handles redeploy).

Refs: wallet.botho.io, #457/#459/#463 (wallet), vite-plugin-pwa.

PR #497, opened 2026-06-21
feat(web-wallet): validate recipient address format in send flow

### LINKED ISSUE #491: web-wallet: validate recipient address format in send flow (block irreversible sends to bad addresses)
The send flow validates fee/insufficient-balance and has a confirm step, but does NOT validate the recipient ADDRESS FORMAT before submit (`features/src/wallet/components/send-modal.tsx` only checks the recipient is non-empty). A malformed or typo'd address can be submitted; sends are irreversible.

Implement:
1. Validate the recipient address format in the send form using `@botho/core`'s `parseAddress`/address validation (the `features` package already imports from `@botho/core`, e.g. `formatBTH`). Show a clear inline error ("Invalid Botho address") and keep the submit button disabled until the address is valid (in addition to the existing amount/balance checks). If you prefer not to couple features→core directly, accept an optional `validateAddress?: (a: string) => boolean` prop wired from `wallet.tsx`; but reusing the existing core dependency is simpler.
2. Self-send guard: if the recipient equals the user's own address, warn (allow with confirmation, or block — pick the safer UX) — pass the own address in if needed.
3. Keep/firm up the pre-send confirmation showing recipient (with contact name if known), amount, and fee.
4. Apply the same validation to the Pay flow's send (`pages/pay.tsx`) where the recipient comes from a link (a malformed link `to` should error clearly).

Acceptance: an invalid address shows an inline error and blocks submit; a valid address passes; self-send is handled; confirmation shows recipient/amount/fee. Unit tests for valid/invalid/self-send. typecheck+build+vitest green; do not reintroduce the per-row Private badge (#468). Scope: web-wallet + @botho/features (+ @botho/core if needed).

rjwalters:
Closes #491

## Problem
The send form only checked the recipient was non-empty — a malformed or typo'd address could be submitted. Sends are irreversible.

## Changes
- **`@botho/features` send-modal.tsx**: validate the recipient with `@botho/core`'s `isValidAddress`. Show an inline "Invalid Botho address" error when the field is non-empty but invalid (empty stays silent), and keep the submit button disabled until the address parses (in addition to the existing amount/balance checks).
- **Self-send guard**: new optional `ownAddress` prop. When provided and the recipient matches, the modal warns ("This is your own address — you can't send to yourself.") and blocks submit. Chose **block** over allow-with-confirmation as the clearer/safer UX for an action that's almost always a mistake.
- **New optional `validateAddress` prop** (defaults to `isValidAddress`) for testability/customisation. Both new props are optional — existing callers are unaffected.
- **`web-wallet/wallet.tsx`**: wires `ownAddress={address}` from `useWallet()`.
- **Confirmation**: the fee summary (recipient name via contact match, amount, total + fee) is preserved.

## Pay flow
`pages/pay.tsx` already validates via `isValidAddress` and renders a clear "invalid recipient address" error for malformed links, plus a self-pay warning — so requirement #5 was already satisfied; no change needed.

## Notes
- Did NOT reintroduce the per-row Private badge (#468).
- `@botho/features` imports only from `@botho/core` (already a dep), never the web-wallet context.
- Reverted stray `.loom-managed` / `pnpm-workspace.yaml` tooling mutations before pushing.

## Validator
`isValidAddress(address: string): boolean` already exists/exported in `@botho/core` (`web/packages/core/src/wallet/address.ts`, wrapping `parseAddress` in try/catch). Reused as-is.

## Verification
- `pnpm install` — clean (ERR_PNPM_IGNORED_BUILDS warning only)
- `pnpm -r typecheck` — green
- `pnpm --filter @botho/web-wallet build` — green
- `pnpm test:run` — 230 passed / 10 skipped, incl. 7 new send-modal tests (invalid blocks, valid passes, self-send warned/blocked, self-send skipped without ownAddress)

### DIFF
--- a/web/packages/features/src/wallet/components/send-modal.test.tsx
+++ b/web/packages/features/src/wallet/components/send-modal.test.tsx
@@ -1,0 +1,128 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { describe, it, expect, beforeEach, vi } from 'vitest'
+import { render, screen, fireEvent, cleanup } from '@testing-library/react'
+import type { Balance } from '@botho/core'
+import { deriveAddress } from '@botho/core'
+import { SendModal } from './send-modal'
+
+// A couple of real, valid testnet addresses derived from fixed mnemonics. Using
+// the real deriver keeps the test honest about what the production validator
+// (`@botho/core`'s isValidAddress) accepts.
+const VALID_ADDRESS = deriveAddress(
+  'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about',
+)
+const OTHER_ADDRESS = deriveAddress(
+  'legal winner thank year wave sausage worth useful legal winner thank yellow',
+)
+
+const BALANCE: Balance = {
+  available: 1_000_000_000_000_000n,
+  pending: 0n,
+  total: 1_000_000_000_000_000n,
+}
+
+function setup(overrides: Partial<React.ComponentProps<typeof SendModal>> = {}) {
+  const onSend = vi.fn().mockResolvedValue({ success: true, txHash: '0xabc' })
+  const props: React.ComponentProps<typeof SendModal> = {
+    isOpen: true,
+    onClose: vi.fn(),
+    balance: BALANCE,
+    estimateFee: vi.fn().mockResolvedValue(4000n),
+    onSend,
+    ...overrides,
+  }
+  render(<SendModal {...props} />)
+  return { onSend, props }
+}
+
+function getRecipientInput(): HTMLInputElement {
+  return screen.getByPlaceholderText(/tbotho:\/\//i) as HTMLInputElement
+}
+
+function getAmountInput(): HTMLInputElement {
+  return screen.getByPlaceholderText('0.00') as HTMLInputElement
+}
+
+function getSubmitButton(): HTMLButtonElement {
+  return screen.getByRole('button', { name: /Send Transaction/i }) as HTMLButtonElement
+}
+
+describe('SendModal address validation', () => {
+  beforeEach(() => cleanup())
+
+  it('does not show an error for an empty recipient, but keeps submit disabled', () => {
+    setup()
+    expect(screen.queryByText('Invalid Botho address')).toBeNull()
+    expect(getSubmitButton().disabled).toBe(true)
+  })
+
+  it('shows an inline error and disables submit for a malformed address', () => {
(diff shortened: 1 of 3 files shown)


rjwalters:
## Approved (LGTM)

Reviewed against issue #491 (validate recipient address format in the send flow). Verified every point and reproduced CI locally in the worktree.

### Validation correctness
- `features/src/wallet/components/send-modal.tsx` validates via `@botho/core`'s `isValidAddress`, which is the real `parseAddress`-based validator (prefix check + base58 decode + strict 64-byte length), not a loose regex. Malformed/garbage recipients are rejected.
- Non-empty invalid recipient shows inline "Invalid Botho address" and keeps submit disabled (`recipientInvalid` + `disabled={... || !recipientValid ...}`).
- Valid `tbotho://` address passes; empty field shows no error and just stays disabled. `handleSend` also defensively re-checks before calling `onSend`.

### Self-send guard
- `recipient == ownAddress` is blocked with a clear message and disabled submit. Comparison is robust — trimmed, case-insensitive (a superset of exact match; base58 is case-sensitive so this can't miss a real distinct address). Guard only activates when `ownAddress` is provided.

### Back-compat & layering
- New props `ownAddress?: string | null` and `validateAddress?` are both optional with sane defaults; callers without them typecheck and behave (self-send guard simply inactive). `wallet.tsx` passes `ownAddress={address}` (`string | null` from `useWallet()`, matches the prop type).
- `features` imports only `@botho/core` and `@botho/ui` — no web-wallet context leak. Layering intact.

### Pay flow
- `pages/pay.tsx` genuinely already validates: `addressValid = isValidAddress(request.to)`, `isSelfPay`, and `canPay = addressValid && amount > 0n && !amountError && !isSending`. Builder's "no change needed" claim confirmed — not an unhandled gap.

### Other
- No "Private" badge reintroduced (#468); `privacyLevel` is an internal hardcoded `'standard'`, no UI badge.
- Tests are meaningful: 7 new tests use the real `deriveAddress` (honest about what the production validator accepts), native matchers, no jest-dom. Cover empty/invalid/valid/self-send/back-compat.
- Scope clean: all changes within `web/`; no Rust/consensus, no .loom-managed or pnpm-workspace mutations.

### CI reproduced locally (worktree)
- typecheck: all 7 packages pass
- build: all packages built
- vitest: 230 passed / 10 skipped (matches builder's report)
- mergeStateStatus: CLEAN

Approving.

Issue #511, opened 2026-06-21
feat(consensus): quorum-health simulator/analyzer for curated FBAS federation (#510 deliverable)


rjwalters:
Deliverable 2 of #510, scoped to **Path A** (curated validator federation + thin clients — see #427). A Rust tool to validate Botho's quorum rule set and monitor the curated federation's safety/liveness buffers as it grows and churns. Fills the gap the research identified: static FBAS analyzers exist (`fbas_analyzer`, `python-fbas`) but no Botho-grounded tool tied to our threshold formula.

## v1 scope (this issue)
A library + small CLI/example (suggested location: `consensus/quorum-sim/` or `tools/quorum-sim/` — verify repo layout) that:
1. **FBAS model**: nodes with threshold-based quorum sets (symmetric top-tier + arbitrary per-node slices). Realistic size N ≤ ~20 (curated federation), so brute-force over subsets is acceptable.
2. **Static analysis** (the `fbas_analyzer` metric set):
   - quorum predicate + **quorum-intersection** check;
   - enumerate **minimal quorums**, **minimal blocking sets** (liveness buffer), **minimal splitting sets** (safety buffer).
3. **Threshold-rule comparison**: for N = 2..12, compare current `effective_threshold = n − floor((n−1)/3)`, **ceil(0.67·n)**, and unanimity → report min blocking-set cardinality (failures-to-halt) and whether disjoint (non-intersecting) quorums exist (fork risk).
4. **Growth/churn dynamics (Path A core)**: simulate curated **admission** (add validator + slice template) and **reactive-shun** (remove validator); report how minimal blocking/splitting-set cardinalities evolve, and **flag any add/remove that breaks quorum intersection** (the coNP-complete check, brute-forced at small N).
5. **Report output**: human-readable table + machine-readable (JSON) for CI/monitoring.

## Correctness tests (must assert known results from #510)
- Current formula degenerates to unanimity at n≤3 → minimal blocking set cardinality 1.
- `2-of-4` admits disjoint quorums {A,B}/{C,D} → quorum intersection FALSE (fork).
- `ceil(0.67·n)`: f=1 tolerance requires n≥4; report the liveness margin m−t+1.
- ≥4 distinct entities required before any nonzero fault tolerance.

## Grounding
- `botho/src/config.rs` `effective_threshold` (the formula under test) + existing `test_quorum_effective_threshold_*`.
[...]

PR #512, opened 2026-06-21
feat(consensus): quorum-health analyzer for curated FBAS federation (#511)

rjwalters:
Closes #511. Deliverable 2 of #510, scoped to Path A (curated validator federation + thin clients; see #427).

## What v1 covers

New crate `consensus/quorum-sim/` (`bth-quorum-sim`), added to the workspace members, with a `quorum-sim` CLI. A static, Botho-grounded re-implementation of the `fbas_analyzer` metric set, tied to Botho's threshold rule `effective_threshold = n − floor((n−1)/3)` (`botho/src/config.rs`).

Modules:
- **nodeset** — `NodeSet` (u32 bitset) for `N ≤ ~20`; lets us brute-force over `2^N` subsets exactly.
- **model** — `Fbas`/`Node`/`QuorumSet`; symmetric top-tier construction (`symmetric_botho` ties to the config formula) AND arbitrary per-node slices; `is_quorum`; curated `admit`/`admit_symmetric` and reactive `shun`/`shun_symmetric` growth dynamics.
- **analysis** — `has_quorum_intersection` (fork check), `minimal_quorums`, `minimal_blocking_sets` (liveness buffer), `minimal_splitting_sets` (safety buffer); serde `HealthReport`.
- **thresholds** — `botho_bft` vs `ceil(0.67·n)` vs `unanimity`, plus supermajority fault tolerance.
- **report** — threshold-rule comparison (N=2..12) and admission/shun churn timeline; human-readable tables + JSON; flags any add/remove that breaks quorum intersection.
- **bin/quorum-sim** — `compare` / `analyze` / `churn` subcommands, all with `--json`.

## Mandatory #510 correctness tests (`tests/correctness.rs`) — all pass

- Current formula degenerates to unanimity at n≤3 → min blocking-set cardinality == 1.
- 2-of-4 admits disjoint quorums {A,B}/{C,D} → `quorum_intersection() == false`.
- `ceil(0.67·n)`: f=1 tolerance requires n≥4; liveness margin `m−t+1` verified.
- <4 distinct entities → no real fault tolerance (blocking cardinality 1).
- Cross-check against `fbas_analyzer` semantics: minimal blocking/splitting sets are top-tier nodes.

## Test results

`cargo test -p bth-quorum-sim`: **29 passed, 0 failed** (24 unit + 5 correctness). `cargo clippy -p bth-quorum-sim --all-targets` clean.

## Explicitly out of scope (deferred follow-up)

[...]

PR #516, opened 2026-06-22
feat(consensus): dynamic message-level SCP simulator — empirical fork/stall detection (#514)

### LINKED ISSUE #514: feat(consensus): dynamic message-level SCP simulator — empirical fork/stall detection (follow-up to #511)
Follow-up to #511 (static analyzer shipped in #512). Build the **dynamic, message-level SCP round simulator** to empirically detect **forks (safety violations)** and **stalls (liveness violations)** under faulty/Byzantine scenarios — the piece the #510 research found does not exist anywhere (static FBAS analyzers exist; a dynamic SCP fork/stall sim does not).

## Why now
The remaining open consensus decision (#427) is the **proposer model**: drop competing-coinbase → explicit leader election (SCP-native hash priority vs VRF). That decision needs *empirical* fork/stall/fairness data under partial synchrony + Byzantine faults, which only a dynamic simulator can provide. This is the validate-before-landing tool for that change (no-hard-forks philosophy).

## Scope
Extend the `bth-quorum-sim` crate (or a sibling) with:
1. **Round-based SCP-ish agreement engine** over the existing FBAS model: nomination (leader-priority) + ballot phases at the level of detail needed to observe agreement/divergence. Synchronous + partially-synchronous (configurable message delay/drop) network model.
2. **Fault injection**: crash (silent), and Byzantine (equivocation — sending different values to different peers; this is the fork-inducing adversary).
3. **Proposer models to compare**: (a) competing-coinbase (every minter nominates its own coinbase), (b) SCP-native hash-priority leader, (c) round-robin leader, (d) VRF-style leader. Measure fork rate (safety), stall rate (liveness), and leadership fairness / grinding susceptibility (cf. arXiv:2501.02970).
4. **Metrics + report**: per-config fork count, stall count, rounds-to-decide distribution, leadership distribution; table + JSON.
5. **Tests**: a known-safe config never forks; an equivocating Byzantine node below the splitting-set threshold never forks but above it can; unanimity below 4 stalls on one crash.

## Grounding
- `consensus/quorum-sim/` (static analyzer to reuse: quorum/blocking/splitting-set logic).
- `botho/src/consensus/` for the real SCP nomination/ballot behavior to mirror at the right abstraction.
- #510 research (Threads A+B) for proposer-model tradeoffs and the grinding/fairness concerns.

## Out of scope
- Production consensus changes (this is a SIMULATOR; the actual proposer-model change is a separate issue gated on this tool's results).

Part of #427. Follow-up to #511.

rjwalters:
Closes #514. Follow-up to #511/#512 (static FBAS analyzer). Extends `bth-quorum-sim` with the **dynamic, message-level SCP round simulator** that empirically detects **forks (safety)** and **stalls (liveness)** under faulty/Byzantine scenarios — the validate-before-landing tool for the open #427 proposer-model decision (drop competing-coinbase → explicit leader election).

## What's new

- **`consensus/quorum-sim/src/sim.rs`** — the dynamic engine. Round-based nomination + **accept-lock** + commit over the existing FBAS model, reusing `Fbas::is_quorum` as the single quorum oracle so dynamic outcomes cross-check the static splitting/blocking-set predictions.
- **`consensus/quorum-sim/tests/dynamic.rs`** — the mandatory acceptance tests.
- **`simulate` CLI subcommand** wired into the existing `quorum-sim` binary (matches the clap structure; sweeps all proposer models by default; `--json`).
- `lib.rs` re-exports, README section, `Cargo.toml` deps (`rand`, `rand_chacha`).

## Proposer models implemented

`competing-coinbase` (Botho's current model), `hash-priority-leader` (SCP-native `Gi(slot,node)`), `round-robin-leader`, `vrf-leader` (seeded-hash VRF approximation).

## Fault types implemented

`crash` (silent) and Byzantine `equivocate` (sends different values to different peers — the fork-inducing adversary).

## Network + determinism

Synchronous and partially-synchronous (seeded per-message delay + drop). All randomness funnels through a single explicitly-seeded `ChaCha8Rng`, so `(config, seed)` is bit-for-bit reproducible (asserted).

## Metrics + report

Per-config fork / stall / agreement counts, rounds-to-decide histogram, leadership distribution + fairness. Human table + serde JSON, matching the existing report style.

## SCP simplifications (kept honest — documented in `src/sim.rs` module docs + README)

- **accept→confirm collapsed into one accept-lock** (a correct node locks the first value it sees a quorum support, pins its vote, never retracts). This is what produces the splitting-set safety guarantee.
- **one slot + one leader per run** (no mid-slot leader rotation, which would inject spurious churn unrelated to safety); leadership **fairness is measured across seeds** (each seed = a distinct slot).
- **no leader-timeout recovery**: a *crashed* leader is survived via a deterministic lowest-value-heard fallback, but a *Byzantine (equivocating) leader stalls* the slot (reported as liveness, not safety). SCP's real leader replacement is out of scope.
- value identity is a small integer standing in for a coinbase/tx_hash; the production combiner is abstracted to "lowest value id wins".

## Test results (all pass)

`cargo test -p bth-quorum-sim`: 32 lib + 5 static `correctness` + 5 `dynamic` acceptance tests.

Mandatory bar:
- `known_safe_below_splitting_never_forks` — n∈{4,7,10}, equivocators **below** the static splitting set, every proposer, partial synchrony, 200 seeds → **0 forks**.
- `equivocation_crosses_static_splitting_threshold` — 1 equivocator (< splitting set 2) never forks; 2 (= splitting set, leader equivocating) **CAN** fork and it is detected. Cross-checks the static analyzer dynamically.
- `unanimity_below_four_stalls_on_crash` — n=3 (3-of-3) + 1 crash → 100% stall.
- `crash_below_blocking_set_stays_live` — n=4 (3-of-4) + 1 crash → stays live.
- `reproducibility_same_seed_same_outcome` — identical outcomes per seed; deterministic aggregate.

`cargo fmt` (pinned nightly-2025-12-03) clean; `cargo clippy -p bth-quorum-sim --all-targets` clean; `cargo build` clean.

## Sample output (informs #427)

```
=== n=4, 2 equivocators (= splitting set), synchronous ===
proposer               n   thr   network  forks  stall  agree  mean_rds fairness
competing-coinbase     4     3      sync      0    300      0         -      n/a
hash-priority-leader   4     3      sync    149    151      0      2.00     0.97
round-robin-leader     4     3      sync    150    150      0      2.00     1.00
vrf-leader             4     3      sync    138    162      0      2.00     0.72
```

Forks appear in leader models exactly when the Byzantine leader equivocates (≈half the seeds), confirming the splitting-set=2 prediction; round-robin shows perfect leadership fairness (1.00) vs VRF's lopsidedness (0.72) at small n.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1463,6 +1463,8 @@ name = "bth-quorum-sim"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "rand 0.8.5",
+ "rand_chacha 0.3.1",
  "serde",
  "serde_json",
 ]
--- a/consensus/quorum-sim/Cargo.toml
+++ b/consensus/quorum-sim/Cargo.toml
@@ -14,6 +14,11 @@ path = "src/bin/quorum-sim.rs"
 serde = { workspace = true, features = ["derive", "alloc"] }
 serde_json = { workspace = true, features = ["std"] }
 clap = { workspace = true, features = ["derive"] }
+# Seeded, reproducible RNG for the dynamic simulator. ChaCha8 is deterministic
+# given a seed and identical across platforms/runs, so simulations reproduce
+# exactly (see the `sim` module). All nondeterminism is funneled through this.
+rand = { workspace = true }
+rand_chacha = { workspace = true }
 
 [features]
 default = []
--- a/consensus/quorum-sim/README.md
+++ b/consensus/quorum-sim/README.md
@@ -1,23 +1,25 @@
 # bth-quorum-sim
 
-Static quorum-health analyzer for Botho's curated FBAS federation (Path A).
+Quorum-health analyzer **and** dynamic SCP simulator for Botho's curated FBAS
+federation (Path A).
 
-This is **Deliverable 2 of #510**, scoped to **Path A** (a curated validator
-federation + thin clients; see #427). It is a small, Botho-grounded
-re-implementation of the static metrics computed by tools such as
-[`fbas_analyzer`](https://github.com/wiberlin/fbas_analyzer) and
-[`python-fbas`](https://github.com/nano-o/python-fbas), tied directly to
-Botho's threshold rule.
+Two layers:
 
-## Scope (v1)
+1. **Static analyzer** (#511/#512) — a small, Botho-grounded re-implementation
+   of the metrics computed by tools such as
+   [`fbas_analyzer`](https://github.com/wiberlin/fbas_analyzer) and
+   [`python-fbas`](https://github.com/nano-o/python-fbas), tied directly to
+   Botho's threshold rule.
+2. **Dynamic message-level SCP simulator** (#514) — a round-based agreement
+   engine that *runs* a nomination + accept-lock + commit protocol over the same
+   FBAS model under faults and a configurable network, and empirically detects
+   **forks** (safety violations) and **stalls** (liveness violations). It is the
+   validate-before-landing tool for the #427 proposer-model decision (drop
+   competing-coinbase → explicit leader election).
 
-Static analysis + threshold/growth comparison. **No** dynamic message-level SCP
-round simulator, Byzantine equivocation injection, or partial-synchrony
-modelling — that is an explicit follow-up.
-
-The curated federation is small (`N ≤ ~20`), so every analysis brute-forces
-over the `2^N` node subsets. This is exponential but exact, and trivially fast
-at the targeted sizes.
+The curated federation is small (`N ≤ ~20`), so every static analysis
+brute-forces over the `2^N` node subsets. This is exponential but exact, and
+trivially fast at the targeted sizes.
 
 ## What it computes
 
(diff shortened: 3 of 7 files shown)


rjwalters:
## Judge review: APPROVE

Reviewed as the validate-before-landing tool for the #427 proposer-model decision, so the bar was correctness of fork/stall detection and honesty of the SCP abstraction.

### Verification (all green)
- `cargo test -p bth-quorum-sim`: 32 lib + 5 static + 5 dynamic acceptance tests pass.
- `cargo clippy -p bth-quorum-sim --all-targets` clean; `cargo fmt --check` clean.
- CI: `cargo build --workspace`, `Generate Coverage Report`, `rustfmt --check` all pass; mergeStateStatus CLEAN.

### Correctness — the parts that matter most
- **Fork detection is sound.** `run_tracked` builds `correct_committed` excluding faulty nodes (`is_faulty(i)` filter), then classifies `distinct.len() >= 2` over the *correct* committers as a Fork. A fork is therefore genuinely "≥2 correct nodes committed different values in a slot" — a real safety violation, not a spurious artifact.
- **The static/dynamic cross-check is real and non-vacuous.** Below the static minimal splitting set, 0 forks across n∈{4,7,10}, every proposer, partial synchrony, 200 seeds. At the splitting set (2 equivocators incl. the leader, n=4), forks appear in ~half the seeds (ran it: round-robin 100/200, hash-priority 98, vrf 89). This dynamically realizes the #512 static splitting-set=2 prediction.
- **Reproducibility is genuinely seeded.** All randomness (delay, drop, vrf_seed) funnels through one `ChaCha8Rng::seed_from_u64`; BTreeMap throughout, no wall clock, no threads. Same `(config, seed)` reproduces bit-for-bit (asserted).
- **Static model is reused, not duplicated.** Commit decisions delegate to `Fbas::is_quorum`; tests read `health_report().min_splitting_set_cardinality` / `min_blocking_set_cardinality`. The dynamic engine never re-derives quorum logic.
- **Simplifications are documented and non-invalidating.** The four declared simplifications (accept→confirm collapsed to one accept-lock; one slot/one leader per run with fairness across seeds; no leader-timeout recovery so a Byzantine leader stalls; integer value identity) are each in the module docs + README and none of them invalidates the fork/stall conclusions. The accept-lock is the correct mechanism for honoring the splitting-set safety bound; the crashed-leader deterministic fallback correctly separates the liveness story from safety.

### One non-blocking caveat for whoever reads the #427 data
The raw `simulate` table can be misread. In the equivocation runs, competing-coinbase shows **0 forks** at/above the splitting set while the leader models fork — which could be read as "competing-coinbase is *safer* than leader election," the opposite of the ratified #427 direction. Two things make that reading wrong, and neither is currently called out next to the numbers:

1. Competing-coinbase's fork-immunity here is partly an **artifact of the value-id ordering**: equivocated values are assigned ids `n + to` (always ≥ n), so under the "lowest-id-wins" combiner an honest node never locks an equivocated value while any honest value (id `0..n`) is in view. The equivocator can split votes (→ stalls) but is structurally barred from inducing a fork. A combiner where the adversary could inject a *lower* id would change this.
2. The honest comparison is: **below the splitting set every model (including the leader models) shows 0 forks** (per `known_safe_below_splitting_never_forks`); the leader-model forks only appear **at/above the splitting set, where no protocol is safe**. So the table does not show leader election to be less safe than competing-coinbase in the regime where safety is actually guaranteed.

The decision-relevant fairness output (round-robin 1.00 vs vrf ~0.72–0.77 vs hash-priority lopsided at small n) is sound and is the right kind of signal for #427. Recommend a one-line note in the README/PR (and ideally a comment by the equivocation-value formula) so a future reader of the fork column doesn't draw the inverted conclusion. Filing as a follow-up is fine — not a blocker for v1 test tooling.

PR #518, opened 2026-06-22
fix(quorum-sim): require confirming quorum of accepters to commit (two-phase federated voting)

### LINKED ISSUE #517: bug(quorum-sim): dynamic SCP simulator forks with ZERO faulty nodes under message delay (accept-lock loses asynchronous safety)
The dynamic SCP simulator (#514/#516) reports **forks (≥2 correct nodes committing different values — a safety violation) with no Byzantine or crashed nodes, purely under message delay.** SCP/federated voting is provably safe under full asynchrony (quorum intersection guarantees agreement regardless of timing), so a correct model must NEVER fork without faulty nodes. This is a fidelity defect.

## Repro (n=7, no faulty nodes, 200 seeds)
```
quorum-sim simulate --n 7                       → forks 0   (perfect sync)
quorum-sim simulate --n 7 --drop-prob 0.1       → forks 0   (drops only)
quorum-sim simulate --n 7 --max-delay 3         → forks 2-4 ← BUG (delay only, zero faults)
```
All four proposer models exhibit it under `--max-delay > 0`.

## Root cause (likely)
The documented v1 simplification "accept→confirm collapsed into a single irrevocable accept-lock" (sim.rs) locks a value on a partial/premature quorum when messages are reordered/delayed, instead of requiring a *confirmed* quorum the way SCP's prepare/commit ballot phases do. Under delay, two correct nodes can each lock different values — which real SCP prevents. The single-phase lock loses the asynchronous safety property.

## Impact
- The simulator's **fork-rate / safety output under partial synchrony is NOT trustworthy** and must not be used to compare proposer-model safety (#427) until fixed.
- Liveness (stall) and fairness outputs appear unaffected and remain usable.
- The existing `known_safe_below_splitting_never_forks` test passes only because its parameters happen to avoid the delay-induced fork; it should be strengthened to assert **zero forks with zero faulty nodes across delay AND drop**, which currently FAILS under delay.

## Fix direction
Make the lock faithful to federated voting: a node may only irrevocably commit a value once it has observed a genuine *confirming* quorum (two-phase: accept-prepared then confirm-committed, or an equivalent that preserves the quorum-intersection safety invariant under arbitrary message reordering). Add a regression test: no faulty nodes ⇒ 0 forks for all proposer models across {sync, delay, drop, delay+drop} and many seeds.

Part of #514 / #427.

rjwalters:
## Summary

The dynamic SCP simulator forked (≥2 correct nodes committing different values — a SAFETY violation) with **ZERO faulty nodes**, purely under message delay/reordering. Federated voting / SCP is provably safe under full asynchrony (quorum intersection guarantees agreement regardless of timing), so this was a model-fidelity defect.

Root cause: the v1 "accept-lock" collapsed SCP's accept→confirm into a single irrevocable lock and committed on a **transient _vote_ quorum**. Under reordering, two correct nodes could each observe a different transient vote-quorum (built from stale, not-yet-retracted votes) and commit different values — exactly the asynchronous fork SCP's two-phase ballot prevents.

## Changes

- **`Message` now carries two federated-voting signals**: the sender's current `vote` and its irrevocably `accept`ed value (`Option`, `None` until locked). Byzantine equivocators also equivocate on the accept signal so the at/above-splitting-set attack is still realizable.
- **Accept (lock)** — `accepted[node]`, pinned: a node locks a value the first time it sees the accept condition: a **quorum of voters** (incl. self) OR a **v-blocking set of accepters**. A correct node accepts at most one value, ever. Added an `is_v_blocking` helper derived directly from the quorum-set definition.
- **Confirm (commit)** — the safety-critical fix: a node irrevocably commits ONLY once a **confirming quorum (incl. self) of _accepters_** exists for the value — not merely voters. Quorum intersection + one-value-per-honest-accepter makes a fork impossible under arbitrary reordering unless the Byzantine set reaches the splitting threshold.
- **Module docs rewritten** — the old "collapsed accept-lock" note was wrong; replaced with the accurate two-phase semantics and the #517 safety argument.
- **Regression tests** (unit `sim.rs::zero_faults_never_fork_under_delay` + integration `dynamic.rs::zero_faults_never_fork_under_any_network`): zero faulty nodes ⇒ 0 forks across {sync, delay-only, drop-only, delay+drop}, all four proposer models, 200 seeds, n ∈ {4,7,10}.

| Zero faulty nodes ⇒ 0 forks across {sync, delay, drop, delay+drop}, many seeds, n∈{4,7,10} | ✅ | New unit + integration tests pass; CLI `--max-delay 5` sweep at n=4/7/10/13 shows forks=0 stall=0 |
| Commit requires a genuinely *confirming* quorum (two-phase) | ✅ | Commit now gated on a quorum of `accept`ers, not voters |
| Safety holds under ARBITRARY reordering | ✅ | Quorum intersection: honest node accepts one value ⇒ no two confirming quorums for distinct values |
| Equivocation AT/above splitting set still forks | ✅ | `equivocation_crosses_static_splitting_threshold` passes; CLI n=4 with `--faulty 0 --faulty 2 --fault equivocate` → 200/200 forks |
| Below splitting set never forks | ✅ | `known_safe_below_splitting_never_forks` passes |
| Unanimity-below-4 stalls on crash | ✅ | `unanimity_below_four_stalls_on_crash` passes |
| Reproducibility (same seed → same outcome) | ✅ | `reproducibility_same_seed_same_outcome` passes |
| Liveness preserved | ✅ | No new stalls; mean rounds-to-decide rises ~2→~6 under delay (expected for two-phase) |

## Test Plan

- `cargo build -p bth-quorum-sim` — clean
- `cargo test -p bth-quorum-sim` — 33 unit + 5 correctness + 6 dynamic, all pass
- `cargo clippy -p bth-quorum-sim --all-targets` — clean
- `cargo +nightly-2025-12-03 fmt --check -p bth-quorum-sim` — clean
- Before fix: `simulate --n 7 --max-delay 3` → forks 2–4/200. After fix → 0/200 (delay and delay+drop).

Closes #517

### DIFF
--- a/consensus/quorum-sim/src/sim.rs
+++ b/consensus/quorum-sim/src/sim.rs
@@ -23,16 +23,30 @@
 //! consensus **dynamics** at the abstraction needed to observe agreement vs.
 //! divergence, and intentionally simplifies real SCP:
 //!
-//! - **Vote + accept-lock + commit, not the full SCP state machine.** Real SCP
-//!   nomination has `voted`/`accepted`/`confirmed` nominate stages and the
-//!   ballot protocol has `PREPARE`/`CONFIRM`/`EXTERNALIZE` with ballot counters
-//!   `b`, `p`, `c`, `h`. Here each correct node casts ONE current vote, then
-//!   **accepts (locks)** a value the first time it observes a quorum (per its
-//!   own quorum set) supporting it, and **commits** the locked value once a
-//!   quorum still supports it. The accept-lock collapses SCP's accept→confirm
-//!   into a single irrevocable lock; it is what produces the splitting-set
-//!   safety guarantee (two correct nodes can only lock different values if
-//!   their quorums share no *honest* node). It is NOT a faithful ballot FSM.
+//! - **Two-phase federated voting (vote → accept → confirm/commit), not the
+//!   full SCP state machine.** Real SCP nomination has `voted`/`accepted`/
+//!   `confirmed` nominate stages and the ballot protocol has `PREPARE`/
+//!   `CONFIRM`/`EXTERNALIZE` with ballot counters `b`, `p`, `c`, `h`. Here each
+//!   correct node casts ONE current vote, then runs the two federated-voting
+//!   steps that actually carry the safety proof. First, **accept (lock)**: a
+//!   node *accepts* and irrevocably locks a value the first time it sees the
+//!   federated-voting accept condition for it — either a **quorum** (incl.
+//!   self) currently *votes* for the value, or a **v-blocking set** of nodes
+//!   have already *accepted* it; the lock is pinned, so a correct node accepts
+//!   at most one value, ever. Second, **confirm (commit)**: a node *irrevocably
+//!   commits* a value only once it observes a **confirming quorum** (incl.
+//!   self) whose members have each *accepted* that value — not merely voted for
+//!   it. This second phase is the fix for the asynchronous-safety defect
+//!   (#517). The earlier v1 simplification collapsed accept→confirm into a
+//!   single lock and committed on a transient *vote* quorum, which under
+//!   message reordering let two correct nodes commit different values with ZERO
+//!   faulty nodes. Requiring a confirming quorum of *accepters* restores the
+//!   quorum-intersection guarantee: two quorums of accepters of different
+//!   values would have to share an honest node, but an honest node accepts only
+//!   one value — so under ARBITRARY delay/reordering, once any correct node
+//!   commits `v`, no correct node can commit `v' ≠ v` unless the Byzantine set
+//!   reaches the splitting threshold. It is still NOT a faithful ballot FSM,
+//!   but the safety invariant is exact.
 //! - **One slot, one leader per simulation.** We simulate agreement on a single
 //!   slot's value, repeated across many seeds — not a growing chain, and (for
 //!   leader models) with ONE leader for the whole slot rather than per-round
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge verdict: APPROVE

Rigorous review of this SAFETY-CRITICAL fidelity fix (informs the #427 consensus decision). The two-phase federated-voting model is now correct, and I verified the safety claims empirically rather than taking the PR body at its word.

### Two-phase commit logic — does it actually guarantee asynchronous safety up to the splitting threshold? YES.
- **Accept (lock)** fires on a quorum of *voters* (incl. self) OR a *v-blocking set* of *accepters* — the textbook federated-voting accept condition. `accepted[node]` is set once and pinned, so an honest node accepts **at most one value, ever**.
- **Commit (confirm)** now requires a *confirming quorum of accepters* (incl. self), not a transient vote quorum. Safety argument holds: if two honest nodes committed `v ≠ v'`, their two accepter-quorums must intersect (quorum intersection, which holds below the splitting threshold), and the shared honest node would have had to accept both values — impossible. Hence no fork under arbitrary reordering unless the Byzantine set reaches the splitting threshold. This is exactly the property the old single-phase lock lost.
- `is_v_blocking` is consistent with the FBAS definitions in `model.rs`/`analysis.rs`: `set` is v-blocking for `node` iff `all \ set` does not satisfy `node`'s quorum set. The `node >= fbas.len()` guard correctly handles out-of-range/equivocator indices.

### Regression test is non-vacuous and decisive — verified by running both trees:
- OLD code (`main`), n=7 `--max-delay 3`, **zero faulty nodes**: forks **2–4 / 200** across all four proposer models (reproduces #517).
- NEW code (PR #518), identical config: **0 forks, 0 stalls**, all 200 agree. So `zero_faults_never_fork_under_any_network` (and the unit twin) would FAIL on old code and PASS now.

### Did NOT over-fix — forks still detected at the splitting threshold:
- Byzantine equivocators now also equivocate their *accept* signal (per-recipient), so the at/above-splitting-set partition attack remains realizable through the two-phase commit.
- `equivocation_crosses_static_splitting_threshold` passes: 2 equivocators (incl. leader) at n=4 (splitting set = 2) still produce forks > 0. Below threshold (1 equivocator) → 0 forks. Safety is conditional, not cheated.

### Liveness / quality:
- No new stalls; mean rounds-to-decide ~3.8 → ~6.3 under delay — the expected two-phase cost, acceptable.
- `cargo test -p bth-quorum-sim`: 33 unit + 5 correctness + 6 dynamic, all pass. `cargo clippy --all-targets`: clean. Module docs rewritten accurately. Reuses the static model.
- CI: rustfmt check PASS; mergeStateStatus CLEAN / MERGEABLE.

🤖 Reviewed by the Loom Judge

PR #520, opened 2026-06-22
feat(quorum-sim): model leader-timeout / view-change to validate round-robin proposer design

### LINKED ISSUE #519: feat(quorum-sim): model leader-timeout / view-change to validate round-robin proposer design
Follow-up to #516/#518. The dynamic sim (now safety-faithful) shows leader proposer models stall ~15% under an equivocating leader ONLY because v1 has no leader-timeout / view-change recovery. The recommended production design (#427) is round-robin leader election WITH view-change, so the sim must model view-change to validate that the recommended design has both good fairness AND no liveness gap.

## Scope
- Add a leader-timeout / view-change mechanism: if the current leader does not drive a decision within a round budget, rotate to the next leader (round-robin) and retry the slot.
- Acceptance: under an equivocating/crashed leader, round-robin + view-change drops the stall rate to ~0 while safety (0 forks below splitting set) is preserved.
- Keep reproducibility + the zero-fault-zero-fork invariant.

## Why
Last empirical gap before the production proposer-model change (competing-coinbase to round-robin+view-change) can be specced with confidence. Validate-first / no-hard-forks.

Part of #427. Follow-up to #514/#516/#518.

rjwalters:
Closes #519

## What

Adds an optional **leader-timeout / view-change** mechanism to the dynamic SCP simulator (`bth-quorum-sim`) so the ratified production proposer design from #427 — **round-robin leader election WITH mandatory leader-timeout / view-change** — can be validated empirically before it is specced as a consensus-breaking change.

## How view-change is modeled

If the current view's leader has not driven the slot to a decision within a per-view round budget, the **view** advances and the leader is **rotated round-robin** to `(base_leader + view) % n`; the slot is retried under the new leader. Undecided, not-yet-locked nodes follow the new leader's value, so a stalled (Byzantine or crashed) leader is rotated out and liveness is recovered. Each view consumes >=1 round, so it always terminates within `--max-rounds` (otherwise: a stall).

- New field `SimConfig::view_change: Option<u32>` — `None` = v1 behavior (no rotation), `Some(budget)` = enabled.
- A `view_budget` of 0 is clamped to 1.
- No effect on the leaderless `competing-coinbase` model.

### CLI

```
# WITHOUT view-change (v1 baseline)
quorum-sim simulate --n 7 --proposer round-robin-leader --faulty 0 --fault equivocate --seeds 200

# WITH view-change
quorum-sim simulate --n 7 --proposer round-robin-leader --faulty 0 --fault equivocate --seeds 200 \
    --view-change --view-budget 4
```

A new `vc` column (`off` / `v<budget>`) reports the setting.

## Safety preserved exactly

View-change only changes *which leader an as-yet-undecided node follows*; it never unwinds an `accept`-lock or a commit. A correct node's vote stays pinned to its accepted value across all views, so the quorum-intersection argument from #517/#518 is unchanged. Two correct nodes still cannot commit different values below the splitting threshold, and the at/above-splitting-set fork is **still observed** (a dedicated test asserts view-change does not mask it). The zero-fault / zero-fork invariant holds with view-change on across {sync, delay, drop, delay+drop}.

## Headline empirical result

Round-robin leader, **equivocating leader**, 200 seeds:

| n | forks | stalls WITHOUT vc | stalls WITH vc |
|---|-------|-------------------|----------------|
| 4 | 0 | ~15% | **0%** |
| 7 | 0 | 29/200 (14.5%) | **0/200 (0%)** |
| 10 | 0 | ~15% | **0%** |

Crashed leader: 0 stalls with and without view-change. This is the empirical validation that the ratified design closes the only liveness gap (Byzantine-leader stall) without weakening safety.

## Tests (all green)

`cargo test -p bth-quorum-sim`: 35 unit + 5 correctness + 12 dynamic.

New acceptance tests in `tests/dynamic.rs`:
- `view_change_eliminates_equivocating_leader_stalls` (headline: <2% over 200 seeds, n in {4,7,10}, asserts the >0 baseline too)
- `view_change_recovers_equivocating_leader_under_delay` (partial synchrony)
- `view_change_preserves_zero_fault_zero_fork` (safety across all networks)
- `view_change_does_not_mask_splitting_set_fork` (not over-fixed)
- `view_change_is_reproducible`
- `view_change_recovers_crashed_leader`

Plus inline `sim.rs` tests for the rotation mechanic and the `view_budget == 0` clamp.

`cargo fmt --check`, `cargo build`, `cargo clippy --all-targets`: all clean (pinned nightly).

Docs updated: crate README (view-change section + comparison commands + result table), module docs in `sim.rs`, and `lib.rs` crate docs.

### DIFF
--- a/consensus/quorum-sim/README.md
+++ b/consensus/quorum-sim/README.md
@@ -48,22 +48,74 @@ rounds-to-decide, and leadership fairness.
   values to different peers — the fork-inducing adversary).
 - **Network**: synchronous (default) or partially-synchronous via `--max-delay`
   / `--drop-prob` (both seeded ⇒ reproducible).
+- **View-change / leader-timeout** (#519): `--view-change` enables SCP-style
+  leader rotation; `--view-budget N` sets the per-view round budget (default 4).
+  Omit `--view-change` to run the v1 behavior (no rotation) for comparison.
 - **Quorum oracle**: commit decisions delegate to the static `Fbas::is_quorum`,
   so dynamic outcomes cross-check the static splitting/blocking-set predictions.
 
 Key empirical guarantees (encoded as tests in `tests/dynamic.rs`): with
 equivocators **below** the static minimal splitting set, no run ever forks; at
-the splitting set, a leader-equivocation fork **is** observed; unanimity below 4
-nodes stalls on a single crash; and a given `(config, seed)` is bit-for-bit
-reproducible.
+the splitting set, a leader-equivocation fork **is** observed (even with
+view-change on — it is not masked); unanimity below 4 nodes stalls on a single
+crash; and a given `(config, seed)` is bit-for-bit reproducible (with and
+without view-change).
+
+### View-change / leader-timeout (#519)
+
+The ratified production proposer design (#427) is **round-robin leader election
+WITH mandatory leader-timeout / view-change**. The v1 engine had no view-change,
+so a *Byzantine (equivocating) leader* could stall its own slot — leader models
+stalled ~15% (29/200) under an equivocating leader. View-change closes exactly
+this gap:
+
+> If the current view's leader has not driven the slot to a decision within
+> `--view-budget` rounds, the **view** advances and the leader is **rotated
+> round-robin** to `(base_leader + view) % n`; the slot is retried under the new
+> leader. Undecided, not-yet-locked nodes follow the new leader's value, so a
+> stalled (Byzantine or crashed) leader is rotated out and liveness is restored.
+
+**Safety is preserved exactly.** View-change only changes which leader an
+as-yet-undecided node follows; it never unwinds an `accept`-lock or a commit. A
+correct node's vote is pinned to its accepted value forever, so once a node
+accepts (let alone commits) it keeps that value across every view. Two correct
+nodes therefore still cannot commit different values unless the Byzantine set
+reaches the splitting threshold — the same quorum-intersection argument as
+without view-change. The fork at/above the splitting set is **still observed**
+with view-change on (it is not papered over). View-change is bounded by
+`--max-rounds` (each view consumes ≥1 round), so the simulation still terminates.
+
+**Empirical result** (n ∈ {4,7,10}, equivocating leader, 200 seeds):
+
+| config                          | forks | stalls         |
+|---------------------------------|-------|----------------|
+| round-robin, **no** view-change | 0     | ~29/200 (~15%) |
+| round-robin, **with** view-change | 0   | **0/200 (0%)** |
+
+Run the comparison yourself:
+
+```
+# WITHOUT view-change — Byzantine leader stalls ~15% of slots
+cargo run -p bth-quorum-sim --bin quorum-sim -- simulate \
+    --n 7 --proposer round-robin-leader --faulty 0 --fault equivocate --seeds 200
+
+# WITH view-change — stalls collapse to ~0, still 0 forks
+cargo run -p bth-quorum-sim --bin quorum-sim -- simulate \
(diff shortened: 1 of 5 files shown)


rjwalters:
✅ **Approved!**

Rigorous review of the view-change / leader-timeout model. The safety-preservation claim — the load-bearing requirement for validating the ratified round-robin+view-change proposer design (#427) — holds at the code level, not just in prose.

## Safety: view-change provably does NOT unwind a lock or commit

The critical invariant is verifiable directly in `consensus/quorum-sim/src/sim.rs`:

- **Locked vote is pinned, leader-independent** (`run_tracked`, vote-derivation): `if let Some(locked) = accepted[node] { locked }` returns the accepted value immediately. The view-change-rotated `leader` is consulted **only in the `else` branch** (`accepted[node].is_none()`). So a node that has accepted can never be pulled onto a different value by a rotated leader.
- **Accept is irrevocable**: the lock step is gated by `if accepted[node].is_none()`, set once, never moved — each correct node accepts at most one value.
- **Commit requires a confirming quorum of *accepters* of the same value** (not voters). With each correct node accepting ≤1 value, two confirming quorums for distinct values must share an honest accepter — impossible below the splitting set. View-change touches neither `accepted` nor `committed`, so the quorum-intersection argument from #517/#518 is unchanged.
- The view-change step itself only mutates `view` / `leader` / `view_deadline`; it never reads or writes lock/commit state. Each view consumes ≥1 round (`view_budget.max(1)`), so it terminates within `max_rounds`.

Conclusion: view-change cannot cause two correct nodes to commit different values. Safety is preserved across views by construction.

## Empirical claims independently reproduced

Ran the CLI comparison and the full suite locally (worktree, commit 079c1bf):

| config | forks | stalls |
|---|---|---|
| round-robin, equiv leader, **no vc**, n=7, 200 seeds | 0 | **29/200 (14.5%)** |
| round-robin, equiv leader, **vc v4**, n=7, 200 seeds | 0 | **0/200 (0%)** |
| 2 equivocators (= splitting set), **vc on**, n=4, 200 seeds | **200/200** | 0 |

- Headline stall-collapse (~15% → 0%) is real and non-vacuous — agreement goes 171→200.
- **Not over-fixed**: at the splitting set the fork is still 200/200 with view-change on. The dedicated `view_change_does_not_mask_splitting_set_fork` test asserts this.
- Zero-fault → 0 forks holds with view-change ON across {sync, delay, drop, delay+drop} (`view_change_preserves_zero_fault_zero_fork`).
- Reproducibility preserved (`view_change_is_reproducible`), crashed-leader recovery covered, `view_budget==0` clamp tested.

## Test Execution

- `cargo test -p bth-quorum-sim` — ✅ 35 unit + 5 correctness + 12 dynamic, all pass
- `cargo clippy --all-targets -p bth-quorum-sim` — ✅ clean
- CLI comparison (above) — ✅ matches PR's reported figures exactly

## Other checks

- `Closes #519` present; merge state CLEAN; CI (rustfmt pinned nightly) passes.
- CLI flags `--view-change` / `--view-budget` fit the existing clap structure; opt-in mapping (`if view_change { Some(view_budget) } else { None }`) is clean.
- `competing-coinbase` unaffected: `base_leader` is `None` for the leaderless model, so `leader_at_view` stays `None` and view-change is a no-op.
- `vc` column added to the report table cleanly.

Minor (non-blocking): a couple of inline comments in the view-change-OFF vote-fallback branch still carry the old "cannot recover from a Byzantine leader" framing — accurate for the vc-off path, and the module/crate docs were updated correctly, so no change needed.

Excellent, careful work. This is solid empirical validation that the ratified design closes the Byzantine-leader liveness gap without weakening safety.

PR #540, opened 2026-06-23
feat(node): detect stalled miner and surface stuck-miner health flag

rjwalters:
## Summary

Adds a stuck-miner detector so a miner that reports `active==true` while producing **0 H/s** self-detects and alarms loudly instead of silently halting the chain. This is the code-side early-warning for the live-testnet halt diagnosed in #537/#539 (the faucet miner ran "active" at 0 H/s for ~50h, silently wedging the chain). Detection + alarm + flag only — no auto-restart (operator policy).

## Changes

- **`botho/src/node/minter.rs`**
  - New thresholds: `STUCK_MINER_SECS = 90`, `STARTUP_GRACE_SECS = 60` (grace so a miner building its ~2 GB RandomX dataset isn't falsely flagged).
  - Pure, testable `evaluate_stall(active, uptime, secs_since_progress, grace, stuck)` verdict.
  - New cloneable `MinterHealth` handle: shares the worker threads' live hash counter, tracks last-progress timestamp in atomics, produces a `MinterHealthSnapshot` (active / totalHashes / hashrate / uptime / stalled), and `check_and_warn()` emits a prominent `warn!` ("RandomX miner stalled: active but 0 H/s for Ns — chain will halt").
  - `Minter` now owns a `MinterHealth`, marks it active on `start()` / inactive on `stop()`, and exposes `health()`.
- **`botho/src/node/mod.rs`**: stable `Arc<RwLock<Option<MinterHealth>>>` published on each `start_minting` and flipped inactive on `stop_minting` (survives start/stop cycles); exposed via `Node::minter_health()`.
- **`botho/src/rpc/mod.rs`**: `RpcState::with_minter_health(...)`; `minting_getStatus` now reports `stalled` (and replaces the hardcoded `hashrate`/`totalHashes` TODOs with live values); `node_getStatus` now reports `minerStalled`. Both default to `false` when no handle is wired (relay nodes / pre-mint).
- **`botho/src/commands/run.rs`**: runs the stall check on the periodic status tick (emits the early-warning log) and wires the health handle into `RpcState`.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| active + 0 hashrate past grace window → flag true + warning path | ✅ | `test_stall_active_zero_hashrate_past_window`; `check_and_warn` emits `warn!` when `snapshot().stalled` |
| active + healthy hashrate → flag false | ✅ | `test_stall_active_healthy_hashrate_not_flagged`, RPC test case (3) |
[...]

Issue #538, opened 2026-06-23
feat(node): stuck-miner detector — warn + health flag when minting is active but hashrate is 0


rjwalters:
Code-side fix split out of #537. The live testnet sat **halted for ~50h** because the faucet miner reported `active:true` while doing **0 hashes (hashrate 0.0)** — and nothing surfaced it. A miner that is active but not hashing should self-detect and alarm loudly instead of silently halting the chain.

## Scope
- In the minter (`botho/src/node/minter.rs` — it already tracks `hashrate`/`totalHashes`/`uptimeSeconds`), detect the condition **`active == true` AND `hashrate == 0` (or totalHashes not advancing) for more than N seconds** (config/const, e.g. 60–120s after a startup grace) and:
  - Emit a prominent `warn!`/`error!` log (RandomX miner stalled — 0 H/s for Ns).
  - Expose a health flag in `minting_getStatus` (e.g. `stalled: bool` / `healthy: bool`) AND surface it in `node_getStatus` (`rpc/mod.rs`) so dashboards/monitoring catch it.
- Do NOT auto-restart the miner here (that's operator policy) — detection + alarm + flag only.

## Acceptance
- Unit test: active + 0 hashrate past the grace window → flag set true + warning path; active + healthy hashrate → flag false; inactive → not flagged.
- Flag present in both `minting_getStatus` and `node_getStatus` payloads.
- Startup grace so a freshly-starting miner isn't flagged during VM/dataset init.

Grounding: `botho/src/node/minter.rs` (miner status tracking), `botho/src/rpc/mod.rs` (`minting_getStatus`, `node_getStatus`).

Part of #537.

rjwalters:
Operator counterpart (diagnose/fix the wedged faucet miner) is #539. This issue (#538) is the code-side early-warning so a future stall surfaces immediately instead of after ~50h.

rjwalters:
Starting work on this issue per user request (code-side early-warning for the live-testnet halt diagnosed in #537/#539).

PR #546, opened 2026-06-23
feat(rpc): report real blocksFound in minting_getStatus

### LINKED ISSUE #543: fix(rpc): minting_getStatus blocksFound hardcoded to 0 — track blocks won by this node
From the 2026-06-22 hardcoded-observability audit. After #540 wired live hashrate/totalHashes, `minting_getStatus` still hardcodes `blocksFound: 0` (rpc/mod.rs:1890, TODO 'track blocks found'). A miner operator can't tell whether their node is actually **winning blocks** (vs just hashing).

## Fix (wireable)
Increment a per-node counter when THIS node's coinbase is the one externalized in a block (the combiner picks the highest-PoW coinbase per slot — botho/src/consensus/service.rs combine_fn / check_externalized). Surface the count in minting_getStatus (and consider node_getStatus). Add a test.

Note: pairs with #540's stuck-miner detector — blocksFound advancing is the positive signal that complements 'hashrate > 0 but 0 blocks = something wrong downstream'.

Found by: hardcoded-observability audit.

rjwalters:
## Summary
`minting_getStatus` hardcoded `blocksFound: 0` (rpc/mod.rs, TODO "track blocks found"). After #540 wired live hashrate/totalHashes through a shared `MinterHealth` handle, a miner operator still couldn't tell whether their node was actually *winning* blocks (vs just hashing). This wires a real per-node "blocks won" counter using that same Arc-shared-atomic pattern.

## Changes
- **`node/minter.rs`**: Added an `Arc<AtomicU64>` `blocks_found` counter plus the minter's captured view/spend public keys to `MinterHealth`; surfaced `blocks_found` in `MinterHealthSnapshot`. New methods: `owns_coinbase(view, spend)` (matches a coinbase's minter keys against this node's address; an unbound/all-zero handle never claims ownership), `increment_blocks_found()`, `blocks_found()`. `Minter::new` now captures the reward address keys into the handle.
- **`commands/run.rs`**: Added `count_won_block()` helper and call it at every site that *successfully* appends a block (local SCP externalize, gossip `NewBlock`, block-sync, compact-block + BlockTxn reconstruction). It increments only when `owns_coinbase` matches — so a won block is counted exactly once on finalize, never on propose, never for a peer's coinbase.
- **`rpc/mod.rs`**: `minting_getStatus.blocksFound` now reads from the shared snapshot (0 when no handle is wired in — relay nodes / tests / before minting).

### Where/how it increments
The winning coinbase (`MintingTx`) carries `minter_view_key` / `minter_spend_key`. The shared `MinterHealth` handle captures this node's address keys at minter construction. On each successful `add_block_from_network`, `count_won_block` compares the block's coinbase keys to the handle's keys via `owns_coinbase`; on a match it calls `increment_blocks_found`. Increment happens on **externalize/finalize** (block added to ledger), gated to **this node's** coinbase only.

| Increment when this node's coinbase externalizes | ✅ | `count_won_block` at every append site, gated by `owns_coinbase`; unit test `test_minter_health_blocks_found_counter` |
| Does NOT increment for another node's coinbase | ✅ | `owns_coinbase` requires both keys match; unit test `test_minter_health_owns_coinbase` (rejects other keys + unbound handle) |
| Reflected in `minting_getStatus` | ✅ | RPC test `test_blocks_found_in_minting_status` |
| 0 when nothing mined / no handle wired | ✅ | RPC test cases (1) no handle => 0, (2) fresh handle => 0 |

## Test Plan
- `cargo build -p botho` (lib + bin) — clean
- `cargo test -p botho --lib minter::` — new tests pass
- `cargo test -p botho --lib -- rpc::tests::test_blocks_found_in_minting_status rpc::tests::test_miner_stalled_flag_in_both_status_payloads` — pass
- `cargo clippy -p botho --lib` — no new warnings; nightly `cargo fmt` applied

Closes #543

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -672,6 +672,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             if let Err(e) = node.add_block_from_network(&block) {
                                 warn!("Failed to add network block: {}", e);
                             } else {
+                                // Count toward blocksFound if this node won the
+                                // slot but the block came back via gossip before
+                                // our own externalize landed it (#543).
+                                count_won_block(&minter_health, &block);
+
                                 // Record block processing time
                                 metrics_updater.observe_block_processing(block_start.elapsed().as_secs_f64());
                                 metrics_updater.inc_blocks_processed();
@@ -960,6 +965,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                                 break;
                                             }
                                             applied_any = true;
+                                            // Count toward blocksFound if a synced
+                                            // block carries this node's coinbase
+                                            // (e.g. re-syncing our own past blocks
+                                            // after a restart) (#543).
+                                            count_won_block(&minter_health, block);
                                             // Record for dynamic timing
                                             consensus.record_block(block.header.timestamp, block.transactions.len());
                                         }
@@ -1090,6 +1100,10 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                         // machine will backfill the gap.
                                         warn!("Failed to add reconstructed block: {} (catch-up handles non-contiguous gaps)", e);
                                     } else {
+                                        // Count toward blocksFound if this node's
+                                        // coinbase won (#543).
+                                        count_won_block(&minter_health, &block);
+
                                         // Record for dynamic timing
                                         consensus.record_block(block.header.timestamp, block.transactions.len());
 
@@ -1225,6 +1239,10 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                         if let Err(e) = node.add_block_from_network(&block) {
                                             warn!("Failed to add completed block: {}", e);
                                         } else {
+                                            // Count toward blocksFound if this
+                                            // node's coinbase won (#543).
+                                            count_won_block(&minter_health, &block);
+
                                             // Record for dynamic timing
                                             consensus.record_block(block.header.timestamp, block.transactions.len());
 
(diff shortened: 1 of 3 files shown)


rjwalters:
## Loom Judge Review — APPROVED

Reviewed against #543. Verified locally on `a3bb56c` (build + tests + clippy) and confirmed CI is green / mergeStateStatus CLEAN.

### Exactly-once counting (the key risk) — VERIFIED
`count_won_block` is invoked at all five block-append sites (local SCP externalize, gossip `NewBlock`, block-sync apply, compact-block reconstruction, BlockTxn completion). In every case the call sits in the **success branch** of `node.add_block_from_network(...)`:
- `add_block_from_network` -> `Ledger::add_block` rejects any block whose height is not exactly `tip + 1` (store.rs:408-415) and any prev-hash mismatch (418-422). A duplicate/already-applied block therefore returns `Err`, so `count_won_block` is never reached for it.
- The sync loop only counts after `applied_any = true` (run.rs:962); an `Err` `break`s out before counting.
- The local externalize path's duplicate-height case is the `if let Err(e)` branch (run.rs:1345) which logs benign-skip and does NOT count.

Net: a single won block is appended exactly once (whichever path wins the race), and is therefore counted exactly once. No cross-path double counting.

### This-node-only — VERIFIED
`owns_coinbase(view, spend)` requires **both** the coinbase's `minter_view_key` and `minter_spend_key` to equal the keys captured at `Minter::new` from the reward address. A peer-won block's coinbase carries the peer's keys -> no match -> not counted. The all-zero/unbound guard returns `false` up front, so a relay node (legacy `MinterHealth::new`, zero keys) never claims even an all-zero coinbase. Unit test covers owned / other-keys / wrong-spend-only / unbound-with-zero-coinbase.

### Externalize-not-propose — VERIFIED
Counting happens only on successful ledger append (finalize), never on propose/mint. Confirmed at each site.

### Restart semantics — ACCEPTABLE
The counter is an in-memory `Arc<AtomicU64>`, resets to 0 on restart; re-syncing our own past blocks re-counts them (restoring the session tally, not double-counting within a process). This is a reasonable "blocksFound" session signal that complements the stuck-miner detector (#538). A persistent lifetime tally would be a separate enhancement — fine to defer.

### Tests — non-vacuous, pass
- `test_minter_health_blocks_found_counter` (0 -> increments -> snapshot reflects)
- `test_minter_health_owns_coinbase` (owned / rejects-other / rejects-wrong-spend / unbound-never-claims)
- `test_blocks_found_in_minting_status` (no handle => 0, fresh handle => 0, after 2 increments => 2)

Local: build clean, `minter::` + RPC tests pass, no new clippy warnings, rustfmt clean. CI: all required checks pass.

Approving and promoting to `loom:pr`.

PR #547, opened 2026-06-23
fix(rpc): network_getPeers surfaces real connected peers (#544)

### LINKED ISSUE #544: fix(rpc): network_getPeers returns empty list stub — surface real connected peers
From the 2026-06-22 hardcoded-observability audit (LOW). `network_getPeers` always returns `"peers": []` (rpc/mod.rs:1921, inline comment 'Return empty for now - would need to get actual peer addresses'). Documented stub, so low risk, but it's an observability dead-end: thin clients can't enumerate peers via RPC (only via DNS seed).

## Fix
Surface the live connected-peer set (ids/addresses) from the network/discovery layer in network_getPeers. Tests. Low priority.

Found by: hardcoded-observability audit.

rjwalters:
Closes #544

## Problem
`network_getPeers` returned a hardcoded `"peers": []` ("Return empty for now — would need to get actual peer addresses"), an observability dead-end: thin clients could not enumerate peers via RPC (only via DNS seed).

## Fix
Surface the live connected-peer set using the **same mechanism `peerCount` already uses**. The discovery peer table is owned by the single-threaded network event loop in `commands::run`, so — exactly as with `peer_count` — the loop publishes a cheap clone of the table into a shared `Arc<RwLock<Vec<PeerInfoSnapshot>>>` on each peer connect/disconnect (and seeds it at startup). The RPC handler reads that snapshot. Read-only, no hot-path locking beyond the existing pattern.

### Peer source
`NetworkDiscovery::peer_table() -> Vec<PeerTableEntry>` — the same struct backing `peer_count()`.

### Fields returned (per peer)
- `peerId` — libp2p peer ID
- `address` — last observed multiaddr (`null` if none)
- `protocolVersion` — advertised version (`null` if not yet identified)
- `versionWarning` — below-minimum-version flag
- `lastSeenSecs` — seconds since last seen (coarse liveness hint)

Plus a top-level `peerCount`. Empty list when no peers are connected.

`direction` (inbound/outbound) and `height` are **not** included — `PeerTableEntry` does not track them today, and adding them would require new network plumbing beyond this issue's scope.

## Tests
`test_get_peers_surfaces_connected_peers`: empty-when-none; populated two-peer case asserting the full field shape (including `null` rendering of absent optionals and the warning flag).

## Verification
- `cargo fmt` (pinned nightly-2025-12-03), `cargo build -p botho`, `cargo clippy -p botho --lib` — clean for changed code (remaining warnings are pre-existing, e.g. unused `libp2p::PeerId` in `network/privacy/broadcaster.rs`).
- `cargo test -p botho --lib rpc::` — 83 passed.

## Files
- `botho/src/rpc/mod.rs` — `PeerInfoSnapshot` type, `RpcState.peers` handle + `with_peers`, rewritten `handle_get_peers`, test.
- `botho/src/commands/run.rs` — `build_peer_snapshot` helper, snapshot handle wired into `RpcState`, refreshed on peer connect/disconnect.

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 535
-# Branch: feature/issue-535
+# Issue: 544
+# Branch: feature/issue-544
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -36,7 +36,8 @@ use crate::{
     node::{MintedMintingTx, Node, SharedLedger},
     rpc::{
         calculate_dir_size, init_metrics, start_metrics_server, start_rpc_server, FaucetState,
-        MetricsUpdater, NodeIdentity, RpcState, WsBroadcaster, DATA_DIR_USAGE_BYTES,
+        MetricsUpdater, NodeIdentity, PeerInfoSnapshot, RpcState, WsBroadcaster,
+        DATA_DIR_USAGE_BYTES,
     },
     transaction::Transaction,
     wallet::Wallet,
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review: APPROVED

LOW-risk observability fix, reviewed pragmatically. Verdict: approve.

### Peer source is real and snapshot-refreshed (confirmed)
- `build_peer_snapshot()` reads `NetworkDiscovery::peer_table()`, which returns `self.peers.values().cloned()` — the exact same `self.peers` map that backs the working `peer_count()` (`self.peers.len()`). Not a stub.
- Snapshot is seeded at startup (`Arc::new(RwLock::new(build_peer_snapshot(&discovery)))`) and refreshed in the network event loop on **both** peer connect and disconnect, adjacent to the existing `set_peer_count` / `ws_broadcaster.peer_connected|disconnected` calls. So it tracks peer churn.
- RPC path is read-only (`read_lock!(state.peers, ...)`), no new hot-path locking — same `Arc<RwLock<..>>` pattern already used for `peer_count`/`sync_status`.

### Consistency
- `network_getPeers.peerCount` is `peers_snapshot.len()`; `node_getStatus`/`network_getInfo` use the separate `state.peer_count` usize. Both derive from the same `discovery.peers` map and are updated in the same handler blocks, so they stay consistent. The two parallel handles are a minor cosmetic redundancy, not a divergence — not a blocker.

### Fields & scope
- `peerId/address/protocolVersion/versionWarning/lastSeenSecs` + top-level `peerCount`; optionals render as JSON `null`; empty list when no peers. Field shape matches `PeerTableEntry`.
- `direction`/`height` deferred — not tracked in `PeerTableEntry` today; noted as out-of-scope follow-on. Acceptable.

### Test
- `test_get_peers_surfaces_connected_peers` is non-vacuous: asserts empty/zero when none, and the full two-peer field shape (including null optionals + warning flag) when populated.

### CI
- `gh pr checks`: all pass (skipped fuzz jobs are path-filtered intentional skips). Merge state CLEAN.
- Local: `cargo test -p botho --lib rpc::` → 83 passed. `cargo clippy -p botho --lib` warnings are all pre-existing and on unrelated lines (none on the added code; new struct/fields are fully documented).

Nice clean reuse of the established snapshot pattern.

PR #548, opened 2026-06-23
feat(network): real bytesSent/bytesReceived + inbound/outbound counters in network_getInfo

### LINKED ISSUE #542: fix(rpc): network_getInfo bytesSent/bytesReceived + inboundCount hardcoded to 0 (need real transport accounting)
From the 2026-06-22 hardcoded-observability audit. `network_getInfo` reports fake traffic/connection counters:
- `bytesSent: 0`, `bytesReceived: 0` (rpc/mod.rs:1909-1910) — always 0 even while actively syncing/relaying.
- `inboundCount: 0` (mod.rs:1907) — always 0; only `outboundCount` is real.

An operator checking 'is data flowing / are we accepting inbound?' sees zeros and may wrongly conclude the network layer is dead.

## Status of the data source
Partial only: a `TrafficPattern` struct (network/transport/fingerprint.rs:55-58) tracks bytes_sent/bytes_received but exists for fingerprinting-resistance testing, not as live node metrics. **Real per-connection byte + inbound/outbound accounting needs to be added in the transport/connection layer**, then surfaced. (This is why it's filed rather than a quick wire-up.)

## Scope
- Add cumulative byte counters (sent/received) and an inbound/outbound connection count in the transport layer; surface them in network_getInfo. Tests.

Found by: hardcoded-observability audit. (Covers audit findings 'bytes' [HIGH] + 'inboundCount' [MED].)

rjwalters:
## Summary

`network_getInfo` previously reported hardcoded `inboundCount: 0`, `bytesSent: 0`, and `bytesReceived: 0` (rpc/mod.rs), so an operator checking "is data flowing / are we accepting inbound?" saw zeros even while the node was actively syncing and relaying. `outboundCount` was the only real field (it equalled `peerCount`).

This adds real, live, lock-free traffic and connection-direction accounting in the network layer and surfaces it via the established shared-`Arc` snapshot pattern (the same one used for `peer_count`, `minter_health`, `sync_status`, and the `network_getPeers` peer snapshot).

## Changes

- **`NetworkStats`** (new, `network/discovery.rs`): four `AtomicU64` counters — `bytes_sent`, `bytes_received`, `inbound_count`, `outbound_count`. Lock-free on the hot path; the closed-connection decrement saturates at 0 (CAS loop) so a spurious close can never underflow.
- **Owned by `NetworkDiscovery`**, exposed via `stats()` (cheap `Arc` clone for the RPC layer) and `stats_ref()` (borrow for the static `broadcast_*` call sites in `commands/run.rs`).
- **Sent bytes** are recorded at every gossipsub send site — `broadcast_block`, `broadcast_transaction`, `broadcast_minting_tx`, `broadcast_scp`, `broadcast_compact_block`, `request_block_txns`, `respond_block_txns`, `broadcast_pex`, `broadcast_block_smart`, `broadcast_upgrade_announcement` — using the serialized payload length, recorded only after a successful `publish`.
- **Received bytes** are recorded in `process_event` at the gossipsub `Message` arm (`message.data.len()`), before any rate-limit drop (the bytes already crossed the wire).
- **Inbound vs outbound** is derived from the libp2p connection endpoint: `ConnectionEstablished` with `endpoint.is_dialer()` => outbound, else inbound; counted once per peer (`num_established == 1`) and decremented when the last connection to that peer closes (`num_established == 0`). This mirrors the existing per-peer `peer_count` accounting.
- **RPC**: `RpcState` gains an optional `network_stats` handle + `with_network_stats(..)` builder; `commands/run.rs` wires in `discovery.stats()`. `handle_network_info` surfaces the real values, falling back to the prior placeholder shape (`inboundCount: 0`, `outboundCount: peerCount`, zero bytes) when no handle is present (tests / relay nodes).

| Byte counters increment when data is sent/received | done | `test_network_stats_byte_counters_accumulate`; sent recorded in each `broadcast_*` after publish, received in `process_event` |
| inboundCount reflects accepted connections; outboundCount dialed | done | `test_connected_point_direction_maps_to_counter` (dialer->outbound, listener->inbound) + `test_network_stats_inbound_outbound_independent` |
| network_getInfo surfaces real values (0 only when genuinely nothing sent / no inbound) | done | `test_network_info_surfaces_real_counters` covers fallback, real values, and genuine-zero cases |
| Lock-free / no per-packet lock | done | `AtomicU64` with `Relaxed` ops; RPC reads a cheap `Arc` snapshot |

## Honest scope / known gaps

These counters track **application-layer gossipsub payload** bytes — the bulk of node traffic (blocks, transactions, SCP consensus, compact blocks, minting txs, PEX, upgrade announcements). Documented on `NetworkStats`:

- **Sync request/response (`request_response`) payload bytes are not counted.** The codec exchanges typed values rather than exposing a serialized byte length at the event boundary, and re-serializing solely to measure would add hot-path cost. Block/tx/SCP/compact-block traffic flows over gossipsub and is counted. A follow-on could instrument the `SyncCodec` read/write to add request/response byte accounting.
- **Transport framing overhead (Noise handshake, yamux framing, TCP headers) is not counted.** These are payload counters, not raw wire counters. libp2p 0.56's only raw-wire hook is `with_bandwidth_metrics`, which writes to a prometheus registry rather than exposing cheap readable atomics; wrapping every yamux stream would be substantially more invasive. A follow-on could add a wire-level counting transport if exact byte-on-wire totals are needed.

## Test Plan

- `cargo +nightly-2025-12-03 fmt` — clean
- `cargo build -p botho` — passes (warnings are pre-existing dead-code in transport/webrtc)
- `cargo test -p botho --lib` — 1030 passed, 0 failed
- `cargo clippy -p botho --lib` — no new warnings introduced by this change (existing warnings are pre-existing: unused `FeeConfig`/`TransactionType` import, `event_tx` never read, etc.)

Closes #542

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -439,7 +439,10 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     .with_identity(node_identity)
     .with_minter_health(minter_health.clone())
     .with_sync_status(sync_status.clone())
-    .with_peers(peers_snapshot.clone());
+    .with_peers(peers_snapshot.clone())
+    // Live traffic / connection-direction counters for `network_getInfo`
+    // (#542). Shares the same handle the network event loop increments.
+    .with_network_stats(discovery.stats());
 
     // Initialize wallet for RPC (balance checking, faucet, etc.)
     if let Some(mnemonic) = config.mnemonic() {
@@ -653,7 +656,9 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                 pending_txs.len()
             );
             for tx in &pending_txs {
-                if let Err(e) = NetworkDiscovery::broadcast_transaction(&mut swarm, tx) {
+                if let Err(e) =
+                    NetworkDiscovery::broadcast_transaction(&mut swarm, discovery.stats_ref(), tx)
+                {
                     debug!("Failed to broadcast pending tx: {}", e);
                 }
             }
@@ -1196,7 +1201,7 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                         indices: missing_indices,
                                     };
 
-                                    if let Err(e) = NetworkDiscovery::request_block_txns(&mut swarm, &request) {
+                                    if let Err(e) = NetworkDiscovery::request_block_txns(&mut swarm, discovery.stats_ref(), &request) {
                                         warn!("Failed to request missing transactions: {}", e);
                                     }
                                 }
@@ -1240,7 +1245,7 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             };
 
                             if let Some(response) = response {
-                                if let Err(e) = NetworkDiscovery::respond_block_txns(&mut swarm, &response) {
+                                if let Err(e) = NetworkDiscovery::respond_block_txns(&mut swarm, discovery.stats_ref(), &response) {
                                     warn!("Failed to send BlockTxn response: {}", e);
                                 }
                             }
@@ -1480,6 +1485,7 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                         let legacy_peers = discovery.legacy_peer_count() > 0;
                                         if let Err(e) = NetworkDiscovery::broadcast_block_smart(
                                             &mut swarm,
+                                            discovery.stats_ref(),
                                             &block,
                                             legacy_peers,
                                         ) {
@@ -1497,7 +1503,9 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                         }
                         ConsensusEvent::BroadcastMessage(msg) => {
                             // Broadcast SCP message to network
-                            if let Err(e) = NetworkDiscovery::broadcast_scp(&mut swarm, &msg) {
+                            if let Err(e) =
+                                NetworkDiscovery::broadcast_scp(&mut swarm, discovery.stats_ref(), &msg)
+                            {
                                 warn!("Failed to broadcast SCP message: {}", e);
                             }
 
@@ -1691,7 +1699,9 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                 // nominate/ballot message. Without this, peers reject our SCP
                 // messages ("Transaction not in cache") and multi-node
                 // nomination never reaches quorum (issue #409).
-                if let Err(e) = NetworkDiscovery::broadcast_minting_tx(&mut swarm, minting_tx) {
+                if let Err(e) =
+                    NetworkDiscovery::broadcast_minting_tx(&mut swarm, discovery.stats_ref(), minting_tx)
+                {
                     debug!("Failed to broadcast minting tx: {}", e);
                 }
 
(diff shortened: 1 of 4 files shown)


rjwalters:
✅ **Approved!**

Real `network_getInfo` traffic/connection counters replacing the hardcoded `0`s from #542. Reviewed against the audit's "don't mislead operators" bar — the counters are correct, the direction accounting is safe, and the partial scope is honestly documented.

## Counter correctness (real + lock-free)
- `NetworkStats` = 4 `AtomicU64` (`Relaxed`), owned by `NetworkDiscovery`, read by RPC via a cheap `Arc` snapshot (`stats()`). No per-packet lock on the hot path. ✅
- `record_sent(len)` fires **after** a successful `publish()` at every gossipsub send site: `broadcast_block`, `broadcast_transaction`, `broadcast_minting_tx`, `broadcast_scp`, `broadcast_compact_block`, `request_block_txns`, `respond_block_txns`, `broadcast_pex`, `broadcast_upgrade_announcement` (and `broadcast_block_smart` via the compact/full paths). Recording after publish means a failed publish doesn't inflate the counter. ✅
- `record_received(message.data.len())` in the gossipsub `Message` arm, before the rate-limit drop — correct, since those bytes already crossed the wire. ✅

## Direction-count safety (cannot drift / double-count)
- `record_connection_opened(!endpoint.is_dialer())` only when `num_established == 1`; `record_connection_closed(...)` only when `num_established == 0`. Mirrors the existing per-peer `peer_count` accounting, so concurrent multi-connection peers count exactly once. ✅
- Close path uses a `compare_exchange_weak` loop that saturates at 0 — a spurious/extra close can never underflow `u64`. Covered by `test_network_stats_close_saturates_at_zero`. ✅
- `test_connected_point_direction_maps_to_counter` pins the dialer→outbound / listener→inbound mapping using real `ConnectedPoint` values, guarding against silent inversion. ✅

## Honest scope (the important part for this audit)
The two gaps are clearly documented on `NetworkStats`, on the `RpcState.network_stats` field, and in the PR body:
1. Sync `request_response` payload bytes are NOT counted.
2. Transport framing (Noise/yamux/TCP) is NOT counted — these are application-gossip-payload counters, not raw wire totals.

An operator reading either the code or the docs will not misread `bytesSent`/`bytesReceived` as total wire traffic. A partial-but-documented counter is strictly better than the prior hardcoded `0`, and the gaps are tracked as follow-ons. Acceptable.

## Test Execution (verified locally in worktree)
- `cargo build -p botho` — passes (only pre-existing dead-code warnings in transport/webrtc).
- `cargo test -p botho --lib network_stats` — 5/5 pass.
- `cargo test -p botho --lib test_network_info_surfaces_real_counters` — passes (fallback shape, real values, genuine-zero all asserted).
- `cargo clippy -p botho --lib` — no new warnings in the changed regions (NetworkStats / handle_network_info).

## CI
- `rustfmt --check` passed. No required checks (`isRequired: null` across the board). Merge state is `UNSTABLE` only because two **non-required** build-sanity jobs (`Compile Benchmarks`, `Fuzz Build Guard (PR)`) are still in progress — neither gates merge nor bears on this change's correctness; build verified locally.

Nice work, including the explicit fallback to the prior placeholder shape (`inboundCount: 0`, `outboundCount: peerCount`) for relay/test nodes with no network loop. Closes #542.

PR #564, opened 2026-06-24
fix(ledger): fail closed on DB errors in consensus double-spend checks

### LINKED ISSUE #559: security(M7): propagate errors on consensus DB checks (double-spend fail-open) [#555]
Split out from #555 (cycle-6 **M7**). Verified still open against `main` on 2026-06-23.

## Problem
Consensus-critical DB lookups swallow errors and default-allow (fail-open):

- `botho/src/ledger/store.rs:1128` — in `verify_transaction`, the double-spend check is `if let Ok(Some(spent_height)) = self.is_key_image_spent(&input.key_image)`. A DB **`Err`** makes the condition false, so the spent-key-image check is silently skipped → a double-spend can pass on a transient DB error.
- `botho/src/ledger/store.rs:1229` — same `if let Ok(Some(..))` pattern in `record_key_image`'s existence check.
- `botho/src/mempool.rs:68` — `ledger.get_cluster_wealth(entry.cluster_id.0).unwrap_or(0)`: a DB error defaults wealth to 0 → lower progressive fee.

## Fix
Propagate the `Result` on consensus checks; never default-allow. The double-spend check at `:1128` must return `Err`/reject on a DB error rather than fall through. Apply the same to `:1229`. For `mempool.rs:68`, surface the error (or fail-closed to a conservative high wealth) rather than `unwrap_or(0)`.

## Notes
- The happy path is unchanged (DB errors should not occur in normal operation), so this is a **correctness/robustness** fix, **not** a consensus-rule change — it can land independently of the #323 reset.
- Add tests that inject a DB error and assert the check rejects (fails closed).

Ref: `audits/2026-06-11-cycle6.md` (M7); parent #555.

rjwalters:
## Summary

Consensus-critical DB lookups swallowed errors and default-allowed (fail-open). A transient DB error could silently skip the double-spend check, record a key image despite a read failure, or lower the progressive fee. This makes all three checks **fail closed**: a DB error is propagated as a node-local error, never converted into an allow.

Critically, a DB error is treated as a **node-local operational failure** (propagated up so the caller aborts/retries), **not** a "block invalid" verdict. This avoids the risk of this node forking from peers that had no DB error. The normal (no-DB-error) path is byte-for-byte identical to before — only the error path changes from allow to propagate.

## Changes

- `botho/src/ledger/store.rs` — `verify_transaction`: replaced `if let Ok(Some(spent_height)) = self.is_key_image_spent(..)` with `if let Some(spent_height) = self.is_key_image_spent(..)?`. A DB `Err` now propagates (`LedgerError::Database`); a genuine spent key image still returns `LedgerError::InvalidBlock` (unchanged).
- `botho/src/ledger/store.rs` — `record_key_image`: the existence check's `if let Ok(Some(..)) = self.key_images_db.get(..)` previously let a DB `Err` fall through to the `put` (recording the key image anyway). Now reads the `Result` explicitly, propagating a DB error and only branding a genuine collision `InvalidBlock`.
- `botho/src/mempool.rs` — `effective_cluster_wealth_from_outputs`: replaced `get_cluster_wealth(..).unwrap_or(0)` (a DB error defaulted wealth to 0, lowering the fee) with error propagation. The function now returns `Result<u64, String>`; the `add_tx` call site maps the error to `MempoolError::LedgerError`. **Choice rationale:** propagating is the least-surprising option that does not lower the fee on error — the tx is rejected from the mempool rather than admitted at a fee computed from bogus zero wealth. (A "conservative HIGH wealth" fallback would invent a fee floor out of thin air; propagating surfaces the real operational fault.)
- Added a shared private `Ledger::open_internal(path, network, max_readers)` so a `#[cfg(test)]` `open_single_reader` can cap the LMDB reader table to 1, plus a `#[cfg(test)]` `read_txn_for_test` accessor — the only practical way to inject a real DB error (`MdbError::ReadersFull`) in a unit test.

## Call-stack trace

`verify_transaction` has exactly one caller: `store.rs` block application (`self.verify_transaction(tx)?`), which already propagates with `?` — the newly-returned `Err` aborts block application cleanly, no caller update needed. `record_key_image` is likewise called with `?` during block application. `effective_cluster_wealth_from_outputs` is called from `Mempool::add_tx` (now mapped to `MempoolError`) and from three tests (updated to `.unwrap()`).

| `verify_transaction` propagates DB error instead of allowing | ✅ | `test_verify_transaction_propagates_db_error_fail_closed` injects `ReadersFull` and asserts `Err(LedgerError::Database)` |
| `record_key_image` existence check propagates DB error | ✅ | Code reads `get(..)` Result and `?`-propagates; build + existing `test_key_image_double_spend_rejected` still green |
| `mempool` wealth lookup does not default to 0 on error | ✅ | `test_effective_cluster_wealth_fails_closed_on_db_error` asserts `Err` |
| Happy path unchanged (valid accepted, truly-spent rejected) | ✅ | `test_verify_transaction_rejects_truly_spent_key_image` returns `InvalidBlock`; full lib suite (1049 tests) passes |
| Callers handle propagated Err sanely (abort, not loop) | ✅ | Sole `verify_transaction`/`record_key_image` callers already use `?`; `add_tx` maps to `MempoolError::LedgerError` |

## Test Plan

- `cargo build -p botho` — passes (only pre-existing warnings).
- `cargo test -p botho --lib` — 1049 passed, 0 failed, 2 ignored (pre-existing).
- New tests injecting a DB error (reader-table exhaustion) confirm fail-closed propagation; happy-path tests confirm valid txs and truly-spent rejection are unchanged.

Closes #559

### DIFF
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -137,11 +137,47 @@ impl Ledger {
         Self::open_for_network(path, Network::Testnet)
     }
 
+    /// Test-only: open a ledger whose LMDB reader table holds exactly one slot.
+    /// Holding a single live `RoTxn` then exhausts the table, so the next
+    /// `read_txn()` (e.g. inside `is_key_image_spent`) fails with
+    /// `MdbError::ReadersFull`. Used by the M7 fail-closed tests to inject a DB
+    /// error on the consensus double-spend check.
+    #[cfg(test)]
+    pub fn open_single_reader(path: &Path) -> Result<Self, LedgerError> {
+        Self::open_internal(path, Network::Testnet, Some(1))
+    }
+
+    /// Test-only: open a read transaction against the private LMDB environment.
+    /// Combined with `open_single_reader`, holding the returned `RoTxn`
+    /// exhausts the reader table so subsequent lookups fail closed (used by
+    /// M7 tests in sibling modules such as the mempool).
+    #[cfg(test)]
+    pub fn read_txn_for_test(&self) -> Result<heed::RoTxn<'_>, LedgerError> {
+        self.env
+            .read_txn()
+            .map_err(|e| LedgerError::Database(format!("Failed to start read txn: {}", e)))
+    }
+
     /// Open or create a ledger at the given path for a specific network.
     ///
     /// The ledger will be initialized with the appropriate genesis block
     /// for the specified network if it's empty.
     pub fn open_for_network(path: &Path, network: Network) -> Result<Self, LedgerError> {
+        Self::open_internal(path, network, None)
+    }
+
+    /// Internal constructor shared by `open_for_network` and test helpers.
+    ///
+    /// `max_readers`, when `Some`, caps the LMDB reader-slot table. Production
+    /// callers pass `None` (LMDB's default of 126). Tests pass a small value so
+    /// they can deterministically exhaust the reader table and force read
+    /// transactions to return `MdbError::ReadersFull` — the only practical way
+    /// to inject a DB failure for the M7 fail-closed tests.
+    fn open_internal(
+        path: &Path,
+        network: Network,
+        max_readers: Option<u32>,
+    ) -> Result<Self, LedgerError> {
         // Create directory if needed
         fs::create_dir_all(path)
             .map_err(|e| LedgerError::Database(format!("Failed to create directory: {}", e)))?;
@@ -154,10 +190,13 @@ impl Ledger {
         // directory first, and storing the Env in the struct which owns it for
         // its lifetime.
         let env = unsafe {
-            EnvOpenOptions::new()
-                .max_dbs(7) // Increased for cluster_wealth_db
-                .map_size(1024 * 1024 * 1024) // 1GB
-                .open(path)
+            let mut opts = EnvOpenOptions::new();
+            opts.max_dbs(7) // Increased for cluster_wealth_db
+                .map_size(1024 * 1024 * 1024); // 1GB
+            if let Some(readers) = max_readers {
+                opts.max_readers(readers);
+            }
+            opts.open(path)
         }
         .map_err(|e| LedgerError::Database(format!("Failed to open environment: {}", e)))?;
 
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review: APPROVED

Reviewed with heightened rigor as this touches the consensus double-spend path. Verdict: **approve**.

### The critical distinction (DB-error vs block-invalid) is correctly implemented

This is the heart of the fix and it is right at every changed site:

- **`verify_transaction`** — `is_key_image_spent(..)?` propagates a DB `Err` as node-local `LedgerError::Database`; only a successful lookup returning `Some(spent_height)` produces the `InvalidBlock` verdict. Confirmed against `is_key_image_spent` (store.rs:1202): it returns `Ok(None)` *only* when the read txn succeeds and the key image is absent, and `Err(Database)` on any read/txn failure. A transient DB error can therefore never be branded "block invalid" — exactly the property that prevents a node-local fork on a transient fault.
- **`record_key_image`** — confirmed the old `if let Ok(Some(..))` let a DB `Err` fall through to the `put` (recording the key image despite a failed existence read = fail-open). New code reads the `Result`, `?`-propagates the DB error, and only an actual collision returns `InvalidBlock`.

### Normal-path behavior is byte-for-byte identical
In the no-DB-error case the accepted/rejected set is unchanged: `Ok(None)` -> continue, `Ok(Some)` -> `InvalidBlock`. Only the previously-swallowed `Err` path changes (allow -> propagate). No change to which valid inputs are accepted. Verified by `test_verify_transaction_rejects_truly_spent_key_image` (still `InvalidBlock`).

### Mempool wealth propagation is sound
`effective_cluster_wealth_from_outputs` now returns `Result<u64, String>`; `unwrap_or(0)` (which lowered the progressive fee on a DB error = fail-open) is replaced with error propagation. At the `add_tx` call site it maps cleanly to `MempoolError::LedgerError(String)` (variant confirmed at mempool.rs:1131). This is relay/mempool policy, not consensus, so rejecting the tx on a transient DB error is acceptable and does not affect block validity. Happy path (`total_value == 0` -> `Ok(0)`, tagged outputs -> weighted wealth) unchanged; the three existing tests were updated to `.unwrap()` with no assertion changes.

### Call-stack correctness
Both consensus call sites are inside `add_block` (store.rs:660, :669) under an open write txn, already invoked with `?`. A newly-propagated `Err` aborts block application cleanly — the `wtxn` is dropped (rolled back, never committed), and `add_block` returns `Err(LedgerError)` to its caller. No loop, no panic, no caller mis-handles the new error.

### Tests genuinely exercise the fail-closed path
The `open_single_reader` / held-`RoTxn` -> `MdbError::ReadersFull` injection is a real DB error (not a mock), and the probe assertion confirms the lookup actually errors before testing `verify_transaction`. Both `open_single_reader` and `read_txn_for_test` are `#[cfg(test)]` and route through the shared private `open_internal`; production `open_for_network` passes `max_readers = None` (LMDB default 126) — no leakage into production. All new tests pass locally:
- `test_verify_transaction_propagates_db_error_fail_closed` -> `Err(Database)`
- `test_verify_transaction_rejects_truly_spent_key_image` -> `InvalidBlock`
- `test_effective_cluster_wealth_fails_closed_on_db_error` -> `Err`

### Validation
- `cargo test -p botho --lib`: **1049 passed, 0 failed, 2 ignored** (run in worktree).
- CI: all required checks green (benchmarks, fuzz build guard, rustfmt all pass; fuzz jobs skipping as expected).
- Merge state: **CLEAN / MERGEABLE**.

No blocking issues. The fail-open hole is closed without altering consensus verdicts on the happy path.

PR #563, opened 2026-06-24
test(network): pass Arc<NetworkStats> to create_sync_behaviour in integration test

rjwalters:
## Summary
The `network_integration` integration-test target stopped compiling on `main` after `create_sync_behaviour` gained a `stats: Arc<NetworkStats>` parameter during the NetworkStats byte-counting work (#548/#551). The test helper `create_test_swarm` still called it with zero arguments, so the target silently failed to build — protecting nothing (a false-green). This PR updates the test to match the current production signature.

## Changes
- `botho/tests/network_integration.rs`:
  - `create_sync_behaviour()` -> `create_sync_behaviour(Arc::new(NetworkStats::new()))` (the only drifted call site, at the old line 64).
  - Added `std::sync::Arc` to the imports.
  - Added `NetworkStats` to the `botho::network` import group (re-exported from `botho::network`).
- Test-only change. No production network code was modified.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Update `create_sync_behaviour` call to pass `Arc<NetworkStats>` | Done | Line now passes `Arc::new(NetworkStats::new())` |
| Sweep file for other signature drift from #542/#548/#549/#551 | Done | `grep` confirmed only one `create_sync_behaviour()` call; full target now compiles, so no other drift remains |
| `cargo test -p botho --test network_integration` builds and passes | Done | Target compiles (only pre-existing warnings); 18/18 tests pass |

## Test Plan
```
cargo test -p botho --test network_integration
...
running 18 tests
test result: ok. 18 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.06s
```
All deterministic and swarm-based tests passed; no flakiness observed on this run.

Closes #561

PR #551, opened 2026-06-23
fix(network): count sync request/response payload bytes in network_getInfo

rjwalters:
## Summary

Closes #549. Completes the observability counter from #542: `network_getInfo`'s `bytesSent` / `bytesReceived` (NetworkStats, added in #548) counted **gossipsub** traffic only — the libp2p `request_response` **sync path** (block/tx sync via `SyncCodec`) was uncounted, so `bytesReceived` substantially under-reported during initial sync.

## What changed

- **`SyncCodec` now optionally carries `Arc<NetworkStats>`** and records payload bytes at the exact boundary where each message is already (de)serialized — no extra serialization pass on the hot path:
  - `write_request` / `write_response` → `record_sent(serialized_len)`
  - `read_request` / `read_response` → `record_received(bytes_read_off_wire)` (counted before deserialize, since the bytes already crossed the wire)
- The handle is an `Option<Arc<_>>` so `SyncCodec` stays `Default` (used by `Behaviour::new` and unit tests). The inner `Arc` is cheaply cloned per substream by libp2p; all clones share the same atomics.
- **`create_sync_behaviour(stats)`** now takes the stats `Arc` and builds the behaviour via `request_response::Behaviour::with_codec`; `discovery::start` clones the shared counters in.
- **`NetworkStats` "Known gaps" doc** updated: dropped the now-covered sync request/response gap; kept the wire-level framing gap (Noise/yamux/TCP headers — **#550**, still open).

## Where bytes are recorded

`botho/src/network/sync.rs`, in the `Codec for SyncCodec` impl: `write_request`, `write_response` (sent) and `read_request`, `read_response` (received). The stats handle reaches the codec via `create_sync_behaviour(Arc<NetworkStats>)` → `SyncCodec::with_stats`, wired from `NetworkDiscovery::start` (`Arc::clone(&self.stats)`).

## Tests

New unit tests in `network::sync::tests`:
- `test_codec_write_request_records_sent_bytes`
- `test_codec_write_response_records_sent_bytes`
- `test_codec_read_request_records_received_bytes`
- `test_codec_read_response_records_received_bytes`
- `test_default_codec_records_nothing` (no stats handle = no-op, still round-trips)
- `test_codec_clones_share_stats` (per-substream clones share atomics)

[...]

PR #565, opened 2026-06-24
ci: compile-check all workspace test targets so integration tests can't silently rot

### LINKED ISSUE #562: ci: build/run integration test targets so they can't silently rot (network_integration was non-compiling on main)
## Problem
The `network_integration` integration-test target silently stopped compiling (see the fix issue) and nothing caught it — CI is green despite a non-building test target. That means CI is **not building/running the integration tests**, or is scoped narrowly enough (`--lib` only) that integration targets are excluded.

## Goal
Ensure integration-test targets cannot silently rot:
- Audit the CI workflow(s) under `.github/workflows/` to see what `cargo test` scope runs (likely `--lib` or a subset).
- Make CI build **and** run the integration targets — e.g. `cargo test --workspace` or an explicit `cargo test -p botho --tests`, so a non-compiling integration target fails the build.
- If some integration tests are intentionally excluded (slow / network-flaky), at minimum add a `cargo test --no-run` compile-check step that includes all targets, so signature drift is caught even when the tests aren't executed.

## Acceptance
- A PR that breaks an integration target's compile (like the #551 `create_sync_behaviour` signature change did) fails CI.
- Document the chosen test scope in the workflow so the intent is explicit.

## Notes
- Depends on / complements the compile fix (the target must build before CI can enforce it).
- Surfaced during the #558/#560 work, 2026-06-24.

Priority: medium (dev-infra; prevents recurrence of silent false-green).

rjwalters:
## Summary
The `network_integration` integration-test target silently stopped compiling on `main` (signature change in #551, fixed in #563) and CI never went red. Root cause: the only PR-gating Rust build job, `workspace-build.yml`, runs `cargo build --workspace`, which does **not** compile `[[test]]`/integration targets. No PR-triggered workflow compiled the integration tests, so they could rot invisibly.

This adds a `cargo test --workspace --no-run` compile-check to the Workspace Build job so that any signature drift in an integration target fails CI.

## Changes
- `.github/workflows/workspace-build.yml`:
  - Add a **Compile-check all test targets (no run)** step: `cargo test --workspace --no-run`. This compiles every test target (including `botho/tests/network_integration.rs` and the other 24 integration suites) without executing them.
  - Broaden the `pull_request` path filter to include `**.rs`. The #551 change touched only `.rs` files (no `Cargo.toml`/`Cargo.lock` edit), so the old path filter would not have triggered this workflow at all — `**.rs` ensures source signature changes are actually exercised.
  - Document the chosen scope and the `--no-run` rationale inline in the workflow header and step comment.

## Why `--no-run` instead of running the tests
Per the issue's CONSTRAINTS: the integration/e2e suites are slow and some are network-flaky. They already run on demand via `e2e-tests.yml` (`workflow_dispatch`). Introducing them into the required PR path would add flakiness. `--no-run` gates on **compile**, which is exactly what catches the rot class (signature drift) while keeping the required check fast and deterministic.

| A PR that breaks an integration target's compile (like #551) now fails CI | ✅ | `cargo test --workspace --no-run` compiles all 25 `botho/tests/*` targets incl. `network_integration`; a signature mismatch there is a compile error → step fails. The `**.rs` trigger ensures the workflow runs for `.rs`-only changes like #551. |
| CI builds all test targets including integration tests under `botho/tests/` | ✅ | Verified locally: the `--no-run` build emits executables for all integration targets (network_integration, byzantine_integration, e2e_consensus_integration, etc.). |
| Chosen scope documented in the workflow | ✅ | Header comment + step comment explain the compile-check intent and the `--no-run` choice. |

## Test Plan
Ran locally on this branch (with #563 merged into main):
- `cargo test -p botho --tests --no-run` → compiles all 25 `botho/tests/*` targets cleanly, including `network_integration`.
- `cargo test --workspace --no-run` → compiles cleanly across the entire workspace (all crates' test targets). This is the exact command CI will run.
- `python3 -c "import yaml; yaml.safe_load(...)"` → workflow YAML is valid.

Closes #562

### DIFF
--- a/.github/workflows/workspace-build.yml
+++ b/.github/workflows/workspace-build.yml
@@ -5,10 +5,23 @@ name: Workspace Build
 # feature-unification break that only manifests on Linux (#361). This job
 # exists to give that bug — and any future Linux-only feature-unification
 # regression — a red/green signal so it can be fixed and stay fixed.
+#
+# It ALSO compile-checks every test target in the workspace (see the
+# `cargo test --workspace --no-run` step). `cargo build --workspace` does NOT
+# compile `[[test]]` / integration targets, so before this an integration test
+# under `botho/tests/` could silently stop compiling and CI stayed green — that
+# is exactly what happened to `network_integration` when the
+# `create_sync_behaviour` signature changed in #551 (fixed in #563). The
+# compile-check below makes that class of signature drift fail CI (#562).
+#
+# Trigger note: `**.rs` is included so a pure source change (no Cargo.toml/lock
+# edit, like #551) still exercises the compile-check. Without it this workflow
+# would not even run for the change that originally rotted the target.
 
 on:
   pull_request:
     paths:
+      - '**.rs'
       - '**/Cargo.toml'
       - '**/Cargo.lock'
       - 'crypto/**'
@@ -54,3 +67,14 @@ jobs:
 
       - name: Build workspace
         run: cargo build --workspace
+
+      # Compile (but do NOT run) every test target in the workspace, including
+      # the integration tests under `botho/tests/` (network_integration,
+      # byzantine_integration, etc.). `cargo build --workspace` above skips
+      # `[[test]]` targets, so this is what actually catches integration-test
+      # signature drift. `--no-run` is deliberate: the integration/e2e suites
+      # are slow and some are network-flaky (they run on demand via
+      # e2e-tests.yml), so we gate on COMPILE here, not execution, to keep this
+      # required check fast and deterministic while still failing on rot (#562).
+      - name: Compile-check all test targets (no run)
+        run: cargo test --workspace --no-run


rjwalters:
## Loom Judge Verdict: APPROVE

Reviewed against issue #562 acceptance criteria. This is a clean, minimal, well-targeted CI fix.

### Verification
1. **Compile-checks test targets (catches #551 class):** Confirmed. `cargo test --workspace --no-run` compiles every test target — including all 25 `botho/tests/*.rs` integration suites (network_integration, byzantine_integration, e2e_consensus_integration, etc.) — and stops before execution. `cargo build --workspace` (the prior only gate) does **not** compile `[[test]]`/integration targets, which is exactly why the `create_sync_behaviour` signature drift in #551 rotted `network_integration` invisibly. A signature mismatch is now a compile error in that step → CI goes red. This would have caught #551.

2. **`**.rs` path filter:** Correct and necessary. The old filter (Cargo.toml/Cargo.lock/crypto/desktop/the-workflow) would not have triggered for a pure-`.rs` change like #551, so the compile-check would never have run for the very change that caused the rot. Adding `**.rs` ensures source signature changes exercise the gate.

3. **`--no-run` rationale:** Sound and documented. Verified `e2e-tests.yml` exists and uses `workflow_dispatch` — the slow/network-flaky e2e suites stay on manual dispatch. Gating on **compile** keeps the required PR check fast and deterministic while still catching the rot class. Rationale is documented in both the workflow header comment and the step comment.

4. **YAML/scope:** YAML parses cleanly; comment style, indentation, and step structure match the existing workflow. Scope is workflow-file-only (24 additions, 0 deletions). No flaky tests introduced into the required path.

### CI status
- Job `cargo build --workspace (Linux)`: **PASS**. Confirmed via run logs that the new **Compile-check all test targets (no run)** step executed (`cargo test --workspace --no-run`) inside that passing job.
- mergeStateStatus: **CLEAN**, mergeable: **MERGEABLE**.

Approving. Good defensive fix that closes a real CI blind spot.

— Loom Judge

Issue #539, opened 2026-06-23
ops(testnet, operator-only): diagnose & fix faucet miner producing 0 hashes — live net halted at height 213


rjwalters:
**OPERATOR-ONLY** (needs SSH/box access — not automatable; no loom build label so sweeps skip it). Split out of #537.

The live testnet is halted at height 213 because the faucet RandomX miner reports `active:true` but **hashrate 0.0 / 0 total hashes over ~50h**. Difficulty was ruled out in #537 (it's the easy initial target, ~341 hashes/block); the miner itself is wedged.

## Steps
1. SSH the faucet box; inspect `journalctl -u botho` (or the service unit) for **RandomX VM/dataset init errors** (~2GB dataset needs RAM / possibly huge-pages; an alloc failure would wedge the mining thread). Confirm whether the mining thread is alive or panicked.
2. Restart the service; watch `minting_getStatus` until `hashrate > 0` and `blocksFound` advances; confirm height climbs past 213 on all 3 nodes.
3. If it recurs, capture RandomX flags/env on the AMI (JIT / huge-pages / memory) — the same AMI/binary runs on seed/seed2, so a root cause there blocks the multi-minter setup too.

## Relation
- **Gates enabling minting on seed/seed2** — do that only after faucet hashes > 0 (else 3 stuck miners, same AMI/binary).
- Overlaps #323 (testnet redeploy/reset); may be resolved as part of it. After any redeploy, verify `minting_getStatus` hashrate > 0 before declaring the net healthy.
- Code-side early-warning for this failure mode: #538 (stuck-miner detector).

Part of #323.

rjwalters:
## Correction to the diagnosis (important)

PR #540 (the #538 detector work) revealed that `minting_getStatus` was **hardcoding `hashrate: 0.0` and `totalHashes: 0`** (TODO placeholders) — so the '0 hashrate / 0 hashes over ~50h' reading this issue cited was likely **the buggy RPC, not ground truth.** The miner may actually have been hashing.

**What's still certain:** the chain is genuinely halted at height 213 (independently confirmed by sampling `chainHeight` — the real ledger tip — 10× over 2.5min; all 3 nodes agree, so a halt, not a fork).

**What's now uncertain:** *why* no blocks are produced. The miner-hashing status can't be trusted from the deployed RPC. Possibilities: miner genuinely stalled, OR hashing but blocks not being proposed/accepted, OR another path.

**Revised operator steps:**
[...]

PR #567, opened 2026-06-24
fix(minter): resilient + truthful RandomX miner — clear active on thread death, fast→light fallback, retry-on-build-failure (code side of #539)

### LINKED ISSUE #566: fix(minter): resilient + truthful RandomX miner — clear active on thread death, fast→light fallback, retry-on-build-failure (code side of #539)
Code-side fixes for the failure mode diagnosed in #539: the live testnet halted at height 213 because the faucet RandomX miner reported `active:true` / `hashrate 0.0` for ~50h. Root cause (confirmed by code reading): the mining thread failed to build the 2 GB fast-mode dataset (`FastHasher::new`, forced by `fast_flags() = recommended | FLAG_FULL_MEM`) on a box without enough RAM/huge-pages, hit the error arm in `mint_loop` (`botho/src/node/minter.rs:608-614`), and `break`ed — but nothing cleared the `active` health flag, so the node reported a healthy miner while the thread was dead.

This issue makes the in-process miner **resilient and truthful** so an undersized box degrades gracefully instead of silently halting the net. It is buildable without box access and should land in the binary deployed by the #323 reset.

## A. Truthful `active` flag (highest value)
`health.set_active(true)` is set in `Minter::start()` (minter.rs:494) and `set_active(false)` is ONLY called in `stop()` (minter.rs:499). Every mint-thread death path leaves `active:true` with 0 hashes:
- dataset-build error `break` (L608-614)
- poisoned `current_work` lock `break` (L570-572)
- channel-closed `break` (L684-694)
- `hash()`'s `.expect("RandomX fast hash failed")` panic (pow.rs:191)

**AC:** when all mint worker threads have exited (for any reason), the miner's `active` health flag AND the `botho_minting_active` metric (run.rs `set_minting_active`) reflect `false`. A node whose mining thread has died must NOT report `active:true`. (Implementation options: have each worker clear/decrement a live-worker count on exit and flip active when it reaches 0; or a lightweight monitor that flips active when `handle.is_finished()` for all handles. Pick the simplest that is correct under panic-exit too.)

## B. Fast→light mining fallback (prevents the halt class)
`fast_flags()` unconditionally sets `FLAG_FULL_MEM` (the 2 GB dataset). There is no degradation path, so a low-RAM box halts entirely. RandomX light mode (~256 MB cache, no dataset) produces the **identical hash** (see pow.rs module docs: "RandomX produces the same hash regardless of mode or CPU-feature flags") — only slower — so a light-mode-mined block verifies on every node. CONSENSUS-SAFE.

**AC:** if `FastHasher::new` fails (e.g. dataset alloc failure), the miner falls back to a light-mode mining hasher and KEEPS HASHING (slower) rather than dying. Add a miner-usable light hasher (analogous to the existing cached light verifier VM but usable for mining arbitrary preimages). Log a clear WARN that the miner is running in degraded light mode. Add a test asserting fast and light hashers produce identical output for the same (seed_key, preimage).

## C. Retry-with-backoff on dataset build failure
A single transient dataset-build failure should not permanently wedge the miner.

**AC:** on `FastHasher::new` failure, retry a small bounded number of times with backoff before falling back to light mode (B). Do not busy-loop. (A+B+C compose: retry fast a few times → fall back to light → keep the `active` flag truthful throughout.)

## Constraints
- Consensus-safe: fast and light RandomX MUST produce identical PoW output — do not change the preimage, seed schedule, or `pow_value` interpretation. This is a resilience/observability change, NOT a consensus-rule change.
- `cargo build` + `cargo test -p botho` green; add the tests noted in A/B.
- Follow CLAUDE.md code-preservation.

## Relation
- Diagnoses/addresses the code side of #539 (operator-only infra halt). The operator should still confirm RAM/huge-pages on the box, but with B the net survives an undersized box.
- Complements #538 (stall detector — already detects this; this issue makes it self-heal + truthful).
- Should land before the #323 reset so the deployed binary is resilient.

Priority: high (gates a safe testnet reset; turns a silent permanent halt into graceful degradation).

rjwalters:
## Summary

Makes the in-process RandomX miner **resilient and truthful** so an undersized
box degrades gracefully instead of silently halting the network. This is the
code-side fix for the live-testnet halt (#539): the faucet miner reported
`active:true` / `hashrate 0.0` for ~50h because the mining thread failed to build
the 2 GB fast-mode dataset and silently died, leaving the active flag stuck true.

All three parts land together because they all touch `mint_loop` and the hasher
path.

### A. Truthful `active` flag (`minter.rs`, `run.rs`)
- Added a `live_workers: Arc<AtomicUsize>` count to `Minter`, seeded to the
  thread count in `start()` (before spawning, before `set_active(true)` to avoid
  a death-during-startup race).
- Each worker thread holds a `WorkerExitGuard` whose `Drop` decrements the count
  and flips the health `active` flag false once the **last** worker exits — fires
  on every death path including a panic (Drop runs during unwind).
- `commands/run.rs` status loop now reconciles the RPC `active` field
  (`minting_active` RwLock) and the `botho_minting_active` metric against the
  health flag: if all workers died while minting was still advertised, it clears
  both (and the WS broadcast). The RPC `active` is sourced from that RwLock
  (`rpc/mod.rs:1979/941`), so this is required for `node_getStatus` /
  `minting_getStatus` to be truthful. No auto-restart (operator policy).

### B. Fast→light mining fallback (`pow.rs`, `minter.rs`)
- Added `LightHasher` (miner-usable light-mode hasher, ~256 MB cache, no
  `FLAG_FULL_MEM`) and a `MinerHasher { Fast | Light }` enum exposing the same
  `hash()/seed_key()` interface.
- `mint_loop` now falls back to light mode and KEEPS HASHING (with a clear WARN)
  instead of breaking when the fast dataset can't be built.
- Consensus-safe: light and fast produce identical PoW. Extended
  `pow::test_light_equals_fast` to also assert `LightHasher == FastHasher ==
  verify_pow_hash` (reuses the single fast dataset build).

### C. Retry-with-backoff (`minter.rs`)
- New `build_mining_hasher` retries `FastHasher::new` up to 3 times with
  exponential backoff (500ms ×1, ×2) before degrading to light mode. Backoff is
  slept in short steps and honors `shutdown` so `stop()` is not delayed.

### Incidental (in scope)
- The work read-lock is now released before the seconds-expensive hasher build
  (it previously held `current_work.read()` across the build, which the
  retry/backoff window would have extended — blocking `update_work`).

| A: all-workers-exited ⇒ health `active` false (incl. panic) | done | `test_active_flips_false_when_last_worker_exits`, `test_active_flips_false_on_worker_panic` |
| A: `botho_minting_active` metric / RPC active kept consistent | done | reconciliation in `run.rs` status loop, clears RwLock+metric+WS |
| B: fast-build failure ⇒ light fallback, keeps hashing | done | `build_mining_hasher` → `MinerHasher::Light`, WARN logged |
| B: fast == light identical output | done | extended `pow::test_light_equals_fast` (LightHasher==FastHasher==verify) |
| C: bounded retry with backoff, no busy-loop | done | `FAST_BUILD_ATTEMPTS=3`, exponential `FAST_BUILD_BACKOFF_BASE_MS` |
| Consensus rules unchanged (preimage/seed/pow_value) | done | no edits to those; known-answer vector test unchanged |
| build + lib tests green | done | `cargo build -p botho`, `cargo test -p botho --lib` |

## Test Plan
- [ ] `cargo build -p botho`
- [ ] `cargo test -p botho --lib` (includes new guard tests + extended fast==light)

Closes #566

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -1543,7 +1543,38 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                 // alarm only — we do NOT auto-restart (operator policy).
                 if let Ok(guard) = minter_health.read() {
                     if let Some(health) = guard.as_ref() {
-                        health.check_and_warn();
+                        let snap = health.check_and_warn();
+
+                        // Truthful-active reconciliation (#566, code side of
+                        // #539). The mint worker threads flip the shared health
+                        // `active` flag false once the LAST of them exits (by any
+                        // path, including a panic). If that has happened while the
+                        // RPC `active` field / `botho_minting_active` metric still
+                        // advertise a live miner, clear them so a node with a dead
+                        // mining thread can never report `active:true`. Idempotent:
+                        // once cleared, later ticks see it already false and skip.
+                        // We deliberately do NOT auto-restart (operator policy,
+                        // matching the stall detector).
+                        if !snap.active {
+                            let still_advertised = minting_active
+                                .read()
+                                .map(|a| *a)
+                                .unwrap_or(false);
+                            if still_advertised {
+                                warn!(
+                                    "All mint worker threads have exited but minting \
+                                     was still advertised active — clearing the active \
+                                     flag/metric to match reality (#566). The miner is \
+                                     no longer producing blocks; investigate the worker \
+                                     death (see logs above) and restart minting."
+                                );
+                                if let Ok(mut active) = minting_active.write() {
+                                    *active = false;
+                                }
+                                metrics_updater.set_minting_active(false);
+                                ws_broadcaster.minting_status(false, 0.0, 0);
+                            }
+                        }
                     }
                 }
             }
--- a/botho/src/node/minter.rs
+++ b/botho/src/node/minter.rs
@@ -5,7 +5,7 @@ use bth_util_from_random::FromRandom;
 use rand_core::OsRng;
 use std::{
     sync::{
-        atomic::{AtomicBool, AtomicU64, Ordering},
+        atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
         mpsc::{channel, Receiver, Sender},
         Arc,
     },
@@ -16,8 +16,20 @@ use tracing::{info, trace, warn};
 
 use crate::{
     block::{calculate_block_reward, MintingTx},
-    pow::{self, FastHasher},
+    pow::{self, FastHasher, LightHasher, MinerHasher},
 };
+
+/// How many times a worker retries building the fast-mode RandomX dataset
+/// before falling back to light mode (#566, part C). A single transient dataset
+/// allocation failure (e.g. momentary memory pressure / huge-pages contention)
+/// should not permanently wedge the miner, but we must not busy-loop either.
+const FAST_BUILD_ATTEMPTS: u32 = 3;
+
+/// Base backoff between fast-mode dataset build retries, in milliseconds. Grows
+/// exponentially (×1, ×2, …) across attempts. Kept small: the goal is to ride
+/// out a transient failure, not to stall for long before degrading to light
+/// mode (#566, part C).
+const FAST_BUILD_BACKOFF_BASE_MS: u64 = 500;
 
 /// Genesis / initial minting difficulty target for **RandomX**.
 ///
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!** Reviewed with heightened rigor as a consensus-touching change. This is a clean, well-documented fix for the #539 silent-halt class of bug.

**Fixed during review:** applied the pinned-nightly `rustfmt` doc-comment rewrapping (`537fb77`) that was failing the `rustfmt --check` CI job — comment rewrapping only, no logic change.

### 1. Consensus safety — HOLDS (no fork risk)
- `LightHasher::new` uses `light_flags()` — **byte-identical** to the verifier path (`verify_pow_hash`/`LIGHT_VM` also use `light_flags()`). `FastHasher` uses `fast_flags()` = same recommended flags `| FLAG_FULL_MEM`. RandomX guarantees identical output across light/fast; flags affect speed only.
- No edits to `pow_preimage`, `seed_key_for_height`, `pow_value`, or the difficulty comparison. The `test_known_answer_vector` and its `KNOWN_ANSWER_HEX` are untouched.
- `test_light_equals_fast` is extended to assert `LightHasher == FastHasher == verify_pow_hash` for the same `(seed_key, preimage)`. **Ran locally (full 2 GB dataset build): passes** in 47s. A light-mined block therefore verifies on every node.

### 2. Truthful `active` — truthful on every surface
- `WorkerExitGuard::Drop` decrements `live_workers` and flips the health flag false only when the **last** worker exits. Verified it fires on panic (`test_active_flips_false_on_worker_panic` — Drop runs during unwind) and on normal/early-`break` exit.
- `start()` seeds the count and sets `active=true` **before** spawning, closing the death-during-startup race.
- The `run.rs` status loop reconciles all three RPC-facing surfaces against the health flag: clears the `minting_active` `RwLock` (the source of the RPC `active` field at `rpc/mod.rs:1979/941`), the `botho_minting_active` metric, and the WS broadcast. Matches the existing set/clear pattern (lines 639/643/645). No path leaves a dead miner reporting `active:true` beyond one status-tick of eventual consistency.

### 3. Retry/backoff (part C) — bounded & shutdown-aware
- `FAST_BUILD_ATTEMPTS=3`, exponential backoff (500ms ×1, ×2). No busy-loop. Backoff is slept in ≤100ms steps that poll `shutdown`, and `shutdown` is checked at the top of each attempt and before the light fallback, so `stop()`/`join()` is not delayed.

### 4. Fallback (part B) — keeps hashing honestly
- On fast-build exhaustion the worker degrades to `MinerHasher::Light` and continues the same hash loop (which increments `total_hashes`), with a loud WARN. Only exits if light **also** fails to build — at which point the exit guard truthfully clears `active`.

### 5. Incidental lock change — correct, no race
- `current_work.read()` is now cloned-and-dropped before the seconds-long hasher build, so `update_work`'s write lock is no longer blocked across the build (or its new retry window). The loop re-checks `work_version` every iteration and re-reads/rebuilds on change, so the early release can at worst mine one iteration on a snapshot that is microseconds-stale — picked up on the next pass. No missed-work or stale-seed hazard (a seconds-long build cannot cross a 2048-block seed epoch).

### CI
All required checks green; merge state **CLEAN**. (The initial `cargo build --workspace` failure was a runner `No space left on device` infra error — re-ran clean in 4m4s.) `rustfmt`, build, fuzz guard, and benchmarks all pass. Guard tests + extended consensus test verified locally.

Excellent post-mortem-driven work — `Closes #566`.

PR #569, opened 2026-06-24
fix(minter): share one RandomX dataset across mining threads (N×2GB caused #539 halt) + RAM-aware thread default

### LINKED ISSUE #568: fix(minter): share one RandomX dataset across mining threads (N×2GB caused #539 halt) + RAM-aware thread default
Root cause of the #539 live-testnet halt (height 213), confirmed via live diagnosis 2026-06-24.

## Problem
When `minting.threads == 0` (the default — `config.rs:519 default_threads() -> 0`), `botho/src/node/mod.rs:172` resolves the count to `num_cpus::get()`. Each mining thread's `mint_loop` (`botho/src/node/minter.rs`) then builds its **own** ~2 GB fast-mode RandomX dataset (`fast_hasher: Option<FastHasher>` is a per-thread local; `FastHasher` owns its own `RandomXDataset`).

So on an N-core box, fast-mode mining demands **N × ~2 GB**. On the faucet box (2 vCPU, 3.8 GB RAM, no huge-pages) that's ~4 GB on a 3.8 GB box → the datasets can't both allocate, the miner thrashes/wedges, reports `active:true` with 0 hashes, and the process went unresponsive for ~3.5 days (net halted at height 213). Live confirmation: botho RSS 2.76 GB, ~766 MB available, journal shows two threads (thread=0 and thread=1) repeatedly logging "Building RandomX fast-mode dataset" within a single seed epoch.

The RandomX dataset is **read-only during hashing** and is explicitly designed to be **shared by many VMs** — allocating one per thread is wasteful and, on realistic boxes, fatal.

## Fix
1. **Share one dataset across all mining threads.** Build a single fast-mode `RandomXDataset` per seed epoch and share it (e.g. `Arc`) across every mining thread's `RandomXVM`. Then N threads cost ~2 GB (one dataset) + N × small per-VM scratchpad, not N × 2 GB. Investigate the `randomx-rs` API for constructing multiple `RandomXVM`s from one shared dataset:
   - If the crate allows a VM to be built from a shared/borrowed dataset (or `RandomXDataset` is internally refcounted / `Clone`), use that.
   - If the crate genuinely cannot share a dataset across VMs, fall back to **capping fast-mode mining to a single dataset+VM** (one mining thread in fast mode) and document why — N×2 GB is never acceptable. (Light mode caches, ~256 MB each, may still be per-thread.)
   The dataset/epoch rebuild logic (rotates every `SEED_ROTATION_INTERVAL = 2048` blocks) must be preserved and the shared dataset rebuilt once per epoch, not per thread.
2. **RAM-aware default thread count.** `num_cpus::get()` ignores the 2 GB/thread dataset cost. With dataset sharing this is far less dangerous, but still pick a sane default (e.g. cap mining threads so total RandomX memory fits comfortably; or document that fast mode shares one dataset so num_cpus VMs is fine once sharing lands). Don't silently spawn a config that can't fit in RAM.

## Constraints
- **CONSENSUS-SAFE.** This is a memory/threading change only. Do NOT alter the PoW preimage (`pow_preimage`), the seed schedule (`seed_key_for_height`), `pow_value`, or difficulty comparison. Hash output must be byte-identical. The existing known-answer/pow-vector test must still pass; add a test that VMs sharing one dataset produce the same hash as a standalone `FastHasher` for the same (seed_key, preimage).
- Composes with #567 (light fallback + truthful active, already merged): if the shared fast dataset fails to build, the existing light fallback still applies.
- `cargo build -p botho` and `cargo test -p botho --lib` green.
- Follow CLAUDE.md code-preservation.

## Relation
- Proper fix for #539 (the operator-only halt; #567 made the miner truthful + light-degradable, this makes fast-mode mining actually scale across cores without N×2 GB).
- Should land before the #323 testnet reset so the deployed binary mines efficiently on modest boxes.

Priority: high (caused the live halt; gates an efficient testnet reset).

rjwalters:
Closes #568

## Problem

Each mining thread built its **own** ~2 GB RandomX fast-mode dataset, so an `N`-core box demanded `N × 2 GB` of RAM. The faucet box (2 cores, 3.8 GB RAM → ~4 GB demand) OOM-halted the live testnet at height 213 (#539). This is the proper fix for that halt.

## Root cause / API finding

`randomx-rs` 1.4.1 makes **true sharing possible** — no fallback-to-single-thread needed:

- `RandomXDataset` is `#[derive(Clone)]` and internally `inner: Arc<RandomXDatasetInner>` (atomic-refcounted, raw C dataset pointer).
- `RandomXVM::new(flags, None, Some(dataset))` *moves a `RandomXDataset` clone into the VM* (`linked_dataset`) and the dataset is **read-only** during hashing.

So one dataset can back many concurrently-reading VMs — exactly how Monero shares a single dataset across all mining threads. Cloning is a cheap `Arc` bump over the same ~2 GB buffer.

## Changes

- **`pow::FastDataset`** (new): builds the ~2 GB dataset **once per seed epoch**; `FastDataset::hasher()` returns a `FastHasher` (VM) over the *same* shared dataset — only a small (~2 MB) per-VM scratchpad. `unsafe impl Send` is justified (read-only after build, `Arc`-backed refcount; each VM owns its own clone and stays `!Send`/thread-local). `FastHasher::new()` now delegates to `FastDataset` (standalone path preserved for single-VM use / tests).
- **`Minter`** holds `Arc<Mutex<Option<FastDataset>>>`. `mint_loop` obtains/rebuilds the shared dataset on seed-epoch rotation (`SEED_ROTATION_INTERVAL = 2048`). The first worker into a new epoch pays the build under the mutex; the rest get cheap VMs over it. The previous-epoch dataset is freed before the new one allocates, so peak is bounded (≤ 2 × 2 GB at a rotation boundary, **independent of `N`**), not `N × 2 GB`.
- **RAM-aware default**: `default_mint_threads(num_cpus)` routes the `threads == 0` auto path, clamped to `[1, MAX_AUTO_MINT_THREADS=16]`. With sharing the count no longer drives memory (~2 GB regardless of `N`), so the clamp is purely an oversubscription guard for many-core boxes (in-process mining vs consensus/RPC, #441/#444). Explicit `minting.threads` bypasses it.

Memory: **startup/steady-state ~2 GB + N×~2 MB** (was `N × 2 GB`).

## Consensus safety

Memory/threading change **only**. `pow_preimage`, `seed_key_for_height`, `pow_value`, the difficulty comparison, and the seed schedule are untouched — hash output is byte-identical. Composes with #567 (`MinerHasher::Light` fallback + `WorkerExitGuard` truthful `active`) unchanged: if the shared fast dataset can't build, the light fallback still applies.

## Tests

- `pow::test_shared_dataset_matches_standalone` (new): builds **one** ~2 GB dataset, derives two VMs over it, and asserts both produce the **byte-identical** canonical (`verify_pow_hash`) hash — proving shared == standalone without multiple 2 GB allocations.
- `minter::test_default_mint_threads_clamps` (new): clamp behavior.
- Existing `test_known_answer_vector`, `test_light_equals_fast` (incl. #566 `LightHasher`) still pass.

`cargo build -p botho` ✅ · `cargo test -p botho --lib` ✅ (1053 passed, 0 failed) · `cargo clippy -p botho --lib` clean for touched files.

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 558
-# Branch: feature/issue-558
+# Issue: 568
+# Branch: feature/issue-568
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/botho/src/node/minter.rs
+++ b/botho/src/node/minter.rs
@@ -445,11 +445,12 @@ pub struct Minter {
     /// The ONE fast-mode RandomX dataset (~2 GB) shared by every mining thread,
     /// bound to the current seed epoch. Built once (by whichever worker first
     /// reaches a new epoch) and reused by all threads via cheap per-VM clones,
-    /// so the miner needs ~2 GB total + a small per-VM scratchpad instead of the
-    /// `N × 2 GB` that OOM-halted the live testnet (#539/#568). Rebuilt only when
-    /// the seed key rotates (every [`pow::SEED_ROTATION_INTERVAL`] blocks). The
-    /// `Mutex` serializes the expensive build so two threads never race to
-    /// allocate two datasets for the same epoch.
+    /// so the miner needs ~2 GB total + a small per-VM scratchpad instead of
+    /// the `N × 2 GB` that OOM-halted the live testnet (#539/#568). Rebuilt
+    /// only when the seed key rotates (every
+    /// [`pow::SEED_ROTATION_INTERVAL`] blocks). The `Mutex` serializes the
+    /// expensive build so two threads never race to allocate two datasets
+    /// for the same epoch.
     fast_dataset: Arc<Mutex<Option<FastDataset>>>,
 }
 
@@ -846,7 +847,8 @@ fn mint_loop(
 /// same dataset (a small per-VM scratchpad). This is what turns the miner's
 /// memory footprint from `N × 2 GB` into ~2 GB + small per-VM scratchpads — the
 /// fix for the live-testnet OOM halt (#539). The previous epoch's dataset is
-/// dropped before the new one is allocated so the cell never holds two datasets.
+/// dropped before the new one is allocated so the cell never holds two
+/// datasets.
 fn obtain_shared_fast_hasher(
     fast_dataset: &Mutex<Option<FastDataset>>,
     seed_key: [u8; 32],
@@ -1162,9 +1164,9 @@ mod tests {
     }
 
     /// The RAM-aware auto thread default (#568): clamps the detected CPU count
-    /// to `[1, MAX_AUTO_MINT_THREADS]`. With one shared dataset this is no longer
-    /// a RAM lever (it was `N × 2 GB` before), so the clamp is purely an
-    /// oversubscription guard.
+    /// to `[1, MAX_AUTO_MINT_THREADS]`. With one shared dataset this is no
+    /// longer a RAM lever (it was `N × 2 GB` before), so the clamp is
+    /// purely an oversubscription guard.
     #[test]
     fn test_default_mint_threads_clamps() {
         // At least one thread even if detection reports an absurd 0.
@@ -1174,7 +1176,10 @@ mod tests {
         assert_eq!(default_mint_threads(2), 2);
         assert_eq!(default_mint_threads(8), 8);
         // The cap value itself passes through.
-        assert_eq!(default_mint_threads(MAX_AUTO_MINT_THREADS), MAX_AUTO_MINT_THREADS);
+        assert_eq!(
+            default_mint_threads(MAX_AUTO_MINT_THREADS),
+            MAX_AUTO_MINT_THREADS
+        );
         // Many-core boxes are clamped to the oversubscription cap.
         assert_eq!(default_mint_threads(128), MAX_AUTO_MINT_THREADS);
         assert_eq!(
--- a/botho/src/node/minter.rs
+++ b/botho/src/node/minter.rs
@@ -7,7 +7,7 @@ use std::{
     sync::{
         atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
         mpsc::{channel, Receiver, Sender},
-        Arc,
+        Arc, Mutex,
     },
     thread::{self, JoinHandle},
     time::Instant,
@@ -16,7 +16,7 @@ use tracing::{info, trace, warn};
 
 use crate::{
     block::{calculate_block_reward, MintingTx},
-    pow::{self, FastHasher, LightHasher, MinerHasher},
+    pow::{self, FastDataset, FastHasher, LightHasher, MinerHasher},
 };
 
 /// How many times a worker retries building the fast-mode RandomX dataset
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge review: CHANGES REQUESTED (one trivial CI blocker)

Reviewed with heightened rigor as a PoW/miner-core change. The implementation is **correct and consensus-safe** — the only thing standing between this and approval is a red CI check.

### Blocker: `rustfmt --check (pinned nightly)` is failing

`gh pr checks` shows the pinned-nightly rustfmt job red and `mergeStateStatus = UNSTABLE`. Confirmed locally with `cargo +nightly fmt --check`: the diffs are **purely cosmetic** (doc-comment re-wrapping in `minter.rs`/`pow.rs` plus splitting one long `assert_eq!(default_mint_threads(MAX_AUTO_MINT_THREADS), MAX_AUTO_MINT_THREADS)` in the clamp test). No logic touched.

Fix: run `cargo +nightly fmt` in the worktree and push. That should turn CI green and this is otherwise ready to approve.

---

### What I verified (all PASS)

**(a) Byte-identical PoW property — HOLDS.** `pow_preimage`, `seed_key_for_height`, `pow_value`, the big-endian difficulty read, and the `SEED_ROTATION_INTERVAL=2048` seed schedule are all untouched. `FastDataset::hasher()` builds the VM via the exact same call `RandomXVM::new(fast_flags(), None, Some(dataset))` as the old `FastHasher::new`; `FastHasher::new` now just delegates to `FastDataset::new()?.hasher()`. `FastHasher::hash` is unchanged. The new `test_shared_dataset_matches_standalone` asserts two VMs over one shared dataset agree with each other AND with the canonical `verify_pow_hash` (light path) ground truth — exactly the right anti-fork assertion. `test_known_answer_vector` / `test_light_equals_fast` unchanged.

**(b) `unsafe impl Send for FastDataset` — SOUND.** Confirmed against the locked randomx-rs **1.4.1** source: `RandomXDataset` is `#[derive(Clone)]` with `inner: Arc<RandomXDatasetInner>` and is documented as a read-only structure; `RandomXVM::new(.., Some(dataset))` moves a *clone* into the VM's `linked_dataset`, so **each mining thread's VM owns its own Arc handle** to the shared ~2 GB buffer. The C `randomx_vm`'s scratchpad lives inside the per-VM object (per-VM, not shared). `FastHasher`/`RandomXVM` stay `!Send` — VMs never cross threads; only the dataset Arc does, and only its atomic refcount is touched on clone/drop. `FastDataset` impls `Send` but not `Sync`, and the shared cell is only read under the `Mutex` guard, so no shared `&FastDataset` race. The safety comment is accurate.

**(3) Memory bound — CORRECT.** `obtain_shared_fast_hasher` does `*guard = None` before `Some(FastDataset::new())`, and `mint_loop` sets `hasher = None` before rebuild, dropping each thread's old VM (and its old-dataset Arc clone) as it rotates. Worst case at an epoch boundary is old buffer (held by lagging VMs) + new buffer = 2×2 GB, **independent of N**. The 2 GB build runs under the mutex (expected stall, single non-nested lock — not deadlock-prone), with poison recovery via `into_inner()`.

**(4) Composes with #567 — INTACT.** `build_mining_hasher` retains the bounded retry + `LightHasher` fallback; `obtain_shared_fast_hasher` returns `Result` and an `Err` flows into the same retry/light-fallback path. `WorkerExitGuard`/`live_workers` untouched.

**(5) RAM-aware default — SANE.** `default_mint_threads` clamps to `[1, 16]`; `node/mod.rs` routes the `minting.threads == 0` path through it; explicit config bypasses. `test_default_mint_threads_clamps` is meaningful and passes locally.

**(6) Build/tests/clippy.** Clamp test passes locally; clippy is clean for the touched files (the 70 warnings are pre-existing in `commands/run.rs`, not introduced here).

Just the fmt fix and this is good to merge.

rjwalters:
## Re-judge verdict: APPROVE

Re-reviewed after the changes-requested cycle. The only blocker on the prior substantive review was the `rustfmt --check` failure; that is now resolved.

**Delta since prior approval = exactly one cosmetic commit (`f718bf0`, "style: cargo +nightly fmt").** Verified via `gh pr diff` / `git show`: every `+`/`-` line is a doc-comment (`///`) re-wrap or the single `assert_eq!(default_mint_threads(MAX_AUTO_MINT_THREADS), ...)` split in `test_default_mint_threads_clamps`. No logic changes.

Confirmed unchanged (load-bearing):
- `pow_preimage`, `seed_key_for_height`, `pow_value`, difficulty comparison, seed schedule — untouched.
- `unsafe impl Send for FastDataset {}` and its soundness justification — unchanged (only the doc comment above it re-wrapped). Each thread still gets its own VM over a read-only-shared, Arc-backed dataset.
- Dataset-sharing logic (`FastDataset::new`/`hasher`, `obtain_shared_fast_hasher`, `RandomXVM::new(..., Some(self.dataset.clone()))`) — unchanged.

**Prior substantive conclusions still hold:**
- **Byte-identical PoW**: HOLDS. Sharing changes memory only; `test_shared_dataset_matches_standalone` pins it against the canonical light-mode verifier.
- **`unsafe impl Send` soundness**: SOUND. Per-thread VM, dataset read-only-shared.
- Memory bound (~2 GB + N x small scratchpads, <=2x2GB across epoch rotation), composition with #567 (Light fallback + WorkerExitGuard), and the RAM-aware thread clamp — all unchanged and verified favorable previously.

**CI**: all required checks green — `rustfmt --check (pinned nightly)` pass, `cargo build --workspace (Linux)` pass, `Compile Benchmarks` pass, `Fuzz Build Guard (PR)` pass. `mergeStateStatus` = CLEAN.

(Posted as a comment, not `gh pr review --approve`, since Loom agents share one account and GitHub blocks self-review.)

Issue #555, opened 2026-06-24
security: verify & close remaining cycle-6 High findings (H1–H4) before public testnet


rjwalters:
Tracking issue so the next push toward public testnet closes the remaining cycle-6 Highs. The 5 Criticals (C1–C5) are now CLOSED (C1–C4 fixed in add_block; C5 = #553). The Highs were NOT re-verified this session — and the C1–C4 lesson is that audit items can already be fixed, so **verify current status against code first** (main-session security review — sub-agent security filter), then file/fix only the genuinely-open ones.

From audits/2026-06-11-cycle6.md:
- **H1** — demurrage + progressive fees are mempool-only policy, not enforced in block validation (add_block / consensus validity). A miner includes own/under-fee txs; wealthy clusters evade demurrage. Prerequisite: M1 (deterministic-input rework) so a naive fix doesn't create chain splits.
- **H2** — spender-chosen decoys defeat the demurrage clock and the cluster factor (fresh high-value background-tagged decoys → factor 1000 → demurrage 0). Needs a decoy-age / quantile / floor policy.
- **H3** — difficulty/emission state persisted in a SEPARATE LMDB txn from the block (node/mod.rs); a crash between them = permanent difficulty divergence ('one unlucky crash = permanent divergence'). Fix: fold emission-state update into the add_block write txn. **Small + high-value — likely do first.**
- **H4** — lottery candidate cap (MAX_LOTTERY_CANDIDATES=10_000, first-N by LMDB key order) excludes most UTXOs at scale and is grindable (vanity-low tx hash buys a permanent seat). Fix: deterministic per-block sampling from the lottery seed.

Also re-check M1/M2/M7 (fail-open consensus DB checks) and I4 (multi-input CLSAG balance looks structurally wrong) noted in the same audit. Consensus-breaking fixes ride #323. Ref: audits/2026-06-11-cycle6.md.

rjwalters:
## Cycle-6 High verification (main-session security review, 2026-06-23)

Re-verified the remaining cycle-6 Highs (+ coupled Mediums and I4) against current `main`. The 5 Criticals are confirmed closed (C1–C4 in `add_block`, C5 = #553). **Unlike C1–C4, none of the Highs were silently fixed — they are all still open.**

| Finding | Status | Evidence (current code) |
|---|---|---|
[...]

PR #571, opened 2026-06-27
security(M1/H1-prereq): make output_penalty deterministic integer-only

### LINKED ISSUE #570: security(M1/H1-prereq): eliminate f64 from output_penalty — last float on the fee-floor consensus path
## Summary
`FeeConfig::output_penalty` (`cluster-tax/src/fee_curve.rs:198-210`) is the **last and only floating-point operation on the per-tx minimum-fee path**. It computes `min(output_count, cap)^exponent` with `f64::powf`:

```rust
let effective_count = std::cmp::min(output_count as u32, self.output_count_cap) as f64;
let exponent = self.output_fee_exponent_scaled as f64 / OUTPUT_FEE_EXPONENT_SCALE as f64;
let penalty = effective_count.powf(exponent);
(penalty * OUTPUT_FEE_EXPONENT_SCALE as f64) as u64
```

Everything else feeding the fee floor is already integer: `ClusterFactorCurve::factor` (sigmoid + linear interpolation, `FACTOR_SCALE`/`SIGMOID_SCALE` fixed-point), `demurrage_charge` (u128, no float), and `compute_fee_with_dynamic_base_and_outputs` (saturating integer mul/div).

## Why this matters (prerequisite for H1)
**H1** (#555) — the progressive/demurrage fee floor is enforced only at mempool admission, not in block validation. The audited fix is to recompute `base_minimum_fee + demurrage_charge` per tx in `Ledger::add_block`. That fix **cannot** promote a float to consensus: `powf` is not bit-reproducible across platforms/compilers, so two honest validators could compute different fee floors and **fork**. This issue removes that blocker — it is the fee-curve analogue of **C5/#553** (which de-f64'd the live difficulty controller).

## Scope (narrow)
Replace the `powf` with deterministic integer math. The input domain is tiny: `effective_count ∈ [0, output_count_cap]` (default cap 10), exponent is a fixed-point rational `exponent_scaled / 1000`.

- For **integer exponents** (default quadratic = 2.0, and any `exponent_scaled % 1000 == 0`): `penalty = (count as u128).pow(k) * OUTPUT_FEE_EXPONENT_SCALE`. This is **bit-identical** to the current float output for every in-domain value (2²→4000, 5²→25000, 10²→100000, cap clamp unchanged) → **value-preserving, NOT consensus-breaking, lands independently of #323.**

## Decision required
Non-integer exponents (`exponent_scaled` not a multiple of 1000) are not the default and are not on any deployed/validated config. Pick one:
- **(A, recommended)** Constrain the consensus fee exponent to integers — validate/clamp `output_fee_exponent_scaled % 1000 == 0`; use `u128::pow`. Simplest, exact, matches the validated quadratic mechanism.
- **(B)** Implement a deterministic integer fixed-point `pow` for fractional exponents (more code, precision-policy ambiguity, no current consumer).

## Acceptance criteria
- [ ] `output_penalty` contains no `f64`/`powf`; deterministic integer-only.
- [ ] Bit-identical results to the current implementation for the default config across the full `[0, cap]` domain (add a test asserting the exact table: 1→1000, 2→4000, 5→25000, 10→100000, 20→100000).
- [ ] Exponent policy from the decision above documented + enforced (reject/clamp configs that would have used fractional powf, or implement the integer pow).
- [ ] `cargo test -p bth-cluster-tax` green; fuzz target `fuzz_cluster_tax_math` still passes (no panic/overflow).
- [ ] Note in the PR that this is value-preserving for default config → lands independently of the #323 reset; it unblocks (does not implement) H1.

## Out of scope
- H1 itself (recomputing the fee floor in `add_block`) — consensus-breaking, rides #323.
- f64 elsewhere in cluster-tax that is NOT on the fee-floor path (entropy_decay, monetary reporting/estimation, age_decay sims) — those are off the consensus path; do not touch here.

Tracking: prerequisite for **H1** in #555. Pattern precedent: **C5/#553**.

rjwalters:
## Summary
`FeeConfig::output_penalty` (`cluster-tax/src/fee_curve.rs`) was the **last and only floating-point operation on the per-tx minimum-fee path** — it computed `min(output_count, cap)^exponent` with `f64::powf`. `powf` is not bit-reproducible across platforms/compilers, so it could not be promoted to consensus without risking validator forks. This replaces it with deterministic integer-only math, removing the blocker for **H1 (#555)**.

## Changes
- `output_penalty` is now integer-only: `penalty = count^exp` via `u128::checked_pow`, staged in `u128` and saturated into `u64`. No `f64`/`powf`.
- **Exponent policy (decision A):** the consensus fee exponent must be an integer. `output_penalty` floors a fractional `output_fee_exponent_scaled` via integer division (defense-in-depth), and a new `FeeConfig::validate()` rejects fractional exponents up front so config loaders can surface the error.
- Tests added:
  - `test_output_penalty_default_table_value_preserving` — exact table across the full `[0, cap]` domain plus the cap clamp (`0→0, 1→1000, 2→4000, 5→25000, 10→100000, 20→100000, ...`), pinned to the previous `powf` output so the rewrite is provably value-preserving.
  - `test_output_penalty_saturates_without_panic` — adversarial config (exponent 64, cap 1e6) saturates to `u64::MAX` instead of panicking/overflowing.
  - `test_exponent_policy_validate_and_clamp` — integer exponents validate; fractional ones are rejected by `validate()` and clamped down by `output_penalty`.

## Value-preserving
For integer exponents the result is **bit-identical** to the previous `f64::powf` implementation across the entire in-domain `[0, cap]` for the default quadratic config (the old truncating `(x * 1000.0) as u64` was already a no-op rounding step). Because no observable fee value changes for the deployed/default config, this **lands independently of the #323 reset** — it is not consensus-breaking on its own.

## Scope
- This **unblocks (does not implement) H1** — recomputing the fee floor in `add_block` (consensus-breaking) still rides #323.
- f64 elsewhere in cluster-tax that is off the fee-floor path (entropy_decay, monetary reporting/estimation, age_decay simulations) is intentionally untouched.
- Precedent: **C5/#553** (de-f64'd the difficulty controller).

## Testing
- `cargo test -p bth-cluster-tax` — green (402 tests).
- `fuzz_cluster_tax_math` fuzz target still compiles; its invariants are unaffected (it exercises the integer fee/demurrage/emission paths, not `output_penalty`).

Closes #570

### DIFF
--- a/cluster-tax/src/fee_curve.rs
+++ b/cluster-tax/src/fee_curve.rs
@@ -219,8 +219,7 @@ impl FeeConfig {
     pub fn output_penalty(&self, output_count: usize) -> u64 {
         // Apply cap. u128 staging keeps `count^exp` from overflowing before the
         // final saturating narrow to u64.
-        let effective_count =
-            std::cmp::min(output_count as u32, self.output_count_cap) as u128;
+        let effective_count = std::cmp::min(output_count as u32, self.output_count_cap) as u128;
 
         // Integer exponent only (see "Exponent policy" above): floor the
         // fixed-point exponent to a whole number via integer division.
@@ -1143,7 +1142,8 @@ mod tests {
 
         // output_penalty clamps the fractional exponent DOWN to 2.0, so it
         // behaves as the floored integer exponent (defense-in-depth).
-        assert_eq!(fractional.output_penalty(3), 9_000); // 3^2 * 1000, not 3^2.5
+        assert_eq!(fractional.output_penalty(3), 9_000); // 3^2 * 1000, not
+                                                         // 3^2.5
     }
 
     #[test]
--- a/cluster-tax/src/fee_curve.rs
+++ b/cluster-tax/src/fee_curve.rs
@@ -188,25 +188,73 @@ impl FeeConfig {
     ///
     /// Returns the multiplier in OUTPUT_PENALTY_SCALE (1000 = 1x).
     ///
+    /// # Determinism (consensus path)
+    /// This is the per-tx minimum-fee path, so it must be **bit-reproducible**
+    /// across platforms/compilers — floating point (`f64::powf`) is not, so it
+    /// is forbidden here. The computation is integer-only.
+    ///
+    /// # Exponent policy
+    /// The consensus fee exponent MUST be an integer (i.e.
+    /// `output_fee_exponent_scaled` MUST be a multiple of
+    /// `OUTPUT_FEE_EXPONENT_SCALE`). A fractional exponent has no current
+    /// consumer and cannot be reproduced bit-identically across platforms, so
+    /// any fractional part is **clamped down** to the nearest integer exponent
+    /// here (integer division of `output_fee_exponent_scaled`). Callers should
+    /// validate configs with [`FeeConfig::validate`] before use; the clamp is a
+    /// defense-in-depth fallback, not a supported configuration mode.
+    ///
+    /// For integer exponents this is **bit-identical** to the previous
+    /// `f64::powf` implementation across the entire in-domain
+    /// `[0, output_count_cap]` (the product is exact, so the old truncating
+    /// `(x * 1000.0) as u64` was already a no-op rounding step).
+    ///
     /// # Examples
     /// With default config (exponent=2.0, cap=10):
+    /// - 0 outputs: 0^2 = 0x (0)
     /// - 1 output: 1^2 = 1x (1000)
     /// - 2 outputs: 2^2 = 4x (4000)
     /// - 5 outputs: 5^2 = 25x (25000)
     /// - 10 outputs: 10^2 = 100x (100000)
     /// - 20 outputs: 10^2 = 100x (capped at 100000)
     pub fn output_penalty(&self, output_count: usize) -> u64 {
-        // Apply cap
-        let effective_count = std::cmp::min(output_count as u32, self.output_count_cap) as f64;
+        // Apply cap. u128 staging keeps `count^exp` from overflowing before the
+        // final saturating narrow to u64.
+        let effective_count =
+            std::cmp::min(output_count as u32, self.output_count_cap) as u128;
 
-        // Compute exponent: exponent_scaled / 1000
-        let exponent = self.output_fee_exponent_scaled as f64 / OUTPUT_FEE_EXPONENT_SCALE as f64;
+        // Integer exponent only (see "Exponent policy" above): floor the
+        // fixed-point exponent to a whole number via integer division.
+        let exponent = self.output_fee_exponent_scaled / OUTPUT_FEE_EXPONENT_SCALE;
 
-        // Compute penalty: count^exponent
-        let penalty = effective_count.powf(exponent);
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Determinism-hardening of `FeeConfig::output_penalty` is correct, well-scoped, and the value-preservation claim holds up under independent verification.

## Verification

**Value-preservation (the load-bearing claim) — confirmed empirically.** I reimplemented both the old `f64::powf` path and the new integer path in a standalone harness and compared outputs across the full input domain `[0, 2000]` for three configs:
- default quadratic (exp=2, cap=10): **0 mismatches**
- linear (exp=1, cap=20): **0 mismatches**
- cubic (exp=3, cap=20): **0 mismatches**

The pinned table in `test_output_penalty_default_table_value_preserving` (`0→0, 1→1000, 2→4000, 5→25000, 10→100000, 20→100000 (cap clamp), 1000→100000`) matches the old `(count.powf(2.0) * 1000.0) as u64` exactly. Because the new path is exact integer math and reproduces the deployed/default config bit-for-bit, this is safe to land independently of the #323 reset, as claimed.

**Overflow/saturation safety — confirmed.** `count^exp` via `u128::checked_pow().unwrap_or(u128::MAX)`, then `saturating_mul(SCALE)`, then `.min(u64::MAX as u128) as u64`. No panic/overflow path: `test_output_penalty_saturates_without_panic` exercises exp=64 / cap=1e6 → `u64::MAX`, and the staging types (`u128` intermediate, `u32` exponent from `u32 / u32`) are sound.

**Exponent policy — coherent.** `validate()` rejects any `output_fee_exponent_scaled` not divisible by `OUTPUT_FEE_EXPONENT_SCALE`; `output_penalty` independently floors a fractional exponent via integer division as defense-in-depth, so there is no path where a fractional exponent silently yields a non-reproducible result. Both branches are covered by `test_exponent_policy_validate_and_clamp`.

**Scope discipline — confirmed.** Diff touches only `output_penalty` + the new `validate()`. The f64 in entropy_decay / monetary reporting / age_decay is untouched, consistent with the PR's stated scope (off the fee-floor path).

## Test Execution
- `cargo test -p bth-cluster-tax` — ✅ 402 passed, 0 failed (matches PR claim). New tests all green.
- Independent value-preservation harness (old powf vs new integer, [0,2000], 3 configs) — ✅ 0 mismatches.

## CI
- `cargo build --workspace (Linux)` — ✅ pass
- `rustfmt --check (pinned nightly)` — ✅ pass
- Merge state: **CLEAN**

## Fixed during evaluation
The initial push failed `rustfmt --check (pinned nightly)` on two formatting nits (a line-join in `output_penalty` and a `wrap_comments` wrap in the new test). I applied the pinned `nightly-2025-12-03` formatter (`cargo fmt -p bth-cluster-tax`, formatting-only, no behavior change) and pushed — CI is now green.

Non-blocking note: `validate()` is added as API but is not yet invoked by any config loader. That wiring belongs to the consensus fee-path work (H1 / #555), which this PR explicitly only unblocks rather than implements — so it is appropriately tracked there rather than here.

Issue #574, opened 2026-06-27
design: consensus-economic fee floor — enforce demurrage/progressive fees at block validation (H1+M1+H2)


rjwalters:
> **Type:** Architecture proposal (design awaiting human approval) — `loom:architect`. READ-ONLY: no source modified by this issue. Decompose into build issues only after human sign-off.
> **Parent:** cycle-6 Highs tracker #555. Rides the #323 coordinated testnet reset (consensus-breaking).

## Why one issue
H1, M1, and H2 from audit cycle 6 are **interlocked** — they cannot be designed independently without baking a known-gameable rule into consensus. This issue asks the architect for ONE coherent design covering all three, which a human approves before any builder touches the consensus path. The cleanly-mechanical follow-ons (H3 #558, H4 #572, H1's f64 prereq #570) are already merged; what's left is genuinely economic-policy.

### H1 — the fee floor is mempool-only policy, not a consensus rule
`Ledger::add_block` (`botho/src/ledger/store.rs`) enforces no per-tx fee floor: it checks double-spend, ring signatures, ring membership, lottery accounting — but never recomputes the progressive minimum fee or the demurrage charge. Those live solely at mempool admission (`botho/src/mempool.rs:686-714`). A miner including own/under-fee txs is accepted by every validator → fee-free self-mining and demurrage evasion (the load-bearing Gini component).

### M1 — the mempool fee base is node-local, so it can't be promoted as-is
The mempool floor multiplies in `dynamic_base = self.dynamic_fee.compute_base(self.at_min_block_time)` (`mempool.rs:682`) — **node-local congestion state**. Two honest validators with different mempools compute different floors. Promoting this to consensus forks the chain. (`#570` already removed the *f64* half of this prerequisite; the *congestion-state* half remains.)

### H2 — the demurrage clock and cluster factor are spender-gameable
`ring_elapsed_centroid` (`cluster-tax/src/demurrage.rs:108`) is the value-weighted **mean** of spender-selected ring-member ages; the cluster factor derives from spender-authored output tags gated only by a cosine-similarity (≥0.7) plausibility test. Fresh, high-value, background-tagged decoys dilute both → factor 1000 → demurrage 0, fees at 1x. `validate_cluster_tag_inheritance` exists but is **not wired** into the node. Enforcing demurrage at consensus (H1) while this stands bakes a gameable rule into consensus.

## Questions the design must answer
[...]

PR #580, opened 2026-06-28
feat(ledger): reject cluster-tag-inflated outputs at block validation

### LINKED ISSUE #576: build(H2-B3): wire validate_cluster_tag_inheritance into add_block
Decomposition item **B3** from design #574 (H2 hardening). **BLOCKED pending design approval of #574.**

`validate_cluster_tag_inheritance` (transaction/core/src/validation/validate.rs:463) is already integer/BTreeMap-clean but is NOT called by the node. Wire it into Ledger::add_block so blocks with tag-inflated outputs are rejected at consensus, not just at mempool. Mechanical wiring of an existing validator.

Determinism contract: validator is already pure-integer; confirm no node-local state enters via the call site. Rides #323. Unblock when #574 design is approved.

rjwalters:
## Summary

Decomposition item **H2-B3** of design #574 (H2 hardening). Wires the existing
conservation-of-mass validator `transaction_core::validation::validate_cluster_tag_inheritance`
into the block-validation path (`Ledger::add_block`) so a block containing a
transaction with **tag-inflated outputs** (outputs claiming more cluster-tag
mass than their inputs can supply) is rejected at consensus, not merely at
mempool admission. Blocks bypass the mempool, so this closes a real consensus
gap: the validator existed but the running node never called it.

This is a **mechanical consensus-wiring** change. It does not alter the
validator's logic, the demurrage quantile (B1), or the cluster-factor floor (B2).

## Changes

- New pure helper `check_cluster_tag_inheritance(input_tagged, outputs)` in
  `botho/src/ledger/store.rs` — ports the upstream integer/`BTreeMap`
  conservation logic to the node's `(ClusterTagVector, value)` /`TxOutput`
  representation (the upstream validator takes transaction-core `TxOut`).
- New method `Ledger::verify_cluster_tag_inheritance(tx)` — resolves each
  transaction's ring members against the committed UTXO set, then applies the
  helper.
- New **C6** guard in `add_block_inner`, immediately after the C3 ring-member
  check, that runs the guard over every transaction and returns
  `LedgerError::InvalidBlock` on inflation.
- Two tests (see below).

## Consensus determinism (fork safety)

The check is a **pure function of the transaction plus committed UTXO state**.
Stated explicitly because proposer and validator must agree:

- **Integer/`BTreeMap` only.** All mass accumulation uses `BTreeMap` (sorted,
  deterministic iteration order) and integer arithmetic. No `HashMap`/`HashSet`
  iteration order, no `f64`, no node-local state enters the computation.
- **Overflow-safe.** Accumulation is in `u128` (vs. the upstream `u64`), since
  summed ring-member mass can exceed `u64::MAX`; a panic on the consensus path
  would halt the node.
- **`decay_rate = 0` (output mass ≤ input mass).** This is the most permissive
  conservation bound. Decay credit legitimately ranges from 0 (no entropy proof
  / proof-required height) up to the base rate, so any nonzero rate could
  false-reject a valid zero-decay-credit block and halt consensus. The guard's
  sole job is to reject inflation, which `decay_rate = 0` captures exactly.
- **Ring members as the input set.** The real input is hidden among decoys, so
  the resolved ring members are the only node-agnostic input set. Real inputs
  are a subset of the ring, so ring-member mass is an upper bound on true input
  mass — the guard never false-rejects a valid block.

**Consensus-breaking; rides the #323 reset.**

| Validator's exact integer logic invoked in block-validation path | ✅ | C6 loop in `add_block_inner` calls `verify_cluster_tag_inheritance` → `check_cluster_tag_inheritance` |
| Block with a tag-inflated-output tx is rejected | ✅ | Both tests assert `InvalidBlock` on inflated outputs |
| Legitimate (decayed-inheritance) block still passes | ✅ | Both tests assert `Ok` on conserving/decayed outputs |
| Determinism stated; integer/BTreeMap only | ✅ | See section above; `u128`/`BTreeMap`, no f64/HashMap/node-local state |
| `cargo build` clean; `cargo test` green for affected crates | ✅ | See Test Plan |

## Test Plan

- `check_cluster_tag_inheritance_rejects_inflation_accepts_decay` — unit test of
  the pure core: decayed inheritance (output mass ≤ input mass) accepted,
  2×-inflated output rejected as `InvalidBlock`.
- `verify_cluster_tag_inheritance_resolves_ring_and_rejects_inflation` —
  end-to-end through the ledger: inserts UTXOs, builds a tx whose ring
  references them, and confirms the guard resolves ring members and rejects an
  inflated tx while accepting a conserving one.
- `cargo test -p botho -p bth-transaction-core`: all library unit tests
  (incl. the 2 new) and all integration suites pass; transaction-core 99 passed.
- Pre-existing/unrelated: 2 doctest failures in
  `botho/src/network/privacy/capacity.rs` (`RelayCapacity::measure` examples
  that don't import `RelayCapacityExt`) — present identically on `origin/main`,
  untouched by this PR (diff is `store.rs` only).

Closes #576

### DIFF
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -1,12 +1,12 @@
 use bth_account_keys::{AccountKey, PublicAddress};
 use bth_cluster_tax::{LotteryCandidate, LotteryDrawConfig, TagVector};
-use bth_transaction_types::{Network, TAG_WEIGHT_SCALE};
+use bth_transaction_types::{ClusterTagVector, Network, TAG_WEIGHT_SCALE};
 use heed::{
     types::{Bytes, U64},
     Database, Env, EnvOpenOptions, RwTxn,
 };
 use rand::Rng;
-use std::{fs, path::Path};
+use std::{collections::BTreeMap, fs, path::Path};
 use tracing::{debug, info, warn};
 
 use super::{ChainState, LedgerError};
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Thorough consensus-critical review of the cluster-tag inflation guard (H2-B3, wiring `check_cluster_tag_inheritance` into `add_block_inner` C6). All make-or-break points hold.

## 1. `decay_rate = 0` consensus bound — SOUND
The rule `output_mass <= input_mass + tolerance` per cluster is the most permissive conservation-of-mass bound and is the correct conservative consensus rule:
- **Rejects inflation**: any output claiming more cluster mass than the inputs supply is rejected.
- **Never false-rejects an honest block**: honest decay only *reduces* mass (`decayed <= input`), so an honestly built block always satisfies `output <= input` regardless of how much decay its wallet applied. `decay_rate = 0` ⇒ `decay_factor = TAG_WEIGHT_SCALE` ⇒ the largest `expected`, i.e. the most permissive ceiling.
- The builder's rationale is correct: decay credit legitimately ranges 0..base depending on the entropy proof, so pinning any nonzero rate at consensus could false-reject a valid zero-decay-credit block and **halt consensus**. Pinning 0 avoids that failure direction entirely.
- The accepted residual gap (a spender who *should* decay can claim up to `input_mass`) is intentional and correct — enforcing the decay *reduction* deterministically at consensus is impossible without the entropy proof; that is economic policy, not a consensus invariant. The guard's job is to stop mass *creation*, which it does.

## 2. Determinism / fork safety — HOLDS
- Integer/`u128`/`BTreeMap` only. No `f64`, no `HashMap`/`HashSet` iteration order, no node-local state.
- **Overflow-safe / no panic path**: `amount: u64`, `weight: u32 (<= 1e6)`, `cluster_id: u64`. Per-entry mass = `value·weight/SCALE <= value <= u64::MAX`; sums accumulate in `u128` (max realistic magnitude ~1e33, far below 3.4e38). This is strictly safer than the upstream `u64` accumulation, whose cross-entry/ring sum could overflow. No `unwrap` on arithmetic; the only `?` is on `get_utxo_by_target_key` (DB error propagated, not panicked).
- Verdict is order-independent (commutative integer sums into a sorted `BTreeMap`), computed purely from the block + the committed *pre-block* UTXO state, so proposer and validator reach an identical accept/reject decision.

## 3. Ring-members-as-input-set — liveness reasoning HOLDS; one documented limitation
Real inputs ⊆ ring ⇒ ring-member mass is an upper bound on true input mass ⇒ never false-rejects. Ring resolution via `get_utxo_by_target_key` is a deterministic DB lookup on unique one-time target keys. I did flag that the bound is the *sum* over all ring members rather than the real input, which permits bounded, decoy-sourced inflation if an attacker uses real cluster-tagged UTXOs as decoys. Crucially this does **not** break the design-named vector ("fresh background-tagged decoys" contribute zero to any tagged cluster's ceiling), it is bounded by ring size, and it is strictly better than the pre-PR status quo. Tracked as a future-tightening follow-up in #581 — not a blocker, especially as B3 lands as part of the coupled B1/B2/B3 H2 set.

## 4. Scope — clean
Only `store.rs` changed; only the C6 guard added. No B1 demurrage-quantile or B2 cluster-factor-floor changes. No scope creep.

## 5. Tests & CI
- `cargo test -p botho --lib cluster_tag_inheritance` → both new tests pass.
- `cargo test -p bth-transaction-core` → 99 passed.
- Tests exercise inflated→reject and conserving→accept at both the pure-core and ledger-resolution (ring-resolving) levels.
- All GitHub CI checks **pass**; `mergeStateStatus` is **CLEAN**.
- Confirmed the 2 `network/privacy/capacity.rs` doctest failures are pre-existing and unrelated: the PR diff touches `store.rs` only, so those doctests are byte-identical to `origin/main`, and CI has no doctest gate (build/fuzz/rustfmt only).

Solid, conservative, fork-safe wiring of an existing validator. Approving.

PR #594, opened 2026-06-30
test(consensus): staggered-start 2-node solo-latch no-fork regression

### LINKED ISSUE #593: test(consensus): integration regression for #427 staggered-start 2-node solo-latch fork (pins #428/#433 gates)
## Summary

Add an **integration regression test** that reproduces the staggered-start 2-node fork documented in #427 (Finding 3) and asserts the `min_peers` participation gate (#428) and transitional-solo guard (#433) prevent it. This closes out **Proposed Direction #2 of #427** ("Fix the solo-latch startup race") at the test level — the *fix* already landed; what is missing is an end-to-end test that pins the exact failure mode so it can never silently regress.

## Background / current state

#427's empirical investigation found that the real 2-minter fragility was **not** an inability to converge, but a **solo-latch startup race**: a freshly-booted minter that mints slot 1 *before* peering completes latches into solo (1-of-1) mode and builds a divergent chain, forking once peers connect. Simultaneous start at N=2/3/4 converged cleanly; only a **staggered** N=2 start forked at every height.

Since that investigation, the mechanism fix has been implemented:

- **#428** — participation gate: `min_peers` in `botho/src/consensus/service.rs` (`set_min_peers`, the propose/externalize gate around lines 386–414 and 813–825). A node that expects peers (`min_peers >= 1`) withholds propose/externalize while `connected_peers < min_peers`.
- **#433** — transitional-solo guard: `is_transitional_solo` (service.rs ~778–842) blocks the solo direct-externalize path when a peer is connected but the quorum still holds 1-of-1.

These are covered by **unit** tests of the guard predicates in isolation:
- `participation_gate_blocks_solo_block_until_peers_connected` (service.rs ~1666)
- `transitional_solo_does_not_directly_externalize` (service.rs ~1726)

**Gap:** there is no *integration* test that drives the #427 staggered-start scenario end-to-end (one node mints/externalizes a slot before its peer connects, then the peer connects) and asserts the two nodes do **not** fork. The existing two-node integration harness (`run_two_services` ~2065, `two_minters_converge_on_single_value_no_fork` ~2134) only exercises the **simultaneous-start** path — the one #427 showed never forked even before the fix.

## Acceptance criteria

- [ ] Add an integration test (alongside `two_minters_converge_on_single_value_no_fork` in `botho/src/consensus/service.rs`, reusing the `run_two_services` / `peered_service` harness) that:
  - [ ] Configures two minters with `min_peers >= 1` (the production gate, not the dev `min_peers == 0` path).
  - [ ] Reproduces a **staggered** start: node A reaches the point where it *would* mint slot 1 while node B is not yet connected.
  - [ ] Asserts node A **withholds** the pre-quorum solo block (participation gate / transitional-solo guard fires) rather than externalizing a divergent solo chain.
  - [ ] Connects node B, lets both advance, and asserts **no fork**: both nodes externalize the same value at each common height (the existing no-fork assertion style).
- [ ] Add a companion negative-control assertion (or doc comment) making explicit that with `min_peers == 0` (genuine single-node) the node still mints solo — i.e. the gate does not regress dev/genesis solo minting.
- [ ] Test is deterministic (no sleeps/wall-clock races) and runs in the existing `cargo test` consensus suite.

## Out of scope

- #427 Direction #3 (explicit round-leader / VRF single-proposer election) — larger design, separate issue.
- #427 Direction #1 (warn/refuse below 4 nodes) — tracked separately (warn-below-4 = #509).
- #421 slot/height drift — separately tracked.

## References

- Parent investigation: #427 (Finding 3 + Proposed Direction #2).
- Implemented fix this test pins: #428 (participation gate), #433 (transitional-solo guard).
- Code: `botho/src/consensus/service.rs` — `min_peers`/`set_min_peers`, `is_transitional_solo`, `run_two_services`, `two_minters_converge_on_single_value_no_fork`.

_Filed from a `/loom:sweep` run on the recommendation of the Champion review of #427._

rjwalters:
## Summary

Adds the missing **integration regression test** for #427 Finding 3 (the staggered-start 2-node solo-latch fork), pinning the #428 participation gate and #433 transitional-solo guard end-to-end. The *fix* already landed; this closes the test gap so the exact failure mode can never silently regress.

The existing two-node harness (`two_minters_converge_on_single_value_no_fork`) only exercised the **simultaneous-start** path — the one #427 showed never forked even pre-fix. There was no test driving the **staggered** start (one node mints a slot before its peer connects) end-to-end.

## Changes

- New test `staggered_start_two_minters_do_not_solo_latch_fork` in `botho/src/consensus/service.rs`, alongside `two_minters_converge_on_single_value_no_fork`, reusing the existing `run_two_services` / `submit_and_register` harness (no new harness).
- Lifecycle driven:
  1. Node A boots holding the transient 1-of-1 startup quorum, `min_peers == 1`, `connected_peers == 0`; mines slot 1; the **participation gate (#428)** withholds the pre-quorum solo block.
  2. Peer connects (`connected_peers == 1`) but quorum still 1-of-1; the **transitional-solo guard (#433)** still withholds.
  3. A reconfigures to 2-of-2, B joins federated; both externalize the **same value** at the common height — no fork. Withheld coinbase is asserted to stay queued (never dropped).
- Negative control: a `min_peers == 0` node still mints solo (no dev/genesis regression).

| Two minters with `min_peers >= 1` (production gate, not dev `min_peers == 0`) | ✅ | Both A and B `set_min_peers(1)` |
| Staggered start: A would mint slot 1 while B unconnected | ✅ | A boots solo 1-of-1, `connected_peers == 0`, mines slot-1 coinbase |
| A withholds the pre-quorum solo block | ✅ | Asserts `externalized.is_none()` under participation gate AND transitional-solo guard |
| Connect B, advance, assert no fork (same value at common height) | ✅ | `run_two_services` then `assert_eq!(a_vals, b_vals)`, one coinbase, chosen ∈ {a,b} |
| Negative control: `min_peers == 0` still mints solo | ✅ | Dedicated block asserts `externalized.is_some()` |
| Deterministic (no sleeps/wall-clock) | ✅ | Drives `propose_pending_values` / `run_two_services` directly; no `sleep` |
| Runs in existing `cargo test` consensus suite | ✅ | `cargo test -p botho --lib consensus::` |

## Test Plan

- `cargo test -p botho --lib consensus::service::tests::staggered_start_two_minters_do_not_solo_latch_fork` → 1 passed.
- `cargo test -p botho --lib consensus::` → 74 passed, 0 failed (no regressions).
- `cargo fmt -p botho -- --check` → clean.

Closes #593

### DIFF
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -2181,6 +2181,181 @@ mod tests {
         );
     }
 
+    /// Issue #593 regression (pins the #428/#433 gates against #427 Finding 3):
+    /// the STAGGERED-start 2-node solo-latch fork must not occur end-to-end.
+    ///
+    /// #427's empirical investigation found the real 2-minter fragility was a
+    /// solo-latch *startup race*: a freshly-booted minter that mints slot 1
+    /// *before* peering completes latches into solo (1-of-1) mode and builds a
+    /// divergent chain, forking once its peer connects. Simultaneous start
+    /// (the existing `two_minters_converge_on_single_value_no_fork` path) never
+    /// forked even pre-fix; only a *staggered* start did. This test drives that
+    /// staggered scenario through the production `min_peers >= 1` gate and
+    /// asserts no fork.
+    ///
+    /// Lifecycle modelled (matches `commands::run` startup):
+    ///   1. Node A boots holding the transient startup solo (1-of-1) quorum,
+    ///      declares it expects a peer (`min_peers == 1`), but B is not yet
+    ///      connected (`connected_peers == 0`). A mines slot 1 and would,
+    ///      pre-fix, directly solo-externalize a divergent block. The
+    ///      participation gate (#428) WITHHOLDS it.
+    ///   2. B connects (`connected_peers == 1`) but A's SCP quorum is still
+    ///      1-of-1 for one more moment. A would now pass the participation
+    ///      gate, but the transitional-solo guard (#433) WITHHOLDS the solo
+    ///      block until the quorum reconfigures.
+    ///   3. A reconfigures to the 2-of-2 quorum (the run loop does this on the
+    ///      peer event) and B joins as a federated 2-of-2 node. Both advance
+    ///      and externalize the SAME value at the common height — no fork.
+    ///
+    /// The withheld slot-1 coinbase is never dropped: it stays queued in
+    /// `pending_values` and is proposed via federated SCP once the quorum
+    /// forms.
+    #[test]
+    fn staggered_start_two_minters_do_not_solo_latch_fork() {
+        let cfg = ConsensusConfig::fixed_timing(0);
+        let prev = [0u8; 32]; // shared genesis tip
+
+        // --- Negative control (no regression to dev/genesis solo minting) ---
+        // A genuine single-node network (`min_peers == 0`) must still mint solo:
+        // the participation gate must not gate a node that declared no peers.
+        {
+            let mut solo = ConsensusService::new(
+                node(1),
+                QuorumSet::new(1, vec![QuorumSetMember::Node(node(1))]),
+                cfg.clone(),
+                genesis_chain_state(),
+            );
+            assert_eq!(
+                solo.min_peers, 0,
+                "genuine single-node has no peer expectation"
+            );
+            assert!(solo.is_solo_mode());
+            let solo_tx = intrinsic_valid_minting_tx(7, prev, 1);
+            let solo_bytes = bincode::serialize(&solo_tx).expect("serialize");
+            solo.submit_minting_tx(solo_tx.hash(), solo_tx.pow_priority(), solo_bytes);
+            solo.propose_pending_values();
+            assert!(
+                solo.externalized.is_some(),
+                "min_peers==0 node must still mint solo (gate must not regress dev/genesis)"
+            );
+        }
+
+        // --- Staggered start of the real (peered) network ---
+        // Node A boots holding the transient startup solo quorum and expects a
+        // peer, but B has not connected yet.
+        let mut a = ConsensusService::new(
+            node(1),
+            QuorumSet::new(1, vec![QuorumSetMember::Node(node(1))]),
+            cfg.clone(),
+            genesis_chain_state(),
+        );
+        a.set_min_peers(1);
+        a.set_connected_peers(0);
+        assert!(
+            a.is_solo_mode(),
+            "A boots holding the transient 1-of-1 startup quorum"
+        );
+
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Strong, load-bearing regression test for the #427 staggered-start solo-latch fork. Pins the #428 participation gate and #433 transitional-solo guard end-to-end.

## Load-bearing verification (the crucial question)

I empirically falsified the tautology concern. After temporarily neutering BOTH gates in `propose_pending_values` (`if false && self.should_propose_this_round()` and `if false && self.is_transitional_solo()`), the test **FAILS** at the step-1 withhold assertion:

```
SOLO-LATCH (#427/#428): A must withhold its pre-quorum slot-1 block, not externalize a divergent solo chain
```

The assertions are genuinely load-bearing, not a tautology. The design is airtight: the negative-control block drives the *same* `propose_pending_values` call with `min_peers==0` and asserts it DOES solo-externalize (`externalized.is_some()`), while steps 1/2 hit the identical code path gated and assert it does NOT (`externalized.is_none()`). Same path, only the gate differs — so the withhold assertions can only pass because the gates fire.

## Acceptance criteria (#593)

- Two minters with `min_peers >= 1` production gate (not dev `min_peers == 0`): ✅
- Staggered start — A would mint slot 1 while B unconnected (`connected_peers == 0`): ✅
- A withholds the pre-quorum solo block under BOTH #428 and #433, withheld coinbase asserted to stay queued in `pending_values` (never dropped): ✅
- Connect B, advance via `run_two_services`, assert no fork (`assert_eq!(a_vals, b_vals)`, exactly one minting tx, chosen ∈ {a,b}): ✅
- Negative control: `min_peers == 0` still mints solo: ✅
- Deterministic — `fixed_timing(0)`, bounded `run_two_services(.., 2000)`, no sleeps/wall-clock: ✅

## Other checks

- **Test-only**: single file, 175 additions, 0 deletions, entirely within `mod tests`. No production code touched.
- **Reuses harness**: `run_two_services` / `submit_and_register` / `intrinsic_valid_minting_tx` / `genesis_chain_state` / `recommended_quorum` — no duplicated harness.
- **Closes #593**: present in PR body.

## Test Execution

- `cargo test -p botho --lib consensus::service::tests::staggered_start_two_minters_do_not_solo_latch_fork` → ✅ 1 passed (local, worktree).
- Gate-disabled falsification → ❌ fails as expected, then reverted clean.

## CI

All checks pass: `cargo build --workspace (Linux)`, `rustfmt --check`, `Compile Benchmarks`, `Fuzz Build Guard (PR)`, `Node Benchmarks`. Merge state: CLEAN.

PR #595, opened 2026-06-30
feat(sim): empirical-gate harness for #577 — decoy-quantile demurrage sweep + ring_elapsed_quantile kernel

rjwalters:
## Empirical gate for #577 (H2-B1) — does NOT close #577

#577's consensus wiring stays **blocked under #323**. This PR delivers the *empirical-gate artifact* the issue requires ("needs sim results + human sign-off; do NOT auto-build") plus the new kernel as a **pure, unwired** function — value-preserving and dead until B1 wires it post-reset, the same pattern as #570's f64 prereq.

### What's here
- `cluster-tax/src/demurrage.rs` — `ring_elapsed_quantile(members, current_height, quantile_bps)`: an integer nearest-rank order-statistic over ring-member **ages only** (value deliberately unread — value-weighting is exactly the H2 dilution vector). `q=10000`→max, `7500`→p75. Pure, deterministic, consensus-safe. Unit-tested (incl. the H2 attack case the mean kernel fails).
- `cluster-tax/src/simulation/decoy_quantile_sweep.rs` + `cluster-tax-sim decoy-quantile-sweep` subcommand — adversarial decoy sweep measuring, per candidate kernel: final Gini, Δgini vs no-demurrage baseline, adversary age-dilution ratio, and honest #314 over-charge ratio.

### Sweep result (default population: 100 poor / 10 honest whales / 10 adversary whales, ring 11, ~25yr, 200bps)
```
| kernel          | final_gini | Δgini   | adv_dilution | honest_overcharge | passes |
|-----------------|------------|---------|--------------|-------------------|--------|
| centroid (mean) |     0.6568 | +0.1716 |       0.0909 |            0.9988 |   no   |
| quantile@p75    |     0.6748 | +0.1535 |       0.0000 |            1.0232 |   no   |
| quantile@p90    |     0.6740 | +0.1544 |       0.0000 |            1.0303 |   no   |
| quantile@max    |     0.4822 | +0.3462 |       1.0000 |            1.0386 |  YES   |
```
- `adv_dilution` = adversary realized charge / charge on TRUE holding age. 1.0 = attack fully blocked; →0 = fresh-decoy attack succeeds.
- `honest_overcharge` = honest realized charge / true accrual (#314 re-check). Must stay ≈1.0.
[...]

Issue #577, opened 2026-06-27
build(H2-B1): ring_elapsed_quantile kernel — replace value-weighted mean (EMPIRICAL-GATED)


rjwalters:
Decomposition item **B1** from design #574 (H2 hardening). **BLOCKED on (a) #574 design approval AND (b) empirical simulation.**

Replace `ring_elapsed_centroid` (value-weighted MEAN, cluster-tax/src/demurrage.rs:108) with a high-quantile/max of ring-member age so spender-chosen fresh decoys can't dilute the demurrage clock.

EMPIRICAL GATE (cannot be a pure code decision): the quantile parameter (e.g. p75 vs p90 vs max) must be chosen against the Gini/decoy simulation, re-confirming Δgini > 0.05 at miner-viable emission and re-running the #314 spend-time-accrual validation against the new kernel. Do NOT auto-build — needs sim results + human sign-off.

Determinism contract: integer-only quantile over public ring ages; pure fn of tx ring members. Rides #323.

rjwalters:
## ✅ Empirical gate SATISFIED (passed-with-condition) — 2026-06-30

The empirical gate this issue requires ("the quantile parameter must be chosen against the Gini/decoy simulation … Do NOT auto-build — needs sim results + human sign-off") has been run and signed off.

**Harness:** PR #595 — a decoy-adversary quantile sweep in the existing `cluster-tax/src/simulation/` framework (NOT a browser/E2E test — this is integer consensus-math), plus the new pure `ring_elapsed_quantile` kernel (unwired, value-preserving until B1 wires it).

**Result (default population, ring 11, ~25yr, 200bps):**
```
| kernel          | final_gini | Δgini   | adv_dilution | honest_overcharge | passes |
| centroid (mean) |     0.6568 | +0.1716 |       0.0909 |            0.9988 |   no   |
| quantile@p75    |     0.6748 | +0.1535 |       0.0000 |            1.0232 |   no   |
| quantile@p90    |     0.6740 | +0.1544 |       0.0000 |            1.0303 |   no   |
| quantile@max    |     0.4822 | +0.3462 |       1.0000 |            1.0386 |  YES   |
```

**Decision — adopt `ring_elapsed_quantile@max` (quantile_bps=10000) for B1:**
[...]

Issue #314, opened 2026-06-11
Re-validate Gini mechanism with spend-time demurrage accrual model


rjwalters:
## Context

The cluster-tilted redistribution mechanism (docs/design/cluster-tilted-redistribution.md) was validated by adversarial simulation (`cluster-tax-sim lottery-experiment`), which modeled demurrage as a **continuous daily charge on balances**. The implementation (cluster-tax/src/demurrage.rs, mempool enforcement) charges **accrual at spend**: `charge = value × rate × (factor−1)/(max−1) × elapsed/blocks_per_year`, where elapsed is the value-weighted centroid of public ring-member creation heights.

These are economically identical for coins that eventually move (total paid is invariant to churn frequency — unit-tested), but diverge for **permanently parked coins**: under spend-time accrual a coin that never moves pays nothing until spent, suffering only emission dilution (~1.25%/yr at the f=50% cap).

The emission-fraction sweep showed demurrage is **load-bearing** for the Δgini > 0.05 criterion at miner-viable emission fractions (experiments/ANALYSIS.md), so this divergence must be measured, not assumed away.

## Task

Extend the `lottery-experiment` harness to model demurrage as implemented:

1. **Spend-time accrual**: charge the spender's elapsed-since-creation at each transaction (and at churn events), instead of the daily balance tick
2. **Ring-centroid anchor**: optionally model the value-weighted ring centroid (including the large-fresh-decoy dilution gaming vector) rather than exact per-UTXO age
3. **New adversary — PermanentParker**: a whale that never transacts, escaping demurrage entirely; measure its wealth-share trajectory under emission dilution alone
4. Re-run the J scenario (cluster-tilted payout + emission + demurrage) at f = 25/50% over 5 years, honest + gamed

## Acceptance Criteria

- [ ] Δgini vs baseline > 0.05 confirmed (or refuted) in the gamed equilibrium with spend-time accrual
- [ ] PermanentParker wealth-share trajectory documented; if parked wealth materially breaks the criterion, evaluate countermeasures (e.g., demurrage catch-up cap, lottery eligibility interactions)
- [ ] Results added to experiments/ANALYSIS.md; design doc Open Question 0 resolved

## References

- docs/design/cluster-tilted-redistribution.md (Open Question 0: marked REQUIRED before mainnet)
- experiments/results/gini_experiment_5yr_f{25,50,100}.txt
- cluster-tax/src/demurrage.rs (implementation + churn-invariance tests)

[...]

PR #597, opened 2026-07-01
feat(wallet): age-similar decoy selection — #314-safe under ring_elapsed_quantile@max

### LINKED ISSUE #596: wallet: enforce age-similar decoy selection — consensus-economic dependency of #577-B1 max demurrage kernel
## Context
The #577 empirical gate (PR #595, decoy-quantile sweep) selected the **`max` order-statistic** as the demurrage-clock kernel (H2-B1): it's the only candidate that fully neutralizes the fresh-high-value-decoy age-dilution attack (adv_dilution 1.0) while delivering Δgini +0.35. See PR #595 for the full sweep.

## Why this issue exists (coupled requirement)
`max` makes the **wallet's decoy-selection policy consensus-economically load-bearing**. Because the demurrage clock now reads the *oldest* ring member's age, an honest spender whose ring contains an unusually old decoy is **over-charged** relative to their true holding time. Empirically (PR #595 sensitivity table):

| honest decoy age-spread | honest over-charge | within #314 tol (≤15%) |
|---|---|---|
| ±5% | 1.039 | ✅ |
| ±10% | 1.077 | ✅ |
| ±20% | 1.158 | ❌ |
| ±35% | 1.277 | ❌ |
| ±50% | 1.388 | ❌ |

So `max` is only #314-safe if honest wallets select **age-similar decoys** (≤ ~±10% spread).

## Asks
- [ ] Define and document the wallet decoy-selection age-similarity policy (age-bucketed / gamma-style recent-biased selection) that keeps honest ring age-spread within the safe band. Note: a recent-biased (Monero-style gamma) decoy distribution likely already satisfies this — quantify it against the sweep.
- [ ] Add the age-similarity constraint to the wallet decoy selector (`botho/src/decoy_selection.rs` and/or the web-wallet path) and a test asserting realized honest over-charge under `max` stays within the #314 bound.
- [ ] Cross-check against the privacy tradeoff: `cluster-tax/src/simulation/constrained_analysis.rs` already models age-similarity vs ring-anonymity — confirm the chosen band doesn't unacceptably degrade effective anonymity bits.
- [ ] **Blocking relationship:** #577-B1 must NOT enforce `ring_elapsed_quantile@max` in consensus (add_block) until this policy is in place, or honest spenders get over-charged on the live chain.

## References
- Empirical gate + sweep: PR #595, issue #577.
- Design parent: #574 (H2 hardening). Privacy/age-similarity model: `cluster-tax/src/simulation/constrained_analysis.rs`.

rjwalters:
## Closes #596 — age-similar decoy selection (dependency of #577-B1)

Makes the wallet decoy selector #314-safe under the `ring_elapsed_quantile@max` demurrage kernel chosen by the #577 empirical gate (PR #595). Wallet-side policy + tests only — does **NOT** wire the kernel into consensus (that's #577-B1, blocked under #323).

### The problem
`@max` reads the OLDEST ring member's age as the demurrage clock. The existing gamma/OSPEAD selector (`decoy_selection.rs`) samples decoy ages from the *global* spend distribution — unbounded relative to the real input — so an honest spender can draw a much-older decoy and be over-charged demurrage.

### The fix
- `AGE_SIMILARITY_SPREAD_BPS = 1000` (±10%): bound every selected decoy's age to `age_similarity_band(real_age) = [real·0.9, real·1.1]`, applying gamma/OSPEAD **within** the band (realistic within-band distribution preserved). Fallback to nearest-age selection when the band is empty.
- **Provable #314 safety:** because `@max` reads the single oldest member, realized honest over-charge ≤ `max_ring_age / real_age` ≤ **1.10** by construction — strictly inside the #314 bound (≤1.15). Matches PR #595's empirical sensitivity (±10% → 1.077).

### Chosen band / measured numbers
| metric | value |
|---|---|
| age band | ±10% relative to real input age |
| honest over-charge under @max | ≤ 1.10 (provable cap); 1.077 empirical |
| #314 tolerance | ≤ 1.15 ✅ |

### Privacy cross-check (via `constrained_analysis`)
| adversary | constrained | unconstrained | reduction |
|---|---|---|---|
| Naive | 3.459 | 3.459 | +0.000 |
| Age-Heuristic | 3.437 | 3.401 | −0.036 |
| Cluster-Fingerprint | 0.036 | 0.045 | +0.010 |
| Combined | 0.045 | 0.057 | +0.012 |

Anonymity cost is negligible: ≤0.036 bits vs an age-heuristic adversary, neutral/slightly-positive vs the stronger cluster-fingerprint and combined adversaries (which already dominate at ~0.045 bits — a pre-existing cluster-tag-leakage property, not caused here).

### Verification
- `cargo test -p botho --lib decoy_selection` → 21 passed (3 new: band bounds, honest-overcharge-under-max, in-band selection).
- `cargo test -p bth-cluster-tax` privacy-cost test green. `cargo fmt` clean.

### DIFF
--- a/botho/src/decoy_selection.rs
+++ b/botho/src/decoy_selection.rs
@@ -232,6 +232,58 @@ const SPEND_HISTORY_SIZE: usize = 10_000;
 const DEFAULT_GAMMA_SHAPE: f64 = 19.28;
 const DEFAULT_GAMMA_SCALE_DAYS: f64 = 1.61;
 
+/// Age-similarity policy: maximum relative spread, in basis points, of a
+/// selected decoy's age around the real input's age (1000 bps = ±10%).
+///
+/// # Why this exists (issue #596, consensus-economic dependency of #577-B1)
+///
+/// The #577 empirical gate (PR #595) selected the `ring_elapsed_quantile@max`
+/// order-statistic as the demurrage clock: the charge is computed from the
+/// **oldest** ring member's age, not the value-weighted centroid. This fully
+/// neutralizes the H2 fresh-high-value-decoy age-dilution attack (an adversary
+/// can no longer drag the elapsed time down with fresh decoys), but it makes
+/// the wallet's decoy age spread consensus-economically load-bearing: an honest
+/// spender whose ring contains a decoy much OLDER than their real input is
+/// over-charged demurrage relative to their true holding time.
+///
+/// PR #595's sensitivity sweep measured the honest over-charge under `@max` as
+/// a function of the ring's decoy age-spread, against the #314 tolerance
+/// (≤15%):
+///
+/// | honest decoy age-spread | honest over-charge | within #314 tol (≤15%) |
+/// |-------------------------|--------------------|------------------------|
+/// | ±5%                     | 1.039              | yes                    |
+/// | ±10%                    | 1.077              | yes                    |
+/// | ±20%                    | 1.158              | no                     |
+/// | ±35%                    | 1.277              | no                     |
+/// | ±50%                    | 1.388              | no                     |
+///
+/// ±10% is the widest empirically-safe band (1.077 ≤ 1.15). We adopt ±10% as
+/// the policy: it maximizes the eligible decoy pool (and hence anonymity) among
+/// the #314-safe bands. Because `@max` reads the single oldest member, the
+/// realized over-charge is bounded above by `max_ring_age / real_age`, which
+/// this band caps at ≤1.10 by construction — strictly inside the #314 bound.
+///
+/// The gamma/OSPEAD age distribution is still applied WITHIN the band, so the
+/// realistic recent-biased spend shape is preserved among age-similar decoys.
+pub const AGE_SIMILARITY_SPREAD_BPS: u64 = 1_000;
+
+/// Compute the inclusive `[min_age, max_age]` decoy-age band around a real
+/// input's age under the [`AGE_SIMILARITY_SPREAD_BPS`] policy.
+///
+/// The lower bound is additionally floored at [`MIN_DECOY_AGE_BLOCKS`] so
+/// decoys are always confirmed. A degenerate real age (below the confirmation
+/// floor) can produce `min_age > max_age`, in which case no candidate is
+/// in-band and callers fall back to nearest-age selection.
+pub fn age_similarity_band(real_input_age: u64) -> (u64, u64) {
+    let delta = (real_input_age as u128 * AGE_SIMILARITY_SPREAD_BPS as u128 / 10_000) as u64;
+    let min_age = real_input_age
+        .saturating_sub(delta)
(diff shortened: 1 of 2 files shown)


rjwalters:
**Approved!** Verified the core safety claim, wallet-side scoping, regression surface, and privacy numbers.

## Core safety claim — genuine, not tautological
`test_honest_overcharge_under_max_within_314_bound` exercises the real path end to end:
- Builds a candidate pool spanning ages 1000..60800 (up to ~2.8x the 21600-block real age) — decoys that WOULD over-charge under `@max` without the band.
- Runs the real `select_decoys_for_input`, reconstructs the ring, and feeds it to the real `ring_elapsed_quantile(.., 10_000)` and real `demurrage_charge`.
- Confirmed `ring_elapsed_quantile(.., 10000)` returns `ages[n-1]` (max) via nearest-rank; `demurrage_charge` is linear in `elapsed_blocks`, so realized over-charge = `max_ring_age / real_age`, which the +/-10% band caps. In-band worst case (~23600/21600 ≈ 1.093) is genuinely > 1.0, so the `<= 1.101` assertion actually tests band enforcement (an unbounded selector would yield ~2.8). Passes locally.

## Band math + fallback verified
`age_similarity_band(21600)=[19440,23760]`, `(50)=[45,55]`, `(11)=[10,12]` (saturating lower floor at `MIN_DECOY_AGE_BLOCKS=10`). Degenerate real ages below the confirmation floor produce `min_age > max_age` (empty band) and are correctly routed to the phase-2 nearest-age liveness fallback rather than reverting to the unbounded global-gamma sample.

## Genuinely wallet-side only
`ring_elapsed_quantile` has NO consensus/add_block caller — only the offline `decoy_quantile_sweep` simulation, the lib re-export, and tests reference it. It stays unwired; #577-B1 remains blocked under #323 as the PR claims. Diff touches only `decoy_selection.rs` and a test in `constrained_analysis.rs`.

## No regression
Selector signature unchanged (`_real_input_age` -> `real_input_age`), so the production caller at `ledger/store.rs:1775` is intact. `cargo test -p botho --lib decoy_selection` -> 21 passed (incl. 3 new); `cargo test -p bth-cluster-tax age_similarity_band_privacy_cost` -> passed.

## Privacy numbers not cooked
`run_comparative_analysis` genuinely applies `max_age_ratio=1.10` in the constrained arm vs unconstrained; test runs and reports the per-adversary reduction. Minor (non-blocking) notes: the symmetric +/-10% band is approximated as `[real/1.1, real*1.1]`, and the assertion is a loose `< 2.0` bit guardrail — both are honestly documented in the test.

## CI / merge
All required checks pass (cargo build --workspace, rustfmt pinned-nightly, benchmarks, fuzz build guard). `mergeStateStatus` = CLEAN. PR body uses `Closes #596`.

Solid, well-documented, and correctly scoped. LGTM.

PR #600, opened 2026-07-01
fix(mempool): fail closed on DB error in key-image double-spend checks

### LINKED ISSUE #598: security(mempool): fail-open double-spend checks swallow DB errors (M7 sibling of #564)
## Summary
Two mempool key-image double-spend checks swallow a `LedgerError` and treat it as "not spent" — the same **M7 fail-open class** that #559/#564 closed on the authoritative block-apply path (`store.rs::verify_transaction`, `record_key_image`). #564 fixed the block-apply gate but did not sweep the sibling **mempool** call sites. Found during a defensive audit of the consensus/mempool validation paths (2026-07-01).

Severity: **Medium** — these are admission/eviction filters, not the authoritative consensus rule. A determined attacker still hits the now-fail-closed `record_key_image` at block-apply. But under DB pressure (LMDB read error / reader-table exhaustion) a node will **admit and gossip**, or **retain**, a double-spend it should have filtered — so this is the last swallowed-error double-spend surface on the mempool side.

## Finding 1 — admission check fails open (`botho/src/mempool.rs:811`)
```rust
// in validate_clsag_inputs
for input in clsag_inputs {
    if let Ok(Some(_)) = ledger.is_key_image_spent(&input.key_image) {
        return Err(MempoolError::KeyImageSpent(input.key_image));
    }
}
```
`is_key_image_spent` returns `Result<Option<u64>, LedgerError>`. On `Err(_)` the `if let Ok(Some(_))` arm is skipped, execution falls through, and the tx is **admitted as if the key image were unspent**. A transient DB error lets an already-spent key image pass the mempool double-spend filter; the tx is then offered for block inclusion.

**Fail-closed fix** (mirror the store.rs M7 pattern):
```rust
for input in clsag_inputs {
    match ledger.is_key_image_spent(&input.key_image) {
        Ok(Some(_)) => return Err(MempoolError::KeyImageSpent(input.key_image)),
        Ok(None) => {}
        Err(e) => return Err(MempoolError::LedgerError(e.to_string())),
    }
}
```

## Finding 2 — eviction pass fails open (`botho/src/mempool.rs:1013-1016`)
```rust
// in remove_invalid
let is_invalid = pending.tx.inputs.clsag().iter().any(|input| {
    matches!(ledger.is_key_image_spent(&input.key_image), Ok(Some(_)))
});
```
`remove_invalid` evicts mempool txs whose inputs were spent on-chain. On `Err(_)`, `matches!(..., Ok(Some(_)))` is `false`, so the tx is deemed **still valid and retained** — the mirror of Finding 1. If the DB errors exactly as a block confirms these inputs, the now-double-spending tx stays in the mempool and can be re-proposed.

**Fail-closed fix** (for an eviction pass, an uncertain result should evict — a false eviction is harmless, the sender can resubmit):
```rust
let is_invalid = pending.tx.inputs.clsag().iter().any(|input| {
    match ledger.is_key_image_spent(&input.key_image) {
        Ok(Some(_)) => true,   // spent on-chain -> evict
        Ok(None) => false,     // unspent -> keep
        Err(_) => true,        // DB error -> fail closed: evict rather than retain a possible double-spend
    }
});
```
(Log the DB error before evicting so the operational failure is visible.)

## Acceptance criteria
- [ ] `validate_clsag_inputs` (mempool.rs:811) distinguishes `Ok(None)` from `Err` and rejects (or propagates) on DB error.
- [ ] `remove_invalid` (mempool.rs:1013) evicts on DB error rather than retaining.
- [ ] Regression tests inject a real `LedgerError` (reuse the `open_single_reader`/`read_txn_for_test` fault-injection harness #564 added at `store.rs:225-238`) and assert: admission REJECTS and eviction EVICTS on DB error.
- [ ] No behavior change on the `Ok(None)` / `Ok(Some)` paths.

## Related / notes
- Same class as **#559/#564 (M7)** — block-apply gate already fail-closed; this is the mempool sibling that fix missed.
- Lower-priority adjacent swallows the audit flagged (NOT included in this issue's ACs, mentioned for the fixer's awareness): `mempool.rs:832` ring-member lookup (guarded by the `!found_any` backstop at :841, so partial-only), and `store.rs:453-460` tip-timestamp read defaulting to 0 on DB error (fail-open timestamp-monotonicity, not double-spend).
- Node-local check → not consensus-breaking; does NOT need to ride #323.

rjwalters:
## Summary
Two mempool key-image double-spend checks swallowed a `LedgerError` and treated it as "not spent" — the M7 fail-open class that #559/#564 closed on the authoritative block-apply path. #564 fixed the block-apply gate but missed the sibling mempool call sites. Under DB pressure (LMDB read error / reader-table exhaustion) a node could admit/gossip or retain a double-spend it should have filtered.

## Changes
- `validate_clsag_inputs` (admission, `botho/src/mempool.rs`): replaced `if let Ok(Some(_))` with a full `match`. On `Err(_)` it now returns `MempoolError::LedgerError(..)` instead of falling through and admitting the tx as if the key image were unspent.
- `remove_invalid` (eviction): replaced `matches!(ledger.is_key_image_spent(..), Ok(Some(_)))` with a `match` that evicts on `Err(_)` (logging the DB error via `warn!`) rather than retaining a possible double-spend. A false eviction is harmless — the sender can resubmit.
- Added two regression tests reusing the #564 reader-exhaustion fault-injection harness (`open_single_reader` / `read_txn_for_test`) that inject a real `LedgerError`.

| `validate_clsag_inputs` distinguishes `Ok(None)` from `Err` and rejects/propagates on DB error | ✅ | `Err(e) => return Err(MempoolError::LedgerError(e.to_string()))`; `test_validate_clsag_inputs_fails_closed_on_db_error` asserts `Err(LedgerError(_))` |
| `remove_invalid` evicts on DB error rather than retaining | ✅ | `Err(e) => { warn!(..); true }`; `test_remove_invalid_evicts_on_db_error` asserts mempool empty after eviction |
| Regression tests inject a real `LedgerError` via the #564 harness | ✅ | Both tests use `open_single_reader` + held `read_txn_for_test` to exhaust the reader table (`MdbError::ReadersFull`) |
| No behavior change on `Ok(None)` / `Ok(Some)` paths | ✅ | `Ok(Some(_)) => KeyImageSpent` / `true` and `Ok(None) => {}` / `false` preserve prior semantics; full mempool suite passes (47 passed) |

Out of scope per issue Notes (not touched): `mempool.rs` ring-member lookup swallow and `store.rs` tip-timestamp read.

## Test Plan
- `cargo build -p botho` — passes (only pre-existing warnings).
- `cargo test -p botho --lib mempool` — 47 passed, 0 failed; includes both new tests.

Closes #598

### DIFF
--- a/botho/src/mempool.rs
+++ b/botho/src/mempool.rs
@@ -807,9 +807,15 @@ impl Mempool {
         }
 
         // Check for double-spends via key images (ledger)
+        //
+        // M7 fail-closed: distinguish `Ok(None)` (genuinely unspent) from a DB
+        // error. Swallowing the error and admitting the tx as if unspent would
+        // let an already-spent key image pass this filter under DB pressure.
         for input in clsag_inputs {
-            if let Ok(Some(_)) = ledger.is_key_image_spent(&input.key_image) {
-                return Err(MempoolError::KeyImageSpent(input.key_image));
+            match ledger.is_key_image_spent(&input.key_image) {
+                Ok(Some(_)) => return Err(MempoolError::KeyImageSpent(input.key_image)),
+                Ok(None) => {}
+                Err(e) => return Err(MempoolError::LedgerError(e.to_string())),
             }
         }
 
@@ -1009,11 +1015,29 @@ impl Mempool {
         let mut to_remove = Vec::new();
 
         for (tx_hash, pending) in &self.txs {
-            // Check if any key image was spent in ledger
-            let is_invalid =
-                pending.tx.inputs.clsag().iter().any(|input| {
-                    matches!(ledger.is_key_image_spent(&input.key_image), Ok(Some(_)))
-                });
+            // Check if any key image was spent in ledger.
+            //
+            // M7 fail-closed: on a DB error, evict rather than retain. Retaining
+            // a possibly-double-spending tx (the pre-fix behavior of
+            // `matches!(..., Ok(Some(_)))`, which treats `Err(_)` as "still
+            // valid") risks re-proposing a double-spend. A false eviction is
+            // harmless — the sender can resubmit.
+            let is_invalid = pending.tx.inputs.clsag().iter().any(|input| {
+                match ledger.is_key_image_spent(&input.key_image) {
+                    Ok(Some(_)) => true, // spent on-chain -> evict
+                    Ok(None) => false,   // unspent -> keep
+                    Err(e) => {
+                        warn!(
+                            "DB error checking key image {} for mempool tx {}; \
+                             evicting (fail closed): {}",
+                            hex::encode(&input.key_image[0..8]),
+                            hex::encode(&tx_hash[0..8]),
+                            e
+                        );
+                        true // DB error -> fail closed: evict
+                    }
+                }
+            });
 
             if is_invalid {
                 to_remove.push(*tx_hash);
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Solid, tightly-scoped M7 fail-closed fix. Both mempool key-image double-spend call sites now distinguish `Ok(None)` from `Err` and fail closed, matching the #559/#564 pattern on the block-apply gate.

## Verification

**Correctness / security**
- `validate_clsag_inputs` (admission, `botho/src/mempool.rs:814`): `if let Ok(Some(_))` replaced with a full `match`; `Err(e) => return Err(MempoolError::LedgerError(e.to_string()))`. An already-spent key image can no longer slip through admission under DB pressure. `Ok(Some(_)) => KeyImageSpent` and `Ok(None) => {}` preserve prior semantics exactly.
- `remove_invalid` (eviction, `botho/src/mempool.rs:1022`): `matches!(..., Ok(Some(_)))` (which treated `Err(_)` as "still valid" and retained the tx) replaced with a `match` that evicts on `Err(_)`, logging via `warn!`. False eviction is harmless (sender resubmits); retaining a possible double-spend is not. `Ok(Some) => true` / `Ok(None) => false` unchanged.
- `warn!` and `hex::encode` were already in scope (used elsewhere in the file); `MempoolError::LedgerError(String)` variant matches the `.to_string()` conversion, consistent with line 936.

**Scope discipline**
- Out-of-scope swallows correctly left untouched: the ring-member lookup `if let Ok(Some(utxo))` at `botho/src/mempool.rs:838` and the `store.rs` tip-timestamp read.

**Tests**
- Two new regression tests reuse the #564 reader-exhaustion harness (`open_single_reader` + held `read_txn_for_test`, both `#[cfg(test)]`) to inject a real `MdbError::ReadersFull` -> `LedgerError`:
  - `test_validate_clsag_inputs_fails_closed_on_db_error` asserts admission returns `Err(LedgerError(_))`.
  - `test_remove_invalid_evicts_on_db_error` asserts the mempool is empty after eviction.
- `cargo test -p botho --lib mempool` -> 47 passed, 0 failed (includes both new tests).

## Test Execution
1. `cargo build --workspace` — ✅ Executed locally: builds clean (only pre-existing warnings)
2. `cargo test -p botho --lib mempool` — ✅ Executed locally: 47 passed, 0 failed

## Note
The original CI run had `cargo build --workspace (Linux)` failing — that was on a stale base. I rebased the branch onto current `main` (clean rebase, force-with-lease pushed) and the re-run is fully green: `cargo build --workspace (Linux)` pass, `rustfmt` pass, `Fuzz Build Guard` pass. Merge state is now CLEAN.

Acceptance criteria all met. Nice work.

PR #601, opened 2026-07-01
fix(consensus): reject/saturate block fee-sum overflow instead of wrapping

### LINKED ISSUE #599: hardening(consensus): unchecked .sum() on block fee totals can wrap/panic (store.rs:864, block.rs:596)
## Summary
Two unchecked `.sum::<u64>()` over attacker-influenced per-tx fees. In release (`overflow-checks = false`) a crafted block whose fees sum past `u64::MAX` **wraps silently**; in debug it **panics**. Found during a defensive audit (2026-07-01).

Severity: **Low.** Neither value feeds the mint/balance decision (per-tx balance is already checked via `checked_add`, #340), so this is not a direct inflation vector. But it's a latent correctness/DoS surface: a debug-build panic on a gossiped block, and a wrapped total that could flip the `total_fees() > 0` gate.

## Sites
1. `botho/src/ledger/store.rs:864` — `let block_fees: u64 = block.transactions.iter().map(|tx| tx.fee).sum();` (used only for logging at ~:1049).
2. `botho/src/block.rs:596-598` — `Block::total_fees()` `self.transactions.iter().map(|tx| tx.fee).sum()` (gates whether lottery validation runs).

## Fix
Use checked accumulation and reject on overflow, matching the pattern already in `verify_ring_signatures` / the #340 balance path:
```rust
let block_fees: u64 = block.transactions.iter()
    .try_fold(0u64, |acc, tx| acc.checked_add(tx.fee))
    .ok_or(LedgerError::/* fee overflow */)?;
```
`total_fees()` can return `Option<u64>` or a saturating variant with a documented caller contract.

## Acceptance criteria
- [ ] Both sites reject (or saturate with a documented rationale) instead of wrapping/panicking on fee-sum overflow.
- [ ] Test: a block whose `tx.fee` values sum past `u64::MAX` is handled deterministically (rejected, not panic/wrap) in both debug and release.

## Notes
- Node-local correctness; does NOT change consensus acceptance semantics, so no #323 dependency.
- Adjacent defense-in-depth the audit suggested (separate, not in scope here): an explicit max-tx-count / max-serialized-size bound inside `add_block_inner` rather than relying solely on the network-layer `MAX_BLOCK_SIZE` gossip cap.

rjwalters:
## Summary
Two unchecked `.sum::<u64>()` reductions over attacker-influenced per-transaction fees could wrap silently under `overflow-checks = false` (release) or panic in debug when a crafted block's fees sum past `u64::MAX`. This hardens both sites so an overflowing fee total is handled deterministically instead. Severity is Low (neither value feeds a mint/balance decision), but this closes a latent debug-panic DoS surface on gossiped blocks and a wrapped-total gate flip.

## Changes
- **`botho/src/ledger/mod.rs`**: add typed `LedgerError::FeeOverflow` variant.
- **`botho/src/ledger/store.rs`**: replace the raw `.sum()` at the fee-accounting line with a new `checked_block_fees` helper (`try_fold` + `checked_add`) that returns `LedgerError::FeeOverflow`. `add_block_inner` now rejects the block on overflow, mirroring the #340 per-tx balance guard.
- **`botho/src/block.rs`**: `Block::total_fees()` now uses `saturating_add` with a documented caller contract — a saturated total only ever fails validation and is separately rejected by the ledger, so the `total_fees() > 0` gate and lottery accounting can no longer panic or wrap.
- Regression tests for both sites (a fee sum past `u64::MAX`), verified in **both** debug and release.

Out of scope (per issue Notes): the adjacent max-tx-count / max-serialized-size bound in `add_block_inner`.

| Both sites reject (or saturate w/ documented rationale) instead of wrapping/panicking | done | `store.rs` rejects via `LedgerError::FeeOverflow`; `block.rs` `total_fees()` saturates with a documented caller contract |
| Test: block whose `tx.fee` sum past `u64::MAX` handled deterministically in debug AND release | done | `test_total_fees_saturates_on_overflow`, `test_checked_block_fees_rejects_overflow` pass under `cargo test -p botho --lib` and `--release` |

## Test Plan
- `cargo build -p botho` — clean (pre-existing dead-code warnings only).
- `cargo test -p botho --lib -- total_fees checked_block_fees` — 5 passed (debug).
- `cargo test -p botho --lib --release -- total_fees checked_block_fees` — 5 passed (release).
- `cargo test -p botho --lib -- block:: consensus::lottery:: ledger::store::` — 86 passed, 0 failed.
- `cargo clippy -p botho --lib` — 0 errors.

Closes #599

### DIFF
--- a/botho/src/block.rs
+++ b/botho/src/block.rs
@@ -592,9 +592,24 @@ impl Block {
         hasher.finalize().into()
     }
 
-    /// Get total fees from all transactions
+    /// Get total fees from all transactions.
+    ///
+    /// Per-transaction fees are attacker-influenced, so this accumulates with
+    /// `saturating_add`: a crafted block whose fees sum past `u64::MAX` clamps
+    /// to `u64::MAX` rather than wrapping silently (release, `overflow-checks =
+    /// false`) or panicking (debug). See issue #599.
+    ///
+    /// Caller contract: this is *not* an authoritative fee total for consensus
+    /// arithmetic. A saturated value only ever fails validation — it cannot
+    /// match a block's declared lottery split — and any block with an
+    /// overflowing fee total is rejected outright by the ledger via a checked
+    /// accumulation in `add_block_inner` (`LedgerError::FeeOverflow`). Callers
+    /// use it only for the "are there any fees at all?" gate and for lottery
+    /// accounting that is separately re-checked against the block's summary.
     pub fn total_fees(&self) -> u64 {
-        self.transactions.iter().map(|tx| tx.fee).sum()
+        self.transactions
+            .iter()
+            .fold(0u64, |acc, tx| acc.saturating_add(tx.fee))
     }
 }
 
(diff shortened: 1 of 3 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped hardening that fully satisfies the acceptance criteria for #599.

## Correctness & Design
- **`store.rs`** — `checked_block_fees` (`try_fold` + `checked_add` → `LedgerError::FeeOverflow`) replaces the raw `.sum()` at the fee-accounting line. Verified the `?` early-returns before the write txn is committed, so an overflowing block is rejected with no state mutation. `block_fees` is still consumed at store.rs:1068 (logging), so no dead binding.
- **`block.rs`** — the `saturating_add` design for `Block::total_fees()` is sound. I traced both consensus callers:
  - `store.rs:816` uses it only in the `total_fees() > 0` gate; a saturated value is still `> 0`, so lottery validation still runs and re-checks the declared split against the block summary. The authoritative fee accounting (`checked_block_fees`) rejects any overflowing block regardless.
  - `lottery.rs:538` (`validate_block_lottery`) feeds it into `compute_pool_accounting`, which is validated against the block's declared lottery summary — a saturated total can only mismatch and fail validation.
  - The documented caller contract on `total_fees()` accurately describes this "cannot feed a mint decision" invariant.

## Tests
- `cargo test -p botho --lib -- total_fees checked_block_fees` — 5 passed in **debug**.
- Same suite with `--release` — 5 passed. Confirmed `overflow-checks = false` in the release profile (Cargo.toml:227), so the release-wrap regression test is genuinely exercising the wrap path the old `.sum()` had.
- Both overflow tests assert the deterministic outcome (saturate to `u64::MAX` / `Err(FeeOverflow)`), and normal-sum tests confirm no behavior regression.

## Quality
- No new clippy warnings on the changed lib (the single `unused doc comment` warning is pre-existing in `gossip/src/messages.rs`, unrelated).
- Error variant, helper, and both test blocks are clearly documented with issue references (#599, #340).
- Out-of-scope items (max-tx-count / max-serialized-size bound) correctly deferred per the issue Notes.

## CI
All GitHub checks pass (`cargo build --workspace`, benchmarks, fuzz build guard, rustfmt). Merge state: CLEAN.

PR description uses `Closes #599`. Nice work.

Issue #581, opened 2026-06-28
Tighten cluster-tag inheritance bound: ring-member-sum input set permits decoy-sourced inflation


rjwalters:
Discovered while evaluating PR #580 (issue #576, H2-B3).

## Context
PR #580 wires `check_cluster_tag_inheritance` into `Ledger::add_block_inner` (C6 guard, `botho/src/ledger/store.rs`). Because the real input is hidden among ring decoys, the guard uses the **sum of all resolved ring members' cluster mass** as the per-cluster input bound (`output_mass <= input_mass + tolerance`). This is correct for liveness (real inputs ⊆ ring ⇒ upper bound ⇒ never false-rejects) and is the only node-agnostic input set available at consensus.

## Limitation
The bound is the sum over ALL ring members, not the real input. An attacker who selects real, on-chain **cluster-X-tagged** UTXOs as decoys can raise cluster-X's input bound to roughly `ring_size ×` the decoy mass, and thereby attribute more cluster-X mass to their own output than their real input supplied — bounded inflation, sourced from decoy selection.

## Why this is NOT a PR #580 blocker
- The specific design-named vector in #574 ("fresh **background**-tagged decoys") does not exploit this: untagged/background decoys contribute zero to any real cluster's bound, so they cannot raise a tagged cluster's ceiling.
- The residual gap requires sourcing real cluster-tagged decoys and is bounded by ring size; it is strictly better than the pre-PR status quo (no check at all, unlimited attribution), which rejected nothing.
- B3 is one of three coupled H2 items (B1 demurrage quantile, B2 cluster-factor floor, B3 this). H2 hardening is intended to land as the set.
- The guard is deterministic and liveness-safe (no false-halt, no fork).

## Possible future tightening
- Constrain the input bound using the minimum (not sum) over ring members, or the real-input commitment if/when a consensus-visible binding exists.
- Reconsider once B1/B2 land — the combined factor-floor + quantile may make residual decoy-sourced tag inflation economically inert.
- Phase-2 Pedersen-committed tags with ZK inheritance proofs (noted in `cluster_tags.rs`) would close it tightly.

Tracking only; surfaced during PR #580 review.

rjwalters:
[...]

Issue #579, opened 2026-06-27
build(M1-B5): mempool relay-policy guard — assert relay only tightens above consensus floor


rjwalters:
Decomposition item **B5** from design #574 (M1). **BLOCKED on B4 (#574-B4) + #574 approval.**

Under the policy-split, congestion (`dynamic_base`, triply node-local: f64 ema_fullness reset-on-restart + wall-clock at_min_block_time + per-node mempool) stays relay-only. Add a guard/assertion + docs that mempool relay policy only ever TIGHTENS above the deterministic consensus floor (never below), and document DynamicFeeBase as relay-only. Mostly guard + docs; depends on B4's consensus floor existing. Rides #323.

Issue #575, opened 2026-06-27
build(H2-B2): floor cluster factor at the integer ring-centroid-implied factor


rjwalters:
Decomposition item **B2** from design #574 (H2 hardening). **BLOCKED pending design approval of #574.**

Replace the f64 cosine plausibility gate with a deterministic integer floor: the spender's claimed cluster factor must be >= the factor implied by the ring centroid. Closes the H2 hole where fresh background-tagged decoys drag the factor to 1000 (demurrage 0). No free/empirical parameter — the floor is derived from public ring data.

Determinism contract: pure fn of (tx ring members, chain UTXO state); integer-only; no HashMap/HashSet iteration order in any consensus output (BTreeMap/sorted Vec). Rides #323. Unblock when #574 design is approved.

PR #602, opened 2026-07-01
build(H1-B4): deterministic consensus fee floor in add_block

### LINKED ISSUE #578: build(H1-B4): deterministic consensus fee floor in add_block (consumes B1+B2)
Decomposition item **B4** from design #574 (H1 enforcement, policy-split model). **BLOCKED on B1 (#574-B1), B2 (#575), and #574 approval.**

Enforce a deterministic per-tx floor in Ledger::add_block: `static_base + cluster_factor + demurrage_charge` recomputed from chain state at block height. Congestion surcharge stays mempool-relay-only (Bitcoin model) — NOT in consensus (dissolves M1). Cuts over from the mempool-only centroid path.

Depends on B1 (quantile demurrage kernel) and B2 (factor floor) landing first so consensus doesn't enforce a gameable rule. Inherits B1's empirical gate. Determinism contract: every input a pure fn of block+chain state; integer-only; enumerate inputs per #574 Q5. Rides #323.

rjwalters:
Closes #578

Item **B4** of approved design #574 (H1 enforcement, policy-split model). Consumes B1 (#577, quantile kernel) and B2 (#575, factor floor). Consensus-breaking — rides the next coordinated reset (pre-mainnet, acceptable per the no-hard-forks-after-mainnet constraint).

## What this does

Adds a new consensus gate **C7** in `Ledger::add_block_inner` that rejects any block containing a transfer tx whose `tx.fee` is below a deterministically recomputed floor:

```
consensus_floor(tx) = base_minimum_fee(tx_size, num_outputs, num_memos,
                                       cluster_wealth, CONSENSUS_FEE_BASE)  // NO congestion
                    + demurrage_charge(output_sum, floored_factor,
                                       ring_elapsed_quantile@max,
                                       rate_bps(height), blocks_per_year)
require: tx.fee >= consensus_floor(tx)
```

This is the sole block-acceptance enforcement of the demurrage stock-level term of the Gini mechanism (a miner could previously include own/under-fee txs and evade demurrage entirely). Minting/coinbase is exempt: the floor runs only over `block.transactions` (transfer txs); the minting tx is a separate `block.minting_tx` field and is never in that list.

## (a) Deterministic base used in place of `dynamic_base`, and why it is congestion-free

The mempool floor multiplies the size fee by `self.dynamic_fee.compute_base(self.at_min_block_time)` — a value carrying an **f64 `ema_fullness`** that is node-local and reset on restart (`dynamic_fee.rs:79-80`), plus a wall-clock-derived `at_min_block_time`. That must never enter consensus (audit cycle 6 M1: a node with a hot vs. cold EMA would accept a different set of blocks).

The consensus floor instead pins the base to a fixed constant `CONSENSUS_FEE_BASE = 1` = `DynamicFeeBase::default().base_min`. This is exactly the value `compute_base` returns whenever the network is uncongested (not at minimum block time, or below the target fullness) — i.e. the "no congestion multiplier" neutral base. Because `compute_base` can only ever return a value `>= base_min`, the mempool's dynamic base can only ever raise a node's local admission threshold **above** this floor, never below it (Bitcoin's min-relay-fee vs. consensus-validity split, design #574 Q1/Q2). Congestion pricing therefore stays a mempool-relay-only policy.

## (b) centroid → quantile@max cutover

The mempool demurrage clock previously used `ring_elapsed_centroid` (value-weighted mean of ring-member ages), the H2 age-dilution vector: a spender pads the ring with fresh high-value decoys to drag the mean toward zero. Both the mempool call site (`mempool.rs:713`) **and** the new consensus floor now use `ring_elapsed_quantile(&members, height, 10_000)` — `quantile_bps = 10000 == max`, an unweighted, value-independent order statistic. The max is the one statistic guaranteed to surface a lone old real input regardless of decoy count (p75/p90 miss a single old input in an 11-ring — need `q > (n-1)/n = 90.9%`). The empirical sweep #577/#595 selected `@max`. Mempool admission and consensus now agree on the demurrage clock.

The factor is floored at the ring-centroid-implied factor via B2's `ring_centroid_floored_factor` (`claimed_factor = fee_config.cluster_factor(cluster_wealth)`), so background-tagged outputs cannot escape a wealthy ring's demurrage.

## (c) Exact insertion point in `add_block_inner`

C7 runs in a per-transfer-tx loop **after** C3 (`verify_ring_members`) and C6 (`verify_cluster_tag_inheritance`), and **before** the lottery gate. It re-resolves the ring members from the committed UTXO set to `(value, created_at)` (for the age quantile) and `(value, &cluster_tags)` (for the factor floor), matching the pattern C6 already uses.

## (d) Proposer and validator see identical chain state for `cluster_wealth`

All validation gates (C1–C7 and the lottery gate) execute **before** the single `add_block_inner` write txn that applies this block's outputs and calls `update_cluster_wealth_for_output`. So every `get_cluster_wealth` read during the floor computation reflects the **pre-block** per-cluster wealth — the same state the proposer saw when building the block. `cluster_wealth` reads are fail-closed (a DB error propagates as `LedgerError`, never defaults to a lower floor — M7). Chain height is `block.height()` (the height being applied), identical on every honest node. The floor applies from genesis (no soft activation-height ramp in the consensus path).

## Determinism contract (design #574 Q5)

Integer-only throughout; the only maps are the `BTreeMap`s inside the reused helpers. No `HashMap`/`HashSet` iteration feeds a consensus value; no `dynamic_fee` EMA, no `at_min_block_time`, no `utxo_pool_size`, no `RING_TAG_*`/`SPARSE_POOL_THRESHOLD` bootstrap constants. `rate_bps(height)` and `blocks_per_year` come from `crate::monetary::mainnet_policy()`.

## Tests (`ledger::store::tests`)

- `consensus_fee_floor_rejects_under_fee_accepts_at_floor` — fee floor−1 rejected (`InvalidBlock`), fee at floor accepted.
- `consensus_fee_floor_is_deterministic_recomputation` — bit-for-bit reproducible floor.
- `consensus_fee_floor_demurrage_at_max_resists_fresh_decoy_dilution` — old wealthy input padded with 10 fresh equal-value decoys still pays demurrage; asserts max-quantile age dwarfs the diluted centroid.
- `consensus_fee_floor_factor_floor_binds_background_outputs` — background-tagged outputs over a wealthy ring still charged demurrage.
- `consensus_fee_floor_is_congestion_independent` — floor equals the congestion-free base recomputation (no dynamic base leaks in).
- `consensus_fee_floor_consults_cluster_wealth` — raising a cluster's global wealth raises the floor.

`cargo build -p botho` clean; `cargo test -p botho --lib` = 1083 passed; `cargo test -p bth-cluster-tax` green.

## Scope

Does not touch B5 (#579 relay-policy guard/docs) or #581's residual tag-bound tightening. `ring_elapsed_centroid` is retained (still used by the decoy-quantile sweep to compare kernels).

### DIFF
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -123,6 +123,33 @@ fn check_cluster_tag_inheritance(
     Ok(())
 }
 
+/// Congestion-free deterministic fee base for the consensus fee floor
+/// (issue #578, design #574 item H1-B4).
+///
+/// The mempool floor multiplies the size fee by a *dynamic* base
+/// (`DynamicFeeBase::compute_base`) that reflects local congestion. That base
+/// carries an f64 EMA that is node-local and reset on restart
+/// (`dynamic_fee.rs:79-80`), so it MUST NOT enter consensus (a node with a hot
+/// vs. cold congestion EMA would otherwise accept a different set of blocks —
+/// audit cycle 6 M1). This constant is the neutral, no-congestion value:
+/// `DynamicFeeBase::default().base_min` (the fee-per-byte floor the dynamic
+/// controller returns whenever the network is uncongested — i.e. not at
+/// minimum block time, or below the target fullness; see
+/// `DynamicFeeBase::compute_base`). Using it makes the consensus floor a
+/// congestion-independent lower bound; the mempool's dynamic base can only ever
+/// raise a node's local admission threshold ABOVE this floor, never below it
+/// (Bitcoin's min-relay-fee vs. consensus-validity split, design #574 Q1/Q2).
+const CONSENSUS_FEE_BASE: u64 = 1;
+
+/// Quantile (in basis points) of the ring members' ages used by the consensus
+/// demurrage clock. `10_000` == the maximum age (issue #578, design #574 item
+/// H2/B1; empirical sweep #577/#595 selected `@max`). The max is the one order
+/// statistic guaranteed to surface a lone old real input regardless of how many
+/// fresh decoys pad the ring — p75/p90 miss it because a single old input in an
+/// 11-ring needs `q > (n-1)/n = 90.9%`. See
+/// `bth_cluster_tax::ring_elapsed_quantile`.
+const CONSENSUS_RING_AGE_QUANTILE_BPS: u32 = 10_000;
+
 /// Emission-controller state to persist atomically alongside a block.
 ///
 /// The difficulty/reward/epoch counters are a pure function of the applied
@@ -806,6 +833,33 @@ impl Ledger {
                     tx_idx, e
                 ))
             })?;
+        }
+
+        // C7 (issue #578, H1-B4): deterministic consensus fee floor.
+        //
+        // Reject any block containing a transfer tx whose `fee` is below the
+        // consensus floor `base_minimum_fee + demurrage_charge` recomputed here
+        // from the transaction and committed chain state at `block.height()`.
+        // This is the sole enforcement of the demurrage stock-level term of the
+        // Gini mechanism (a miner could otherwise include own/under-fee txs and
+        // evade demurrage entirely). The floor is CONGESTION-FREE: the mempool's
+        // node-local dynamic fee base (f64 EMA, restart-reset) is deliberately
+        // excluded, so a node with a hot vs. cold congestion EMA accepts the
+        // SAME set of blocks (design #574 Q1/Q2, audit cycle 6 M1). Runs after
+        // C3 (ring resolution) and C6 (tag inheritance), reusing the same
+        // committed-UTXO ring resolution, and BEFORE this block's outputs are
+        // applied — so the proposer and every validator compute identical floors
+        // (no fork). See `verify_consensus_fee_floor`.
+        for (tx_idx, tx) in block.transactions.iter().enumerate() {
+            self.verify_consensus_fee_floor(tx, block.height())
+                .map_err(|e| match e {
+                    LedgerError::InvalidBlock(msg) => LedgerError::InvalidBlock(format!(
+                        "Transaction {} fee-floor validation failed: {}",
+                        tx_idx, msg
+                    )),
+                    // Propagate DB errors unchanged (fail-closed, M7).
+                    other => other,
+                })?;
         }
 
         // Validate lottery results and compute the new carryover pool.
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!** — Consensus-critical review complete (H1-B4, design #574).

This is a consensus-breaking change; verified against the design's determinism/fork-safety contract (#574 Q1/Q3/Q5). All priorities cleared.

## Determinism / fork-safety (the whole point)
- **No f64 anywhere in the floor path.** `CONSENSUS_FEE_BASE=1` is a constant; `cluster_factor`/`sigmoid_approx` are integer LUT (i128/u64, post-#571); `demurrage_charge`, `ring_elapsed_quantile` (ages-only integer sort), and `demurrage_rate_bps` are all integer.
- **No node-local state.** No `dynamic_fee` EMA, no `at_min_block_time`, no `utxo_pool_size`, no `RING_TAG_*`/`SPARSE_POOL` bootstrap constants reach the consensus path.
- **No HashMap/HashSet feeds a consensus value.** `effective_cluster_wealth_from_outputs` and `ring_centroid_floored_factor` iterate ordered Vecs; the only maps are BTreeMaps inside reused helpers.
- **Proposer/validator chain-state agreement (the subtlest fork risk) — correct.** C7 (`store.rs:854`) runs before the lottery gate and before the single `add_block_inner` write txn (`:910`) that inserts this block's outputs and calls `update_cluster_wealth_for_output` (`:956`+). Every `get_cluster_wealth` / `get_utxo_by_target_key` read reflects PRE-block state — identical to what the proposer saw. Height is `block.height()`.

## Fail-closed (M7)
`get_cluster_wealth` returns `Err(LedgerError::Database)` on a real read error (only `Ok(0)` for genuinely-absent clusters), propagated via `?` through `effective_cluster_wealth_from_outputs` and `ring_centroid_floored_factor`. The `add_block_inner` match arm re-wraps `InvalidBlock` but passes DB errors through unchanged. A read error can never silently lower the floor.

## Floor math — CONSENSUS_FEE_BASE=1 is correct, not a collapse
Scrutinized whether pinning the base to 1 collapses the size/output static component. It does not. In `compute_fee_with_dynamic_base_and_outputs` (`fee_curve.rs:520`) the base is a **multiplier** over `size × cluster_factor × output_penalty` (plus `max(fee_per_memo, base×100)×num_memos`), not a separate additive `static_base`. `base=1` is exactly `base_min`, which is exactly what `compute_base` returns whenever uncongested — so the consensus floor is the full size/factor/penalty/memo structure minus congestion scaling. The design's `static_base + cluster_factor + demurrage` prose maps onto this multiplicative curve; the policy-split (congestion stays mempool-relay-only, can only tighten) is faithfully implemented. Not under-charging vs intent.

## Minting/coinbase exemption
`Block.minting_tx` (`block.rs:426`) is a separate typed field from `transactions: Vec<Transaction>` (`:428`); C7 iterates only `block.transactions`, so coinbase is never fee-floor-rejected.

## quantile@max cutover
`mempool.rs:713` cut over centroid→`ring_elapsed_quantile(@10000)`; consensus uses the same kernel — relay and consensus agree on the demurrage clock. `ring_elapsed_centroid` retained only for tests/sweeps (no live-path caller).

## Tests
6 new tests cover under-fee reject / at-floor accept, deterministic recomputation, demurrage@max anti-dilution, factor floor binding background outputs, congestion independence, and cluster-wealth consultation. Local: `cargo test -p botho --lib` = 1083 passed; `cargo test -p bth-cluster-tax` green.

## Fixed during review
Applied `cargo +nightly fmt` to `store.rs` (doc-comment wrapping + a `fold`/`push` reflow) — formatting-only, no behavior change — to clear the failing `rustfmt --check (pinned nightly)` CI job. All checks now pass; merge state CLEAN.

PR #603, opened 2026-07-01
build(M1-B5): assert mempool relay fee only tightens above consensus floor

### LINKED ISSUE #579: build(M1-B5): mempool relay-policy guard — assert relay only tightens above consensus floor

rjwalters:
## Summary

Item **B5** of design #574 (M1 handling). Adds a guard + docs that enforce the relay-only-tightening invariant: the mempool admission fee threshold is always `>=` the deterministic consensus fee floor. Congestion pricing stays a relay-only policy and never enters consensus. Dependency B4 (#578) is merged (PR #602), so this is unblocked.

**No consensus change; no mempool admission-behavior change** beyond the added assertion.

## What changed

1. **Guard** (`botho/src/mempool.rs`, `Mempool::add_tx`): a `debug_assert!` at the admission fee-check site verifies `minimum_fee >= ledger.consensus_fee_floor(&tx, chain_height)`. Both sides compute the SAME base curve and the SAME demurrage kernel (`ring_elapsed_quantile@max` + `ring_centroid_floored_factor`) at the SAME `chain_height`; the only structural difference is the relay congestion multiplier `dynamic_base`, which `compute_base` clamps to `>= base_min == CONSENSUS_FEE_BASE`. So the invariant holds structurally.
   - `chain_height` is hoisted out of the demurrage block so both the demurrage clock and the assertion use the identical height (avoids a false-fire from an off-by-one-block demurrage age).
   - On a DB error recomputing the floor, the assertion is skipped (not an invariant violation).

   **Chosen mechanism: `debug_assert!` (not a runtime check).** It enforces the contract in tests/CI and during development without adding an extra hot-path ledger read (ring-UTXO resolution) to release builds. If it ever fires, a relay-policy change has broken the never-fall-below-consensus contract and must be fixed before ship. Admission still decides on `tx.fee < minimum_fee` exactly as before.

2. **Docs**:
   - `cluster-tax/src/dynamic_fee.rs`: module-level + `DynamicFeeBase` type doc-comments marking it **RELAY-ONLY** — its `f64 ema_fullness` (reset on restart) and wall-clock `at_min_block_time` inputs are node-local and must never enter consensus; captures why congestion stays out of the consensus floor and how the clamp guarantees relay only tightens.
   - `botho/src/mempool.rs`: a note at the `dynamic_base` site cross-referencing the consensus floor and the invariant.

3. **Test** (`relay_minimum_never_below_consensus_floor_hot_and_cold`): a HOT congestion EMA and a COLD EMA both produce a relay minimum `>=` the consensus floor (hot strictly stricter); the uncongested/cold minimum equals the floor exactly (only difference is the 1x congestion multiplier); and both nodes accept the same BLOCKS (block validation via `verify_consensus_fee_floor` is congestion-independent — a tx at the floor is valid, one below is rejected, regardless of EMA). Reuses the store.rs seeding pattern via the public snapshot path.

4. Added a `#[cfg(test)] verify_consensus_fee_floor_for_test` wrapper on `Ledger` so the sibling mempool test can exercise the real consensus block-validity gate rather than re-deriving it.

## Scope discipline

- Did NOT modify `add_block` / `consensus_fee_floor` (B4 is merged).
- Did NOT touch #581's tag-bound tightening.

## Testing

- `cargo build -p botho` — clean (pre-existing warnings only)
- `cargo test -p botho --lib mempool` — 48 passed (incl. the new test)
- `cargo test -p botho --lib consensus_fee_floor` — 6 passed
- `cargo test -p bth-cluster-tax dynamic_fee` — 16 passed

Closes #579

### DIFF
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -1648,6 +1648,19 @@ impl Ledger {
         Ok(())
     }
 
+    /// Test-only entry point to the private block-validity gate
+    /// [`Ledger::verify_consensus_fee_floor`], so tests in sibling modules
+    /// (e.g. the mempool relay-only-tightening invariant, issue #579) can
+    /// exercise the real consensus acceptance path rather than re-deriving it.
+    #[cfg(test)]
+    pub fn verify_consensus_fee_floor_for_test(
+        &self,
+        tx: &BothoTransaction,
+        block_height: u64,
+    ) -> Result<(), LedgerError> {
+        self.verify_consensus_fee_floor(tx, block_height)
+    }
+
     /// Compute the deterministic consensus fee floor for a transfer tx.
     ///
     /// Split out from [`Ledger::verify_consensus_fee_floor`] so tests (and any
--- a/botho/src/mempool.rs
+++ b/botho/src/mempool.rs
@@ -678,7 +678,18 @@ impl Mempool {
         // Count outputs with encrypted memos for fee calculation
         let num_memos = tx.outputs.iter().filter(|o| o.has_memo()).count();
 
-        // Get the current dynamic fee base (adjusts based on congestion)
+        // Get the current dynamic fee base (adjusts based on congestion).
+        //
+        // RELAY-ONLY policy: `DynamicFeeBase` carries a node-local f64 EMA (reset
+        // on restart) and a wall-clock `at_min_block_time` flag, so its output is
+        // NOT deterministic across nodes. It must NEVER enter consensus — see the
+        // type docs on `bth_cluster_tax::DynamicFeeBase` and the consensus floor
+        // `Ledger::consensus_fee_floor` (which deliberately pins the base to the
+        // neutral `CONSENSUS_FEE_BASE`, omitting congestion). Because
+        // `compute_base` is clamped to `>= base_min == CONSENSUS_FEE_BASE`, this
+        // term can only ever RAISE the local relay threshold above the consensus
+        // floor (never below it) — the relay-only-tightening invariant asserted
+        // just below where `minimum_fee` is computed.
         let dynamic_base = self.dynamic_fee.compute_base(self.at_min_block_time);
 
         // Use dynamic fee calculation for minimum (with actual output count)
@@ -707,9 +718,13 @@ impl Mempool {
         let demurrage_factor =
             self.ring_centroid_floored_factor(&ring_tags, claimed_factor, ledger)?;
 
+        // Current chain tip height. Used both for the demurrage clock below and
+        // for the relay-only-tightening invariant check (M1-B5), so the mempool
+        // and the consensus floor evaluate the demurrage at the identical height.
+        let chain_height = ledger.get_chain_state().map(|s| s.height).unwrap_or(0);
+
         let demurrage = {
             let policy = crate::monetary::mainnet_policy();
-            let chain_height = ledger.get_chain_state().map(|s| s.height).unwrap_or(0);
             // H2/B1 (issue #578, design #574): use the max-quantile order
             // statistic of ring-member ages, NOT the value-weighted mean. The
             // mean lets a spender pad the ring with fresh high-value decoys to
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge review — APPROVED

Item B5 of design #574 (M1 handling). This is a guard + docs change that enforces the relay-only-tightening invariant: mempool admission `minimum_fee` is always `>=` the deterministic `Ledger::consensus_fee_floor`. Reviewed against the four review priorities.

### No behavior change to consensus or admission — confirmed
- `add_block` / `consensus_fee_floor` untouched (the only store.rs change is the `#[cfg(test)] verify_consensus_fee_floor_for_test` wrapper).
- The guard is a `debug_assert!`, so it compiles out of release builds — no extra hot-path `consensus_fee_floor` ledger read (ring-UTXO resolution) in production. Admission still decides on `tx.fee < minimum_fee` exactly as before; the assertion has no side effect on the decision.

### False-fire risk (the priority concern) — no false-fire path found
Both sides compute their threshold from identical inputs except the congestion base:
- **Base curve:** both call `FeeConfig::default().minimum_fee_dynamic_with_outputs(Hidden, size, cluster_wealth, num_outputs, num_memos, BASE)`. Mempool passes `dynamic_base`; consensus pins `CONSENSUS_FEE_BASE == 1`. `compute_base` is clamped to `>= base_min == CONSENSUS_FEE_BASE`, so mempool base >= consensus base always.
- **cluster_wealth:** the mempool module fn and the ledger method `effective_cluster_wealth_from_outputs` are byte-for-byte identical in math (same u128 accumulation, same division order, same fail-closed-on-DB-error). No divergence.
- **Demurrage kernel:** both use `ring_elapsed_quantile` at quantile `10_000` (mempool literal; consensus `CONSENSUS_RING_AGE_QUANTILE_BPS == 10_000`), the same `demurrage_charge`, the same `mainnet_policy().demurrage_rate_bps(height)` and `blocks_per_year`.
- **Demurrage factor:** both route through `Ledger::ring_centroid_floored_factor(claimed_factor, ring_refs, &FeeConfig::default().cluster_curve)`.
- **Ring resolution:** both resolve members via `get_utxo_by_target_key` on the committed UTXO set.
- **Height:** the hoisted `chain_height` feeds both the mempool demurrage clock and the `consensus_fee_floor` call, so there is no off-by-one between the two demurrage evaluations. Good catch by the builder — this was the most plausible false-fire source.

The DB-error branch returning `true` (skip) is correct: a transient read failure is not an invariant violation.

### Docs — accurate
The RELAY-ONLY rationale on `DynamicFeeBase` (module + type doc) is correct and clear: `f64 ema_fullness` (reset on restart, non-associative FP) and wall-clock `at_min_block_time` are node-local and would fork consensus (audit cycle 6 M1) if they gated block validity. The clamp-to-`base_min` argument for why relay only tightens is sound.

### Test validity — reasonable
`relay_minimum_never_below_consensus_floor_hot_and_cold` exercises the real consensus path (`consensus_fee_floor` + `verify_consensus_fee_floor_for_test`, not a mock): hot and cold EMA both `>=` floor, cold `==` floor exactly, and block-validity is congestion-independent (at-floor accepted, floor-1 rejected). The `#[cfg(test)]` test-only wrapper is a reasonable affordance to reach the private block-validity gate from a sibling module.

### Determinism — unaffected
No new consensus input is introduced; the congestion term stays strictly relay-side.

### CI
The only failure was the pinned-nightly `rustfmt --check` (formatting-only, confined to the two touched files). Per the review guidance I ran `cargo +nightly fmt` on the touched files (import grouping / line wrapping / doc reflow only — no logic change), committed, and pushed (06aa539). All checks now pass and merge state is CLEAN.

Note: I could not execute the unit tests locally (the machine is at 100% disk / <1.5G free; the workspace needs several GB to compile). CI's `cargo build --workspace` passes but does not run this crate's `--lib` tests, so the new tests were validated by source inspection rather than execution. The invariant holds structurally as analyzed above.

Approving. Relabeling loom:review-requested -> loom:pr.

Issue #606, opened 2026-07-01
infra: coordinated testnet reset + redeploy for H1 consensus fee floor (protocol bump; successor to #323)


rjwalters:
## Summary
Successor to **#323** (the last coordinated reset, closed 2026-06-28, deployed protocol `2.0.0`). Current `main` now enforces **new block-acceptance rules that the running 2.0.0 testnet does not**, so the two are consensus-incompatible. A fresh coordinated reset + redeploy (new genesis) is required before these fixes take effect — and to keep third-party miners/relayers safe once the testnet is public.

**This is the deployment vehicle for the cycle-6 H1/H2 consensus work (design #574, now fully implemented).** Those changes are merged to `main` but undeployed; without this reset they silently sit inert.

## Why a reset is required (not a rolling upgrade)
The **H1 consensus fee floor (#578 / PR #602, merged 2026-07-01T17:55)** landed *after* #323 closed (2026-06-28T20:39). A node on current `main` rejects any block whose transfer txs fall below the deterministic consensus floor (`add_block_inner` C7 gate) — including the fee-free/under-fee blocks the existing 2.0.0 chain has been producing. New-code nodes and old-chain history are mutually invalid ⇒ no clean migration ⇒ coordinated reset.

## Consensus-affecting changes on `main` since the #323 (2.0.0) deploy
| PR | Change | Block-acceptance impact |
|---|---|---|
| **#602** (H1-B4) | deterministic consensus fee floor in `add_block` + `ring_elapsed_centroid`→`ring_elapsed_quantile@max` cutover | **NEW reject rule** — the reset-defining change |
| #601 | reject/saturate block fee-sum overflow | new reject path for pathological fee sums (benign for honest blocks) |
| #564 / #598 / #600 | M7 fail-closed double-spend (store + mempool) | store.rs path is consensus; fail-closed on DB error |

(Already deployed at 2.0.0 via #323: #580 B3 tag-inheritance, #582 B2 factor-floor helper — but the floor they feed only became *enforced* via #602, so the effective rule is new.)

## Tasks
- [x] **Bump `PROTOCOL_VERSION`** (`botho/src/network/discovery.rs:68`) from `2.0.0` → `2.1.0` (or `3.0.0`) and set `MIN_SUPPORTED_PROTOCOL_VERSION` to match, so 2.0.0 peers are disconnected (the cycle-6 protocol-version peer-disconnect mechanism). Document the bump rationale like the existing 2.0.0 comment (I4).
[...]

PR #618, opened 2026-07-03
fix(decoy): guard degenerate age band to stop young-input clamp panic

### LINKED ISSUE #611: panic in OSPEAD decoy selection: age_similarity_band clamp asserts min<=max for young real inputs
## Summary
Found live during the #606 reset smoke test (2026-07-02) on the fresh 3.0.0 testnet.

`age_similarity_band` (botho/src/decoy_selection.rs:278) explicitly documents the degenerate case:

> A degenerate real age (below the confirmation floor) can produce `min_age > max_age`, in which case no candidate is in-band and **callers fall back** to nearest-age selection.

But the caller at decoy_selection.rs:712 does:
```rust
let target_age_blocks = (target_age as u64).clamp(min_age, max_age);
```
Rust's `clamp` **asserts** `min <= max` — so instead of falling back, the process panics:
```
thread 'main' panicked at library/core/src/cmp.rs:1090:9:
assertion failed: min <= max
```

## Repro (live)
With `AGE_SIMILARITY_SPREAD_BPS = 1_000` (10%) and `MIN_DECOY_AGE_BLOCKS = 10`:
- real input age = 0 (spend a just-confirmed UTXO): min_age = max(0-0, 10) = 10, max_age = 0 + 0 = 0 → clamp(10, 0) → panic.
- Any real age < ~10 blocks hits it (need `age * 1.1 >= 10` to be safe).

Hit in practice by building a spend of a 0-block-old faucet dispensation via `Wallet::create_private_transaction` on the fresh chain.

## Impact
- Any tx build whose real input is younger than ~10 blocks panics the calling thread.
- In-node exposure: the faucet's `handle_faucet_request` calls `create_private_transaction` on the RPC path — if the faucet's freshest UTXO is selected while young (e.g., high faucet traffic right after reset, spending its own change), the RPC worker panics. #583's InsufficientDecoys cold-start guard catches the typed error path but NOT this panic.
- Wallet-side exposure: same selector is used by wallet flows (#596).

## Fix sketch
In the phase-1 in-band draw, guard the degenerate band as the doc comment promises:
```rust
if min_age > max_age {
    // degenerate band: no candidate can be in-band; skip phase 1
    // and fall through to nearest-age (phase 2) selection
    break; // or bypass the in-band loop entirely
}
```
or use `target_age_blocks = target_age.min(max_age).max(min_age)` only when `min_age <= max_age`. Add a regression test spending an age-0 input on a fresh chain (the #583 cold-start scenario plus ~30 warm blocks so decoys exist but the real input is new).

Related: #596 (age-similar decoy selection, introduced the band), #583 (cold-start InsufficientDecoys guard — panics bypass it).

Found during: #606 reset verification.

rjwalters:
## Summary

Spending a real input younger than `MIN_DECOY_AGE_BLOCKS` (age 0-9) panicked the calling thread inside `GammaDecoySelector::select_decoys_for_input`. `age_similarity_band` returns an inverted band (`min_age = 10 > max_age = real_age`) for such inputs — its lower bound is floored at the confirmation depth while its upper bound tracks the young real age. The phase-1 in-band loop then called `u64::clamp(min_age, max_age)` unconditionally, and Rust's `clamp` asserts `min <= max`, so the process panicked (`assertion failed: min <= max`) instead of falling back to nearest-age selection as `age_similarity_band`'s own doc comment promises.

On the live 3.0.0 testnet (found during the #606 reset smoke test) this crashed the faucet RPC worker whenever its freshest change UTXO was selected while young.

## Changes

- `botho/src/decoy_selection.rs`: wrap the entire phase-1 in-band draw in `if min_age <= max_age { ... }`. Degenerate bands skip phase 1 and fall through to the existing, unchanged phase-2 nearest-age fallback, which fills the ring from confirmed candidates. No RNG is drawn in the degenerate case, so the non-degenerate path is byte-identical.
- Added regression tests `test_young_real_input_no_panic` (ages 0, 9, 10) and `test_select_decoys_deterministic_with_seed`.

| AC-1 no panic on degenerate band | ✅ | `test_young_real_input_no_panic` calls with `real_input_age` 0 and 9; both return `Ok` (previously panicked) |
| AC-2 phase-2 fills the ring | ✅ | Test asserts `decoys.len() == count` for ages 0 and 9, every decoy age ≥ `MIN_DECOY_AGE_BLOCKS` |
| AC-3 non-degenerate path byte-identical | ✅ | Fix only adds an outer `if min_age <= max_age` guard; the loop body and RNG draw order are unchanged. `test_age_similar_selection_stays_in_band` and `test_honest_overcharge_under_max_within_314_bound` still pass |
| AC-4 regression test at age 0 | ✅ | `test_young_real_input_no_panic` age-0 branch |
| AC-5 boundary tests at ages 9 and 10 | ✅ | Age 9 = last degenerate (phase 2); age 10 = first non-degenerate, band `[10, 11]`, asserts every decoy in-band (phase 1 ran) |
| AC-6 determinism preserved | ✅ | `test_select_decoys_deterministic_with_seed` runs twice with `ChaCha8Rng::seed_from_u64(...)` and asserts identical decoy sets/order |
| AC-7 faucet RPC smoke | ⏸️ Deferred to operator | Live-testnet step; unit tests fully cover the code path. Recommend spending an age-0/1 faucet UTXO via the faucet RPC on the 3.0.0 testnet and confirming no panic in node logs |

## Test Plan

`cargo build -p botho` — clean. `cargo test -p botho --lib decoy_selection` — 23 passed, 0 failed (2 new tests included).

## Scope / Safety

Liveness-only. Neither `consensus/validation.rs`, `block.rs`, nor any `add_block` path references decoy selection, so there is no consensus-rule, block-validation, or network-compatibility change and no fork risk. Single production caller affected; the `botho-wallet` crate does not contain `age_similarity_band`.

Closes #611

### DIFF
--- a/botho/src/decoy_selection.rs
+++ b/botho/src/decoy_selection.rs
@@ -707,25 +707,39 @@ impl GammaDecoySelector {
         // Phase 1: draw from WITHIN the band. Sample a target age from the gamma
         // spend distribution, clamp it into the band, and pick the nearest-aged
         // in-band candidate. This keeps the realistic within-band distribution.
-        for _ in 0..count {
-            let target_age: f64 = gamma.sample(rng);
-            let target_age_blocks = (target_age as u64).clamp(min_age, max_age);
+        //
+        // Skip phase 1 entirely when the band is degenerate (`min_age > max_age`).
+        // `age_similarity_band` produces this for a real input younger than
+        // `MIN_DECOY_AGE_BLOCKS`: the lower bound is floored at the confirmation
+        // depth while the upper bound tracks the (young) real age, so no candidate
+        // can ever be in-band. Its doc comment promises callers fall back to
+        // nearest-age selection here — this is that fallback. Entering the loop
+        // would hit `u64::clamp(min_age, max_age)`, which panics when
+        // `min_age > max_age` (issue #611). No RNG is drawn in the degenerate
+        // case, so the non-degenerate path is unaffected and stays deterministic.
+        if min_age <= max_age {
+            for _ in 0..count {
+                let target_age: f64 = gamma.sample(rng);
+                let target_age_blocks = (target_age as u64).clamp(min_age, max_age);
 
-            let best = eligible
-                .iter()
-                .filter(|c| {
-                    c.age_blocks >= min_age
-                        && c.age_blocks <= max_age
-                        && !used_keys.contains(&c.output.target_key)
-                })
-                .min_by_key(|c| (c.age_blocks as i64 - target_age_blocks as i64).unsigned_abs());
+                let best = eligible
+                    .iter()
+                    .filter(|c| {
+                        c.age_blocks >= min_age
+                            && c.age_blocks <= max_age
+                            && !used_keys.contains(&c.output.target_key)
+                    })
+                    .min_by_key(|c| {
+                        (c.age_blocks as i64 - target_age_blocks as i64).unsigned_abs()
+                    });
 
-            if let Some(candidate) = best {
-                selected.push(candidate.output.clone());
-                used_keys.push(candidate.output.target_key);
-            } else {
-                // Band exhausted; remaining slots handled by the fallback below.
-                break;
+                if let Some(candidate) = best {
+                    selected.push(candidate.output.clone());
+                    used_keys.push(candidate.output.target_key);
+                } else {
+                    // Band exhausted; remaining slots handled by the fallback below.
+                    break;
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed as the Loom Judge for issue #611 (loom:urgent OSPEAD young-input clamp panic). This is a clean, minimal, correctly-scoped liveness fix. Approving.

### Highest-stakes check — AC-3 non-degenerate path is byte-identical (verified)
`git diff -w` on `botho/src/decoy_selection.rs` shows the *only* production-code change is wrapping the existing phase-1 loop in `if min_age <= max_age { ... }`, plus one rustfmt reflow of the `.min_by_key(...)` closure onto multiple lines (semantically identical — forced by the extra indentation level). With the guard `true`:
- the loop body is character-for-character the pre-fix code,
- `gamma.sample(rng)` is still called once per iteration in the same order,
- selection order and `used_keys` bookkeeping are unchanged.
No refactor perturbs RNG consumption. Normal-transaction decoy selection is unaffected network-wide.

### AC-1/AC-2 — degenerate path (verified)
With `min_age > max_age` the entire `if` block is skipped, so **zero RNG is drawn in phase 1** (no `gamma.sample`) and execution falls straight through to the unchanged phase-2 nearest-age fallback. Since pre-fix behavior here was a panic, there is no compatibility constraint on the degenerate path. Phase 2 seeds `used_keys` from `exclude_keys` and filters `!used_keys.contains(...)`, so it respects both `exclude_keys` and inter-decoy uniqueness while filling the full ring from confirmed (`age >= MIN_DECOY_AGE_BLOCKS`) candidates; it returns a typed `InsufficientCandidates` error rather than a short ring if the pool is thin.

### AC-5 — boundary semantics (verified)
Tests pin both sides: age 9 -> band `[10, 9]` degenerate -> phase 2; age 10 -> band `[10, 11]` non-degenerate -> phase 1, with an explicit `assert_eq!((min10, max10), (10, 11))` and an in-band assertion over 50 draws against a pool stocked only at ages 10/11.

### AC-6 — determinism (verified)
`test_select_decoys_deterministic_with_seed` runs the non-degenerate path twice with `ChaCha8Rng::seed_from_u64(...)` on identical inputs and asserts full `Vec<[u8;32]>` equality (membership AND order). Combined with the byte-identical diff, RNG-draw ordering is provably preserved.

### Local verification
- `cargo test -p botho --lib decoy_selection`: **23 passed, 0 failed**, including the new `test_young_real_input_no_panic` and `test_select_decoys_deterministic_with_seed`, plus the pre-existing `test_age_similar_selection_stays_in_band` and `test_honest_overcharge_under_max_within_314_bound`.
- `cargo fmt -p botho -- --check`: clean.
- CI: all checks green (rustfmt, `cargo build --workspace (Linux)`, Fuzz Build Guard, benchmarks). mergeStateStatus **CLEAN**.

### Notes (non-blocking)
- AC-7 (live faucet RPC smoke on the 3.0.0 testnet) is correctly deferred to the operator by design; the unit tests fully cover the code path.
- Scope is liveness-only: no `consensus/validation.rs`, `block.rs`, or `add_block` path references decoy selection — no consensus/fork risk. Confirmed.

LGTM.

PR #627, opened 2026-07-05
feat(cluster-tax): log-domain cluster-factor curve in picocredits (#626 PR 1/3)

rjwalters:
## Summary

PR 1 of 3 for #626. Replaces `ClusterFactorCurve`'s linear-domain sigmoid (calibrated in simulator units, `w_mid=10M`) with the **normative log-domain construction** from the architect's sign-off-approved spec (last comment on #626), consumed in on-chain **picocredits**. This fixes the step-function collapse where all 463 live-testnet clusters were pinned at ~6x, and de-quantizes the factor (old code snapped to whole multipliers and capped at 5000, `fee_curve.rs:675`).

Implemented **exactly** as specified — the integer algorithm IS the curve definition.

## What changed

**`ClusterFactorCurve` (consensus curve)**
- `log2_fp(w)`: `127 - leading_zeros` integer part + top-16 mantissa bits (Q16), pure `const fn`, no floats
- `z_fp = log2_fp(W) - LOG2_WMID_FP`; `x = (z_fp * 1000) / 131_072` (i64 truncating division, normative); shared 7-knot sigmoid LUT with **tails changed to exact saturation** (`x ≤ −6000 → 0`, `x ≥ 6000 → 65536`)
- output `1000 + (5000 * σ) / 65536`; `W = 0 → 1000` exactly; midpoint (100k BTH) `→ 3500` exactly
- fields `w_mid`/`steepness` → `w_mid_pico: u128`/`log_width_fp: u64`
- **compile-time locks**: `W_MID_PICO == 100_000 * PICO_PER_BTH` (=1e17) and `LOG2_WMID_FP == 3_695_429`

**API widened u64 → u128.** `ClusterFactorCurve::factor` and `ZkFeeCurve::factor` take u128 picocredit wealth. Call sites (mempool / lottery / demurrage / store) pass `w as u128` for now; `FeeConfig` wrappers stay u64 and cast internally. **The `cluster_wealth_db` u128 widening is PR 2 and is NOT in this PR.**

**`ZkFeeCurve` re-anchored** to 5 segments at power-of-two multiples of `W_MID_PICO`: boundaries `[0, W>>8, W>>3, W<<3, W<<8, u128::MAX]`, factors `[1000, 1090, 2071, 4928, 5909, 6000]` (sampled from the sigmoid). Boundaries/`SegmentParams` widened to u128; ZK prover (`committed_tags.rs`, Phase-2 provisional) casts `wealth as u128` and saturates diffs.

[...]

Issue #626, opened 2026-07-04
fix(M2-prereq): cluster-factor curve w_mid is in simulator units, not picocredits — on-chain sigmoid collapses to a step function


rjwalters:
## Summary
Found while evaluating #605 (2026-07-04). The progressive-fee sigmoid is effectively a **step function at dust level** on the live chain, because the curve midpoint constant was calibrated in simulator wealth units but consumes on-chain picocredit values.

## The mismatch
- On-chain accumulation: `update_cluster_wealth_for_output` (botho/src/ledger/store.rs:2218) stores `amount × weight / TAG_WEIGHT_SCALE` — **picocredits** (1 BTH = 10^12 pico).
- Curve midpoint: `ClusterFactorCurve::default_params()` (cluster-tax/src/fee_curve.rs:639) sets `w_mid = 10_000_000` = **0.00001 BTH**, steepness 5M pico.
- Consequence: a single fully-tagged transfer of 0.001 BTH is already 100× past the midpoint → any economically active cluster is pinned at ~6× immediately. The intended smooth 1×–6× progressivity never exists on-chain; behavior is binary (background 1× vs tagged 6×).
- The Gini validation (#314, experiments/ANALYSIS.md) ran the same `default_params()` inside `cluster-tax/src/bin/sim.rs`'s **abstract unit space**, where 10M is mid-scale wealth — so the validated curve shape has never actually operated on-chain.

## Consumers affected (all read the same curve)
- Mempool progressive fee (`mempool.rs cluster_factor`)
- Consensus fee floor C7 (`ring_centroid_floored_factor`, #602)
- Tilted lottery weighting (`consensus/lottery.rs:1296`)

## Why this blocks #605
The decay-vs-document decision (#605) is meaningless under a step function: decay wouldn't unpin factors (decayed wealth still ≫ w_mid), and 'document' would be documenting an accident. Calibration must be fixed/decided first.

## What 'fixed' needs
1. Decide the intended real-unit midpoint (e.g. what BTH-denominated cluster size should sit at 3.5×?) — economic design choice, informed by supply (~611M BTH) and expected cluster-size distribution.
2. A consensus-safe recalibration (constants only — but changing them changes the fee floor ⇒ consensus-breaking ⇒ rides a testnet reset; free pre-mainnet).
3. Re-run the Gini sweep with unit-consistent calibration (this run doubles as the #605 empirical input).
[...]

PR #607, opened 2026-07-02
fix(M3): cluster-wealth rebuild uses saturating_add (determinism parity)

rjwalters:
## Summary

Fixes a determinism/parity hazard between the two cluster-wealth accumulation paths.

- Incremental path `update_cluster_wealth_for_output` (botho/src/ledger/store.rs) clamps per-cluster wealth with `current.saturating_add(contribution)`.
- Rebuild path `rebuild_cluster_wealth_index` previously used an unchecked `+=` on the in-memory `HashMap<u64, u64>` accumulator.

Because progressive-fee and demurrage inputs read `cluster_wealth_db`, a rebuilt node must produce **byte-identical** cluster wealth to an incrementally-accumulated one. At the u64 saturation boundary the unchecked `+=` would panic in debug or wrap in release, diverging from the incremental path.

## Change

`rebuild_cluster_wealth_index` now uses `saturating_add`, matching the incremental path exactly:

```rust
let e = cluster_wealths.entry(entry.cluster_id.0).or_insert(0u64);
*e = e.saturating_add(contribution);
```

Below the saturation boundary the result is unchanged (saturating_add == add when no overflow).

## Test

Adds `test_rebuild_matches_incremental_cluster_wealth_at_saturation`:
- Drives the incremental path (`update_cluster_wealth_for_output`) and the rebuild path (`rebuild_cluster_wealth_index`) over the same set of outputs.
- Cluster 1 receives three ~0.75·u64::MAX contributions, forcing saturation on both paths.
- Cluster 2 stays below the boundary to confirm no behavior change.
- Asserts `get_all_cluster_wealth()` is byte-identical across both ledgers, cluster 1 == u64::MAX on both, cluster 2 == exact unsaturated sum on both.

Note: the saturating_add clamp-at-MAX property is order-independent for non-negative contributions, so the result is robust to utxo_db iteration order.

## Acceptance criteria

1. rebuild_cluster_wealth_index uses saturating_add — done.
2. Test: rebuild vs incremental produce identical cluster_wealth_db contents, including a near-saturation-boundary case — done.
3. No behavior change below the saturation boundary — done (verified by cluster 2 assertions and existing cluster-wealth tests still passing).

## Test results

- `test_rebuild_matches_incremental_cluster_wealth_at_saturation` ... ok
- All 10 `cluster_wealth`-matching tests pass; no regressions.

Closes #604

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #604, opened 2026-07-01
fix(M3): cluster-wealth rebuild uses unchecked += vs incremental saturating_add (determinism parity)


rjwalters:
## Summary
Cycle-6 audit finding **M3** (carried, re-verified 2026-07-01 against current `main`). Follow-on from the #555 cycle-6 Highs tracker.

The two cluster-wealth accumulation paths use **different overflow arithmetic**, so a rebuilt node can diverge from an incrementally-accumulated one at the saturation boundary — a determinism/fork hazard for a consensus-adjacent value.

- **Incremental path** — `update_cluster_wealth_for_output` (`botho/src/ledger/store.rs:2242`): `let new_wealth = current.saturating_add(contribution);`
- **Rebuild path** — `rebuild_cluster_wealth_index` (`botho/src/ledger/store.rs:2448`): `*cluster_wealths.entry(entry.cluster_id.0).or_insert(0) += contribution;` (unchecked `+=` on `u128`).

At `u128` the boundary is astronomically far (the accumulator was widened from the `u64` the audit originally flagged), so this is **low severity today** — but it is a free correctness/parity fix: a node that rebuilds its index (`rebuild_cluster_wealth_index`) must produce byte-identical wealth to one that accumulated incrementally, or progressive-fee/demurrage inputs (which read `cluster_wealth_db`) could differ between nodes.

## Fix
Use `saturating_add` in the rebuild path so both paths have identical overflow semantics:
```rust
let e = cluster_wealths.entry(entry.cluster_id.0).or_insert(0u128);
*e = e.saturating_add(contribution);
```

## Acceptance criteria
- [ ] `rebuild_cluster_wealth_index` uses `saturating_add` (or the identical checked/saturating discipline as `update_cluster_wealth_for_output`).
- [ ] Test: rebuild vs incremental produce identical `cluster_wealth_db` contents, including a case engineered near the saturation boundary (inject large synthetic contributions).
- [ ] No behavior change below the saturation boundary (the common case).

## Notes
Node-local index arithmetic; deterministic. Related: M2 (#605, why wealth is cumulative-forever and thus can approach the boundary), #555 (parent tracker).

PR #628, opened 2026-07-05
feat(ledger): widen cluster_wealth_db accumulator u64 → u128 (#626 PR 2/3)

rjwalters:
## Summary

PR 2 of 3 for #626. Widens the per-cluster wealth accumulator (`cluster_wealth_db`) from **u64 → u128** so cumulative tagged wealth can exceed the former `u64::MAX` pico ceiling (**18.44M BTH**), which the log-domain factor curve (PR 1, #627, merged) now maps across the full supply range.

**Consensus-breaking — rides the protocol-4.0.0 fresh-genesis testnet reset** (bundled per the #606 precedent). No in-place migration: fresh genesis is assumed.

### Storage & determinism contract (`botho/src/ledger/store.rs`)
- **16-byte LE u128 serialization**, pinned in a test and centralized in one `decode_cluster_wealth` helper.
- **Reject-legacy (fail closed, M7 discipline):** a wrong-width value (e.g. a legacy 8-byte `u64`) is a hard `LedgerError` on read, never silently reinterpreted.
- **Incremental** (`update_cluster_wealth_for_output`) and **rebuild** (`rebuild_cluster_wealth_index`) paths keep full-u128 math with `saturating_add` — byte-identical semantics preserved at the new width (M3 / #604 / #607 parity contract).
- Readers now return u128: `get_cluster_wealth`, `get_all_cluster_wealth`, `compute_cluster_wealth_for_utxos` / `ClusterWealthInfo`, the consensus lottery-tilt path, and the snapshot create/load path (`UtxoSnapshot` cluster wealth widened to u128).

### Consumer seam (u128 accumulator, u64 fee/centroid APIs until PR 3)
`FeeConfig::cluster_factor` and `ring_centroid_implied_factor` stay u64 (their widening is PR 3), so call sites feeding them (mempool relay fee, ring-centroid consensus fee floor, CLI send estimate) **clamp to u64 deterministically** — the ceiling simply persists there until PR 3, no fork. The **lottery tilt** uses the un-clamped u128 directly (`factor()` accepts u128 as of PR 1).

### RPC (u128 exceeds JS number range → string-encoded, per the `totalMined` precedent)
`cluster_getWealth`, `cluster_getWealthByTargetKeys`, `cluster_getAllWealth`, `estimateFee`, `entropy_estimateFee` now emit wealth fields as **decimal strings** and accept the `cluster_wealth` param as a string (numeric still accepted for back-compat).

### RPC consumers checked
- **Web** (`web/packages/adapters/src/{local,remote,types}.ts`): already model `clusterWealth` as `bigint` and send it via `.toString()`, reading back only the fee — **already string-compatible, no change needed.**
- **Mobile:** no consumers of these wealth fields found.
- **Wallet** (`botho-wallet`): `FeeEstimate` / `NetworkFeeRate` do not deserialize any cluster-wealth field, so no parser change; the estimateFee mock fixture was updated to the string contract for fidelity.

## Tests
- Rebuild-vs-incremental **u128 parity summing PAST `u64::MAX`** (exact sum, no saturation; byte-compares full DB contents).
- **16-byte LE serialization pin.**
- **Legacy 8-byte rejection** on every reader (single + iterate).
- **u128-boundary `saturating_add` discipline** (astronomically unreachable, kept for discipline).

## Verification
`cargo build --workspace`, `cargo test -p botho --lib ledger` (+ cluster-wealth/snapshot/mempool/rpc), `cargo test -p botho-wallet`, `cargo fmt --all -- --check` — all green.

Updates #626. PR 3 follows (widen the remaining `FeeConfig` / `ring_centroid_implied_factor` / transaction-crate consumer APIs to u128 end-to-end; ZkFeeCurve re-anchoring; sim harness).

### DIFF
--- a/botho-wallet/src/rpc_pool.rs
+++ b/botho-wallet/src/rpc_pool.rs
@@ -810,7 +810,9 @@ mod tests {
             "clusterFactorDisplay": "1.00x",
             "recommendedFee": 2000u64,
             "highPriorityFee": 4000u64,
-            "clusterWealth": 0u64,
+            // clusterWealth is now a STRING (u128 pico, #626); FeeEstimate does
+            // not consume it, but the fixture mirrors the real string contract.
+            "clusterWealth": "0",
             "params": {
                 "amount": 100u64,
                 "txType": "transfer",
--- a/botho/src/commands/send.rs
+++ b/botho/src/commands/send.rs
@@ -97,11 +97,14 @@ pub fn run(
                 .entries
                 .iter()
                 .map(|entry| {
+                    // get_cluster_wealth is now u128 (16-byte accumulator, #626).
                     let global = ledger.get_cluster_wealth(entry.cluster_id.0).unwrap_or(0);
-                    entry.weight as u128 * global as u128
+                    entry.weight as u128 * global
                 })
                 .sum();
-            (weighted / bth_transaction_types::TAG_WEIGHT_SCALE as u128) as u64
+            // estimate_typical_fee takes a u64 wealth (widening is PR 3); clamp.
+            (weighted / bth_transaction_types::TAG_WEIGHT_SCALE as u128).min(u64::MAX as u128)
+                as u64
         })
         .max()
         .unwrap_or(0);
--- a/botho/src/ledger/snapshot.rs
+++ b/botho/src/ledger/snapshot.rs
@@ -111,7 +111,7 @@ impl UtxoSnapshot {
         chain_state: ChainState,
         utxos: Vec<Utxo>,
         key_images: Vec<([u8; 32], u64)>, // (key_image, spent_height)
-        cluster_wealth: Vec<(u64, u64)>,  // (cluster_id, wealth)
+        cluster_wealth: Vec<(u64, u128)>, // (cluster_id, wealth) — u128 pico (#626)
     ) -> Result<Self, SnapshotError> {
         // Compute UTXO Merkle root
         let utxo_hashes: Vec<[u8; 32]> = utxos
@@ -178,7 +178,7 @@ impl UtxoSnapshot {
     }
 
     /// Decompress and deserialize the cluster wealth data.
-    pub fn get_cluster_wealth(&self) -> Result<Vec<(u64, u64)>, SnapshotError> {
+    pub fn get_cluster_wealth(&self) -> Result<Vec<(u64, u128)>, SnapshotError> {
         let decompressed = decompress(&self.cluster_wealth_data)?;
         bincode::deserialize(&decompressed).map_err(|e| SnapshotError::Serialization(e.to_string()))
     }
@@ -444,7 +444,7 @@ mod tests {
     fn test_snapshot_creation() {
         let utxos = vec![test_utxo(1), test_utxo(2), test_utxo(3)];
         let key_images = vec![([1u8; 32], 100u64), ([2u8; 32], 200u64)];
-        let cluster_wealth = vec![(1u64, 1000u64), (2u64, 2000u64)];
+        let cluster_wealth = vec![(1u64, 1000u128), (2u64, 2000u128)];
 
         let snapshot = UtxoSnapshot::new(
             1000,
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -201,13 +201,48 @@ pub struct Ledger {
     /// tx_index u32) Maps transaction hashes to their location for fast
     /// lookups (exchange integration).
     tx_index_db: Database<Bytes, Bytes>,
-    /// cluster_wealth: cluster_id (8 bytes) -> wealth (8 bytes)
+    /// cluster_wealth: cluster_id (8 bytes) -> wealth (16 bytes, u128 LE)
     /// Tracks total value per cluster tag across all UTXOs for progressive fee
     /// calculation. Note: This is an approximation - with ring signatures,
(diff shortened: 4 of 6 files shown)


rjwalters:
✅ **Approved!** — u64→u128 cluster-wealth widening (PR 2/3, #626)

Thorough review against the determinism contract. All priorities check out.

## M3 parity discipline (highest stakes) — PASS
- Incremental (`update_cluster_wealth_for_output`) and rebuild (`rebuild_cluster_wealth_index`) use **identical** contribution math at full u128 (`amount×weight/SCALE`, no down-cast) and `saturating_add` on u128 in both paths.
- Order-independence holds: with non-negative addends, `saturating_add` yields `min(true_sum, u128::MAX)` regardless of accumulation order, so the incremental (output-order) and rebuild (utxo_db-iteration-order) paths cannot diverge.
- `test_rebuild_matches_incremental_cluster_wealth_past_u64_max` byte-compares the full DB via sorted `get_all_cluster_wealth` (decode is bijective with the 16-byte value → equal u128 ⟺ equal bytes) **and** asserts cluster 1 == `(u64::MAX/4*3)*3` which is `> u64::MAX` (no saturation). Genuinely exercises the past-u64 case. Verified passing locally.

## Serialization contract — PASS
- 16-byte LE pinned by `test_cluster_wealth_serialization_is_16_byte_le_u128` (value `> u64::MAX`).
- Reject-legacy is fail-closed (M7 discipline): centralized `decode_cluster_wealth` returns a hard `LedgerError` on any non-16-byte value — no truncation, no zero-pad fallback. `test_cluster_wealth_rejects_legacy_8_byte_value` covers both the single reader and the iterate reader; because every reader (single/iterate/snapshot-create/lottery-scan) routes through the one helper, all readers fail closed by construction. Fresh-genesis-under-4.0.0 assumption documented at the helper and the db field.

## u64 clamp seam (ruling: ACCEPT the deferral) — PASS
- (a) Every clamp is a pure `.min(u64::MAX as u128)` — deterministic, no env dependence, identical on every node (send.rs, mempool.rs, `ring_centroid_floored_factor`, both RPC estimate handlers). Lottery tilt uses the un-clamped u128.
- (b) No consensus-forking inconsistency: the lottery-vs-fee-floor width mismatch only affects a **single** cluster whose cumulative wealth exceeds the u64 pico ceiling (18.44M BTH). Every node clamps the fee-floor input identically, so there is no cross-node divergence — the only effect is a deterministic, network-wide conservative under-statement of the C7 floor factor above that ceiling. That is a policy conservativeness, not a fork.
- (c) Honestly documented at every clamp site and in the PR body's "Consumer seam" section.
- **Ruling:** the 18.44M-BTH single-cluster ceiling is astronomically far for the testnet cycle, so deferring seam-closing is acceptable. The tracking requirement is already met: the PR body explicitly scopes PR 3 to "widen the remaining FeeConfig / ring_centroid_implied_factor / transaction-crate consumer APIs to u128 end-to-end" — so the seam-closing is not silently deferred to the sim harness; it is a named PR 3 deliverable. Please keep that line in PR 3's description so it doesn't drift.

## Overflow audit at u128 — PASS
- Mempool: both operands are clamped to `u64::MAX` before `value_fraction * global_wealth`, and `value_fraction ≤ output.amount ≤ u64::MAX`, so each term `< 2^128`. This exactly restores the pre-PR bound (global_wealth was a u64 before), so no regression in the accumulation sum either.
- store.rs centroid/lottery paths (`weight × wealth`): `weight ≤ TAG_WEIGHT_SCALE = 2^20`, real wealth ≤ total supply (~611M BTH ≈ 2^69 pico) ⇒ ~2^89 ≪ 2^128; `saturating_add` guards the sums. Overflow would require wealth ~14 orders of magnitude beyond total supply — unreachable. The only path that seeds `u128::MAX` (the saturation test) drives the incremental accumulator, whose per-term math is `amount×weight/SCALE ≤ 2^84`. No reachable overflow.

## RPC string encoding — PASS
- All 5 handlers emit wealth as decimal strings; `parse_cluster_wealth_param` accepts string-or-number (back-compat). Wallet fixture `"clusterWealth": "0"` mirrors the node's actual `wealth.to_string()` emission for 0 (the #610 fixture-fidelity lesson). Spot-checked `web/packages/adapters/src/local.ts`: sends `clusterWealth?.toString()` and reads back only `data.fee` — genuinely no change needed. `botho-wallet` rpc_pool fixture test (`test_fee_estimate_deserializes_from_node_json`) passes with the string contract.

## Flaky-test investigation
Ran `cargo test -p botho --lib rpc` twice: **88 passed / 0 failed** both times. Could not reproduce the 2 intermittent failures the builder mentioned. The rpc/mod.rs diff is purely string-encoding of wealth fields, fully covered by the passing tests. Consistent with pre-existing intermittent infra flakes, not introduced here.

## Test / CI results
- `cargo build --workspace`: clean
- `cargo test -p botho --lib ledger`: 53 passed / 0 failed
- `cargo test -p botho --lib cluster_wealth`: 12 passed / 0 failed (parity, 16-byte pin, legacy-rejection, u128-saturation all green)
- `cargo test -p botho-wallet` (incl. lib): 112 passed / 0 failed
- `cargo fmt --all -- --check`: clean
- `gh pr checks 628`: all pass (conditional fuzz jobs correctly skip on PRs); mergeStateStatus **CLEAN**

Solid, disciplined work — the fail-closed reader, the parity test that actually crosses u64::MAX, and the honestly-documented seam are exactly right.

Issue #605, opened 2026-07-01
design(M2): cluster-wealth is monotonic cumulative volume — decay schedule vs accept-and-document


rjwalters:
## Summary
Cycle-6 audit finding **M2** (re-verified 2026-07-01 against current `main`). Follow-on design item from the #555 cycle-6 Highs tracker — the one re-check item from #555's scope that is neither closed nor mechanically fixable.

`cluster_wealth_db` is **monotonically non-decreasing**: every output ever created adds to a cluster's tracked wealth (`update_cluster_wealth_for_output`, `botho/src/ledger/store.rs:2218`), and **nothing ever decrements it** — confirmed by code search (no decay/decrement/sub path exists). Decrement is essentially impossible under ring privacy (you can't tell when tagged value is spent). So "cluster wealth" is really *lifetime cumulative tagged volume*, not current holdings.

## Why it matters
- Active-commerce clusters ratchet monotonically toward the max progressive-fee factor (~6x) regardless of their actual current holdings — the progressive fee increasingly mis-prices them.
- It interacts with H2/#581: cumulative-forever wealth is what makes decoy-sourced tag inflation a (bounded) griefing lever, and it's why the M3 saturation boundary is theoretically reachable at all.
- Deterministic, so **economic-model-only** — not a fork/consensus-safety bug. But it shapes the validated Gini mechanism, so the mispricing compounds over time.

## The decision this issue asks for
One of:
1. **Decay schedule** — introduce a deterministic, integer, consensus-safe decay of cluster wealth over time/height (e.g. exponential half-life in block heights) so wealth tracks recent activity rather than all-time volume. Must be a pure function of block+chain state (no float, no node-local state), and re-validated against the Gini sweep (#314) so it doesn't break redistribution.
2. **Accept and document** — declare "cumulative tagged volume" the intended semantic, document it in `docs/design/cluster-tilted-redistribution.md`, and confirm via simulation that the monotonic ratchet doesn't degrade the validated Gini outcome at realistic horizons.

## Deliverable
An architect design proposal recommending (1) or (2) with rationale, the determinism contract if (1), and the empirical check to run. No code until sign-off (mirrors the #574 process). Consensus-breaking if (1) → rides the next testnet reset.

## Related
#555 (parent), #581 (H2 residual, same cumulative-wealth root), M3 (#604, saturation parity), #314 (Gini re-validation).

rjwalters:
**Champion Review: LEFT FOR HUMAN DECISION**

[...]

PR #629, opened 2026-07-05
feat(cluster-tax): close u64 cluster-wealth seam + real-curve M2 sim harness (#626 PR 3/3)

### LINKED ISSUE #626: fix(M2-prereq): cluster-factor curve w_mid is in simulator units, not picocredits — on-chain sigmoid collapses to a step function

rjwalters:
## Summary

**PR 3 of 3 for #626.** Closes the u64 seam PR 2 (#628) left open and closes the sim-harness gap from spec §7. **Resumes a partial build** — the prior builder hit its session limit ~174 tool-calls in; this PR verified, kept, and finished that work.

**Consensus-breaking (C7 fee-floor width) — rides the protocol-4.0.0 fresh-genesis testnet reset**, bundled with PR1 (#627, log-domain curve) and PR2 (#628, u128 accumulator).

## A. Close the u64 seam (end-to-end u128 wealth)

The log-domain curve (PR1) and the u128 accumulator (PR2) were still throttled by u64 fee/factor APIs. This widens them and removes PR2's deterministic clamps:

- `FeeConfig::{cluster_factor, estimate_typical_fee, estimate_fee_with_outputs, compute_fee*, …}` now take `cluster_wealth: u128`.
- `ring_centroid_implied_factor(&[(u64, u128)])` — the **C7 consensus fee-floor** path (#602) in `ledger/store.rs` — takes u128 member wealth.
- **Removed PR2's `.min(u64::MAX as u128)` clamps** at the three named sites: mempool relay fee (`mempool.rs`), C7 ring-centroid floor (`store.rs`), CLI estimate / RPC `estimateFee` (`send.rs`, `rpc/mod.rs`).
- **Re-derived the multiply bound each clamp protected** (explicit comments): `value` (u64, ≤1.8e19) × cumulative `wealth` (u128, → full supply ~6.11e20 pico) exceeds u128 only at astronomically distant wealth, so the centroid math uses `saturating_mul`/`saturating_add`, pinning overflow to `u128::MAX` → factor 6000 (max floor) — the conservative anti-gaming direction, **identical on every node → no fork**. Deterministic, no floats.

Note: `transaction/core`'s `ClusterWealthProvider` trait (a separate in-memory library API, no consensus callers — the live C7 path is `store.rs`) keeps its u64 lookup; those call sites cast `as u128` at the FeeConfig boundary. Full widening of that trait is out of scope for the three named seam sites.

## B. Sim-harness gap closure (spec §7)

New library harness `cluster-tax/src/simulation/m2.rs` exercises the **REAL production `ClusterFactorCurve`** (not the #314 hardcoded 1.0/2.0/6.0 factors): population declared in **BTH**, converted via `PICO_PER_BTH` at the curve boundary, per-agent factors derived from simulated cumulative cluster wealth through the same compiled `factor()` the node runs.

- Legacy `add_owner_with_factor` (the #314 entry point) **kept but marked deprecated**; new work uses `add_owner_via_production_curve`.
- **#605 run matrix as CLI subcommands** (`--help` + `experiments/M2_RUNBOOK.md`):
  - `m2-cumulative` (run set 1): 10/20yr horizons, honest + gamed, emits Δgini vs >0.05, merchant mispricing, whale progressivity.
  - `m2-decay` (run set 2): deterministic epoch-halving (`w >>= 1` per epoch, half-lives {2,5,10}yr — pure function of height, M3 lesson), plus **wash-trading evasion** (<20% gate; prior art 94–99%) and **privacy ring-ID rate** (<50% gate; prior art 78.7%) per `experiments/ANALYSIS.md`.
- **Full suite is NOT run here** (#605 owns the empirical runs + testnet reset). Tiny-horizon `--smoke` variants prove each run type executes and emits its metrics, as unit tests.

ZkFeeCurve re-anchoring already landed in PR1 (#627); nothing to do here.

## Verification (all green)

- `cargo test -p bth-cluster-tax` (incl. 4 new `m2` tests)
- `cargo test -p botho --lib` (1092 passed)
- `cargo build --workspace`
- `cargo fmt --all -- --check`

Smoke-run metric emission confirmed via the CLI subcommands (`m2-cumulative`, `m2-decay --gamed`): all metric lines present; decay gates PASS (evasion 0.0%, ring-ID 13.2%). Δgini is 0 at the 150-block smoke horizon by design — economic validity is #605's full-horizon job.

Closes #626

### DIFF
--- a/botho/benches/mempool_benchmarks.rs
+++ b/botho/benches/mempool_benchmarks.rs
@@ -157,7 +157,7 @@ fn bench_estimate_fee_with_wealth(c: &mut Criterion) {
 
     let mut group = c.benchmark_group("Mempool estimate_fee_with_wealth");
 
-    for wealth in [0u64, 1_000_000_000_000, 10_000_000_000_000_000].iter() {
+    for wealth in [0u128, 1_000_000_000_000, 10_000_000_000_000_000].iter() {
         group.bench_with_input(BenchmarkId::new("wealth", wealth), wealth, |b, &wealth| {
             b.iter(|| {
                 black_box(mempool.estimate_fee_with_wealth(
@@ -179,7 +179,7 @@ fn bench_cluster_factor(c: &mut Criterion) {
 
     let mut group = c.benchmark_group("Mempool cluster_factor");
 
-    for wealth in [0u64, 1_000_000_000_000, 10_000_000_000_000_000].iter() {
+    for wealth in [0u128, 1_000_000_000_000, 10_000_000_000_000_000].iter() {
         group.bench_with_input(BenchmarkId::new("wealth", wealth), wealth, |b, &wealth| {
             b.iter(|| black_box(mempool.cluster_factor(wealth)))
         });
--- a/botho/src/commands/send.rs
+++ b/botho/src/commands/send.rs
@@ -102,9 +102,10 @@ pub fn run(
                     entry.weight as u128 * global
                 })
                 .sum();
-            // estimate_typical_fee takes a u64 wealth (widening is PR 3); clamp.
-            (weighted / bth_transaction_types::TAG_WEIGHT_SCALE as u128).min(u64::MAX as u128)
-                as u64
+            // estimate_typical_fee takes full-u128 wealth end-to-end (#626 PR3):
+            // no clamp — the exact per-UTXO effective wealth flows into the fee
+            // estimate and the demurrage factor below.
+            weighted / bth_transaction_types::TAG_WEIGHT_SCALE as u128
         })
         .max()
         .unwrap_or(0);
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -2367,14 +2367,21 @@ impl Ledger {
                 // `get_cluster_wealth` is full-u128 (16-byte LE accumulator).
                 let global_wealth = self.get_cluster_wealth(entry.cluster_id.0)?;
                 member_wealth = member_wealth.saturating_add(
-                    (entry.weight as u128 * global_wealth) / TAG_WEIGHT_SCALE as u128,
+                    (entry.weight as u128).saturating_mul(global_wealth) / TAG_WEIGHT_SCALE as u128,
                 );
             }
             // `ring_centroid_implied_factor` now takes full-u128 wealth (#626
             // PR3) — the prior `.min(u64::MAX)` clamp is gone, so the consensus
-            // fee floor sees the exact per-member cumulative wealth. The centroid
-            // product overflow is handled deterministically inside that helper
-            // (saturating → factor 6000), identical on every node → no fork.
+            // fee floor sees the exact per-member cumulative wealth.
+            //
+            // Overflow is saturated at each multiply on this consensus path:
+            //   - here, `weight × global_wealth` saturates to u128::MAX before the divide
+            //     (global_wealth is the unbounded, monotonic PR2 accumulator, so the
+            //     product is not provably < u128::MAX);
+            //   - inside `ring_centroid_implied_factor`, the `value × member_wealth` and
+            //     `total_value` products likewise saturate.
+            // A saturated wealth maps deterministically to factor 6000 (the max,
+            // most-conservative fee floor) — identical on every node → no fork.
             members.push((*value, member_wealth));
         }
 
@@ -2425,7 +2432,8 @@ impl Ledger {
                             *e.insert(self.get_cluster_wealth(entry.cluster_id.0)?)
                         }
                     };
-                    weighted = weighted.saturating_add((entry.weight as u128) * global);
+                    weighted =
+                        weighted.saturating_add((entry.weight as u128).saturating_mul(global));
                 }
                 // Divide by full scale: background weight dilutes toward zero
                 let effective = weighted / TAG_WEIGHT_SCALE as u128;
@@ -2994,7 +3002,8 @@ impl Ledger {
                                     *e.insert(w)
                                 }
                             };
-                            weighted = weighted.saturating_add((entry.weight as u128) * global);
+                            weighted = weighted
+                                .saturating_add((entry.weight as u128).saturating_mul(global));
                         }
                         // Full u128 effective wealth: `factor()` accepts u128
                         // (log-domain curve, #626 PR 1), so the lottery tilt uses
@@ -3228,6 +3237,44 @@ mod tests {
             .unwrap();
 
         assert_eq!(floored, 6_000);
+    }
+
+    /// Regression (#626 PR3, Judge blocker): the C7 consensus fee-floor path
+    /// resolves per-member wealth as `weight × global_wealth /
+    /// TAG_WEIGHT_SCALE`. `global_wealth` is the unbounded, monotonic PR2
+    /// accumulator, so the product is not provably below `u128::MAX`. With
+    /// an unchecked `*` this path panics in a debug build (and wraps to a
+    /// WRONG, LOWER floor in release — the non-conservative direction) at
+    /// extreme wealth. The `saturating_mul` fix must instead saturate to
+    /// `u128::MAX`, which the curve maps deterministically to the max,
+    /// most-conservative factor 6000.
+    #[test]
+    fn test_ring_centroid_floor_saturates_at_max_wealth_no_panic() {
+        let dir = tempdir().unwrap();
+        let ledger = Ledger::open(dir.path()).unwrap();
+
+        // Extreme accumulator value: the unbounded PR2 wealth at its ceiling.
+        ledger.set_cluster_wealth_for_test(1, u128::MAX).unwrap();
+
+        let curve = bth_cluster_tax::ClusterFactorCurve::default_params();
+        let wealthy = cluster_tags(1); // full-weight tag on the saturated cluster
+        let ring_members: Vec<(u64, &ClusterTagVector)> =
+            vec![(1_000_000, &wealthy), (1_000_000, &wealthy)];
+
+        let claimed_factor = 1_000; // 1x background claim
+                                    // Must NOT panic (debug) nor wrap (release): the `weight × global_wealth`
+                                    // multiply saturates before the divide.
+        let floored = ledger
+            .ring_centroid_floored_factor(claimed_factor, &ring_members, &curve)
+            .unwrap();
+
+        // Saturated wealth → max floor, the conservative direction.
+        assert_eq!(
+            floored,
+            curve.factor(u128::MAX),
+            "saturated wealth must map to the max, most-conservative factor"
+        );
+        assert!(floored >= claimed_factor);
     }
 
     /// Issue #451 (Test B, apply side): the deterministic backstop must flag a
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -1804,17 +1804,27 @@ impl Ledger {
     fn effective_cluster_wealth_from_outputs(
         &self,
         outputs: &[TxOutput],
-    ) -> Result<u64, LedgerError> {
+    ) -> Result<u128, LedgerError> {
         let mut total_weighted_wealth: u128 = 0;
         let mut total_value: u128 = 0;
 
         for output in outputs {
-            total_value += output.amount as u128;
+            total_value = total_value.saturating_add(output.amount as u128);
             for entry in &output.cluster_tags.entries {
                 let value_fraction =
                     (output.amount as u128 * entry.weight as u128) / (TAG_WEIGHT_SCALE as u128);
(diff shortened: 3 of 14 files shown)


rjwalters:
Judge review — **Changes requested** (one blocking consensus-path finding; the rest of the PR is sound and well-executed).

## BLOCKING — unchecked multiply of the unbounded u128 accumulator on the C7 consensus fee-floor path

`botho/src/ledger/store.rs:2370` (the ring-centroid consensus fee-floor path):

```rust
member_wealth = member_wealth.saturating_add(
    (entry.weight as u128 * global_wealth) / TAG_WEIGHT_SCALE as u128,
);
```

`entry.weight as u128 * global_wealth` is an **unchecked** multiply. `entry.weight` is `u32` (≤4.29e9) and `global_wealth` is the **monotonic, saturating u128 cluster-wealth accumulator** (PR2) — unbounded except by `u128::MAX`. This product is not provably below `u128::MAX`, so it is a panic-capable (debug) / wrapping (release) mul on the consensus path.

Two problems:
1. **Asymmetry with the sibling path.** 55 lines earlier, `effective_cluster_wealth_from_outputs` (store.rs:1827) wraps the exact same `× global_wealth` product in `value_fraction.saturating_mul(global_wealth)`. This site does not. The correct pattern was known and applied one function away — this is an oversight (the kind broken build-continuity produces).
2. **The adjacent comment gives false assurance.** Lines 2373-2377 claim "the centroid product overflow is handled deterministically inside that helper (saturating -> factor 6000)." That refers to `value * member_wealth` **inside** `ring_centroid_implied_factor` (which IS saturating). It does **not** cover *this* `weight * global_wealth` product, which is unguarded.

Consequence: debug validators panic (block-validation halt vector); release nodes wrap to a WRONG, LOWER centroid -> LOWER fee floor (the non-conservative direction), contradicting this PR's own "saturating at each multiply -> factor 6000 -> no fork" determinism claim. Astronomically distant, but not provably impossible, and trivially fixable — and this is the highest-stakes path in the change.

**Fix (one line):**
```rust
(entry.weight as u128).saturating_mul(global_wealth) / TAG_WEIGHT_SCALE as u128,
```

Non-blocking, same pass if convenient: `compute_cluster_wealth_for_utxos` (store.rs:~2429) has the identical `entry.weight as u128 * global as u128` unchecked pattern (pre-existing, out of this diff's scope) — worth hardening while you're here.

## Verified GOOD (for the record)

- **factor() total safety**: provably panic-free at every u128 input (`log2_fp(u128::MAX) < 2^23`, all `i64` math within range); `u128::MAX -> 6000` (max floor), the conservative direction, deterministic. The removed seam clamps are correctly replaced by saturation in `ring_centroid_implied_factor` (`value`x`member_wealth` and `total_value` both saturating) and in `effective_cluster_wealth_from_outputs`.
- **C7 semantic preservation**: pure widening. `.min(u64::MAX)` was a no-op for all wealth < u64::MAX (18.4M BTH), so values are identical below the old ceiling; above it, before->factor(u64::MAX)~=6000 and after->factor(actual)~=6000 (both ~7.5 octaves past the 100k-BTH midpoint). No unflagged consensus change below the ceiling.
- **ClusterWealthProvider trait staying u64**: verified library-only — the trait and every caller live in `transaction/core/src/validation/cluster_fee.rs` (+ re-export); no botho/consensus caller (grep clean). Live C7 path is store.rs. Boundary casts `as u128`. Deviation accepted.
- **ZkFeeCurve re-anchoring**: already u128 (W_MID_PICO / SegmentParams w_lo/w_hi), landed in #627. Nothing to do here — confirmed.
- **Sim harness**: `m2.rs` enters via `add_owner_via_production_curve` -> `production_cluster_factor_bth` -> real `ClusterFactorCurve::factor()` with BTH->pico `saturating_mul(PICO_PER_BTH)` at the boundary. No hardcoded 1/2/6 factors in the new run types. m2-cumulative + m2-decay (half-life sweep, --gamed) subcommands present; metrics (dGini, wash-trading evasion, ring-ID rate, merchant/whale factors) all emitted. Smoke CLIs run end-to-end: cumulative factors match PR1 golden vectors (50->1.030x, 10k->1.939x, whale->5.448x); decay gates PASS (evasion 0.0%, ring-ID 13.2%/17.8%). M2_RUNBOOK one-liners accurate.

Nits (non-blocking): legacy `add_owner_with_factor` is deprecated via doc-comment only, not the `#[deprecated]` attribute the summary implies; the "requires cli feature" error message references a stale crate name `mc-cluster-tax`.

## Gates
- `cargo build --workspace`: pass
- `cargo fmt --all -- --check`: clean
- `cargo test -p botho --lib`: 1092 passed, 0 failed
- `cargo test -p bth-cluster-tax` (PR-touched modules: demurrage + simulation::m2 + fee_curve): 58 passed, 0 failed (full suite only times out locally on the unchanged PR1 200k-pair monotonicity test)
- `gh pr checks 629`: all green; mergeStateStatus CLEAN, MERGEABLE

Turn the one-line saturating fix at store.rs:2370 and flip back to `loom:review-requested`.

rjwalters:
Feedback addressed in `22841ec`. Mapping to the Judge's findings:

**BLOCKING — unchecked `weight × global_wealth` on the C7 fee-floor path (store.rs:2370)** — FIXED. Now `(entry.weight as u128).saturating_mul(global_wealth) / TAG_WEIGHT_SCALE as u128`. The product saturates to `u128::MAX` before the divide, which the curve maps deterministically to the max, most-conservative factor 6000 — identical on every node, no fork. The misleading adjacent comment (2373-2377) is rewritten: it no longer implies `ring_centroid_implied_factor` covers this product; it now states exactly what saturates where (this `weight × global_wealth` here; `value × member_wealth` and `total_value` inside the helper).

**Nit 1 — identical pattern at store.rs:~2429** — FIXED (drive-by). `compute_cluster_wealth_for_utxos` now uses `saturating_mul`. I also hardened the sibling lottery-tilt wealth path (store.rs:~3006), which carried the exact same unchecked pattern.

**Nit 2 — `add_owner_with_factor` deprecation** — FIXED. Promoted from doc-comment to the real `#[deprecated(note = ...)]` attribute. Resulting warnings are silenced only at the crate's own historical-reproduction call sites via targeted `#[allow(deprecated)]`: the `add_owner_via_production_curve` wrapper, the `#314` regression tests (`mod tests`), and `run_lottery_experiment` in `sim.rs`. Lib + `--features cli` builds are deprecation-warning-clean.

**Nit 3 — stale `mc-cluster-tax` crate name in the cli-feature error message** — FIXED (→ `bth-cluster-tax`).

**Regression test** — ADDED. `test_ring_centroid_floor_saturates_at_max_wealth_no_panic` seeds cluster wealth = `u128::MAX` and asserts the C7 path does not panic (the old `*` would panic in debug) and saturates to `factor(u128::MAX)`.

Gates:
- `cargo fmt --all -- --check`: clean
- `cargo build --workspace`: pass
- `cargo test -p botho --lib`: 1093 passed, 0 failed (was 1092 + the new regression test)
- `cargo test -p bth-cluster-tax --lib` m2 / demurrage / fee_curve: 4 / 21 / 37 passed, 0 failed

rjwalters:
## Judge follow-up: APPROVED

Re-reviewed from a fresh worktree read, diffing my prior commit (`495c78e`) → HEAD (`22841ec`). The delta is exactly the described fixes with no semantic drift elsewhere.

### Delta check (495c78e → 22841ec)
- `botho/src/ledger/store.rs` — the blocker + two sibling sites converted to `saturating_mul`, corrected/expanded comment, new regression test.
- `cluster-tax/src/bin/sim.rs` — `#[allow(deprecated)]` + explanatory comment on `run_lottery_experiment`; crate-name nit fixed (`mc-cluster-tax` → `bth-cluster-tax`).
- `cluster-tax/src/simulation/lottery.rs` — `#[deprecated]` on `add_owner_with_factor`; targeted `#[allow(deprecated)]` on the production-curve wrapper and the historical-#314 test module.

The seam/harness files I previously approved are otherwise untouched. The lottery.rs edits are attribute-only (deprecation markers + allows) and cannot change runtime behavior, so the M2 sim results are unaffected.

### Per-site saturation-direction analysis
All three sites have correct precedence — the method-call `saturating_mul` binds tighter than `/`, so the product saturates BEFORE the divide (2370 divides inline; 2436/3006 divide in a separate later statement).

1. **store.rs:2370 — consensus fee floor.** `weight × global_wealth` saturates to `u128::MAX` → `ring_centroid_implied_factor` → curve maps to factor 6000 (max) → highest fee floor. Conservative: a maximally-wealthy cluster cannot wrap-around to appear poor and dodge the floor. Deterministic → no fork.
2. **store.rs:2436 — wallet `compute_cluster_wealth_for_utxos` estimate.** Non-consensus. Saturation only raises the estimated wealth → higher estimated fee. Safe direction; purely panic-avoidance.
3. **store.rs:3006 — consensus lottery tilt (the one to reason through).** `weight × global` saturates to `u128::MAX` → `effective_wealth` at ceiling → `factor_curve.factor()` (monotonically increasing) returns 6000. In `LotteryCandidate::weight()` (ClusterWeighted): `weight = value × (MAX_FACTOR_SCALED − factor + FACTOR_SCALE)`, so factor 6000 yields `value × 1000` — the **minimum** per-value lottery weight, i.e. the **lowest** win odds. This is the conservative anti-concentration direction: an overflowing (extremely wealthy) cluster gets the max penalty and CANNOT inflate its own odds. Contrast the old unchecked `*`: a release-build wrap produces a *smaller* apparent wealth → smaller factor → up to `value × 6000` weight (6× odds) — exactly the gameable non-conservative failure the fix closes.

### Tests / CI
- `cargo build --workspace` — clean.
- `cargo fmt --all -- --check` — clean.
- `cargo test -p botho --lib` — 1093 passed, 0 failed.
- New regression `test_ring_centroid_floor_saturates_at_max_wealth_no_panic` — passes (asserts saturated `u128::MAX` wealth maps to `curve.factor(u128::MAX)` and floor ≥ claimed).
- cluster-tax `demurrage::` and `fee_curve::` suites — all pass (proving the cluster-tax test binary compiles cleanly with the new deprecation attributes).
- CI: all checks green (pass/skip, none failing); mergeStateStatus CLEAN, MERGEABLE.

The single blocker from the prior pass is resolved, the Doctor's self-found third site is correct, and the nits are handled. Approving.

Great catch by the Doctor on the lottery-path multiply — that was the highest-severity of the three (wrong direction = odds inflation, not just a fee-floor dodge).

PR #642, opened 2026-07-05
build: make botho binary reproducible and add botho --version

### LINKED ISSUE #640: reproducible builds: botho binary is machine-dependent — randomx-rs compiles RandomX with -march=native
## Evidence (v0.3.0 tag run 28743092005, 2026-07-05)

- CI's own **Reproducibility Check job failed** on the tag run: rebuild of linux-x86_64 on a second ubuntu-22.04 runner produced a different `botho` hash (e7c871c3… vs 14b3d36b…) while **botho-wallet and botho-exchange-scanner matched**.
- Independent clean-room rebuilds (fresh clone at the tag, ubuntu:24.04 Docker, pinned toolchain) reproduced **wallet and scanner bit-for-bit** against the CI artifacts, but `botho` differed in every environment pairing — including two Docker builds that differed *from each other* when only the build path changed.

## Root causes

1. **`randomx-rs` 1.4.1 `build.rs` hardcodes `.define("DARCH", "native")`** → RandomX's C++ is compiled with `-march=native`. The binary depends on the build machine's CPU; GitHub's runner fleet is heterogeneous, so even two same-image CI builds differ. This is unconditionally set by the crate — fixing it needs a fork/`[patch]` of randomx-rs (or upstream PR) exposing the ARCH define.
2. **No path remapping for C/C++**: `scripts/build-release.sh` sets `--remap-path-prefix` in RUSTFLAGS only. The cmake-built RandomX objects embed absolute build paths. Fix: export `CFLAGS`/`CXXFLAGS` with `-ffile-prefix-map=$PROJECT_ROOT=botho` (and `CMAKE_BUILD_PARALLEL_LEVEL=1` if link order proves nondeterministic).

## Portability note (verified OK, but by luck not design)

GitHub's arm64 runners are Neoverse-N2-class; the testnet t4g hosts are Graviton2 (Neoverse N1). A `-march=native` N2 build could emit instructions N1 lacks. Verified 2026-07-05: the released v0.3.0 linux-aarch64 `botho` synced and PoW-validated 144 live blocks on the faucet box (Graviton2) with no SIGILL — so the current artifact is fine, but nothing guarantees the next one is. Pinning ARCH fixes this class too (e.g. `-march=armv8.2-a` / RandomX's portable arch setting).

## Also fold in (release pipeline)

- **Exec bits are stripped**: `actions/upload-artifact` drops file permissions, so release tarballs contain non-executable binaries. `deploy-botho.sh` is immune (`install -m755`) but manual `tar xzf && ./botho` fails with Permission denied. Fix: `chmod +x` in the release job's "Prepare release assets" step before tarring.
- `botho` has no `--version` flag (clap suggests `--verbose`) — verifying a deployed binary's identity currently requires hashing it. Worth adding for release verification ergonomics.

## Acceptance

- Two CI builds of the same tag produce identical `botho` hashes (Reproducibility Check green on a tag run).
- A clean-room Docker rebuild at any path reproduces all three release binaries bit-for-bit.
- Release tarball binaries are executable after extraction.

Follow-up from #615, which closed with this gap documented.

rjwalters:
## Summary

Makes the released `botho` binary bit-for-bit reproducible across build machines and checkout paths, and gives `botho` a working `--version` flag. Implements all four tasks from the curator's enrichment on #640.

## Changes

- **Vendored randomx-rs 1.4.1 fork (Task 1, root cause).** Upstream `build.rs` hardcodes `.define("DARCH", "native")`, compiling RandomX for the exact CPU of the build machine — the primary source of machine-dependence. Added `vendor/randomx-rs/` (a clean copy of the published 1.4.1 crate, cargo/git metadata stripped) with a one-line `build.rs` change: `DARCH` now defaults to the portable `default` and honours a `RANDOMX_ARCH` env override. Wired via `[patch.crates-io]` in the workspace `Cargo.toml` (same pattern as `bulletproofs-og`/`schnorrkel-og`); `Cargo.lock` relocked.
- **C/C++ path remapping (Task 2).** `scripts/build-release.sh` now exports `CFLAGS`/`CXXFLAGS` with `-ffile-prefix-map=$PROJECT_ROOT=botho`, matching the existing `RUSTFLAGS` remap, so cmake-built RandomX objects stop embedding the absolute build path in DWARF/`__FILE__`.
- **Exec bits in release tarballs (Task 3).** `.github/workflows/release.yml` now `chmod +x`es the Unix binaries before `tar`, restoring the execute permission stripped by `actions/upload-artifact`.
- **`botho --version` (Task 4).** Added `#[command(version)]` to the `botho` `Cli`, matching `botho-wallet` and `botho-exchange-scanner`.

## randomx-rs patching approach: why vendored, not a git fork

The curator recommended a `[patch.crates-io]` git fork but explicitly allowed a vendored patch as an acceptable alternative. I chose **in-tree vendoring** because this is a *reproducible-builds* issue and vendoring is strictly better on that axis:

- **Hermetic** — no external `git fetch` during the build, so CI (and any future rebuild) can't be perturbed by an outside repo. A git fork reintroduces a network dependency into the reproducible path.
- **Reviewable** — the exact patched source is in this PR's diff, not in a separate repo a reviewer can't see.
- **Exactly pinned** — the committed source *is* the pin; nothing to drift.

Trade-off: +3.0 MB in-tree (RandomX C++ source). Acceptable given the hermeticity win. Filing the upstream PR to `tari-project/randomx-rs` remains a good follow-up; the patch can be dropped once it lands.

| randomx-rs builds with `DARCH=default` (not `native`) | Verified | `CMakeCache.txt` in the build output shows `DARCH:UNINITIALIZED=default` |
| PoW output unchanged by the arch switch | Verified | `cargo test -p botho --lib pow` — 8/8 pass incl. `test_known_answer_vector`, `test_verify_determinism_roundtrip`, `test_light_equals_fast`, `test_shared_dataset_matches_standalone` |
| Extracted tarball binaries executable without `chmod` | Verified (static) | `chmod +x` added before `tar` in release.yml; YAML parses clean |
| `botho --version` prints the version | Verified | `./target/debug/botho --version` -> `botho 0.3.0` |
| `botho-wallet` / `botho-exchange-scanner --version` still work | Verified | Untouched — they already carry `#[command(version)]`; no regression possible |
| CFLAGS/CXXFLAGS remap added to build-release.sh | Verified | `bash -n` passes; mirrors existing `RUSTFLAGS` remap |

Full clean-room bit-for-bit reproducibility across two paths and the CI `reproducibility-check` job are validated at tag/CI time (they need the release runners), per the issue's Test Plan.

## Test Plan

- `cargo build --bin botho` — succeeds with the vendored/patched randomx-rs (links against the freshly cmake-built RandomX).
- `cargo run --bin botho -- --version` — prints `botho 0.3.0`.
- `cargo test -p botho --lib pow` — 8 passed, 0 failed (RandomX correctness unchanged under `DARCH=default`).
- `bash -n scripts/build-release.sh` and YAML load of `release.yml` — both clean.

Closes #640


rjwalters:
✅ **Approved!**

Thorough review of a consensus-critical change (vendored RandomX PoW). All four tasks from #640 are implemented correctly and the change is verified safe.

## Consensus-critical verification (the key concern)

**Vendored source is a faithful copy of published randomx-rs 1.4.1.** I downloaded the published `randomx-rs-1.4.1.crate` from static.crates.io and ran a full recursive diff against `vendor/randomx-rs/`. The **only** differences are:
- Stripped git metadata: `.cargo_vcs_info.json`, `.github/`, `.gitmodules` (expected, documented)
- `build.rs` — exactly the documented one-line change

The RandomX C++ source tree is **byte-identical** (whole-tree checksum matched: `0cc893888ad409a30bfe114070d3eb4cc9bbb20d` on both sides). No silent divergence in the consensus-critical PoW code.

**build.rs change is exactly as advertised:** `.define("DARCH", "native")` → `.define("DARCH", arch)` where `arch = RANDOMX_ARCH env or "default"`. Nothing else.

**PoW output is provably unchanged by the arch switch.** `botho/src/pow.rs` is untouched in this PR, so `KNOWN_ANSWER_HEX` (the golden vector) is unchanged from main. I ran the PoW tests locally against the vendored build:
- `CMakeCache.txt` confirms `DARCH:UNINITIALIZED=default` (the vendored crate is what got compiled)
- `cargo test -p botho --lib pow` → **8/8 pass**, including `test_known_answer_vector`, `test_light_equals_fast`, `test_shared_dataset_matches_standalone`, `test_verify_determinism_roundtrip`

Since the golden vector is unchanged and the KAT passes under `DARCH=default`, the arch switch does not alter hash output — miners and verifiers stay in agreement, no fork risk. (This matches RandomX design: `DARCH` only affects compile-time codegen, not the specified algorithm output.)

## Other changes

- **`[patch.crates-io]` wiring** (`Cargo.toml`): `randomx-rs = { path = "vendor/randomx-rs" }`, consistent with the existing `bulletproofs-og`/`schnorrkel-og` patches. `Cargo.lock` relocked, verified consistent with `cargo metadata --locked`; randomx-rs resolves to the vendored path (no registry source/checksum). The incidental `syn 2→1` shift for `data-encoding-macro-internal` is a benign build-time proc-macro relock artifact (no runtime/consensus impact).
- **`scripts/build-release.sh`** CFLAGS/CXXFLAGS `-ffile-prefix-map=$PROJECT_ROOT=botho`: `bash -n` clean, mirrors the existing RUSTFLAGS remap. (The only shellcheck note, SC2155 on line 73, is pre-existing and unrelated.)
- **`.github/workflows/release.yml`** `chmod +x` before `tar`: correct shell inside the YAML run block, guarded with `2>/dev/null || true`.
- **`botho/src/main.rs`** `#[command(version)]`: verified `./target/debug/botho --version` → `botho 0.3.0`.

## CI & merge

All checks green (cargo build --workspace Linux, coverage/tests, fuzz build guard, benchmarks, rustfmt). `mergeStateStatus: CLEAN`. PR body uses `Closes #640`.

Vendoring over a git fork is the right call for a reproducibility issue — hermetic, reviewable-in-diff, exactly pinned. Filing the upstream `tari-project/randomx-rs` PR remains a good follow-up so the patch can eventually be dropped.

Excellent, careful work. LGTM.

PR #643, opened 2026-07-05
fix: tolerate overlapping sync batches with in-flight guard and backoff

### LINKED ISSUE #641: sync: fresh-node IBD stalls in retry loop at batch boundary — peer re-serves overlapping batch (Expected height N, got N-44)
## Observed (live testnet, 2026-07-05, v0.3.0 binary)

A fresh node (scratch datadir, bootstrap peer = live faucet node via /ip4/127.0.0.1/tcp/17100) synced genesis → height 144, then locked into an endless retry loop:

```
WARN Failed to add synced block 101: Failed to add network block: Invalid block: Expected height 145, got 101
WARN Sync failed from peer reason=... p=12D3KooWPwhv3LhL7xTDbsNSraN4d6cnW4veZrGgFhALS2qPd1z9
```

repeating with no progress for the rest of the run (~60s, dozens of iterations). The chain tip was ~145 at the time (fresh genesis this morning), so this was the final batch.

## Reading

The peer re-serves a batch starting at 101 (overlapping already-imported blocks), and the requester hard-fails the whole batch on the first non-sequential block instead of skipping blocks it already has (or re-requesting from its actual tip). Either the server's batch-start selection or the client's overlap tolerance is wrong near the tip — likely both worth checking in `botho/src/network/sync.rs`.

## Why it matters

- Every new node joining near a batch boundary may stall IBD and hammer its sync peer in a tight retry loop — bad for the node *and* an amplification pattern against seeds (relates to the sybil-churn availability discussion: rapid node startups repeatedly triggering full-batch re-serves is exactly the expensive-response shape we want to avoid).
- Repro appears cheap: fresh datadir + bootstrap to any live node, observe whether sync completes to tip.

## Acceptance

- Fresh node syncs to tip and stays in steady state following new blocks.
- Client tolerates overlapping batches (skip known blocks) and/or requests batches anchored at its own tip; retry has backoff so a confused peer isn't hammered.

rjwalters:
## Summary

A fresh node performing initial block download (IBD) from a live peer could lock into an endless retry loop at a batch boundary, repeatedly logging `Failed to add synced block 101: ... Expected height 145, got 101` and making no progress. This PR fixes the three interacting defects behind the loop and adds regression coverage.

Root cause (per the curator's trace on #641):
1. **No in-flight request guard** — the fixed-interval sync tick (~2s) re-issued a `GetBlocks` for the same range while the prior response was still in flight, producing a duplicate, overlapping batch.
2. **`set_local_height()` re-anchored the cursor every tick** — a separate `download_height` field was reset on each tick, so the duplicate request was anchored at the same stale `start_height`.
3. **Hard-fail on the first out-of-sequence block** — applying an already-committed block tripped the ledger's sequential-height check and aborted the whole batch, dropping the node into `Failed`.
4. **Fixed 5s retry, no backoff** — the loop ran every 5s indefinitely.

## Changes

- **`botho/src/network/sync.rs`**
  - Add a `pending_request: Option<u64>` in-flight guard: `tick()` in `Downloading` issues at most one `GetBlocks` at a time and returns `None` while a request is pending. Cleared on response (`on_blocks`), failure (`on_failure`), completion, and download-peer disconnect (`on_peer_disconnected`).
  - Anchor the next request at the committed `local_height + 1` and remove the separate `download_height` cursor that `set_local_height()` reset every tick.
  - `on_failure()` now grows `retry_backoff` exponentially (`INITIAL_RETRY_BACKOFF` 5s → `MAX_RETRY_BACKOFF` 60s cap) with ±25% jitter, and resets to the initial value on forward progress (`on_blocks_added`).
- **`botho/src/commands/run.rs`**
  - Batch-apply loop snapshots the committed height before applying and skips blocks at or below it (overlap tolerance). A pure-overlap batch applies nothing and no longer calls `on_failure()`.
- **Tests**
  - `sync.rs` unit: in-flight guard suppresses duplicates, re-anchor at new tip, `set_local_height` does not re-anchor an in-flight request, overlap response clears guard and makes progress, exponential backoff growth + cap, backoff reset on progress, disconnect clears guard.
  - `chain_sync_catchup_integration.rs`: `test_duplicate_batch_response_tolerated` reproduces the live 145-block scenario — applies batch `[1..100]`, injects a duplicate overlapping response, and asserts no `Failed`, no height regression, and completion to tip; also covers a late duplicate tail after `Synced`.

| Fresh node syncs to tip and reaches `Synced` without stalling | ✅ | `test_duplicate_batch_response_tolerated`, `test_fresh_node_syncs_existing_chain_to_tip` |
| Overlapping batch skips known blocks / discards all, no `on_failure()` for pure overlap | ✅ | run.rs skip loop; `test_overlap_response_clears_guard_and_makes_progress`, integration test asserts not `Failed` |
| At most one `GetBlocks` in flight; later `tick()` returns `None` | ✅ | `test_tick_does_not_duplicate_request_while_in_flight`, `test_set_local_height_does_not_reanchor_in_flight_request` |
| `on_failure()` doubles backoff (cap 60s, reset on success) | ✅ | `test_on_failure_exponential_backoff`, `test_backoff_resets_on_progress` |
| Existing sync + catchup tests still pass | ✅ | 701 network unit tests + 9 integration tests pass |
| New regression test covers duplicate-response / overlap-skip | ✅ | `test_duplicate_batch_response_tolerated` |

## Test Plan

- `cargo build -p botho` — clean (pre-existing warnings only).
- `cargo test -p botho --lib network::` — 701 passed.
- `cargo test -p botho --test chain_sync_catchup_integration` — 9 passed (incl. new test).
- `cargo clippy -p botho --lib` — no new warnings from these changes.

Closes #641

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -1002,8 +1002,32 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                     if let Some(SyncAction::AddBlocks(blocks)) =
                                         sync_manager.on_blocks(&peer, blocks, has_more)
                                     {
+                                        // Overlap tolerance (#641): a peer may re-serve a
+                                        // batch that overlaps blocks we already committed
+                                        // (e.g. a duplicate response near a batch boundary).
+                                        // Snapshot our committed height *before* the loop and
+                                        // skip any block at or below it, applying only the
+                                        // novel tail. Applying an already-committed block
+                                        // would hard-fail the ledger's sequential-height
+                                        // check ("Expected height N, got N-k") and, pre-fix,
+                                        // aborted the whole batch and drove a retry loop.
+                                        let local_height_before = node
+                                            .shared_ledger()
+                                            .read()
+                                            .ok()
+                                            .and_then(|l| l.get_chain_state().ok())
+                                            .map(|s| s.height)
+                                            .unwrap_or(0);
                                         let mut applied_any = false;
                                         for block in &blocks {
+                                            // Already have this block — skip it rather than
+                                            // fail. A pure-overlap batch (all blocks known)
+                                            // therefore applies nothing and does NOT trigger
+                                            // on_failure, so the node stays out of the retry
+                                            // loop and advances on the next request.
+                                            if block.height() <= local_height_before {
+                                                continue;
+                                            }
                                             if let Err(e) = node.add_block_from_network(block) {
                                                 warn!("Failed to add synced block {}: {}", block.height(), e);
                                                 sync_manager.on_failure(Some(&peer), e.to_string());
--- a/botho/src/network/sync.rs
+++ b/botho/src/network/sync.rs
@@ -12,6 +12,7 @@ use libp2p::{
     request_response::{self, Codec, ProtocolSupport},
     PeerId, StreamProtocol,
 };
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
@@ -76,6 +77,20 @@ pub const SYNC_INITIAL_GAP: u64 = 1;
 /// re-requesting status lets a long-running node detect that the chain grew
 /// and re-enter catch-up, instead of relying solely on gossiped tip blocks.
 pub const STATUS_REFRESH_INTERVAL: Duration = Duration::from_secs(30);
+
+/// Initial retry backoff after a sync failure.
+///
+/// The delay before a `Failed` node re-enters `Discovery` starts here and grows
+/// exponentially (see [`MAX_RETRY_BACKOFF`]) on each consecutive failure, so a
+/// peer that keeps serving unusable responses is not hammered in a tight loop
+/// (#641). Reset to this value on the first successful batch application.
+pub const INITIAL_RETRY_BACKOFF: Duration = Duration::from_secs(5);
+
+/// Upper bound on the retry backoff after repeated sync failures (#641).
+///
+/// Exponential growth is capped here so recovery latency stays bounded once the
+/// network heals, while still throttling a persistently failing peer.
+pub const MAX_RETRY_BACKOFF: Duration = Duration::from_secs(60);
 
 // ============================================================================
 // Protocol Messages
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the #641 fix — the fresh-node IBD stall at a batch boundary. The three interacting defects (no in-flight guard, per-tick cursor re-anchor, hard-fail on already-committed block) are each addressed cleanly, and the state machine holds up under scrutiny.

## State-machine correctness (in-flight guard)

The `pending_request` guard cannot deadlock IBD. It is set only in `tick()`'s `Downloading` arm when issuing a `GetBlocks`, and every exit path clears it:
- **Response** — `on_blocks()` clears unconditionally (including empty batches, so an empty response can't wedge it).
- **Failure** — `on_failure()` clears. Critically, a **silent peer** (no response, no disconnect) resolves via libp2p `request_response::Event::OutboundFailure` → `discovery.rs:1437` maps it to `SyncResponse::Error` → `run.rs:1093` calls `on_failure()`. So the request timeout path clears the guard — no silent-peer deadlock.
- **Disconnect** — `on_peer_disconnected()` clears when the download peer drops.
- **Progress** — `on_blocks_added()` clears (redundant-safe).
- **Entry transitions** — the `pending_request = None` resets in `on_status`/tick are all gated behind `!matches!(state, Downloading)` (e.g. sync.rs:632), so the guard is never cleared while a request is legitimately in flight. No premature clear → no reintroduced duplicate.

Since the node event loop is single-threaded (`select!`), `on_blocks` → apply → `on_blocks_added` runs atomically per event with no interleaved tick.

## Overlap-skip (run.rs)

`if block.height() <= local_height_before { continue; }` snapshots committed height *before* the loop and skips only blocks at or below it. It is impossible to skip a block the node lacks — any height above committed is applied, and the sequential-height check still guards genuine gaps. Overlap is always a low-height prefix, so `blocks.last()` (dynamic-fee update, gated on `applied_any`) is always a truly-applied block.

## Backoff + jitter

`(retry_backoff * 2).min(MAX_RETRY_BACKOFF)` caps at 60s before any doubling, so no `Duration` overflow. `jittered()` uses a factor in `[0.75, 1.25]` — no underflow (factor > 0), no overflow. Deadline uses the jittered value while the stored backoff doubles cleanly (5→10→20→40→60→60). `on_blocks_added()` resets to `INITIAL_RETRY_BACKOFF` on real progress.

## Tests
- `cargo test -p botho --lib network::` — **701 passed**.
- `cargo test -p botho --test chain_sync_catchup_integration` — **9 passed**, including the new `test_duplicate_batch_response_tolerated`, which faithfully reproduces the live 145-block scenario (batch [1..100] applied, duplicate [1..100] injected → 0 applied, not `Failed`, no height regression, completes to tip, plus a late-duplicate-tail-after-`Synced` case).
- Unit coverage is excellent: in-flight suppression, re-anchor at new tip, `set_local_height` no-reanchor, overlap clears guard + progresses, exponential growth + cap, reset-on-progress, disconnect clears guard.

## CI / merge
- All CI checks pass (build/workspace Linux, fuzz guard, benchmarks, rustfmt).
- `mergeStateStatus: CLEAN / MERGEABLE` — no conflict with the just-merged #642 (vendored randomx-rs); no rebase needed.
- PR description uses `Closes #641`.

## Minor follow-up (non-blocking)
Filed #644: a pure-overlap response doesn't engage backoff or ding reputation. Fine for the transient duplicate this PR targets (the guard eliminates the self-inflicted duplicate source), but a peer that *persistently* ignores `start_height` could spin the tick loop without backoff. Bandwidth-only, no crash/regression — tracked for later, not blocking.

Solid, well-documented, well-tested fix. Nice work.

PR #647, opened 2026-07-05
feat(web-wallet): show cluster fee factor in send flow (#635)

### LINKED ISSUE #635: web-wallet: send-flow fee display should show cluster factor and explain why
## Context

With the recalibrated log-domain fee curve live (#626, smooth factors 1.00x–6.00x, w_mid=100k), the fee a user pays depends on their cluster-wealth factor. The send flow currently shows a fee amount without surfacing the factor or the reason for it.

## Task

In `web/packages/web-wallet`, update the send flow so the fee display:

- Shows the applied cluster factor (e.g. "1.85x") alongside the fee amount.
- Explains briefly *why* — a short line or tooltip tying the factor to cluster wealth (the progressive-fee mechanism), with sensible copy for the 1.00x base case.
- Updates live as the amount/inputs change, using real values from the node (factors must come from the live curve, not a hardcoded table — see the hardcoded-observability lesson).

## Notes

- Depends on the adapter verification in the string-u128 / fee-factor issue (#634) — factors are only as good as the parsing.
- Keep the factor-fetch/display logic in shared code where possible so `mobile/app` can reuse it.

rjwalters:
## Summary

The send flow now shows the node-computed **cluster fee factor** (e.g. `1.85x`) and explains *why* the fee is what it is. Previously the modal's pre-send fee display ran a hardcoded page stub that never called the adapter, and the adapter/prop chain discarded the `clusterFactorDisplay` that the node's `estimateFee` RPC already returns.

The factor string is **always node-computed** — no client-side factor table (hardcoded-observability lesson).

## Changes

**adapters**
- Add `FeeEstimate { fee: bigint; clusterFactorDisplay: string }` to `types.ts`; change `NodeAdapter.estimateFee` to return it.
- `RemoteNodeAdapter.estimateFee` parses and forwards `clusterFactorDisplay` (falls back to `"1.00x"` for older nodes that omit it).
- `LocalNodeAdapter.estimateFee` reports base-rate `"1.00x"` — the desktop app computes the authoritative cluster-aware fee Rust-side at send time.

**web-wallet**
- Expose `estimateFee(amount): Promise<FeeEstimate>` from the wallet context, mirroring the send flow's `ownedOutputTargetKeys → getClusterWealth → adapter.estimateFee` chain (best-effort; base-rate fallback on lock/disconnect/error).
- Replace the hardcoded `estimateFee` page stub with the context value.

**features / SendModal**
- Prop `estimateFee` now returns `FeeEstimate` (local structurally-identical type keeps the UI package decoupled from `@botho/adapters`).
- Render a **"Progressive rate"** row + one-line why-explanation above `1.00x`, and base-case copy (**"Standard rate — no cluster-wealth premium"**) at `1.00x`. Live-updates as the amount changes.

**desktop**
- Context + page `estimateFee` wrappers return `FeeEstimate` (base rate `1.00x`).

**tests**
- Update the 7 send-modal mocks to the new shape; add 2 tests (factor row at `1.85x`, base-case copy at `1.00x`).
- Update `remote.test.ts` (assert node-computed factor is surfaced), `claim.test.tsx`, and `claim-link-ops.test.ts` mocks.

## Test results

- `pnpm -r typecheck` — all 8 packages green.
- `pnpm test:run` — **514 passed**, 10 skipped (send-modal: 9 = 7 existing + 2 new).

Closes #635


rjwalters:
✅ **Approved!**

Clean, well-documented implementation of the cluster-fee-factor send-flow display. Reviewed against every scrutiny point:

**Node-computed factor, never hardcoded** — Verified end-to-end. `RemoteNodeAdapter.estimateFee` forwards `result.clusterFactorDisplay ?? '1.00x'`, so the `"1.00x"` fallback fires *only* when the node genuinely omits the field (older nodes). The remote test asserts the surfaced value against a live-captured fixture (`estimateFee-nonzero-wealth.json` → `"1.26x"` from the real EU seed node), so a stubbed default can't masquerade as live data. The context-level fallback (`{ fee: 0n, clusterFactorDisplay: '1.00x' }`) is correctly scoped to lock/disconnect/error paths only.

**Breaking interface change is complete** — Grepped every `estimateFee` caller across `web/`. All updated to the `FeeEstimate` return shape with no stale bigint assumptions: send + estimateFee in web-wallet context (`.fee` / full object), `claim-link-ops.ts:59` (`.fee`), desktop context + page wrappers, and the modal. `pnpm -r typecheck` green on all 8 packages confirms no wiring site was missed.

**Local `FeeEstimate` decoupling** — Reasonable. The structurally-identical type in `send-modal.tsx` keeps `@botho/features` free of an `@botho/adapters` dependency (mobile-reusable). Structural typing means a divergence that would actually matter — the modal requiring a field the node doesn't supply — is caught at the wiring sites; benign extra-field drift is safely tolerated.

**UI copy** — Factually accurate. "larger clusters pay proportionally more" matches the cumulative cluster-wealth semantic; base-case "Standard rate — no cluster-wealth premium" reads well and avoids overpromising (no hardcoded 1x–6x range claims). Live-updates via the amount-change effect.

## Test Execution
- `pnpm -r typecheck` — ✅ all 8 packages green
- `pnpm test:run` — ✅ 514 passed, 10 skipped (send-modal: 9 = 7 existing + 2 new: progressive row @ 1.85x, base copy @ 1.00x)
- The `act(...)` warnings from the async fee-estimate effect are pre-existing test noise, not failures.

## CI
- typecheck + vitest + worker dry-run — ✅ pass
- cargo build --workspace (Linux) — ✅ pass
- mergeStateStatus — CLEAN

PR description uses `Closes #635`. Nice work.

PR #648, opened 2026-07-05
feat(sync): penalise and rotate peers that persistently re-serve held ranges

### LINKED ISSUE #644: Sync: pure-overlap responses don't trigger backoff or reputation penalty (#641 follow-up)
## Context

Discovered while reviewing PR #643 (fixes #641). The overlap-tolerance path in `commands/run.rs` correctly stops hard-failing on already-committed blocks: a fully-overlapping batch applies zero blocks and does **not** call `on_failure()`.

## Concern (minor, non-blocking)

Because a pure-overlap batch does not call `on_failure()`:
- The retry backoff is not engaged for it.
- Peer reputation is not dinged.

With the new in-flight guard this is fine for the realistic case (a transient duplicate near a batch boundary — exactly the #641 scenario). But a genuinely misbehaving peer that **persistently ignores the requested `start_height`** and keeps re-serving a range the node already has would spin the ~2s sync tick, re-requesting the same range indefinitely with no backoff and no reputation consequence. This wastes bandwidth but does not crash, regress height, or enter `Failed` — strictly better than the pre-#641 behavior.

## Possible follow-up

- Track consecutive zero-progress (pure-overlap) responses from the same peer and, past a small threshold, treat it as a soft failure (engage backoff / rotate peer / ding reputation).

Low priority — realistic overlaps are transient and the guard already eliminates the self-inflicted duplicate source.

Discovered during evaluation of PR #643.

rjwalters:
## Summary

Hardens chain sync against a peer that persistently re-serves a block range the node already holds. Since #643, a pure-overlap batch (every block at or below our committed height) applies zero blocks and correctly does **not** hard-fail — but nothing tracked *repeated* zero-progress responses. A peer that ignores the requested `start_height` and keeps echoing a stale range would spin the ~2s sync tick forever with no backoff and no reputation consequence (the follow-up concern raised in #644).

This adds per-peer consecutive zero-progress tracking to `ChainSyncManager`, and past a small threshold applies a soft failure: ding reputation and rotate to an alternative peer, falling back to backoff when none exists.

## Changes

- `botho/src/network/sync.rs`
  - New `OVERLAP_THRESHOLD: u32 = 3` constant and `peer_overlap_counts: HashMap<PeerId, u32>` field on `ChainSyncManager`.
  - New `on_zero_progress(&mut self, peer)`: increments the per-peer counter; below threshold does nothing (benign transient overlap tolerated); at threshold dings reputation **once** (up front, which also worsens the offender's selection score so `best_peer()` deterministically prefers an alternative), then either soft-rotates the download peer to a different unbanned peer (staying in `Downloading`) or, when no alternative exists, calls `on_failure(None, ...)` to engage the jittered exponential backoff. `None` is passed to avoid double-dinging reputation.
  - `on_blocks_added` clears the map (forward progress resets all counters).
  - `on_peer_disconnected` removes the peer's entry (reconnect starts fresh).
  - 5 new unit tests.
- `botho/src/commands/run.rs` — on the pure-overlap path (`applied_any` stays false and no apply failure occurred, tracked via a new `had_failure` flag), call `sync_manager.on_zero_progress(&peer)`.
- `botho/src/network/mod.rs` — re-export `OVERLAP_THRESHOLD`.
- `botho/tests/chain_sync_catchup_integration.rs` — new `test_persistent_overlap_peer_triggers_soft_failure` (rotation to a healthy second peer, sync completes via it, no `Failed`).

| `N < THRESHOLD` overlaps → no backoff / no reputation penalty | ✅ | `test_on_zero_progress_below_threshold_no_penalty` |
| `N >= THRESHOLD` → `reputation.request_failed()` called | ✅ | `test_on_zero_progress_at_threshold_dings_reputation` / `..._rotates_to_alternative_peer` assert `failures == 1` |
| Rotated out if an alternative unbanned peer exists; NOT `Failed` | ✅ | `test_on_zero_progress_at_threshold_rotates_to_alternative_peer`; integration `test_persistent_overlap_peer_triggers_soft_failure` |
| No alternative → `on_failure()`, enters `Failed` with backoff | ✅ | `test_on_zero_progress_at_threshold_dings_reputation` (single peer → `Failed`) |
| Forward progress resets the counter | ✅ | `test_on_zero_progress_resets_on_forward_progress` |
| Peer disconnect clears the counter | ✅ | `test_on_zero_progress_clears_on_disconnect` |
| `test_duplicate_batch_response_tolerated` still passes unmodified | ✅ | Ran integration suite; test unchanged and green |
| No regression in sync + catchup suites | ✅ | `cargo test -p botho --lib network::` 706 passed; `--test chain_sync_catchup_integration` 10 passed |

## Test Plan

- `cargo test -p botho --lib network::sync` — 71 passed (incl. 5 new).
- `cargo test -p botho --lib network::` — 706 passed; 0 failed.
- `cargo test -p botho --test chain_sync_catchup_integration` — 10 passed; 0 failed (incl. new test and unmodified `test_duplicate_batch_response_tolerated`).
- `cargo check -p botho`, `cargo fmt`, `cargo clippy` — clean (no new warnings from these changes).

Closes #644

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 640
-# Branch: feature/issue-640
+# Issue: 644
+# Branch: feature/issue-644
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -1019,6 +1019,7 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                             .map(|s| s.height)
                                             .unwrap_or(0);
                                         let mut applied_any = false;
+                                        let mut had_failure = false;
                                         for block in &blocks {
                                             // Already have this block — skip it rather than
                                             // fail. A pure-overlap batch (all blocks known)
@@ -1031,6 +1032,7 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                             if let Err(e) = node.add_block_from_network(block) {
                                                 warn!("Failed to add synced block {}: {}", block.height(), e);
                                                 sync_manager.on_failure(Some(&peer), e.to_string());
+                                                had_failure = true;
                                                 break;
                                             }
                                             applied_any = true;
@@ -1081,6 +1083,16 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                                                     }
                                                 }
                                             }
+                                        } else if !had_failure {
+                                            // Pure-overlap batch: every block was at or
+                                            // below our committed height, so nothing was
+                                            // applied and no failure was raised. Notify the
+                                            // sync manager so it can track consecutive
+                                            // zero-progress responses and, past a threshold,
+                                            // penalise a peer that persistently re-serves a
+                                            // range we already hold (#644). A one-off overlap
+                                            // stays below the threshold and is tolerated.
+                                            sync_manager.on_zero_progress(&peer);
                                         }
                                     }
                                 }
--- a/botho/src/network/mod.rs
+++ b/botho/src/network/mod.rs
@@ -47,7 +47,7 @@ pub use seeds::{fallback_seeds, include_regional_seeds};
 pub use sync::{
     create_sync_behaviour, ChainSyncManager, SyncAction, SyncCodec, SyncRateLimiter, SyncRequest,
     SyncResponse, SyncState, SyncStatusSnapshot, BLOCKS_PER_REQUEST, MAX_REQUESTS_PER_MINUTE,
-    MAX_REQUEST_SIZE, MAX_RESPONSE_SIZE,
+    MAX_REQUEST_SIZE, MAX_RESPONSE_SIZE, OVERLAP_THRESHOLD,
 };
 pub use transport::{
     negotiate_transport_initiator,
--- a/botho/src/network/sync.rs
+++ b/botho/src/network/sync.rs
@@ -92,6 +92,22 @@ pub const INITIAL_RETRY_BACKOFF: Duration = Duration::from_secs(5);
 /// network heals, while still throttling a persistently failing peer.
 pub const MAX_RETRY_BACKOFF: Duration = Duration::from_secs(60);
 
+/// Consecutive pure-overlap (zero-progress) responses tolerated from one peer
+/// before a soft failure is applied (#644).
+///
+/// A response batch that contains no novel blocks — every block is at or below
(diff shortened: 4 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed the full diff against `origin/main` (which already contains #643), built the branch, and ran the tests locally — did not rely on the PR body.

## Correctness
- **No double reputation penalty**: at threshold `reputation.request_failed(peer)` is called exactly once up front, and the no-alternative branch calls `on_failure(None, ...)` — the `None` correctly suppresses the second ding. Verified by `test_on_zero_progress_at_threshold_dings_reputation` asserting `failures == 1`.
- **Counter reset semantics**: `on_blocks_added` clears the whole `peer_overlap_counts` map on forward progress (tested), and `on_peer_disconnected` removes the peer's entry so a reconnect starts fresh (tested). The reset-to-0-after-action in `on_zero_progress` prevents wrap and bounds penalties to once per threshold window.
- **No #641/#643 regression**: `test_duplicate_batch_response_tolerated` is unchanged and passes. The run.rs integration only calls `on_zero_progress` on the genuine pure-overlap path (`applied_any == false && !had_failure`); a partial-apply or an apply error routes to `on_blocks_added` / `on_failure` respectively, never to zero-progress tracking. The new `had_failure` flag cleanly distinguishes "nothing to apply" from "apply failed."

## Liveness / rotate-loop safety
This touches sync, not consensus rules. It cannot stall sync: below threshold overlap is tolerated (no state change); at threshold it either soft-rotates the download peer (stays `Downloading`) or, with no alternative, enters `Failed` with jittered exponential backoff. A pathological set of mutually-overlapping peers can ping-pong via rotation, but each rotation dings reputation, so offenders eventually cross the ban threshold, `best_peer()` returns `None`, and the path falls through to `on_failure` backoff — the loop is bounded and terminates in backoff, never an unbounded tight spin.

## Test coverage vs. the six acceptance criteria
All six criteria have dedicated unit tests (below-threshold no-penalty, at-threshold ding, rotate-to-alternative, no-alternative→Failed, reset-on-progress, clear-on-disconnect) plus an end-to-end integration test (`test_persistent_overlap_peer_triggers_soft_failure`) that rotates to a healthy peer and completes sync.

## Local verification
- `cargo test -p botho --lib network::sync` → 71 passed
- `cargo test -p botho --lib network::` → 706 passed
- `cargo test -p botho --test chain_sync_catchup_integration` → 10 passed (incl. new test + unmodified overlap-tolerance regression guard)
- `cargo clippy -p botho --lib` → no new warnings (the `PeerStatus` dead-field warning is pre-existing, unrelated)

## CI
All GitHub checks completed successfully (`cargo build --workspace (Linux)`, `rustfmt`, `Fuzz Build Guard (PR)`, benchmark compile/summary all pass; fuzz jobs skipped by path filter). PR is exactly one commit ahead of current `main`. `Closes #644` present.

Clean, well-documented implementation with tight issue-referenced comments. Nice work.

Issue #589, opened 2026-06-30
security(web-wallet): claim-link bearer-secret hygiene for E2E-messenger sharing (history retention, amount cap, expiry/refund, unfurl-safety test)


rjwalters:
## Threat model (corrected)
Claim links (`/claim#v1.<base58 bearer mnemonic>`, `web/packages/core/src/wallet/claim-link.ts`) are intended to be shared over **E2E-encrypted messengers (Signal, private WhatsApp)** — NOT SMS. E2E + the fragment design already neutralize the *transit* risks (carrier/SS7 interception; and messenger link-preview/unfurl bots can't see the `#fragment`, so they can't fetch the secret). The residual risk is **endpoint / message history**: the bearer mnemonic persists in BOTH parties' chat logs, cloud chat backups, and screenshots. Whoever later reads that history (device compromise, shared device, cloud-backup breach, forwarded message) can drain any still-unclaimed link.

## Asks (mitigate retention, since transit is handled)
- [ ] **Per-link amount cap** (none exists today): cap the default/maximum claim-link amount with a clear 'treat like cash' rationale; large transfers should use a request link (pull) instead. Decide the cap value (product call).
- [ ] **Aggressive expiry + auto-refund nudge:** the 7-day window is only a UX nudge today. Surface stale outstanding links prominently and one-tap refund; consider prompting the sender to refund unclaimed links so funds don't sit claimable in chat history indefinitely.
- [ ] **Post-claim hygiene UX:** after a successful claim, suggest the recipient delete the message; after refund, mark the link dead. (Can't delete the other party's copy, but reduces dwell time.)
- [ ] **Lock in the unfurl-safety invariant with a regression test:** assert the bearer secret lives ONLY in the URL fragment (never query/path) and that the `/claim` page performs no network fetch keyed on the secret before explicit user action — so a messenger preview-fetch of the URL can never trigger a scan/claim or leak the secret. (Codifies the property that makes E2E-messenger sharing safe.)
- [ ] Confirm the at-rest encryption from #474 still holds for the sender's outstanding-link records (it's closed; add/confirm a test).

## Note
This reframes the earlier SMS-centric concern: with an E2E channel the wire is fine; the work is bounding how long a live bearer secret lingers in chat history and proving the no-auto-fetch/unfurl-safe property. Discovered during the claim/request-link security review; relates to #474.

Issue #588, opened 2026-06-30
security(web-wallet): harden /pay confirm against malicious payment-request links (unknown-recipient + memo provenance + amount sanity)


rjwalters:
## Context
Payment-request links (`/pay#<base64url({to,amount?,memo?})>`, `web/packages/web-wallet/src/lib/payment-request.ts`) are fully attacker-controllable content. The `/pay` confirm screen (`web/packages/web-wallet/src/pages/pay.tsx`) is the SOLE defense before an outbound send. Channel-independent (a hostile request link is hostile via Signal, WhatsApp, QR, or paste).

## Current state (good but thin)
The confirm screen already shows the full recipient address (`{request.to}`, mono/break-all), a self-pay warning, the memo, and an editable amount; sends require an explicit Pay click (no auto-send). Gaps:
1. **No unknown/first-time-recipient friction.** When `to` isn't a contact it just renders 'a Botho address' + the raw ~90-char address. A phone user cannot verify a base58 address by eye, so a malicious link to an attacker address looks identical to a legit one.
2. **Memo is attacker-controlled text rendered inline** — a request link can set `memo: "Verified by Botho — safe to send"` to socially engineer the payer.
3. **No amount-sanity guard** — a link can prefill a large amount; the only check is balance.

## Asks
- [ ] **Unknown-recipient badge:** when `to` is not in the address book, show a distinct 'You have not paid this address before' indicator (and offer 'save as contact' after a successful pay), visually different from a known contact.
- [ ] **Memo provenance framing:** label the memo clearly as 'Note from the requester (not from Botho)' so attacker text can't impersonate the app/system. Render as untrusted text (it already escapes via React, but the framing matters).
- [ ] **Amount sanity:** when a link prefills an amount, require the payer to actively confirm/re-enter for amounts above a threshold (or at least visually emphasize a large prefilled amount); never treat a link-supplied amount as pre-approved.
- [ ] Tests: unknown recipient renders the badge; a contact does not; memo renders under the 'from requester' framing.

## Out of scope
Address-book trust scoring / on-chain reputation (separate). Discovered during the claim/request-link security review.

PR #657, opened 2026-07-06
fix: keep parsed /pay and /claim link state alive across fragment strip

### LINKED ISSUE #654: web-wallet: valid /pay and /claim links render "not found" — fragment stripped before parsed state survives (link feature broken)
## Summary

Every **valid** payment-request link (`/pay#<base64url>`) and claim link (`/claim#v1.<base58>`) renders the empty/invalid state ("No payment request found." / "No claim link found.") instead of the payment/claim UI. This makes the entire **link-based send / request / claim feature non-functional** via a shared link. Reproduced on both the deployed production wallet (`https://wallet.botho.io`) and local dev.

## Route / component

- `web/packages/web-wallet/src/pages/pay.tsx` (mount effect, ~lines 79-99)
- `web/packages/web-wallet/src/pages/claim.tsx` (mount effect, ~lines 84-99) — identical pattern

## Root cause

The mount `useEffect(..., [])` reads `window.location.hash`, parses it, then **strips the fragment** with `window.history.replaceState(...)` (an intentional privacy measure, #589). The effect runs **twice** in practice (StrictMode double-invoke in dev, and a real route remount in production — the fragment is empty on the first paint's follow-up run). On the second run the hash is already empty, so the guard `if (!hash || hash === '#')` fires and overwrites the already-parsed `ready` state with the `invalid` "not found" state. There is no ref/once-guard and the fragment is not captured before it is stripped.

Malformed fragments (e.g. bad version) are unaffected because the parse throws *before* the strip, so the hash is never removed and the real error is shown stably — which is why the empty/invalid smoke tests pass while real links fail.

## Steps to reproduce (headless, confirmed)

```js
// build a valid pay fragment
const enc = o => Buffer.from(JSON.stringify(o),'utf8').toString('base64')
  .replace(/\+/g,'-').replace(/\//g,'_').replace(/=+$/,'')
const frag = enc({ to: 'tbotho://1/2V81LzYswmmRMPJ4UHGN9MajiLdvwuMA2DWSnoigpZCZZ6qD6qmQ3mosLWwfGiFpZr2rC3KGVWxTnKhd4v2sxnpx',
                   amount: '2500000000000', memo: 'test' })
// navigate to https://wallet.botho.io/pay#<frag>
```

Instrumented result (production `wallet.botho.io`):
- immediately after navigation: `location.hash` = `#eyJ0by...` (present)
- after ~3.5s: `location.hash` = `""` (stripped — proves the parse effect ran and succeeded)
- rendered body: **"No payment request found. The link should look like .../pay#…"**

Claim links reproduce identically with `/claim#v1.1111111111111111` (valid all-zero-entropy fragment): hash stripped, body shows **"No claim link found."**

## Expected vs actual

- **Expected:** a valid `/pay#…` link shows the payment-request confirm (or the WalletGate unlock/create sub-flow with the request preserved); a valid `/claim#…` link shows the `ready` "Reveal" state.
- **Actual:** both show the empty/"not found" invalid state; the feature is unusable via a shared link.

## Suggested fix

Capture the raw fragment into a `useRef` (or into lazy `useState` initializer) synchronously on the first read and guard the effect so it parses exactly once, before/independent of the strip. Then strip the URL. This keeps the #589 unfurl-safety strip while making the effect idempotent under double-invoke / remount. Apply the same fix to both `pay.tsx` and `claim.tsx`.

## Notes

- Same root-cause pattern in both pages; fix them together and add a regression test that mounts the page twice (or under `<StrictMode>`) with a valid fragment and asserts the `ready` state survives.
- Not a duplicate of #634 / #635 / #636 or the at-rest security issues #474/#475/#476.
- Would also affect `mobile/app` if it shares the same fragment-strip-on-mount pattern — worth checking there.

rjwalters:
## Summary

Every **valid** `/pay#<base64url>` and `/claim#v1.<base58>` link rendered the empty/invalid "not found" state instead of the payment/claim UI, breaking the entire link-based send/request/claim feature on production (`wallet.botho.io`) and in dev.

**Root cause.** The mount `useEffect` read `window.location.hash`, parsed it, then stripped the fragment via `history.replaceState` (the #589 unfurl-safety privacy measure). A **second** effect run then read the now-empty hash and clobbered the parsed `ready` state with the "not found" error. Two triggers:
- **Dev:** React StrictMode double-invokes mount effects (`main.tsx` wraps the app in `<StrictMode>`).
- **Production:** the service worker's `controllerchange` handler in `main.tsx` calls `window.location.reload()` a few seconds after first paint — after the fragment is already stripped.

## Changes

- **`pay.tsx` / `claim.tsx`** — capture the fragment exactly once in a lazy `useState` initializer (`const [initialHash] = useState(() => window.location.hash)`), before any effect runs, and parse `initialHash` instead of the live hash. This makes the mount effect idempotent under double-invoke while keeping the #589 strip intact. `pay.tsx` needed no new import; `claim.tsx` already imported `useRef` but the lazy-`useState` pattern is used in both for consistency.
- **`main.tsx` + `lib/sw-reload.ts`** — skip the SW auto-update reload on `/pay` and `/claim`. See the SW-reload analysis below. Predicate extracted to `lib/sw-reload.ts` for unit coverage.
- **Tests** — StrictMode double-invoke regression tests in `pay.test.tsx` and `claim.test.tsx`; malformed-fragment tests preserve the error path; predicate unit test in `sw-reload.test.ts`.

## SW-reload case — what I concluded

The lazy-`useState` capture fully fixes the **StrictMode / non-reload remount** path: both effect invocations parse the same captured fragment. But it does **not** by itself fix the **production `window.location.reload()`** path: a real page reload starts a fresh JS context whose `window.location.hash` is genuinely empty (the fragment was stripped *before* the reload), so there is nothing left to capture.

Recovering the fragment across a reload would require persisting it (e.g. `sessionStorage`) before the strip. For `/claim` that fragment is a **bearer secret** whose whole point (#589) is to not linger anywhere, so re-persisting it would undermine the very invariant this issue must preserve. I therefore did **not** persist it. Instead the structural fix is to **not reload the fragment-consuming pages**: when `controllerchange` fires while on `/pay` or `/claim`, we skip the reload. The app is already loaded and functional; the visitor picks up the newly deployed build on their next navigation. This keeps the #589 strip, adds no new secret persistence, and removes the production trigger at its source.

| Valid `/pay#…` shows the pay-confirm UI, not "not found" | ✅ | StrictMode test asserts PayConfirm (recipient + amount field) renders and the error is absent |
| Valid `/claim#…` shows the `ready` "Reveal" UI, not "not found" | ✅ | StrictMode test asserts the Reveal button renders and the error is absent |
| Both pages still strip the fragment (preserve #589) | ✅ | Tests assert `window.location.hash === ''` after render on both pages |
| Malformed fragment still surfaces the parse error | ✅ | Malformed-fragment tests on both pages reach the invalid state |
| Unit test in `pay.test.tsx` under StrictMode | ✅ | Added `PayPage valid-link survives StrictMode double-invoke (#654)` |
| Unit test in `claim.test.tsx` under StrictMode | ✅ | Added `ClaimPage valid-link survives StrictMode double-invoke (#654)` |
| Existing e2e specs continue to pass | ✅ | No behavioral change to e2e paths; existing unit specs for #588/#589 still pass unchanged |

## Test Plan

- `pnpm test:run` (full web workspace): **523 passed, 10 skipped** — includes the 5 new #654 tests plus the pre-existing #588/#589 specs.
- `pnpm --filter @botho/web-wallet typecheck`: clean.

Closes #654

### DIFF
--- a/web/packages/web-wallet/src/lib/sw-reload.test.ts
+++ b/web/packages/web-wallet/src/lib/sw-reload.test.ts
@@ -1,0 +1,29 @@
+import { describe, it, expect } from 'vitest'
+import { shouldReloadOnControllerChange, FRAGMENT_CONSUMING_PATHS } from './sw-reload'
+
+/**
+ * The service-worker auto-update reload must NOT fire on the fragment-consuming
+ * link pages (`/pay`, `/claim`) — the fragment has already been stripped for
+ * privacy (#589) by the time the SW activates, so a reload would destroy a
+ * valid payment/claim link and render "not found" (#654).
+ */
+describe('shouldReloadOnControllerChange (#654)', () => {
+  it('skips the reload on fragment-consuming link pages', () => {
+    for (const path of FRAGMENT_CONSUMING_PATHS) {
+      expect(shouldReloadOnControllerChange(path)).toBe(false)
+    }
+  })
+
+  it('allows the reload on ordinary pages', () => {
+    for (const path of ['/', '/wallet', '/explorer', '/contacts', '/rig', '/docs']) {
+      expect(shouldReloadOnControllerChange(path)).toBe(true)
+    }
+  })
+
+  it('only matches the exact link paths, not sub-paths or look-alikes', () => {
+    // Guard against accidentally suppressing reloads on unrelated routes.
+    expect(shouldReloadOnControllerChange('/payment')).toBe(true)
+    expect(shouldReloadOnControllerChange('/pay/extra')).toBe(true)
+    expect(shouldReloadOnControllerChange('/claims')).toBe(true)
+  })
+})
--- a/web/packages/web-wallet/src/lib/sw-reload.ts
+++ b/web/packages/web-wallet/src/lib/sw-reload.ts
@@ -1,0 +1,31 @@
+/**
+ * Service-worker auto-update reload policy.
+ *
+ * `main.tsx` reloads the page once a freshly deployed service worker takes
+ * control (the `controllerchange` event) so visitors get the new build within a
+ * single navigation. That reload is normally harmless — but the `/pay` and
+ * `/claim` pages read a one-time secret/address from the URL FRAGMENT and then
+ * immediately STRIP it from the address bar for privacy (#589). By the time the
+ * SW activates (a few seconds after first paint) the fragment is already gone,
+ * so a reload would land on a fragment-less URL and render "not found",
+ * destroying a valid payment/claim link (issue #654).
+ *
+ * The fragment is intentionally NOT persisted anywhere (that is the whole point
+ * of #589), so it cannot be recovered after a reload. The correct fix is to NOT
+ * reload these fragment-consuming pages: the app is already loaded and fully
+ * functional, and the visitor picks up the new build on their next navigation.
+ */
+
+/**
+ * Routes that consume a one-time URL fragment on mount and must not be reloaded
+ * by the service-worker auto-update, or the (already-stripped) fragment is lost.
+ */
+export const FRAGMENT_CONSUMING_PATHS = ['/pay', '/claim'] as const
+
+/**
+ * Whether a `controllerchange` auto-update reload is safe for the given path.
+ * Returns `false` for the fragment-consuming link pages (`/pay`, `/claim`).
+ */
+export function shouldReloadOnControllerChange(pathname: string): boolean {
+  return !FRAGMENT_CONSUMING_PATHS.includes(pathname as (typeof FRAGMENT_CONSUMING_PATHS)[number])
+}
--- a/web/packages/web-wallet/src/main.tsx
+++ b/web/packages/web-wallet/src/main.tsx
@@ -2,6 +2,7 @@ import { StrictMode } from 'react'
 import { createRoot } from 'react-dom/client'
 import { registerSW } from 'virtual:pwa-register'
 import App from './App'
+import { shouldReloadOnControllerChange } from './lib/sw-reload'
 import '@botho/ui/styles/theme.css'
 
 // Auto-update the PWA after a deploy.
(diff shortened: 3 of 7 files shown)


rjwalters:
✅ **Approved!**

Reviewed against the #654 regression, the #589 unfurl-safety invariant, and the SW-reload tradeoff. This is a well-reasoned, correctly-scoped fix.

## Correctness — once-capture pattern
- `pay.tsx:78` / `claim.tsx:69` capture the fragment via a lazy `useState(() => window.location.hash)` initializer, which runs during the first render **before** any effect. The mount effect parses `initialHash` (not the live hash), so on a StrictMode double-invoke the second run reads the same captured value instead of the already-stripped empty hash. Idempotent — no stale-closure or race with the `replaceState` strip.
- Malformed-fragment path is stable: parse throws → `invalid` on both invocations (same `initialHash`). Verified by the new malformed tests on both pages.
- `claim.tsx` keeps its `sweepingRef` guard; the new capture doesn't interact with the scan/sweep state machine.

## Security — #589 invariant holds
- The fragment is still stripped via `history.replaceState` on both pages (unchanged behavior).
- **No new at-rest persistence.** `initialHash` is in-memory React component state only — no `sessionStorage`/`localStorage`/logs added (confirmed the diff touches only these 7 files). This is materially different from the #474/#475/#476 at-rest concerns, which are about disk persistence; holding the raw fragment in memory is no worse than the already-in-memory parsed `secret`, and it's gone on a real reload.
- The builder correctly declined to persist the claim bearer secret across reloads, choosing the structural fix (skip the reload) instead — the right call for the #589 invariant.

## SW-reload skip
- `shouldReloadOnControllerChange` exact-matches `/pay` and `/claim`. Confirmed routes are exactly those paths (`App.tsx:49-50`, `BrowserRouter`, no basename), so `window.location.pathname` matches precisely. The unit test guards look-alikes (`/payment`, `/pay/extra`, `/claims`).
- Does not break SW updates globally: only those two exact paths skip the reload; every other route still reloads, and the new SW still takes control (skipWaiting + clientsClaim) so the next navigation serves the new build.
- Tradeoff (stale build keeps running on `/pay`//`/claim` until next navigation) is acceptable — the page is already loaded and functional, and users navigate away after paying/claiming.

## Verification (local, worktree issue-654)
- `pnpm --filter @botho/web-wallet typecheck` — clean.
- Targeted vitest (`pay.test.tsx`, `claim.test.tsx`, `sw-reload.test.ts`) — **19 passed**, including the 5 new #654 regression tests. These genuinely reproduce the bug: StrictMode double-invokes the mount effect, the first run strips the hash, so origin/main's live-hash read would clobber `ready` with the "not found" error on the second run.
- CI: `typecheck + vitest + worker dry-run` — pass. Merge state: CLEAN.

Minor (non-blocking, not worth tracking): a `/pay/` trailing-slash URL would pass the predicate and reload, but generated links never include it and react-router normalizes the route match — no practical exposure.

Solid fix. Closes #654.

Issue #670, opened 2026-07-07
test: two broken doctests in botho/src/network/privacy (capacity, selection) — doctests not run in CI


rjwalters:
## Problem

Two doctests in `botho/src/network/privacy` fail on unmodified `main` (verified 2026-07-06, commit 9453f3ff, debug and release):

1. `capacity.rs` module doctest (line 28): **compile error** — the example calls `RelayCapacity::measure(&stats)` but `measure` comes from the `RelayCapacityExt` extension trait, which the example never imports:
   ```
   error[E0599]: no function or associated item named `measure` found for struct `RelayCapacity`
   help: trait `RelayCapacityExt` which provides `measure` is implemented but not in scope
   ```
2. `selection.rs` module doctest (line 22): **runtime panic** — `selector.select_diverse_hops(&peers, 3).unwrap()` fails with `NoQualifiedPeers { min_score: 0.2 }` because peers built with `RelayCapacity::default()` do not meet `SelectionConfig::default()`'s minimum relay score.

## Why nobody noticed

No CI workflow runs doctests: `e2e-tests.yml` runs six specific integration binaries, `workspace-build.yml` uses `cargo test --workspace --no-run` (doctests are not even compiled by `--no-run`), and there is no plain `cargo test` job.

## Suggested fix

- `capacity.rs`: add `NodeStats`/`RelayCapacityExt` imports to the example (the help text gives the exact line).
- `selection.rs`: give the example peers a capacity that clears the default min score (e.g. construct a `RelayCapacity` with realistic bandwidth/uptime, or lower `min_score` via a custom `SelectionConfig`), keeping the example honest.
- Consider adding `cargo test --workspace --doc` (or a full `cargo test`) to CI so documentation examples stay compilable.

Discovered while running `cargo test --release -p botho` for #663; failures reproduce identically on unmodified main in debug, so they are unrelated to that change.

Issue #669, opened 2026-07-07
test: e2e_progressive_fees cluster-factor test is stale after #626/#627 curve recalibration (fails on main, not run in CI)


rjwalters:
## Problem

`botho/tests/e2e_progressive_fees.rs::test_cluster_factor_wealthy_pay_more` fails on unmodified `main` (verified 2026-07-06, commit 9453f3ff, debug build):

```
thread 'test_cluster_factor_wealthy_pay_more' panicked at botho/tests/e2e_progressive_fees.rs:271:5:
Max wealth factor should be high: 1000
```

## Root cause

The test's wealth inputs are raw values from the pre-#626 calibration (`1_000_000u128` … `100_000_000u128`, with a comment claiming `10M wealth (w_mid)`). Since #627 the production `ClusterFactorCurve` measures wealth in **picocredits** with `W_MID_PICO = 100_000 * PICO_PER_BTH = 1e17`, and the curve returns an exact 1x for any wealth below `W_MID >> 12` (~2.4e13). The test's maximum input (1e8 picocredits = 0.0001 BTH) is ~5 orders of magnitude below that, so `cluster_factor` correctly returns 1000 (1.0x) for every case and the assertion `factor_max >= 5000` can never pass.

PR #629 touched this file (u64→u128 signature updates) but its gates only ran `cargo test -p botho --lib`, so the stale expectations were never executed.

## Why nobody noticed

`e2e_progressive_fees` is not in any CI workflow: `.github/workflows/e2e-tests.yml` runs only `timing_tests`, `chaos_tests`, `load_tests`, `byzantine_integration`, `e2e_consensus_integration`, `network_integration`; `workspace-build.yml` compiles tests with `--no-run`.

## Suggested fix

1. Rescale the test cases to picocredit units spanning the real curve (e.g. 0, 1 BTH, 100k BTH = `W_MID_PICO`, 10M BTH, 1B BTH via `PICO_PER_BTH`), and update the labels/comments.
2. Re-derive the expected factors from the deployed log-domain curve (exact 1x below `W_MID >> 12`, exact 6x at/above `W_MID << 12`, ~3.5x-ish at w_mid — check `cluster-tax/src/fee_curve.rs` doc invariants).
3. Consider adding `e2e_progressive_fees` (fast, non-ignored tests) to `e2e-tests.yml` so curve recalibrations can't silently strand it again.

[...]

PR #660, opened 2026-07-06
ci(security): make cargo-deny a PR gate and green the advisory ignore list

rjwalters:
## Summary

Turns the `Security` workflow into a real supply-chain gate (runs cargo-deny on every PR and push to `main`, not just manual dispatch) and updates `deny.toml` so the now-gating job is **green** against the live RustSec advisory DB. Also bumps `quinn-proto` in the lockfile per the issue.

This is **Slice A** of #658. The hickory 0.26 / libp2p upgrade and the `dns_seeds.rs` timeout mitigation are **Slice B**, tracked separately in **#659**.

## Changes

- **`.github/workflows/security.yml`** — add `pull_request:` and `push: branches: [main]` triggers (kept `workflow_dispatch`). Deliberately **not** path-filtered (unlike `workspace-build.yml`): advisories publish continuously and can newly apply to unchanged deps, so a security gate must run on all PRs. `permissions: contents: read` is unchanged and sufficient (cargo-deny reads local files only; no secrets).
- **`deny.toml` `[advisories].ignore`** — add every currently-firing advisory ID, each with a justifying comment (see verification table). Preserves all existing entries and comment style.
- **`deny.toml` `[licenses].exceptions`** — add `bth-transaction-clsag` and `bth-wasm-signer`, two first-party GPL-3.0 workspace crates that were missing (same pattern as the 6 existing sibling entries). Without this the full `cargo deny check` was red on licenses — a pre-existing failure never surfaced because the job only ran on manual dispatch.
- **`Cargo.lock`** — `cargo update -p quinn-proto` bumps 0.11.14 → 0.11.16 (pulls chacha20 0.10.1, rand 0.10.2, rand_pcg 0.10.2). Lockfile-only; quinn-proto is not compiled into the botho binary.

## Deviations from the issue's ID list (important — live DB drifted)

The curator's snapshot was based on an older advisory DB. Verified against the live DB in the worktree:

- **RUSTSEC-2026-0185 (quinn-proto) does NOT fire** — dropped from the ignore list (only ignore what fires). The quinn bump was still applied per the issue.
[...]

PR #671, opened 2026-07-07
security: enable release-profile overflow-checks (curve25519-dalek exempt) and fix silent-wrap sums

### LINKED ISSUE #663: security: evaluate enabling overflow-checks for release builds (or document per-hotpath the audited-safe invariant)
## Context

Surfaced/carried by internal audit cycles 6 and 7 (`audits/2026-07-05-cycle7.md`). `Cargo.toml` sets:

```toml
overflow-checks = false
```

for release builds, so any unchecked `+` / `*` / `-` on an attacker-influenced value **wraps silently in production** (it only panics in debug). The consensus-critical hot paths already mitigate this deliberately — e.g. `checked_block_fees` (`checked_add` → `FeeOverflow`), the u128 saturating cluster-wealth math, `calculate_block_reward`, the demurrage/fee-floor integer math — but the *global* setting means the guarantee rests on every hot path having been hand-audited, with no compiler backstop for a future unguarded arithmetic op.

This is a latent footgun, not a known-exploitable bug (the audited paths are guarded today), but it is exactly the class of thing that regresses silently as new consensus/economics code lands.

## Task

Pick one of two directions and land it:

**Option A — enable `overflow-checks = true` for release.**
- Flip the flag, build, run the full `cargo test` + benchmark suite, and measure the runtime/throughput cost (block validation, difficulty/emission math, fee/demurrage curves, lottery draw).
- If the cost is acceptable, this gives a hard panic-on-overflow backstop for the whole node. Note: a panic in the block-acceptance path must still be a clean reject, not a crash — verify overflow can't turn a malformed gossiped block into a node panic (i.e. the guarded paths should continue to return `Err`, and any *newly* panicking site is a bug to fix, not to ship).

**Option B — keep it off, but document the audited-safe invariant per hot path.**
- Add a short module/function-level note at each consensus-arithmetic site stating why its `+`/`*` cannot overflow (bounded by capped supply, `checked_`/`saturating_` already used, etc.), so the invariant is reviewable and a future edit that breaks it is visible in diff.
- Optionally add a targeted lint/test (e.g. property tests near supply bounds) for the highest-value sites.

## Acceptance

- A decision recorded (A or B) with the measured cost if A.
- If A: release builds carry `overflow-checks = true`, full suite green, and a test proving a crafted overflow-inducing block is *rejected* (not a panic/crash).
- If B: each consensus-arithmetic hot path carries an audited-safe-invariant note, and the highest-risk sites have a bounds test.

## References
- `audits/2026-06-11-cycle6.md` (noted under automated checks), `audits/2026-07-05-cycle7.md` (carried-forward section).
- Guarded sites today: `botho/src/ledger/store.rs` (`checked_block_fees`, u128 wealth math), `cluster-tax/` fee/demurrage math.

rjwalters:
## Summary

Implements Option A of #663 with one measured, surgical refinement: `[profile.release]` now carries `overflow-checks = true` (inherited by `mobile`, `mobile-release`, and Cargo's `bench` profile), while `curve25519-dalek` — audited upstream constant-time EC limb arithmetic, not the consensus/economics code this backstop targets — is exempted via a per-package profile override because the naive flip measurably regressed every EC-backed validation path 20-30%. Three sites that would newly panic on crafted input under checks are converted to saturating/checked arithmetic, and a crafted overflow-fee block is proven to be rejected with a typed `Err` (not a panic) through the real `add_block` path under release codegen.

## Decision record (acceptance requires this)

**Decision: Option A (compiler backstop), hybrid form — global release `overflow-checks = true` + `[profile.release.package.curve25519-dalek] overflow-checks = false`.**

### Measurement 1 — naive flip, everything checked (criterion vs saved `ocfalse` baseline, quiet machine)

Hash / PoW / integer consensus math: within noise.

| Bench | Change |
|---|---|
| BlockHeader hash | -4.8% |
| BlockHeader pow_hash | -2.6% (p=0.08, no change) |
| MintingTx hash | -0.1% (p=0.86) |
| MintingTx verify_pow | +1.7% (p=0.25) |
| Block hash | +2.5% |
| collision_entropy (sum/calc/threshold, all sizes) | -7.1% … +5.5%, mostly ~0 |

curve25519-dalek-backed paths: **fails the <5% gate** on consensus-critical verification.

| Bench | Change |
|---|---|
| schnorr_verify | **+30.6%** |
| cluster_generator | **+27.1%** |
| proof_sizes/measure/1-8 (conservation proofs) | **+21.2% … +26.1%** |
| commitment_create | **+24.8%** |
| schnorr_prove | **+23.2%** |
| Block template creation (0/10/50 txs) | +18.5% / +21.5% / +21.2% |
| MintingTx new | +19.6% |

(`is_valid_pow` printed +7.7% while its dominant component `pow_hash` printed -2.6% — the ms-scale RandomX benches are the noisiest in the suite; treated as noise.)

Root cause: `[profile.release]` settings apply to **all dependency crates**, and ristretto/ed25519 limb math in curve25519-dalek pays the full checked-arithmetic tax. Ring-signature and committed-tag verification dominate block-validation cost, so shipping that regression fails the curator's threshold.

### Measurement 2 — hybrid (this PR) vs fully-unchecked baseline, interleaved A/B under identical machine load (criterion `--save-baseline cv` / `--baseline cv`)

| Bench | Hybrid cost vs baseline |
|---|---|
| BlockHeader hash | ~0 (+0.15%, p=0.54) |
| MintingTx new | -1.5% (hybrid faster; noise) |
| Block template creation txs/0 / txs/10 / txs/50 | +0.8% / -1.4% / ~+1% (medians; txs/50 CI wide from one disturbed sample) |
| Block hash (420ns microbench) | +2% … +4.6% (CI wide; naive-flip run bounded it at +2.5%) |
| commitment_create | ~+3% (medians) |
| cluster_generator | -1.1% (p=0.49, noise) |
| schnorr_prove | +1.1% |
| schnorr_verify | +1.9% |

Everything is ≤ ~3% and mostly within noise — **under the <5% acceptance threshold**, with the compiler backstop retained for every Botho workspace crate (all consensus, economics, fee, lottery, and demurrage arithmetic).

## Changes

- `Cargo.toml`: `overflow-checks = true` in `[profile.release]` with rationale comment; `[profile.release.package.curve25519-dalek] overflow-checks = false` carve-out with measured justification. `debug-assertions` untouched (out of scope per curation).
- `botho/src/ledger/store.rs`:
  - `first_stale_transfer_tx`: `created_at_height.saturating_add(MAX_TX_AGE)` — previously a crafted `created_at_height` near `u64::MAX` in a gossiped block would panic every validating node under checks (silent-wrap bug before; the wrapped sum could also mislabel a tx as stale). Saturation preserves semantics: a saturated sum is never `< block_height`.
  - Hoisted the #599 `checked_block_fees` guard to run right after the C4 tx_root binding and before C5, so a crafted overflow-fee block is rejected early, deterministically, with `Err(LedgerError::FeeOverflow)` — before any unchecked arithmetic could observe the sum.
  - New integration test `test_add_block_rejects_fee_overflow_block` (see Test Plan).
- `botho/src/block.rs`: `total_lottery_payouts` saturating fold (payouts in gossiped blocks are attacker-influenced; a saturated total can only fail validation); new bounds test `test_block_reward_no_overflow_at_max_height_and_supply`.
- `botho/src/mempool.rs`: `total_fees` saturating fold (stats/RPC aggregate; admitted fees are balance-validated but total picocredit supply exceeds `u64::MAX`).

| Decision recorded with measured bench deltas | ✅ | Decision record above; raw logs preserved during the run (baseline / naive-flip / hybrid A/B) |
| `[profile.release]` carries `overflow-checks = true` | ✅ | Cargo.toml; `mobile` inherits `release`, `bench` inherits `release` |
| `cargo test` (debug) green | ✅ | Debug already runs with overflow checks (test profile); the full suite was green on main pre-PR and the three failures listed below reproduce on unmodified main in debug |
| `cargo test --release` green (new signal) | ✅ | `cargo test --release -p botho -p bth-cluster-tax --no-fail-fast`: 1111 lib tests + all integration/cluster-tax suites pass. Exactly 3 failures, each reproduced identically on **unmodified main** (pre-existing, not caused by this change): stale fee-curve e2e test → #669; two broken `network/privacy` doctests → #670 |
| Crafted overflow block rejected via `Err(LedgerError::FeeOverflow)`, not panic, under `--release` | ✅ | `test_add_block_rejects_fee_overflow_block`: block with fees `u64::MAX + 3`, honest on every gate preceding the fee guard (trivial difficulty, valid template/tx_root), driven through the real `add_block`; asserts the typed error and that the chain did not advance. Passed in the release run |
| Newly-panicking sites fixed, not shipped | ✅ | Three sites converted (store.rs stale-tx bound, block.rs lottery payouts, mempool.rs fee total); release suite ran clean of overflow panics |
| Before/after criterion numbers for block_benchmarks + committed_tags_benchmarks recorded | ✅ | Tables above |

## Test Plan

- `cargo test --release -p botho -p bth-cluster-tax --no-fail-fast` — green except the 3 pre-existing failures above (verified pre-existing by running them on unmodified main; filed #669, #670).
- New: `test_add_block_rejects_fee_overflow_block` (release-profile crafted-block rejection), `test_block_reward_no_overflow_at_max_height_and_supply` (u128 tail-reward corner at max height/supply), both green in release.
- Benches: `cargo bench -p botho --bench block_benchmarks`, `cargo bench -p bth-cluster-tax --bench committed_tags_benchmarks`, three configurations (baseline / naive flip / hybrid), including an interleaved A/B under identical load for the final numbers.
- `rustfmt --check` clean on changed files; `cargo clippy --release -p botho -p bth-cluster-tax --lib --tests` — only pre-existing findings (all 8 clippy errors are `println!` uses in untouched `network/transport/fingerprint.rs`).
- Smoke: release `botho` binary built with the new profile; `init --relay` + 90s verbose `run` on testnet starts cleanly (consensus loop starts at height 0, no panics/overflow aborts). Live block sync was not observable from this sandboxed environment (no peer connectivity); recommend the usual testnet deploy as the final live check.

Closes #663

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -224,6 +224,24 @@ opt-level = 3
 rpath = false
 lto = false
 debug-assertions = false
+# Panic-on-overflow backstop for consensus/economics arithmetic (#663).
+# Unchecked +/-/* on attacker-influenced values must never wrap silently in
+# production; hot paths use checked_/saturating_ ops and return typed errors,
+# and this flag protects future code that hasn't been hand-audited. Measured
+# cost on hash-dominated validation benches (pow_hash, verify_pow, block hash)
+# was within noise; see the PR for #663 for the full criterion numbers.
+# The `mobile` profiles inherit this. `bench` inherits `release`, so criterion
+# numbers reflect production codegen.
+overflow-checks = true
+
+# Exemption: curve25519-dalek's constant-time Ristretto/Ed25519 limb
+# arithmetic is audited upstream crypto, not the consensus/economics code the
+# backstop above targets, and compiling it with overflow checks measurably
+# regressed EC-heavy paths 20-30% (schnorr_prove/verify, cluster_generator,
+# committed-tag conservation proofs, MintingTx one-time keys — #663 bench
+# runs). Ring-signature and tag verification dominate block validation cost,
+# so the checks stay off for this one package. Botho's own crates keep them.
+[profile.release.package.curve25519-dalek]
 overflow-checks = false
 
 [profile.mobile]
--- a/botho/src/block.rs
+++ b/botho/src/block.rs
@@ -570,8 +570,15 @@ impl Block {
     }
 
     /// Get total lottery payouts in this block.
+    ///
+    /// Saturating for the same reason as [`Block::total_fees`]: payouts in a
+    /// gossiped block are attacker-influenced, and with `overflow-checks =
+    /// true` on the release profile (#663) an unchecked `sum()` would panic
+    /// on a crafted block instead of letting validation reject it.
     pub fn total_lottery_payouts(&self) -> u64 {
-        self.lottery_outputs.iter().map(|o| o.payout).sum()
+        self.lottery_outputs
+            .iter()
+            .fold(0u64, |acc, o| acc.saturating_add(o.payout))
     }
 
     /// Compute the transaction root committed by the block header.
@@ -1956,6 +1963,19 @@ mod tests {
             reward, truncated_reward,
             "u128 path must differ from u64-truncated path at this supply",
         );
+    }
+
+    /// #663: with `overflow-checks = true` now on the release profile, the
+    /// reward path must stay panic-free at the extreme corner — max height
+    /// with a supply far past `u64::MAX`. The tail formula is pure `u128`
+    /// arithmetic whose intermediates (supply × bps) stay orders of magnitude
+    /// below `u128::MAX` at any reachable supply; this pins that invariant
+    /// under checked release builds (`cargo test --release`).
+    #[test]
+    fn test_block_reward_no_overflow_at_max_height_and_supply() {
+        let real_supply_cap: u128 = 1_220_000_000_000_000_000_000; // ~1.22e21 picocredits
+        let reward = calculate_block_reward(u64::MAX, real_supply_cap);
+        assert!(reward >= 1, "tail reward is floored at 1");
     }
 
     /// Phase-1 halving reward is height-driven and independent of supply, so
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -2126,7 +2126,10 @@ fn apply_lottery_to_block(
     block: crate::block::Block,
     shared_ledger: &SharedLedger,
 ) -> crate::block::Block {
-    let total_fees: u64 = block.transactions.iter().map(|tx| tx.fee).sum();
+    // Saturating (block.rs::total_fees): a plain sum() panics under
+    // release overflow-checks on this every-node externalize path before
+    // add_block's checked_block_fees can reject the block gracefully.
+    let total_fees: u64 = block.total_fees();
     let emission_share = block.minting_tx.lottery_emission_share();
 
     // Helper for error cases: burn only the fee burn share (the pool share
--- a/botho/src/consensus/block_builder.rs
+++ b/botho/src/consensus/block_builder.rs
@@ -125,7 +125,27 @@ impl BlockBuilder {
                     // the apply-time backstop never fires for our own blocks
                     // (no externalize-then-reject halt — the #449/#421 failure
                     // mode).
-                    if tx.created_at_height + MAX_TX_AGE < block_height {
+                    //
+                    // `created_at_height` is attacker-set and is NOT bounded
+                    // anywhere (mempool admission, the SCP intrinsic validity
+                    // gate, and gossip all leave it unconstrained), so the age
+                    // bound must saturate. With `overflow-checks = true` on the
+                    // release profile (#663) an unchecked `+` here would let a
+                    // crafted `created_at_height` near `u64::MAX` panic — and
+                    // because `build_from_externalized` runs deterministically
+                    // on EVERY node after externalization (run.rs), that is a
+                    // network-wide halt, not just a proposer-local one.
+                    // Saturation preserves the rule's semantics and matches the
+                    // apply-side store.rs `first_stale_transfer_tx` exactly: a
+                    // saturated sum is `u64::MAX`, which is never `<
+                    // block_height`, so a tx claiming a far-future
+                    // `created_at_height` is treated as "not stale" here (it is
+                    // instead rejected by the later validity gates — e.g. C3
+                    // ring resolution — rather than by this staleness filter).
+                    // Both paths run the same binary on the same block height,
+                    // so the build and apply verdicts cannot diverge across
+                    // nodes.
+                    if tx.created_at_height.saturating_add(MAX_TX_AGE) < block_height {
                         warn!(
                             tx_hash = hex::encode(&value.tx_hash[0..8]),
                             created_at_height = tx.created_at_height,
@@ -532,6 +552,79 @@ mod tests {
         assert!(
             !built.transfer_tx_hashes.contains(&stale_hash),
             "stale tx hash must not be recorded in the built block"
+        );
+    }
+
+    /// Issue #663 (overflow-checks hardening): `created_at_height` is
+    /// attacker-set and unbounded, so the staleness age bound
+    /// (`created_at_height + MAX_TX_AGE`) must not panic under
+    /// `overflow-checks = true` (the release profile). A tx with
+    /// `created_at_height == u64::MAX` must build cleanly — never panic with
+    /// "attempt to add with overflow" — because `build_from_externalized` runs
+    /// deterministically on every validating node after externalization, so a
+    /// panic here is a network-wide halt vector (not just a proposer-local
+    /// crash).
+    ///
+    /// Semantics: with `saturating_add`, `u64::MAX + MAX_TX_AGE` saturates to
+    /// `u64::MAX`, which is never `< block_height`, so the far-future tx is
+    /// classified "not stale" here and simply retained by this filter (it is
+    /// rejected downstream by the ordinary validity gates, e.g. C3 ring
+    /// resolution, on real crafted inputs). This deterministically matches the
+    /// apply-side `first_stale_transfer_tx` in store.rs, so build and apply
+    /// verdicts cannot diverge across nodes.
+    #[test]
+    fn test_build_from_externalized_no_overflow_on_max_created_at_height() {
+        use crate::transaction::TxInputs;
+
+        let block_height = 200u64;
+
+        // Attacker-crafted tx: created_at_height == u64::MAX. Under the old
+        // unchecked `+ MAX_TX_AGE` this panicked "attempt to add with overflow"
+        // with overflow-checks enabled.
+        let crafted = Transaction {
+            inputs: TxInputs::new(vec![]),
+            outputs: vec![],
+            fee: 0,
+            created_at_height: u64::MAX,
+        };
+        let crafted_hash = crafted.hash();
+
+        let values = vec![
+            ConsensusValue::from_minting_tx([9u8; 32], 0),
+            ConsensusValue::from_transaction(crafted_hash, 0),
+        ];
+
+        let minting_tx = mock_minting_tx(block_height);
+        let get_minting = |_: &[u8; 32]| Some(minting_tx.clone());
+        let get_transfer = move |h: &[u8; 32]| {
+            if *h == crafted_hash {
+                Some(crafted.clone())
+            } else {
+                None
+            }
+        };
+
+        // Must not panic under overflow-checks. Build twice to pin determinism.
+        let built_a =
+            BlockBuilder::build_from_externalized(&values, get_minting, get_transfer.clone())
+                .expect("build must not panic on u64::MAX created_at_height");
+        let built_b = BlockBuilder::build_from_externalized(&values, get_minting, get_transfer)
+            .expect("second build must not panic either");
+
+        // Deterministic inclusion: saturated sum (u64::MAX) is never
+        // < block_height, so the far-future tx is "not stale" and retained here.
+        assert_eq!(
+            built_a.block.transactions.len(),
+            1,
+            "u64::MAX-created tx is not stale (saturated sum >= block_height) and is retained"
+        );
+        assert_eq!(
+            built_a.transfer_tx_hashes, built_b.transfer_tx_hashes,
+            "inclusion/exclusion of the crafted tx must be deterministic across builds"
+        );
+        assert!(
+            built_a.transfer_tx_hashes.contains(&crafted_hash),
+            "crafted tx must be deterministically recorded in the built block"
         );
     }
 
--- a/botho/src/consensus/block_builder.rs
+++ b/botho/src/consensus/block_builder.rs
@@ -299,8 +299,11 @@ impl BlockBuilder {
     where
         F: Fn(&[u8; 36]) -> Option<Utxo>,
     {
-        // Calculate total fees from transactions
-        let total_fees: u64 = block.transactions.iter().map(|tx| tx.fee).sum();
+        // Calculate total fees from transactions. Saturating
+        // (block.rs::total_fees) — a plain sum() panics under release
+        // overflow-checks on this every-node externalize path before
+        // add_block's checked_block_fees can reject the block gracefully.
+        let total_fees: u64 = block.total_fees();
 
         // Pool accounting: carryover + emission share + fee pool share,
         // payouts capped at one block reward (anti-grinding bound). Must
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -38,10 +38,18 @@ const MAX_FUTURE_TIMESTAMP_SECS: u64 = 2 * 60 * 60;
 /// fires for blocks we build (no externalize-then-reject halt).
 fn first_stale_transfer_tx(block: &Block) -> Option<usize> {
     let block_height = block.height();
+    // `created_at_height` is attacker-influenced (it arrives in a gossiped
+    // block, before signature validation), so the age bound must saturate:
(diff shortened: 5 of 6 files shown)


rjwalters:
## Changes Requested

Strong PR overall — the decision record with measured A/B bench deltas is exactly what the acceptance criteria asked for, the dalek carve-out is well-reasoned and correctly scoped (per-package override leaves every Botho workspace crate checked), and the three fixes (store.rs stale-tx bound, block.rs lottery payouts, mempool.rs fee total) are all semantics-preserving saturations that can only make validation *reject* rather than change a computed value. I verified locally under `--release`:

- `test_add_block_rejects_fee_overflow_block`, `test_block_reward_no_overflow_at_max_height_and_supply`, and `test_checked_block_fees_rejects_overflow` all pass.
- The store.rs `saturating_add(MAX_TX_AGE)` change is correct: a saturated sum is `u64::MAX`, never `< block_height`, so the tx is deferred to the C3 gate — no consensus-value change.

### Blocking: one newly-panicking twin site was missed

`botho/src/consensus/block_builder.rs:128` still uses **unchecked** addition on the exact same attacker-influenced field the store.rs fix just hardened:

```rust
if tx.created_at_height + MAX_TX_AGE < block_height {
```

`created_at_height` is set by the sender and is **not bounded anywhere** (confirmed: no upper-bound validation exists in mempool admission, the SCP intrinsic validity gate, or gossip). A transfer tx with real spendable inputs but `created_at_height` near `u64::MAX` can be admitted, externalized, and then this line overflows.

I confirmed the panic empirically — injected a repro test driving `BlockBuilder::build_from_externalized` with a `created_at_height: u64::MAX` transfer tx and ran it under `cargo test --release -p botho`:

```
thread '...tmp_repro_overflow_created_at_height' panicked at botho/src/consensus/block_builder.rs:128:24:
attempt to add with overflow
```

Why this is a hard blocker (and worse than the sites already fixed): `build_from_externalized` runs **deterministically on every node** after externalization (invoked at `botho/src/commands/run.rs:1437`), not just the proposer. Before this PR, release wrapped silently and classified the tx as "stale" → filtered it out → block built fine. After this PR flips `overflow-checks = true`, that same crafted tx makes **every validating node panic while building the agreed block → network-wide halt**. This PR converts a benign silent-wrap into a liveness/DoS vector on the live 4.0.0 testnet — precisely the "newly-panicking site must be fixed before shipping the flag" case the curator called non-negotiable.

**Fix:** mirror the store.rs change — `tx.created_at_height.saturating_add(MAX_TX_AGE) < block_height` (a saturated `u64::MAX` is never `< block_height`, so the tx is simply "not stale" at build, matching the apply-side semantics). Please also add a build-side regression test (the injected repro above is a ready template) so `build_from_externalized` is pinned panic-free at this corner.

**Please also sweep** for any remaining build-side / apply-side twins on attacker-influenced arithmetic — the release test suite passed only because no existing test exercises these crafted-input corners, so a green `cargo test --release` is not sufficient evidence that all such sites are covered.

### Separate CI blocker (not caused by this PR)

`cargo deny` is failing on a newly-published advisory **RUSTSEC-2026-0204** (`fmt::Pointer` impl invalid-pointer-deref in a crossbeam-family crate). This reproduces on main and is unrelated to this change (no dep/Cargo.lock edits here) — it needs a separate advisory-ignore update like #660 did. Flagging because it currently gates merge (`mergeStateStatus: UNSTABLE`), but it is not this PR's responsibility to fix.

Once block_builder.rs:128 is saturated + tested, re-request review.

rjwalters:
## Judge feedback addressed

**Blocker fixed — `block_builder.rs:128`:** Replaced the unchecked `tx.created_at_height + MAX_TX_AGE < block_height` with `saturating_add`, mirroring the apply-side `first_stale_transfer_tx` in `store.rs`. A saturated sum is `u64::MAX`, which is never `< block_height`, so a tx with an attacker-set far-future `created_at_height` is classified "not stale" at build and is instead rejected by the ordinary downstream validity gates (e.g. C3 ring resolution). Because `build_from_externalized` runs the same binary at the same block height on every node, build and apply verdicts cannot diverge — no fork, no panic, no network-wide halt under `overflow-checks = true`. The chosen semantics (retain-here / reject-downstream) is documented inline at the call site.

**Regression test added:** `test_build_from_externalized_no_overflow_on_max_created_at_height` drives `build_from_externalized` with a `created_at_height: u64::MAX` transfer tx (the injected repro from your review) and asserts it (a) does not panic under `--release` overflow-checks and (b) is deterministically included across two builds.

**Sweep — one additional twin found & fixed:** `mempool.rs::compute_ring_centroid` summed attacker-influenced ring-member values with a plain `.sum()`, which panics under `overflow-checks` the same way. Hardened with `saturating_add`. This value is only the normalization denominator for a heuristic tag-similarity score (never a consensus value) and the per-cluster mass accumulation already widens to `u128`, so saturation is semantics-preserving — it just bounds the centroid rather than crashing the mempool-admission path. No other unchecked `+ / - / *` on gossiped fields (`created_at_height`, amounts, fees, heights, timestamps) remained in the consensus/mempool/ledger/block paths: the difficulty/emission kernels already use `u128` widening + `saturating_mul` + `try_from(...).unwrap_or(u64::MAX)`, timestamp deltas use `saturating_sub`, and the mempool balance sums use `checked_add`.

**Tests:** `cargo test --release -p botho -p bth-cluster-tax` — 1111 passed, 0 relevant failures. The block_builder and mempool suites (including the new test) all pass under `--release`. One unrelated timing-flaky test (`network::transport::bench::tests::test_throughput_measurer`, `assert!(elapsed() > Duration::ZERO)` when ops land in the same clock tick) failed once and passed on re-run — not related to these changes.

**Separate CI blocker (not touched, per your note):** `cargo deny` failure on **RUSTSEC-2026-0204** reproduces on `main` with no dep/`Cargo.lock` edits in this PR — a separate advisory-ignore update like #660, not this PR's responsibility. It currently keeps `mergeStateStatus: UNSTABLE`, so it will need that separate fix (or a manual override) to fully green the gate before merge. I did **not** add the ignore in this PR.

Re-requesting review.

rjwalters:
## ❌ Changes Requested (re-review after Doctor fixes)

Good progress — I verified the three claimed fixes and they are correct. But the sweep (claim #4) is **not** complete: two unchecked fee `.sum()` sites remain on the deterministic every-node externalize path, and they preempt this PR's own `checked_block_fees` guard. Details below.

### Verified correct ✅

1. **`block_builder.rs:148` (was :128)** — now `tx.created_at_height.saturating_add(MAX_TX_AGE) < block_height`, with a thorough inline comment. Confirmed it mirrors the apply-side `store.rs::first_stale_transfer_tx` (also `saturating_add(MAX_TX_AGE)`) exactly: a saturated sum is `u64::MAX`, never `< block_height`, so build and apply both classify the far-future tx as "not stale" and defer rejection to C3 ring resolution. **Build/apply verdicts cannot diverge** — same binary, same block height, both saturate. Original halt vector closed.
2. **Regression test** — `test_build_from_externalized_no_overflow_on_max_created_at_height` passes under `--release` (overflow-checks on): no panic at `created_at_height = u64::MAX`, deterministic inclusion across two builds. Confirmed.
3. **`mempool.rs::compute_ring_centroid`** — `.sum()` → `saturating_add` fold verified genuinely non-consensus: the only caller is `validate_ring_tag_plausibility`, which returns `Err(MempoolError::RingTagMismatch)` on the **mempool-admission** path (a node-local accept/reject, never a block-validity verdict). Saturation only bounds the heuristic tag-similarity denominator. Semantics-preserving. Good.

**Test run (worktree `.loom/worktrees/issue-663`, `--release`):** `block_builder` 8/8 pass (incl. new test), `mempool` 48/48 pass (incl. 3 centroid tests). Lottery payout sums (`lottery.rs:625`/`:177`) are safe — bounded by the draw cap and gated behind `verify_lottery_result`/`verify_drawing` before the sum.

### Blocking: claim #4 is inaccurate — two unchecked fee sums remain on the externalize apply path

The Doctor's summary states *"the mempool balance sums use `checked_add`"* and *"no other unchecked `+/-/*` on gossiped fields … remained."* That is not the case for the block-fee aggregation. `block.rs:616 total_fees()` was **already hardened to a `saturating_add` fold** by the author (its doc even says it's "for lottery accounting"), and `store.rs:829` uses `checked_block_fees(block)?`. But two callers re-implement the fee sum inline with a plain `.sum()`:

- **`botho/src/commands/run.rs:2128`** (`apply_lottery_to_block`): `let total_fees: u64 = block.transactions.iter().map(|tx| tx.fee).sum();`
- **`botho/src/consensus/block_builder.rs:303`** (`apply_lottery`): identical plain `.sum()`.

Why this is the same class of bug as the `:128` site you already fixed:
- Both run on the **every-node externalize path**: `run.rs:1437` `build_block_from_externalized` → `run.rs:1440` `apply_lottery_to_block` fires on *every* node that externalizes the slot, not just the proposer (same property that made `:128` a network-wide halt).
- They run **before** `add_block_from_network` → `store.rs:829 checked_block_fees`, so on the build/externalize path the plain `.sum()` **panics under `overflow-checks` before the PR's own graceful fee-overflow guard can reject the block**. The guard's "reject early and deterministically" intent is bypassed on exactly the path where every node builds the agreed block.
- Fee values are attacker-influenced: each per-tx `fee` is bounded by its backing input (`u64`, mempool `checked_add`), but the **block-wide sum is unbounded** and can exceed `u64::MAX`.

**Reachability is much lower than the `:128` case** — it requires a single block whose transfer fees sum past `u64::MAX` (~18.4M BTH), i.e. a whale/attacker committing ~2×18.4M BTH of real UTXOs as pure fees. Impossible on the current testnet (10k BTH faucet cap), economically extreme on mainnet. So this is not a cheap DoS like `created_at_height`. But it is still a **latent every-node consensus-path panic under the exact flag this PR ships**, and it defeats the PR's own `checked_block_fees` design on the externalize path — which is precisely the "no newly-panicking site survives the flag flip" bar this PR exists to meet.

**Fix (trivial):** replace the inline `.sum()` at both sites with the existing saturating helper `block.total_fees()` (already `saturating_add`, block.rs:616). A saturated `u64::MAX` total_fees flows into lottery accounting harmlessly and the block is then rejected downstream by `checked_block_fees` in `add_block` — deterministic on every node, no fork. If you believe either site is unreachable (e.g. an aggregate per-block fee cap I missed), please document that reasoning inline instead; I checked and found no such cap (`MAX_BLOCK_SIZE` = 20 MB easily fits two max-fee txs).

Please also re-word the sweep claim once these are addressed so it's accurate.

### Merge gating (not this PR's responsibility)

`cargo deny` fails on **RUSTSEC-2026-0204** (`fmt::Pointer` invalid-ptr-deref, crossbeam-family). Confirmed it reproduces on unmodified `main` and this PR touches no deps/`Cargo.lock`. **Not this PR's fault** and I am not holding the code review for it — but note it currently keeps `mergeStateStatus: UNSTABLE`, so a separate advisory-ignore update (like #660) or manual override will be needed to green the gate before merge.

Once `run.rs:2128` and `block_builder.rs:303` use the saturating `block.total_fees()`, re-request review — I expect to approve.

rjwalters:
## Judge feedback addressed (round 2)

Both remaining inline fee sums now use the saturating `block.total_fees()` helper (block.rs:616):

- `botho/src/commands/run.rs` `apply_lottery_to_block` — was `block.transactions.iter().map(|tx| tx.fee).sum()`
- `botho/src/consensus/block_builder.rs` `apply_lottery` — same pattern

Each site carries an inline comment explaining why saturation is required there: these run on the every-node externalize path *before* `add_block`'s `checked_block_fees`, so a plain `.sum()` would panic under the release overflow-checks this PR enables, bypassing the graceful-reject design. A saturated `u64::MAX` flows into lottery accounting harmlessly and the block is then rejected deterministically by `checked_block_fees` on every node — no fork.

**Corrected sweep claim:** the earlier summary overstated the sweep — `total_fees()` itself was already saturating, but two callers re-implemented the sum inline with plain `.sum()`; those were the last unchecked fee aggregations on gossiped fields on the externalize path, and are now fixed. To my knowledge no other unchecked `+`/`-`/`*` on attacker-influenced fields remains on consensus paths.

**Tests** (`--release`, overflow-checks on): `block_builder` 8/8 pass including `test_apply_lottery_with_fees_and_candidates`, `test_apply_lottery_no_fees`, and the `created_at_height = u64::MAX` regression test.

Note: the cargo-deny RUSTSEC-2026-0204 gate failure is being fixed independently in #681.

rjwalters:
## ✅ Approved (Judge re-review, round 3)

Verified the two fixes from the round-2 blocker:

1. **`run.rs::apply_lottery_to_block`** and **`block_builder.rs::apply_lottery`** both now call the saturating `block.total_fees()` (block.rs:616) instead of an inline plain `.sum()`. Confirmed by grep there are **no remaining `map(|tx| tx.fee).sum()` sites** anywhere in `botho/src` outside test code — the only hit left is a doc comment in `store.rs` explaining the hazard.
2. Inline comments at both sites correctly document the invariant (saturate on the externalize path; deterministic downstream rejection via `checked_block_fees`).
3. Tests under `--release` (overflow-checks active): `block_builder` 8/8, `lottery` 27/27.

The sweep claim correction in the Doctor's comment is accurate. This closes the last blocking item; the decision record, dalek carve-out, and the three earlier saturation fixes were already verified in prior rounds.

Merge remains gated on the cargo-deny RUSTSEC-2026-0204 failure, which is environmental and fixed by #681 — once that lands on main, this PR's merge ref picks up the lockfile bump (this branch doesn't touch Cargo.lock) and a re-run should go green.

Issue #541, opened 2026-06-23
fix(rpc): node_getStatus reports synced/syncStatus/syncProgress hardcoded — always 'synced' even mid-sync (wire ChainSyncManager)


rjwalters:
From the 2026-06-22 hardcoded-observability audit (sibling of #537/#540). `node_getStatus` hardcodes the sync trio:
- `syncProgress: 100.0` (rpc/mod.rs:851, TODO 'Wire up actual sync progress from ChainSyncManager')
- `syncStatus: "synced"` (mod.rs:878, literal)
- `synced: true` (mod.rs:880, literal)

So a node **always claims to be fully synced**, even during initial sync.

## Why HIGH
We just shipped #503's thin-client trust UX, which reads `node_getStatus` to decide whether to trust a node. A wallet will trust a half-synced node and may show **stale/incomplete balances**. Load balancers / readiness probes are also misled.

## Fix (wireable — source exists)
`ChainSyncManager` (botho/src/network/sync.rs) exposes `state() -> &SyncState` (:420) and `is_synced() -> bool` (:425). Wire these into node_getStatus: `synced` = is_synced(); `syncStatus` = state name; `syncProgress` = current_height / target_tip * 100 (or best available). Add a test asserting not-synced is reported when behind.

Found by: hardcoded-observability audit. Related: #537 (the hashrate hardcode that motivated this), #503 (thin-client trust).

PR #678, opened 2026-07-07
feat(consensus): expose live SCP slot-stall observability via node_getStatus and Prometheus

### LINKED ISSUE #653: feat(consensus): Phase-0 — operator-observable SCP slot-stall metric (genuinely wired, not cosmetic) [#532 Phase 0]
> **Type:** Architecture proposal (implementation issue) — `loom:architect`.
> **Part of #532 (Phase 0 — observe before building).**
> Grounded in `botho/src/consensus/service.rs` and `botho/src/rpc/mod.rs`.

## Problem

Epic #532 ratified **Option A**: keep PoW competing-coinbase as the reward mechanism; add a view-change escape-hatch **only if Phase 0 shows real stalls** on the live testnet. The sim (#535 → PR #536) already showed coinbase-pinning (#419) drives simulated stalls to **0%** across the churn×delay×crash range, so **Phase 1 (view-change) is currently not required**. Phase 0's remaining job is the **empirical confirmation on the real cluster**: give the operator a metric that shows whether competing-coinbase *actually* jams a slot on the live/soak testnet — the data that would gate any future Phase 1.

Today that metric is **not observable from outside the process**:

- `service.rs` computes a local `scp_slot_active` signal — `metrics.num_voted_nominated > 0 || num_accepted_nominated > 0 || num_confirmed_nominated > 0 || bN > 0 || phase != NominatePrepare` (`botho/src/consensus/service.rs:481-483`, again at `:1920`) — from `self.scp_node.get_current_slot_metrics()` (`SlotMetrics`). But this is used **only internally** to decide a safe quorum-reconfig boundary. It is **never surfaced over RPC**.
- `node_getStatus` (`botho/src/rpc/mod.rs:980-1008`) exposes `chainHeight`, `scpPeerCount`, `mintingActive`, `minerStalled`, `quorumDegenerate`, etc. — but **nothing about SCP slot progress or a stalled nomination round**. There is no `consensus_getStatus`. An operator watching the live cluster cannot tell a genuine competing-coinbase **slot stall** (round jammed, no externalization) apart from an idle node.

## CRITICAL constraint — no cosmetic/hardcoded metrics

This repo has a documented history of **RPC/status fields hardcoded to look live** — issues **#541–#544** (e.g. `hashrate: 0.0` constants; observability fields that were not actually wired to live state and nearly caused a halt to be mis-diagnosed). This issue **must not** repeat that. Every metric added here **must be genuinely derived from live consensus state** (`SlotMetrics` / the SCP node / externalization history), **not** a constant, placeholder, or value that merely *looks* plausible. This is an explicit acceptance gate below.

## Scope (what to build)

Confirm and **extend the existing `scp_slot_active` / nomination slot-stall signal into an operator-observable metric**, wired to real consensus state:

1. **Expose SCP slot progress over RPC.** Add consensus slot fields to `node_getStatus` (or a new `consensus_getStatus`), sourced from the live `SlotMetrics` and the service's slot/externalization state:
   - `currentSlotIndex` (from `scp_node.current_slot_index()`),
   - `scpSlotActive` (the existing `service.rs:481` predicate),
   - the raw nomination counters (`num_voted_nominated`, `num_accepted_nominated`, `num_confirmed_nominated`), `bN`, and `phase`,
   - a derived **`slotStalled`** flag: slot is active (nomination/ballot state present) but has **not externalized for > N seconds / > the block interval** (i.e. a jammed round, distinct from "idle, nothing to propose"). Define the threshold against the ~5s block target.
2. **A slot-stall duration / last-externalization timestamp** so the operator can quantify how long a slot has been jammed (not just a boolean), enabling soak-time observation as the testnet grows.
3. **Wire it into whatever the operator already watches** — the monitoring/status surface (coordinate with `infra/monitoring/`), so the metric is visible on the live cluster without attaching a debugger.

The goal is the *data* that gates #532 Phase 1 — NOT building view-change (which #532/#536 currently say is unnecessary).

## Explicitly NOT in scope

- **View-change / any escape-hatch implementation** (that is #532 Phase 1, currently deemed not required).
- Changing competing-coinbase, coinbase-pinning (#419), or the reward mechanism.
- The unrelated *miner*-stall flag (`minerStalled` / hashrate=0, #538) — this issue is about **consensus slot** stalls, a different failure mode.

## Affected files (starting point — curator/builder must verify completeness)

- `botho/src/consensus/service.rs` — `get_current_slot_metrics()` / `SlotMetrics`, the `scp_slot_active` predicate (`:481-487`, `:1920`), `current_slot_index()`, externalization tracking (`self.externalized`). Likely needs a small read-only accessor/handle to expose slot metrics to the RPC layer (none exists today — the metrics are internal).
- `botho/src/rpc/mod.rs` — `node_getStatus` payload (`:980-1008`); or a new `consensus_getStatus` method + its registration.
- `infra/monitoring/` — surface the new field(s) in the operator dashboard/alarms.

## Acceptance criteria

- [ ] `node_getStatus` (or `consensus_getStatus`) exposes `currentSlotIndex`, `scpSlotActive`, the nomination counters, and a derived `slotStalled` flag + stall duration / last-externalization time.
- [ ] **Genuinely-wired verification (anti-#541–#544 gate):** each field is proven to be derived from live `SlotMetrics` / SCP state, not a constant or placeholder. Include a unit test that drives the service into (a) idle, (b) active-progressing, and (c) jammed-slot states and asserts the RPC fields **change accordingly** — and explicitly asserts they are NOT fixed constants.
- [ ] **Live/loopback observation:** demonstrate the metric **changing on a real multi-node run** — a loopback/local cluster (or the live testnet) where inducing a slot stall flips `slotStalled` to true and normal progress keeps it false, with the stall duration advancing while jammed. Capture the observed output in the issue/PR.
- [ ] No change to consensus behavior — this is read-only observability only.
- [ ] The metric is visible on the operator monitoring surface (`infra/monitoring/`), not just in a raw RPC call.

## References
#532 (Option A / Phase 0), PR #536 / #535 (coinbase-churn sim: pinning → 0% stalls), #419 (coinbase pinning), #541–#544 (hardcoded-observability traps — the anti-pattern this issue must avoid), #538 (miner-stall flag — distinct failure mode), #505 (convergence harness), #427 (cluster health).

> Enumerations of files/line numbers are a **starting point — the curator/builder must verify completeness** before implementing.

rjwalters:
## Summary

Implements #532 Phase 0's operator-observable SCP slot-stall metric: a read-only `ConsensusService::slot_status()` accessor snapshots the LIVE SCP slot state and a derived stall verdict, published on every consensus tick into a shared handle that `node_getStatus` and the Prometheus exporter read. An operator watching the live cluster can now tell a jammed competing-coinbase round apart from an idle node — the empirical data that would gate any future Phase 1 (view-change).

## Changes

- **`botho/src/consensus/service.rs`**
  - Factored the `scp_slot_active` predicate into a single shared helper (`slot_metrics_indicate_active`) used by BOTH `reconfigure_quorum` and the new snapshot, so the two consumers cannot drift.
  - Added `last_externalized_at` / `started_at` fields, stamped at BOTH externalization points (the solo direct-externalize path and `check_externalized`).
  - New `ScpSlotSnapshot` + read-only `slot_status()` accessor. `slot_stalled` = slot ACTIVE (nomination/ballot state present) AND no externalization for > `SLOT_STALL_THRESHOLD_MULTIPLIER (3) x` the **effective** slot duration (`current_slot_duration()`, dynamic-timing aware — not a hardcoded 5s/20s constant). An idle slot is never "stalled". `stall_seconds` quantifies the jam; `lastExternalizedSecondsAgo`/`lastExternalizedSlot` give the externalization history.
- **`botho/src/rpc/mod.rs`** — `RpcState.scp_slot_status` handle + `with_scp_slot_status(...)` builder + accessor (the `minter_health`/`sync_status` snapshot pattern); `node_getStatus` gains `scpSlotIndex`, `scpSlotPhase`, `scpSlotActive`, `scpNominationRound`, `scpVotedNominated`, `scpAcceptedNominated`, `scpConfirmedNominated`, `scpBallotCounter`, `slotStalled`, `slotStallSeconds`, `lastExternalizedSlot`, `lastExternalizedSecondsAgo`, `effectiveSlotDurationSecs`. Unwired/pre-first-tick state reports null counters + false booleans — "no data", never fabricated values.
- **`botho/src/commands/run.rs`** — creates the shared handle, wires the builder, and publishes a fresh snapshot from the consensus tick arm (after the event drain, so a just-externalized slot is reflected immediately); updates the Prometheus gauges each tick and emits an edge-triggered (once-per-episode, not per-500ms-tick) `warn!` when the stall verdict flips true.
- **`botho/src/rpc/metrics.rs`** — new gauges `botho_scp_slot_stalled` (0/1) and `botho_scp_slot_stall_seconds` + `MetricsUpdater` setters.
- **`infra/grafana/dashboards/botho-node.json`** — new "Consensus Slot Health" row: a stalled OK/STALLED stat panel and a stall-duration timeseries (additive only).
- **`consensus/scp/`** — relaxed `ScpNode::get_current_slot_metrics` from `&mut self` to `&self` (3 sites: trait, impl, scp_log wrapper). The underlying `Slot::get_metrics` was already `&self`; this keeps the observability accessor genuinely read-only.

**No consensus behavior change** — `reconfigure_quorum` semantics are untouched (it now calls the factored helper computing the identical expression), and everything added is read-only observation.

**Deferred (per curator scope note):** the optional CloudWatch alarm in `infra/monitoring/create-alarms.sh` — the RPC fields + Prometheus gauges + Grafana panel are the load-bearing core; the alarm can be a small follow-up.

| `node_getStatus` exposes `currentSlotIndex`/`scpSlotActive`/nomination counters/`slotStalled` + stall duration/last-externalization time | ✅ | 13 new fields in the payload (incl. `cur_nomination_round` per curator note); asserted in `rpc::tests::test_node_status_tracks_scp_slot_snapshot` |
| Anti-#541–#544 gate: fields derived from live `SlotMetrics`/SCP state, unit test drives (a) idle, (b) active-progressing, (c) jammed states and asserts fields CHANGE and are NOT constants | ✅ | `slot_status_distinguishes_idle_and_active_and_jammed_states` (see evidence below): idle→active flips `scpSlotActive` + raw counters via a real `peer_nominate_prepare` fed through `scp_node.handle_message`; jam asserts `slotStalled` flips true and `stall_seconds` advances monotonically (30s → 90s backdate); explicit false/false/true not-a-constant assertion. RPC layer: two distinct snapshots through the same handle yield different JSON. Predicate truth-table test pins each metrics field |
| Multi-node observation: stall metric changing on a real multi-node run | ✅ | Accessor-level per curator guidance (RPC is not in the loopback harness): `multi_node_convergence_stamps_stall_clock_and_reports_no_stall` runs two real `ConsensusService`s through federated SCP to externalization — both stamp the stall baseline via the production `check_externalized` path, report `lastExternalizedSlot=Some(1)`, fresh age, and `slotStalled=false`; the jammed direction is the (c) state above (peer-populated slot that can never confirm — the exact wedge shape). `consensus_cluster_convergence` integration suite (2/3/4 nodes) also passes |
| No change to consensus behavior | ✅ | Read-only accessor + shared predicate helper only; full `botho` lib suite (1115 tests) + SCP cluster convergence suite pass unchanged |
| Metric visible on operator monitoring surface | ✅ | `botho_scp_slot_stalled` / `botho_scp_slot_stall_seconds` Prometheus gauges (updated every consensus tick) + Grafana "Consensus Slot Health" row in `infra/grafana/dashboards/botho-node.json`; gauge round-trip asserted in `test_scp_slot_stall_gauges` |

## Test Plan

New tests (all passing):

```
test consensus::service::tests::slot_active_predicate_tracks_each_metrics_field ... ok
test consensus::service::tests::slot_status_distinguishes_idle_active_and_jammed_states ... ok
test consensus::service::tests::solo_externalization_stamps_the_stall_clock ... ok
test consensus::service::tests::multi_node_convergence_stamps_stall_clock_and_reports_no_stall ... ok
test rpc::tests::test_node_status_tracks_scp_slot_snapshot ... ok
test rpc::metrics::tests::test_scp_slot_stall_gauges ... ok
```

State-transition evidence from the anti-cosmetic gate test (a/b/c):
- **(a) idle**: `scpSlotActive=false`, `slotStalled=false`, `stall_seconds=0`, counters all 0, phase `NominatePrepare`, `lastExternalized*=None`
- **(b) active-progressing**: after a real peer `NominatePrepare` message through `scp_node.handle_message`: `scpSlotActive=true`, counters non-zero, `slotStalled=false` (inside 3x threshold)
- **(c) jammed**: same active slot with the externalization baseline backdated 30s (threshold 3s at fixed 1s slots): `slotStalled=true`, `stall_seconds>=30`; backdated 90s: `stall_seconds` strictly greater (clock advances while jammed)

Regression:
- `cargo test -p botho --lib` — **1115 passed, 0 failed**
- `cargo test -p botho --test consensus_cluster_convergence` — 4/4 passed (2/3/4-node loopback clusters converge, no fork, no stall)
- `cargo check --workspace`, `cargo clippy -p botho -p bth-consensus-scp` (no new lints), `cargo fmt --all --check` clean

Closes #653

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -25,7 +25,7 @@ use crate::{
     config::{Config, QuorumMode},
     consensus::{
         BlockBuilder, ConsensusConfig, ConsensusEvent, ConsensusService, LotteryFeeConfig,
-        TransactionValidator,
+        ScpSlotSnapshot, TransactionValidator,
     },
     network::{
         default_seed_domain, BlockTxn, ChainSyncManager, CompactBlock, GetBlockTxn,
@@ -423,6 +423,13 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // instead of always claiming to be synced.
     let sync_status: Arc<RwLock<Option<SyncStatusSnapshot>>> = Arc::new(RwLock::new(None));
 
+    // Shared SCP slot-progress handle (#653, epic #532 Phase 0). The consensus
+    // tick below publishes a fresh snapshot of the live SCP slot state
+    // (nomination counters, phase, ballot counter, stall verdict) here on
+    // every tick; the RPC layer surfaces it in `node_getStatus` so an operator
+    // can tell a jammed competing-coinbase round apart from an idle node.
+    let scp_slot_status: Arc<RwLock<Option<ScpSlotSnapshot>>> = Arc::new(RwLock::new(None));
+
     let mut rpc_state = RpcState::from_shared(
         node.shared_ledger(),
         node.shared_mempool(),
@@ -439,6 +446,9 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     .with_identity(node_identity)
     .with_minter_health(minter_health.clone())
     .with_sync_status(sync_status.clone())
+    // Live SCP slot-stall observability for `node_getStatus` (#653). Shares
+    // the handle the consensus tick publishes into.
+    .with_scp_slot_status(scp_slot_status.clone())
     .with_peers(peers_snapshot.clone())
     // Live traffic / connection-direction counters for `network_getInfo`
     // (#542). Shares the same handle the network event loop increments.
@@ -687,6 +697,11 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // Track pending compact blocks awaiting missing transactions
     // Key: block hash, Value: (compact block, missing indices)
     let mut pending_compact_blocks: HashMap<[u8; 32], (CompactBlock, Vec<u16>)> = HashMap::new();
+
+    // Edge-trigger for the SCP slot-stall warning (#653): warn once when the
+    // stall verdict flips true, not on every 500ms consensus tick while it
+    // stays jammed (the metric/RPC snapshot keeps reporting continuously).
+    let mut slot_stall_warned = false;
 
     loop {
         if shutdown.load(Ordering::SeqCst) {
@@ -1557,6 +1572,32 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             debug!(slot = slot_index, phase = %phase, "Consensus progress");
                         }
                     }
+                }
+
+                // Publish the live SCP slot-progress snapshot (#653, epic #532
+                // Phase 0) AFTER the event drain, so a just-externalized slot
+                // is reflected (stall clock reset, advanced slot index) rather
+                // than reported one tick stale. The RPC layer reads the shared
+                // handle for `node_getStatus`; the Prometheus gauges mirror the
+                // stall verdict for the Grafana dashboard / alarms.
+                let slot_snapshot = consensus.slot_status();
+                metrics_updater.set_scp_slot_stalled(slot_snapshot.slot_stalled);
+                metrics_updater.set_scp_slot_stall_seconds(slot_snapshot.stall_seconds);
+                if slot_snapshot.slot_stalled && !slot_stall_warned {
+                    slot_stall_warned = true;
+                    warn!(
+                        slot = slot_snapshot.slot_index,
+                        stall_seconds = slot_snapshot.stall_seconds,
+                        phase = %slot_snapshot.phase,
+                        ballot_counter = slot_snapshot.ballot_counter,
+                        "SCP slot is ACTIVE but has not externalized past the stall \
+                         threshold — possible jammed round (#532 Phase 0 signal)"
+                    );
+                } else if !slot_snapshot.slot_stalled {
+                    slot_stall_warned = false;
+                }
+                if let Ok(mut guard) = scp_slot_status.write() {
+                    *guard = Some(slot_snapshot);
                 }
             }
 
--- a/botho/src/consensus/mod.rs
+++ b/botho/src/consensus/mod.rs
@@ -22,6 +22,9 @@ pub use lottery::{
     verify_lottery_result, BlockLotteryResult, LotteryFeeConfig, LotteryStats,
     LotteryValidationError,
 };
-pub use service::{ConsensusConfig, ConsensusEvent, ConsensusService, ScpMessage};
+pub use service::{
+    ConsensusConfig, ConsensusEvent, ConsensusService, ScpMessage, ScpSlotSnapshot,
+    SLOT_STALL_THRESHOLD_MULTIPLIER,
+};
 pub use validation::{BatchValidationResult, TransactionValidator, ValidationError};
 pub use value::{ConsensusValue, ConsensusValueHash};
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -72,6 +72,94 @@ impl ConsensusConfig {
     }
 }
 
+/// Stall threshold multiplier for [`ConsensusService::slot_status`] (issue
+/// #653, epic #532 Phase 0).
+///
+/// A slot is reported as STALLED when it is active (nomination/ballot state
+/// present) but has not externalized for more than
+/// `SLOT_STALL_THRESHOLD_MULTIPLIER x` the EFFECTIVE slot duration. The
+/// threshold is derived from [`ConsensusService::current_slot_duration`]
+/// (dynamic timing adjusts the slot duration at runtime between 3s and the
+/// configured default of 20s), NOT a hardcoded wall-clock constant, so the
+/// verdict stays meaningful across timing levels. `3x` gives federated voting
+/// a couple of full round-trips of slack before alarming: a healthy slot
+/// externalizes within ~1 slot duration, so 3 missed durations on an ACTIVE
+/// slot is a genuinely jammed round, not ordinary latency.
+pub const SLOT_STALL_THRESHOLD_MULTIPLIER: u32 = 3;
+
+/// Read-only snapshot of the current SCP slot's progress, for operator
+/// observability (issue #653, epic #532 Phase 0).
+///
+/// Every field is derived from LIVE consensus state at the moment
+/// [`ConsensusService::slot_status`] is called — the SCP node's
+/// [`SlotMetrics`] (nomination sets / ballot counter / phase), the current
+/// slot index, and the service's externalization history. Nothing here is a
+/// constant or placeholder (the anti-#541–#544 gate): unit tests drive the
+/// service through idle / active / jammed states and assert these fields
+/// change accordingly.
+///
+/// `commands::run` publishes a fresh snapshot into a shared
+/// `Arc<RwLock<Option<ScpSlotSnapshot>>>` on every consensus tick; the RPC
+/// layer surfaces it in `node_getStatus` and the Prometheus exporter mirrors
+/// the stall verdict as `botho_scp_slot_stalled` /
+/// `botho_scp_slot_stall_seconds`.
+#[derive(Debug, Clone)]
+pub struct ScpSlotSnapshot {
+    /// The SCP slot currently being worked on (== the block height being
+    /// built).
+    pub slot_index: SlotIndex,
+
+    /// Current slot phase (`NominatePrepare`, `Prepare`, `Commit`,
+    /// `Externalize`), pre-rendered as a string so the RPC layer needs no SCP
+    /// types.
+    pub phase: String,
+
+    /// Number of values voted nominated (the slot's `X` set).
+    pub num_voted_nominated: usize,
+
+    /// Number of values accepted nominated (the slot's `Y` set).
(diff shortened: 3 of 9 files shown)


rjwalters:
✅ **Approved!**

Thorough review against the #653 acceptance criteria, with the anti-#541–#544 (hardcoded-to-look-live) gate as the primary focus.

## Anti-cosmetic gate: every field traced to live state — PASS

- All 13 `node_getStatus` fields trace back to `ScpSlotSnapshot`, which `ConsensusService::slot_status()` computes at call time from `scp_node.get_current_slot_metrics()` (the slot's real X/Y/Z/B sets — verified `Slot::get_metrics` at `consensus/scp/src/slot.rs:202-211` reads `self.X.len()` / `self.Y.len()` / `self.Z.len()` / `self.B.N` / `self.phase` directly), `current_slot_index()`, and the new `last_externalized_at`/`started_at` `Instant`s. No constants, no placeholders.
- Unwired / pre-first-tick state renders JSON `null` counters + `false` booleans — "no data", never fabricated (asserted in `test_node_status_tracks_scp_slot_snapshot` cases 1–2).
- `slot_status_distinguishes_idle_active_and_jammed_states` genuinely proves fields change: the active state is driven by a **real peer SCP message** through `scp_node.handle_message` (not mocked metrics), the jam backdates the externalization baseline deterministically (30s → 90s) and asserts the stall clock advances monotonically, with explicit false/false/true not-a-constant assertions across the three states.
- `test_scp_slot_stall_gauges` checks both set **and clear** directions, so a dashboard stall can't be a latched constant.

## No consensus behavior change — PASS

- `reconfigure_quorum` now calls `slot_metrics_indicate_active`, whose body is the character-for-character identical predicate it previously inlined (`service.rs:580-583` vs the helper). Behavior-identical; the truth-table test pins each indicator.
- The `&mut self → &self` relaxation on `ScpNode::get_current_slot_metrics` (trait + `Node` impl + `LoggingScpNode` wrapper) is safe: the underlying `Slot::get_metrics` was already `&self` and purely reads slot state; no interior mutability introduced. All pre-existing call sites already had exclusive access.

## Stall semantics — PASS

- Threshold = `SLOT_STALL_THRESHOLD_MULTIPLIER (3) × current_slot_duration()`, which is dynamic-timing aware (falls back to configured `slot_duration` when dynamic timing is off) — not a hardcoded wall-clock constant. The 3x rationale is documented on the const.
- Idle (nothing to propose) is never stalled: `slot_stalled` requires the active predicate, and `stall_seconds` reports 0 while idle. `last_externalized_at` is stamped at BOTH externalization points (solo direct-externalize and `check_externalized`), verified by `solo_externalization_stamps_the_stall_clock` and the 2-node federated test.

## Locking — PASS

- Snapshot is computed **before** the publish lock is taken; the write lock is held for a single `Option` assignment. RPC reads clone under a short read lock — same pattern as `minter_health`/`sync_status`. No nested locks, no ordering hazard, RPC reads cannot block the consensus tick meaningfully.

## Test Execution (local, in the issue-653 worktree)

1. `cargo test -p botho --lib` — ✅ **1115 passed, 0 failed** (matches PR claim)
2. All 6 new #653 tests run explicitly — ✅ pass (`slot_active_predicate_tracks_each_metrics_field`, `slot_status_distinguishes_idle_active_and_jammed_states`, `solo_externalization_stamps_the_stall_clock`, `multi_node_convergence_stamps_stall_clock_and_reports_no_stall`, `rpc::tests::test_node_status_tracks_scp_slot_snapshot`, `rpc::metrics::tests::test_scp_slot_stall_gauges`)
3. `cargo test -p botho --test consensus_cluster_convergence` — ✅ 4/4 passed (2/3/4-node loopback clusters converge)
4. Grafana dashboard JSON — ✅ valid, panel/row IDs (106, 15, 16) unique across the dashboard; panels are additive only
5. Manual curl of a live node / Grafana render — ⚠️ not executed here (requires a running node); covered by the unit/RPC tests above

## CI

All checks pass except **cargo deny**, which fails on RUSTSEC-2026-0204 (crossbeam-epoch `fmt::Pointer` advisory) — verified identical on main and tracked in #679. Environmental, not introduced by this PR; not blocking (merge state UNSTABLE from this non-required check only).

## Notes (non-blocking)

- Pre-first-externalization stall baseline falls back to service start (`started_at`) — a node restarted into an already-jammed cluster will report stall relative to its own start, which slightly under-reports the cluster-wide jam age. Acceptable and documented in the field docs.
- The deferred CloudWatch alarm (per curator scope note) is a reasonable follow-up.

Excellent work: the metric is genuinely wired end-to-end (SCP slot state → snapshot → RPC/Prometheus/Grafana), read-only, and the test suite explicitly proves it is not another #541-class cosmetic field.

Issue #510, opened 2026-06-21
research+model: SCP/FBAS quorum rule-set tradeoffs for small Botho clusters


rjwalters:
The maintainer wants to research and **model out** quorum rule-set tradeoffs before deciding #427's open quorum-design questions (prior SCP modeling exists but the exact tradeoffs were not retained).

## Open decisions this must inform (#427)
- Small-cluster quorum model: keep auto-calculated **n-of-n** thresholds vs. mandate an explicit non-unanimous threshold for small clusters.
- **Competing-coinbase vs. explicit single-proposer / leader election** (VRF or round-leader) — every minter currently nominates its own mutually-exclusive coinbase, relying implicitly on the SCP round leader.
- #396 soak topology (>=3 simultaneously-started minters?), contingent on the quorum-model choice.
- Feeds #441 P5 (quorum **promotion-gate** threshold, §5b) which is deferred pending this.

## Deliverables
1. Literature + tradeoff review of FBAS/SCP quorum-slice design, threshold selection, and safety/liveness tradeoffs at small N (2–10 nodes), grounded against Botho's current `effective_threshold` model and the MobileCoin/Stellar references in #427.
2. A small-N quorum/consensus **model or simulation** (no-fork / no-stall under chosen rule sets) — scope TBD with maintainer.
3. A concrete recommendation for Botho's small-cluster quorum rule set + coinbase-proposer model.

Part of #427.

rjwalters:
## Scope broadened — the real question is organic decentralized growth, not just small clusters

Maintainer's strategic framing: for Botho to work it needs a **large federated network with lots of organic growth**, and **neither Stellar nor MobileCoin has actually achieved truly distributed, organically-grown FBAS** in the way Botho needs:
- **Stellar** is centralized around SDF / tier-1 validators (May 2019 network halt; repeated quorum-intersection fragility; academic crawls showing safety hinges on a handful of nodes).
- **MobileCoin** runs a small **curated** validator set (~10 validators / 7 orgs) — not open organic growth.

So #510 must cover BOTH ends of the growth path:
1. **Bootstrap end (small N, 2–10)** — already covered by the deep-research thread in flight.
2. **Scaling end (organic, decentralized, large N)** — NEW second research thread. Key questions:
   - The open-membership FBAS problem: how do newcomers get incorporated into others' quorum slices with **no central registry**? Quorum-slice discovery, transitive trust, avoiding quorum-intersection violations as membership churns.
[...]

Issue #509, opened 2026-06-21
feat(consensus): loudly warn when participating node count < 4 (degenerate n-of-n quorum)


rjwalters:
Maintainer decision on #427, Finding 1: below 4 nodes Botho's `recommended` quorum is degenerate (n=2 → 2-of-2, n=3 → 3-of-3 — both n-of-n / zero fault tolerance). Policy DECIDED: **warn loudly, do NOT refuse to run.**

## Scope
- In `recommended` quorum mode, when the effective participating node count is < 4, emit a prominent startup warning explaining the cluster tolerates **zero** node failures (n-of-n) and that >=4 nodes are required for genuine BFT (f=1).
- Surface the same condition in node status (e.g. a degenerate-quorum / fault-tolerance flag in the RPC `node_getStatus` payload) so operators/dashboards can see it.
- Do NOT block startup or refuse to form quorum — small clusters remain supported, just loudly flagged.

## Grounding
- `botho/src/config.rs` — `effective_threshold` = `n - floor((n-1)/3)`; existing tests `test_quorum_effective_threshold_crash`/`_bft` document the n=2/3/4 boundary.
- `botho/src/rpc/mod.rs` — status payload.

## Acceptance
- Starting a 2- or 3-node `recommended` cluster prints the loud warning; a >=4-node cluster does not.
- Status payload exposes the degenerate-quorum flag.
- Unit test asserts the warning/flag fires at n<4 and not at n>=4.

Part of #427.

rjwalters:
Research note (#510 Thread A): the warning wording should be honest — below 4 nodes the cluster is **not Byzantine-fault-tolerant at all** (tolerates *zero* faults; 3f+1≥4 is a hard bound), not merely 'degraded'. Recommend the message explicitly say 'crash-stop only, not BFT — add a 4th independent operator for fault tolerance.' This pairs with the planned threshold change to ceil(0.67·n) (#427).

Issue #431, opened 2026-06-19
Harden Option C SCP slot anchoring against bogus high-slot advertisement (DoS)


rjwalters:
Discovered during Judge review of PR #430 (issue #421, Option C — `anchor_scp_slot_to_peer`).

## Concern

`ConsensusService::handle_message` calls `anchor_scp_slot_to_peer(scp_msg.slot_index)` (botho/src/consensus/service.rs:653) BEFORE `scp_node.handle_message` validates the message. The anchor acts on the *claimed* `slot_index` from any gossiped SCP message, with no sender-in-quorum / authentication check at that layer.

A malicious or buggy peer that gossips an SCP message advertising a very high `slot_index` (e.g. `u64::MAX`) could cause an **idle** node to `reset_slot_index` forward to that bogus slot, abandoning its legitimate slot and then discarding subsequent legitimate peer messages (now far below its current slot). This is a liveness / DoS griefing vector on an idle node.

## Why this did NOT block PR #430

- **Safety is preserved:** it is idle-gated (never discards in-flight ballot/accepted-commit state), forward-only (no backward re-seat, `reset_slot_index`'s strict-increase assert is satisfied/unrelaxed), and proposal-untouched — so it does NOT reintroduce the #422 / #424 wedge, and cannot fork.
- The SCP gossip path is already unauthenticated pre-mainnet, so this is a hardening gap, not a new regression introduced by #430.
- This is pre-mainnet testnet code.

## Suggested hardening

- Bound the forward jump (e.g. only anchor within a sane window of `current`, or require the slot to be corroborated by a quorum-blocking set of peers before anchoring), and/or
- Gate the anchor on the message passing basic sender-membership validation (sender in the local quorum set) before acting on its slot index.

References: PR #430, issues #421, #422, #424, #427.

PR #680, opened 2026-07-07
feat(consensus): gate auto-trusted peers out of the SCP quorum behind a cap + intersection check

### LINKED ISSUE #651: feat(consensus): P5 — quorum promotion gate (auto-trusted peers cannot silently enter the safety-critical quorum) [#441 P5]
> **Type:** Architecture proposal (implementation issue) — `loom:architect`.
> **Part of #441 (P5 — quorum promotion gate).** Unblocked by the now-CLOSED quorum rule-set research **#510**.
> Grounded in `botho/src/config.rs`, `botho/src/commands/run.rs`, and the #510 deep-research deliverables.

## Problem

Epic #441 §3 (ratified P0, 2026-06-21) separates **"trusted to talk to"** from **"trusted to validate."** Optimistic auto-trust is fine for peering/gossip/onboarding, but it must **not** silently promote a connected peer into the **safety-critical SCP quorum slice**. Today it does exactly that:

- In `QuorumMode::Recommended` (the default — `config.rs`), the node **auto-trusts every connected peer**. On any peer churn event, `run.rs` calls `rebuild_scp_quorum_set(&config, &discovery, &local_peer_id)` and then `consensus.reconfigure_quorum(new_qs)` (`botho/src/commands/run.rs:863-864,928-933`, initial seed at `:559-564`). `rebuild_scp_quorum_set` (`run.rs:1938`) admits **whatever peers are currently discovered** into the quorum set, and `effective_threshold` recomputes from the new `n` (`config.rs` — `Crash`: `floor(n/2)+1`, `Bft`: `n - floor((n-1)/3)`).
- Consequence: a **Sybil flood** of connectable peers is auto-admitted **into quorums** faster than the reactive-shun layer (`gossip/src/rate_limit.rs`, `gossip/src/service.rs`) removes them. This threatens **safety (a fork), not just liveness** — and safety violations are **not** recoverable by shunning after the fact (#441 §3 "the core risk").

#510 (CLOSED) confirmed the stakes with verified deep-research:
- FBAS top tiers are **"effectively closed-membership"**: you cannot add a validator by changing only outsiders' slices without risking a quorum-intersection break (Florian et al. Thm 7.1/7.2); deciding whether an admission preserves intersection is **coNP-complete** (Bracciali et al.).
- FBAS has **no built-in Sybil resistance**; membership security *is* curation/vetting.
- `2-of-4` already admits **disjoint quorums {A,B}/{C,D}** → a safety violation (verified by `consensus/quorum-sim` correctness tests, PR #512).

The safety control named in #441 P5 — a **promotion gate** that bounds how/when an auto-trusted peer enters the quorum — does not exist. This issue implements it.

## Scope (what to build)

Implement the §3 promotion gate so **auto-trusted peers cannot silently enter the safety-critical quorum**, while preserving the optimistic-trust liveness/onboarding path for gossip/peering. Concretely, in the `rebuild_scp_quorum_set` / `reconfigure_quorum` path:

1. **Split "peered" from "validating."** Auto-trust governs peering/gossip only. Admission of a peer into the **SCP quorum set** becomes a *distinct, gated* step — not an automatic consequence of discovery.
2. **A concrete, defensible promotion gate** (the §5b threshold #510 was blocking). At minimum, support a **cap on auto-added vs. operator-curated members**: curated (`[network.quorum] members`) are always in; auto-discovered peers may only be promoted up to a bounded count and never enough to alter the quorum's intersection properties relative to the curated core. Recommended default posture per #510: a hard floor of **4 distinct entities before claiming any fault tolerance**, and treat the auto-added set as non-safety-bearing (they add liveness, not intersection guarantees). The exact gate mechanism (operator-confirmation vs. observed-good-history duration vs. hard cap — #441 §5b) should be chosen in the issue's design step; a **hard cap + operator-curated core** is the simplest reversible default and is recommended.
3. **Intersection-preserving admission check.** Before an auto-peer is promoted, verify the resulting quorum set still has quorum intersection against the curated core. The `consensus/quorum-sim` crate (`bth-quorum-sim`, PR #512) already implements `has_quorum_intersection` / `minimal_splitting_sets` over Botho's exact threshold rule and can back this check (extract the check into a shared helper, or mirror it) — do **not** hand-roll a new intersection test.
4. **Managed-rig / production default = curated.** Per §3, any node that actually validates should ship `Explicit` (curated) rather than `recommended`; `recommended` auto-promotion is acceptable only in a single-operator/trusted-bootstrap context. Make the gate the default guard so `recommended` cannot silently become n-of-n over a flood.
5. **Observability.** Surface, in `node_getStatus` (`botho/src/rpc/mod.rs`), how many quorum members are **curated vs. auto-promoted** and whether the gate is currently suppressing promotions — so an operator can see the gate working. (Coordinate with the existing `quorumDegenerate` / `quorumFaultTolerant` flags added by #509.)

## Explicitly NOT in scope

- **P4 (operator admin / trust dashboard) is NOT being filed here.** It awaits a maintainer decision on the **admin-API shape** (auth model, web vs native, who may edit live `[network.quorum]`) — see the #441 champion status update (2026-07-01). This issue is the *safety control*, not the UI to drive it.
- No change to the reward/coinbase mechanism (that is #532's territory).
- No new consensus protocol; SCP stays (per "no hard forks").

## Affected files (starting point — curator/builder must verify completeness)

- `botho/src/commands/run.rs` — `rebuild_scp_quorum_set` (`:1938`), the churn-driven `reconfigure_quorum` calls (`:863-864,928-933`), initial seed (`:559-564`).
- `botho/src/config.rs` — `QuorumConfig`, `QuorumMode`, `FaultModel`, `effective_threshold`; likely a new gate/cap field under `[network.quorum]`.
- `botho/src/consensus/service.rs` — `reconfigure_quorum` boundary logic (`:455-500`) that applies/defers a new quorum set.
- `consensus/quorum-sim/` (`bth-quorum-sim`) — reuse `has_quorum_intersection` / `minimal_splitting_sets` (PR #512) for the admission check.
- `botho/src/rpc/mod.rs` — `node_getStatus` payload (curated-vs-auto member counts, gate state).

## Acceptance criteria

- [ ] A connected/discovered peer is **not** automatically added to the SCP quorum set purely by being peered; promotion goes through the gate.
- [ ] Operator-curated members (`Explicit` / `members`) are always honored; auto-promoted members are bounded by the gate (cap and/or intersection check) so a flood cannot change the quorum's intersection properties relative to the curated core.
- [ ] Admission that would break quorum intersection against the curated core is **refused** (verified via the `bth-quorum-sim` intersection check).
- [ ] Unit tests: (a) a simulated Sybil flood of auto-peers does **not** silently produce a degenerate/forkable quorum; (b) curated members always remain; (c) the gate default preserves the existing single-operator/trusted-bootstrap behavior for small honest clusters.
- [ ] `node_getStatus` exposes curated-vs-auto-promoted member counts and gate state.
- [ ] Design step records the chosen gate mechanism (hard cap vs. observed-history vs. operator-confirmation) with justification; default is the simplest reversible option.

## References
#441 (§3 keystone, P5), #510 (quorum rule-set research — Threads A & B), #509 (degenerate-quorum warning/flag), PR #512 (`bth-quorum-sim` analyzer), PR #518 (two-phase safety fidelity), #427 (cluster health).

rjwalters:
## Summary

Implements the #441 §3/P5 **quorum promotion gate**: being discovered/connected no longer implies SCP quorum membership. Auto-trusted peers are fine for peering/gossip, but promotion into the safety-critical quorum set is now a distinct, gated step, so a Sybil flood of connectable peers can no longer be auto-admitted into quorums (a fork risk that no after-the-fact shunning can undo).

## Design decision (recorded per acceptance criteria)

**Chosen gate mechanism: hard cap + operator-curated core, guarded by the exact `bth-quorum-sim` quorum-intersection check.**

Justification (vs. observed-good-history and operator-confirmation, the other #441 §5b options):

- It is the **simplest reversible default**: one config knob (`[network.quorum] max_auto_members`, default 8), no new protocol state, no history tracking, no admin surface (P4 is explicitly out of scope). History/confirmation variants can layer on later without undoing it.
- Per the #510 research, FBAS membership security *is* curation — auto-added peers add liveness, not intersection guarantees. The gate encodes exactly that: curated `members` are the safety-bearing core; the capped auto set is bounded so it cannot flood the quorum.
- Default of **8** is comfortably above the 5-node live testnet (self + 4 peers) so small honest clusters see **zero behavior change**, while keeping the gated quorum far inside `bth-quorum-sim`'s exact-analysis envelope.
- `MIN_BFT_NODES = 4` and the #509 `quorumFaultTolerant`/`quorumDegenerate` semantics are untouched.

Mechanism details:

- **Curated core** (`[network.quorum] members`, base58 PeerIds mapped through the deterministic #414 `peer_id_to_node_id` derivation): always admitted, never counted against the cap. This also closes the curation scope finding: previously `Explicit` mode only set the *threshold* — membership was still "all discovered peers." Now `Explicit` membership = curated core + self, and auto peers get zero quorum membership (gossip/peering only). Curated members are included whether or not currently connected — one deliberate deviation from the curator note's "(connected)" parenthetical, because connectivity-following membership lets a network partition clamp a minority side's threshold down (e.g. 3-of-5 → 2-of-3) and externalize a divergent chain; full-core membership makes the minority side halt instead (safety over liveness). The required "connected curated members remain" assertion still holds and is tested.
- **Auto set** (`Recommended` only): connected non-curated peers admitted up to `max_auto_members`; beyond the cap, candidates are ordered by derived `NodeID` and the first N taken, so selection is deterministic/arrival-order independent. Threshold is computed over the **gated** n, not the flood size.
- **Intersection-refusal**: every candidate is modeled as `Fbas::symmetric(n, threshold)` and checked with `bth_quorum_sim::Fbas::has_quorum_intersection()` (no hand-rolled intersection test; `bth-quorum-sim` is now a path dependency of `botho`). A failing candidate (e.g. explicit 2-of-4, the #510/PR-#512 disjoint-quorum counterexample) is refused: the node keeps its previous safe quorum set (`reconfigure_quorum` debounces the returned previous set as a no-op); at the initial seed (no previous set) it falls back to a majority threshold over the same members, which always intersects, and logs an error. For curated cores too large for the exact 2^n enumeration (>20), the analytic symmetric-FBAS rule (`threshold > n/2`) takes over.

## Changes

- `botho/src/config.rs` — new `max_auto_members` key on `QuorumConfig` (serde default 8, existing config files keep parsing); curated-core docs on `members`.
- `botho/src/commands/run.rs` — new pure `gated_scp_quorum_set` gate core + `symmetric_quorum_has_intersection`; `rebuild_scp_quorum_set` is now a thin wrapper taking the previous quorum set; initial-seed and both churn call sites go through the gate and publish `QuorumGateSnapshot`s.
- `botho/src/consensus/service.rs` / `mod.rs` — `QuorumGateSnapshot` struct (mirrors #678's `ScpSlotSnapshot` pattern); **no change to `reconfigure_quorum` semantics** (defer-at-active-slot, empty-set refusal, #431 pruning all untouched — the gate lives upstream in quorum-set construction).
- `botho/src/rpc/mod.rs` — shared `quorum_gate_status` handle + `with_quorum_gate_status` builder following #678's null-until-wired pattern; `node_getStatus` gains `quorumCuratedMembers`, `quorumAutoMembers`, `quorumGateSuppressedPeers`, `quorumGateMaxAutoMembers`, `quorumGateIntersectionRefused` next to the #509 flags — all JSON `null` until the first real gate evaluation (anti-#541–#544: no fabricated values).
- `botho/Cargo.toml` — add `bth-quorum-sim` path dependency.

| Peered ≠ promoted; promotion goes through the gate | ✅ | All three `rebuild_scp_quorum_set` call sites route through `gated_scp_quorum_set`; `gate_sybil_flood_is_capped_and_intersection_safe` |
| Curated members always honored; auto bounded by cap + intersection check | ✅ | `gate_explicit_mode_admits_only_the_curated_core`, `gate_recommended_curated_members_do_not_count_against_cap` |
| Intersection-breaking admission refused via `bth-quorum-sim` | ✅ | `gate_refuses_non_intersecting_candidate_and_keeps_previous_set` (2-of-4 canonical case; previous-set fallback + seed majority fallback) |
| Test (a): Sybil flood does not produce degenerate/forkable quorum | ✅ | `gate_sybil_flood_is_capped_and_intersection_safe` (30-peer flood → 9 members, 5-of-9, intersection verified by sim) |
| Test (b): curated members always remain | ✅ | `gate_explicit_mode_admits_only_the_curated_core` |
| Test (c): small honest clusters preserved exactly | ✅ | `gate_preserves_small_honest_clusters_exactly` (k=0..=8: identical members **and order** and threshold vs the ungated construction; covers the 5-node testnet shape and solo 1-of-1 bootstrap) |
| `node_getStatus` exposes curated/auto counts + gate state | ✅ | `test_node_status_tracks_quorum_gate_snapshot` (null-until-published, live not constant) |
| Design step records chosen mechanism + justification | ✅ | This PR body, "Design decision" section |
| Explicit membership honors `members` (curation scope finding) | ✅ | `gate_explicit_mode_admits_only_the_curated_core` |
| Deterministic auto selection | ✅ | `gate_selection_is_deterministic_across_arrival_orders` |
| Snapshot follows #678 null-until-wired pattern, no fabricated values | ✅ | `test_node_status_tracks_quorum_gate_snapshot` steps (1)/(2) |
| `bth-quorum-sim` consumed as path dep, no reimplementation | ✅ | `botho/Cargo.toml`; gate calls `Fbas::symmetric(...).has_quorum_intersection()` |
| `reconfigure_quorum` boundary semantics unchanged | ✅ | `service.rs` logic untouched (only the `QuorumGateSnapshot` struct added); all `consensus::service` tests green |

Not included (listed as optional in curation): Prometheus gauges for the gate — can follow as a tiny separate change if wanted. P4 admin UI/API, reward mechanism, protocol changes: out of scope per the issue.

## Test Plan

- `cargo test -p botho --lib` — **1125 passed, 0 failed** (includes the 8 new gate tests, new config serde-default tests, and the new RPC snapshot test)
- `cargo test -p bth-quorum-sim` — **all green** (61 tests across units/integration)
- `cargo test -p botho --test consensus_cluster_convergence` — **4/4 passed** (2/3/4-node clusters converge, no fork, no stall — proves no consensus regression)
- `cargo clippy -p botho --lib` — no new warnings from this change (remaining warnings pre-exist on main)
- `cargo fmt -p botho -- --check` — clean

Closes #651

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 644
-# Branch: feature/issue-644
+# Issue: 651
+# Branch: feature/issue-651
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -790,6 +790,7 @@ dependencies = [
  "bth-crypto-pq",
  "bth-crypto-ring-signature",
  "bth-gossip",
+ "bth-quorum-sim",
  "bth-transaction-clsag",
  "bth-transaction-core",
  "bth-transaction-types",
--- a/botho/Cargo.toml
+++ b/botho/Cargo.toml
@@ -58,6 +58,9 @@ bth-common = { path = "../common", features = ["std"] }
 bth-consensus-scp = { path = "../consensus/scp", features = ["test_utils"] }
 bth-consensus-scp-types = { path = "../consensus/scp/types", features = ["test_utils"] }
 bth-gossip = { path = "../gossip" }
+# Quorum promotion gate (#651): exact FBAS quorum-intersection check used to
+# refuse quorum-set candidates that could admit disjoint quorums (a fork).
+bth-quorum-sim = { path = "../consensus/quorum-sim" }
 bth-crypto-digestible = { path = "../crypto/digestible" }
 libp2p = { workspace = true, features = [
     "tokio",
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -22,10 +22,10 @@ use std::collections::HashMap;
 
 use crate::{
     block::MintingTx,
-    config::{Config, QuorumMode},
+    config::{Config, QuorumConfig, QuorumMode},
     consensus::{
         BlockBuilder, ConsensusConfig, ConsensusEvent, ConsensusService, LotteryFeeConfig,
-        ScpSlotSnapshot, TransactionValidator,
+        QuorumGateSnapshot, ScpSlotSnapshot, TransactionValidator,
     },
     network::{
         default_seed_domain, BlockTxn, ChainSyncManager, CompactBlock, GetBlockTxn,
@@ -430,6 +430,14 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // can tell a jammed competing-coinbase round apart from an idle node.
     let scp_slot_status: Arc<RwLock<Option<ScpSlotSnapshot>>> = Arc::new(RwLock::new(None));
 
+    // Shared quorum-promotion-gate handle (#651, epic #441 §3/P5). The quorum
+    // rebuild path below publishes a fresh snapshot of the gate's evaluation
+    // (curated vs auto-promoted member counts, suppressed peers, intersection
+    // refusals) at the initial seed and on every peer-churn rebuild; the RPC
+    // layer surfaces it in `node_getStatus` so an operator can see the gate
+    // working. `None` until the first evaluation (never fabricated).
+    let quorum_gate_status: Arc<RwLock<Option<QuorumGateSnapshot>>> = Arc::new(RwLock::new(None));
+
     let mut rpc_state = RpcState::from_shared(
         node.shared_ledger(),
         node.shared_mempool(),
@@ -449,6 +457,9 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // Live SCP slot-stall observability for `node_getStatus` (#653). Shares
     // the handle the consensus tick publishes into.
     .with_scp_slot_status(scp_slot_status.clone())
+    // Quorum promotion gate observability for `node_getStatus` (#651). Shares
+    // the handle the quorum rebuild path publishes into.
+    .with_quorum_gate_status(quorum_gate_status.clone())
     .with_peers(peers_snapshot.clone())
     // Live traffic / connection-direction counters for `network_getInfo`
     // (#542). Shares the same handle the network event loop increments.
@@ -571,7 +582,16 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
     // solo. A genuine lone node (no peers connected) still yields a 1-of-1 solo
     // quorum and mints solo — no regression for `min_peers == 0`.
     let scp_quorum_set = if discovery.peer_count() > 0 {
-        rebuild_scp_quorum_set(&config, &discovery, &local_peer_id)
+        // Initial seed goes through the promotion gate (#651) like every
+        // later churn rebuild; there is no previous quorum set yet, so an
+        // intersection refusal falls back to a majority threshold inside the
+        // gate. Publish the gate snapshot so `node_getStatus` reflects the
+        // seed evaluation.
+        let (qs, gate_snapshot) = rebuild_scp_quorum_set(&config, &discovery, &local_peer_id, None);
+        if let Ok(mut guard) = quorum_gate_status.write() {
+            *guard = Some(gate_snapshot);
+        }
+        qs
     } else {
         build_scp_quorum_set(&quorum, &local_peer_id)
     };
@@ -852,7 +872,10 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             if let Ok(mut count) = peer_count.write() {
                                 *count = new_peer_count;
                             }
-                            // Update SCP peer count (all peers participate in consensus)
+                            // Update SCP peer count. NOTE: this is the raw connected-peer
+                            // count used by the #509 quorumFaultTolerant/quorumDegenerate
+                            // flags; actual quorum MEMBERSHIP is bounded by the promotion
+                            // gate (#651) and surfaced separately via QuorumGateSnapshot.
                             if let Ok(mut count) = scp_peer_count.write() {
                                 *count = new_peer_count;
                             }
@@ -873,9 +896,23 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             consensus.set_connected_peers(new_peer_count);
 
                             // Reconfigure the consensus quorum to include the
-                            // newly connected peer. This is what lifts a node
+                            // newly connected peer, THROUGH the promotion gate
+                            // (#651): curated members are always admitted,
+                            // auto-discovered peers only up to the configured
+                            // cap, and a candidate that fails the quorum-
+                            // intersection check is refused (the gate returns
+                            // the previous set, which reconfigure_quorum
+                            // debounces as a no-op). This is what lifts a node
                             // out of latched solo mode without a restart.
-                            let new_qs = rebuild_scp_quorum_set(&config, &discovery, &local_peer_id);
+                            let (new_qs, gate_snapshot) = rebuild_scp_quorum_set(
+                                &config,
+                                &discovery,
+                                &local_peer_id,
+                                Some(consensus.quorum_set()),
+                            );
+                            if let Ok(mut guard) = quorum_gate_status.write() {
+                                *guard = Some(gate_snapshot);
+                            }
                             if consensus.reconfigure_quorum(new_qs) {
                                 info!(
                                     threshold = consensus.quorum_set().threshold,
@@ -916,7 +953,10 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             if let Ok(mut count) = peer_count.write() {
                                 *count = new_peer_count;
                             }
-                            // Update SCP peer count (all peers participate in consensus)
+                            // Update SCP peer count. NOTE: this is the raw connected-peer
+                            // count used by the #509 quorumFaultTolerant/quorumDegenerate
+                            // flags; actual quorum MEMBERSHIP is bounded by the promotion
+                            // gate (#651) and surfaced separately via QuorumGateSnapshot.
                             if let Ok(mut count) = scp_peer_count.write() {
                                 *count = new_peer_count;
                             }
@@ -943,8 +983,19 @@ async fn run_async(config: Config, config_path: &Path, mint: bool) -> Result<()>
                             // is gone. rebuild_scp_quorum_set always yields a
                             // non-empty set (local node is always a member) and
                             // recomputes a satisfiable threshold, so churn can
-                            // never produce an empty or unreachable quorum.
-                            let new_qs = rebuild_scp_quorum_set(&config, &discovery, &local_peer_id);
+                            // never produce an empty or unreachable quorum. The
+                            // rebuild goes through the promotion gate (#651);
+                            // a suppressed auto peer may be promoted here to
+                            // backfill the departed member's slot.
+                            let (new_qs, gate_snapshot) = rebuild_scp_quorum_set(
+                                &config,
+                                &discovery,
+                                &local_peer_id,
+                                Some(consensus.quorum_set()),
+                            );
+                            if let Ok(mut guard) = quorum_gate_status.write() {
+                                *guard = Some(gate_snapshot);
+                            }
                             if consensus.reconfigure_quorum(new_qs) {
                                 info!(
                                     threshold = consensus.quorum_set().threshold,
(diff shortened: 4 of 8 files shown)


rjwalters:
❌ **Changes Requested** — one blocking item; everything else in this PR verified clean under maximum scrutiny.

## Blocking: `MAX_EXACT_SIM_NODES = 20` stalls the network event loop for up to ~47 seconds per churn event

The doc comment on `symmetric_quorum_has_intersection` says the exact check is "exact and fast" up to 20 nodes. I measured it (release build, `Fbas::symmetric(n, t).has_quorum_intersection()`):

| n | threshold | elapsed |
|---|---|---|
| 9 | 5 | 86 µs |
| 13 | 7 | 4.9 ms |
| 15 | 8 | **733 ms** |
| 17 | 9 | **3.3 s** |
| 18 | 10 | **9.2 s** |
| 19 | 10 | **26.4 s** |
| 20 | 11 | **46.6 s** |

The blow-up is not the 2^n enumeration — it's `minimal_quorums()`' O(|quorums|²) subset scan plus the pairwise intersection loop (`consensus/quorum-sim/src/analysis.rs`); at n=20/t=11 there are ~432k quorums and ~168k minimal ones.

Why this is blocking, not a note:
- The check runs **synchronously in the main network event loop** (`run.rs` peer connect/disconnect handlers), so a slow evaluation starves consensus ticks and all network event processing for its duration — a liveness stall introduced by the safety gate itself. Debug builds are substantially worse.
- It is reachable through **legitimate operator config**, and specifically through the configs this feature exists to encourage: Recommended mode with ≥6 curated members (n = 1 + curated + 8 ≥ 15 → 733 ms+), or an Explicit core of 14–19 members (n ≤ 20 → seconds to ~47 s). In Explicit mode the identical candidate is re-checked on **every** churn event, so a 15-member federation pays this repeatedly for an unchanging answer.
- Default config and the live 5-node testnet are unaffected (n ≤ 9 → ≤ 86 µs), which is why tests pass — but "default doesn't hit it" isn't a defense for a consensus-path stall.

**Fix (small):** lower `MAX_EXACT_SIM_NODES` to 13 or less (≤ 5 ms worst case). Nothing is lost: for a flat symmetric quorum the analytic rule `threshold > n/2` is *exact* at every n (quorums are precisely the member subsets of size ≥ threshold, so intersection ⟺ 2·threshold > n) — the sim call at small n is belt-and-braces validation, not added precision. Please also update the doc comment (which currently claims n ≤ 20 is cheap) and adjust `symmetric_intersection_check_matches_known_cases` so it still exercises both sides of the new boundary. If you'd rather keep exactness deeper, memoizing the (n, threshold) result would also work, but the constant change is simpler and sufficient.

## Non-blocking notes

1. **Explicit-mode zero-peer seed**: when `discovery.peer_count() == 0` at startup, the seed still routes to `build_scp_quorum_set` → 1-of-1 solo instead of the curated core — briefly inconsistent with this PR's own (correct) "membership must not follow connectivity" rationale. It's mitigated today: Explicit hardcodes `min_peers_to_wait = 1`, so the participation gate stays closed at 0 peers, and the first peer-connect rebuilds through the gate synchronously before the next consensus tick. Since you're touching the file anyway, consider routing the Explicit seed through `gated_scp_quorum_set` (it handles an empty `connected_peers` correctly) or adding a comment explaining why the solo seed is safe there. Recommended mode's 0-peer path must stay untouched (criterion c).
2. When a churn candidate is refused, the published snapshot describes the **refused candidate's** curated/auto/suppressed counts while the applied set is the previous one — acceptable since `quorumGateIntersectionRefused` disambiguates, just noting it's intentional.
3. `previous` is the *applied* set (`consensus.quorum_set()`), not a pending deferred set; a refusal while a different set is pending would re-pend the active set. Unreachable today (Recommended thresholds always intersect; Explicit candidates are connectivity-independent, so constant), so no change requested.

## What I verified (all clean)

- **Testnet preservation (critical)**: `gate_preserves_small_honest_clusters_exactly` asserts members **in order** plus threshold identical to the ungated construction for k=0..=8. I traced the code path independently: under the cap in Recommended mode, membership iterates `connected_peers` in peer-table order after self, threshold is `effective_threshold(n-1)`, `QuorumSet::new` does no normalization, and the intersection check passes (crash majority) so the candidate is returned unmodified. For today's 5-node testnet (self + 4, Recommended/Crash, cap 8) the output is byte-identical to pre-PR: 3-of-5, same member order. Deploying this PR cannot change the live cluster's quorum sets.
- **Explicit-mode deviation is sound**: including curated members whether or not connected keeps `n` (and thus the clamped threshold) fixed under partition; connectivity-following membership would let a minority side clamp into a small always-intersecting-locally-but-globally-forkable config. Minority now halts instead — safety over liveness, correct for a safety gate. It IS a behavior change for Explicit-mode operators (membership was previously all-discovered-peers with only the threshold explicit), but that is exactly the curator's scope finding, and it's documented in `config.rs` and the PR body. Live testnet runs Recommended → no live impact.
- **Model faithfulness**: `Fbas::symmetric(n, threshold)` (n nodes sharing one flat threshold-of-n slice) matches exactly what `reconfigure_quorum` installs; the check is local-view, introducing no new global-agreement assumption — SCP's tolerance of heterogeneous slices is unchanged.
- **Fallback safety**: seed fallback `floor(n/2)+1` always intersects; churn fallback returns the previous set which `reconfigure_quorum` debounces via order-insensitive equality (verified `QuorumSet::PartialEq` sorts before comparing). Recommended thresholds (crash `floor(n/2)+1`, BFT `n−floor((n−1)/3)`) always exceed n/2, so refusal fires only on Explicit misconfigs — deterministic given config, no flapping.
- **Determinism**: over-cap trimming sorts by derived NodeID; `gate_selection_is_deterministic_across_arrival_orders` is meaningful because `QuorumSet` equality is order-insensitive.
- **`reconfigure_quorum` semantics untouched** (service.rs gains only the snapshot struct); defer-at-active-slot / empty-set refusal / debounce all as before.
- **Observability (anti-#541–#544)**: handle genuinely wired via `with_quorum_gate_status`; RPC test proves null-until-published and live-not-constant.
- **Self-dedupe fix** (operator listing own PeerId in `members`) is a nice bonus — a duplicate NodeID would have made the set invalid.

## Test execution

1. `cargo test -p botho --lib` — ✅ 1125 passed, 0 failed (one initial failure of the #678 timing test `multi_node_convergence_stamps_stall_clock_and_reports_no_stall` was a flake caused by my concurrent 46-second benchmark saturating the CPU; passes in isolation and in a clean full run).
2. `cargo test -p bth-quorum-sim` — ✅ 61 passed.
3. `cargo test -p botho --test consensus_cluster_convergence` — ✅ 4/4 passed.
4. CI: build/coverage/fuzz-guard/benchmarks/rustfmt green; **cargo deny fails on RUSTSEC-2026-0204 (crossbeam-epoch)** — pre-existing on main, tracked in #679, not counted against this PR.

This is a strong implementation — the blocking item is a one-line constant plus doc/test touch-ups. Happy to re-review immediately.

rjwalters:
Addressed the blocking finding.

**Fix:** lowered `MAX_EXACT_SIM_NODES` from 20 to **13** in `symmetric_quorum_has_intersection` (`botho/src/commands/run.rs`). At n<=13 the brute-force sim is <=5 ms; above it the code falls back to the analytic rule `threshold > n/2`, which is *exact* (not an approximation) for a flat symmetric FBAS — its minimal quorums are exactly the threshold-sized member subsets, so quorum intersection holds iff `2*threshold > n`. This caps the worst-case synchronous cost on the network event loop at ~5 ms instead of the ~47 s measured at n=20.

**Doc:** the `MAX_EXACT_SIM_NODES` comment now records your measured release-build timings (86 µs @ n=9 ... 733 ms @ n=15 ... 46.6 s @ n=20) and notes the cost is `minimal_quorums()`' O(|quorums|²) subset scan, explaining why 13 is the bound and why the analytic fallback is exact.

**Tests:**
- Moved the boundary asserts in `symmetric_intersection_check_matches_known_cases` to straddle the new n=13/14 boundary (sim path at n=13, analytic path at n=14), keeping the n=25 analytic cases.
- Added `analytic_rule_matches_sim_for_flat_symmetric_quorums`: property check asserting the sim and analytic paths agree for every `(n, threshold)` with 1<=n<=13 — this guards the constant against future divergence.

Nothing else changed (per your pass on testnet byte-identity, Explicit deviation, fallback semantics, observability wiring).

**Test results:**
- `cargo test -p botho --lib` — 1126 passed, 0 failed (incl. the #678 timing test `multi_node_convergence_stamps_stall_clock_and_reports_no_stall`, green on the first run).
- `cargo test -p bth-quorum-sim` — all passed.

Ready for re-review.

rjwalters:
✅ **Approved!** (re-review after fdd6f090)

The single blocking finding from the previous pass is fully addressed. Verified each claim independently:

1. **`MAX_EXACT_SIM_NODES` lowered 20 → 13** in `symmetric_quorum_has_intersection` (`botho/src/commands/run.rs`). Worst case on the sim path is now ~5 ms (release) instead of ~47 s; the synchronous consensus-path stall is gone. The doc comment records the measured release-build timings (86 µs @ n=9 … 4.9 ms @ n=13 … 733 ms @ n=15 … 46.6 s @ n=20) and correctly attributes the blow-up to `minimal_quorums()`' O(|quorums|²) subset scan.
2. **Analytic fallback verified exact, not approximate.** The `else` branch is `threshold > n / 2`. For a flat symmetric FBAS the quorums are precisely the member subsets of size ≥ threshold, so minimal quorums are the threshold-sized subsets, and two t-subsets of n must intersect iff 2t > n. Integer `t > n/2` is equivalent to 2t > n for both even and odd n (checked both parities). So raising the boundary loses nothing and lowering it costs nothing.
3. **Tests:** `symmetric_intersection_check_matches_known_cases` now straddles the new boundary (sim path at 13/7 true, 13/6 false; analytic path at 14/8 true, 14/7 false; n=25 cases retained). New property test `analytic_rule_matches_sim_for_flat_symmetric_quorums` exhaustively checks all 91 (n, threshold) pairs for 1 ≤ threshold ≤ n ≤ 13 — sim and analytic agree everywhere; runs in 0.45 s even in a debug build, and it guards the constant against a future raise past a divergence point.
4. **Scope:** fdd6f090 touches exactly one file (`botho/src/commands/run.rs`, +57/−11), all within the doc comment, constant, and tests. No gate call sites, no `reconfigure_quorum`, no observability changes — everything approved in the first pass is untouched.

**Fixed during evaluation** (trivial, formatting-only): CI's pinned-nightly rustfmt flagged one over-wrapped line in the new property test; I applied `cargo fmt` and pushed 33cb019c (1 file, +1/−2, zero behavior change). rustfmt now passes.

## Test Execution
1. `cargo test -p botho --lib` — ✅ 1126 passed, 0 failed, 2 ignored (53 s).
2. Gate + intersection tests specifically — ✅ all 8 `gate_*` tests, `symmetric_intersection_check_matches_known_cases`, `analytic_rule_matches_sim_for_flat_symmetric_quorums`, and `test_node_status_tracks_quorum_gate_snapshot` pass.

## CI
All checks green (build, coverage, fuzz guard, benchmarks, rustfmt) except **cargo deny**, which fails on RUSTSEC-2026-0204 (crossbeam-epoch) — pre-existing environmental failure tracked in #679; the branch already contains main's ignore-list fix (eea51325) and the advisory post-dates it. Fails on main too; not counted against this PR. Merge state is UNSTABLE solely due to that check.

Ready to merge.

Issue #673, opened 2026-07-07
web-wallet: /pay WalletGate shows Create (not Unlock) for an existing locked wallet, and link-flow create silently overwrites the stored wallet


rjwalters:
## Summary

Two related defects around `/pay`'s `WalletGate` (`web/packages/web-wallet/src/pages/pay.tsx`) that combine into a wallet-overwrite (funds-loss) hazard for users who already have a wallet:

### 1. Cold-loading a pay link with an existing LOCKED wallet shows the CREATE form, with no way to unlock

`WalletGate` initializes its mode once, at mount:

```tsx
const [mode, setMode] = useState<SetupMode>(isLocked ? 'unlock' : 'create')
```

On a cold page load the wallet context has not hydrated yet, so the gate mounts with `hasWallet=false / isLocked=false` → `mode='create'`. When the context then hydrates to `hasWallet=true / isLocked=true`:

- the heading flips to **"Unlock your wallet to pay this request."** (it reads live `isLocked`),
- but `mode` stays `'create'`, so the body renders a **freshly generated mnemonic + "I've written down my recovery phrase" + "Create & Continue"**,
- and the Create/Import toggle is hidden (it renders only when `!isLocked`), so there is **no path to the unlock form at all**.

**Reproduced** on the live testnet (2026-07-07, local dev @ 4.0.0): create an encrypted wallet, generate a payment-request link from RequestModal, open the link in a new tab (full-page navigation always drops the in-memory key, so the wallet is locked). Observed page: "Unlock your wallet to pay this request." heading directly above a new-wallet mnemonic and "Create & Continue". Since a full-page navigation is the ONLY way to arrive at `/pay` (nothing links to it in-app), **every encrypted-wallet user who opens a pay link hits this state — the pay flow is effectively unusable for them**.

### 2. Proceeding with "Create & Continue" (or the `/claim` "Create one" flow) silently OVERWRITES the existing stored wallet

`createWallet`/`saveWallet` write `botho-wallet-mnemonic` / `botho-wallet-address` unconditionally — there is no "a wallet already exists here" guard in:
- `pay.tsx` `WalletGate` `handleCreate` / `handleImport`, and
- `claim.tsx` `handleSweep` (`await saveWallet(createdMnemonic)`).

A user funneled into the state from defect 1 (or a claim recipient who forgot they have a wallet on this device) replaces their stored (possibly un-backed-up, encrypted) seed with the new one. The old seed is unrecoverable afterwards → loss of any funds it held.

[...]

PR #682, opened 2026-07-07
security(web-wallet): enforce required-password policy on /pay and /claim inline wallet flows

### LINKED ISSUE #672: web-wallet: /pay and /claim inline wallet create/import persist the seed in PLAINTEXT — bypasses the #475 required-password policy
## Summary

The main wallet setup (`/wallet` → `WalletSetup`) enforces the #475 policy: a password is REQUIRED and the seed is always encrypted at rest (`wallet.tsx` gates `onCreate`/`onImport` behind `isPasswordValid`). But both link-entry funnels bypass it entirely:

1. **`/pay` → `WalletGate`** (`web/packages/web-wallet/src/pages/pay.tsx`): `handleCreate` calls `onCreate(newMnemonic)` and `handleImport` calls `onImport(seedPhrase)` — **no password argument, no password fields in the UI at all**.
2. **`/claim` → "Don't have a wallet? Create one"** (`web/packages/web-wallet/src/pages/claim.tsx`, `handleSweep`): `await saveWallet(createdMnemonic)` — no password.

`saveWallet(mnemonic /* no password */)` (`web/packages/core/src/wallet/index.ts:188`) then writes the **raw 12-word mnemonic in cleartext** to `localStorage['botho-wallet-mnemonic']` with `botho-wallet-encrypted='false'`. Per its own doc comment, the no-password path "is preserved only for backward compatibility and migration" — these two flows use it for brand-new wallets.

## Reproduced against the live testnet (2026-07-07, local dev server @ protocol 4.0.0)

- `/pay#<valid fragment>` in a fresh browser → gate shows Create → reveal → confirm checkbox → "Create & Continue" (**zero password inputs on the page**) → `localStorage`:
  - `botho-wallet-encrypted` = `false`
  - `botho-wallet-mnemonic` = plaintext 12-word phrase
- `/claim#<funded fragment>` → claimable → "Create one" → Claim → same plaintext result after the sweep persists the wallet.

## Expected

Link-flow wallet creation/import applies the same required-password policy as `WalletSetup` (#475) — or, if friction is deemed unacceptable mid-claim, at minimum:
- an explicit, visible opt-in to "no password (unprotected)" with the same warning language, and
- an immediate, prominent "Set a password" step right after the claim/pay completes (the dashboard's security card alone is easy to miss).

## Actual

A user whose first touch of Botho is a claim link or payment-request link (the primary sharing funnels) ends up with a plaintext seed at rest — exactly what #474/#475 eliminated for the main setup path.

## Pointers

- `web/packages/web-wallet/src/pages/pay.tsx` — `WalletGate` `handleCreate` / `handleImport` (no password plumbed)
- `web/packages/web-wallet/src/pages/claim.tsx` — `handleSweep` (`saveWallet(createdMnemonic)`)
- `web/packages/core/src/wallet/index.ts:188` — `saveWallet` plaintext branch

Context: regression-class finding for #474/#475 (those issues are CLOSED and fixed for the main setup; do not reopen — this is the same policy escaping via side doors added later, #460/#470). Found during the #637 survey.

## Acceptance criteria

- [ ] Creating a wallet through `/pay`'s WalletGate persists an encrypted seed (or an explicit, warned plaintext opt-out consistent with product policy).
- [ ] Importing a wallet through `/pay`'s WalletGate does the same.
- [ ] The `/claim` "Create one" wallet is persisted encrypted (or explicit opt-out + post-claim password nudge).
- [ ] A test asserting `botho-wallet-encrypted === 'true'` (or the opt-out marker) after each of the three flows.

rjwalters:
## Summary

Closes #672 (loom:urgent). Both link-entry funnels bypassed the #475 required-password policy and persisted the raw 12-word seed to `localStorage` in **plaintext** — for exactly the users whose first touch of Botho is a shared pay/claim link.

- **`/pay` → `WalletGate`**: both create and import modes now render the shared `PasswordFields` (same #475 policy as `WalletSetup`: min 8 chars, confirm field, strength hint). The Create/Import buttons stay disabled until the password is valid, and the password is plumbed through to `createWallet`/`importWallet` (which already forward it to `saveWallet`).
- **`/claim` → "Create one"**: the in-flow wallet panel now includes the password fields; the Claim button stays disabled until the password is valid, and `handleSweep` persists with `saveWallet(mnemonic, password)` so the stored seed is an encrypted vault blob. Copy updated to say the wallet is saved *encrypted*.

No changes to the wallet context or `saveWallet` were needed — both already accept and honor a password; the bug was purely the two call sites omitting it.

## Acceptance criteria (from #672)

- [x] `/pay` WalletGate create persists an encrypted seed — password required, plumbed to context
- [x] `/pay` WalletGate import — same
- [x] `/claim` "Create one" wallet persisted encrypted — `saveWallet(mnemonic, password)`
- [x] Tests asserting the encrypted-at-rest outcome for each of the three flows

## Tests

- `pay.test.tsx` (+2): create and import flows — buttons blocked without a valid password (including too-short), password reaches the context call.
- `claim.test.tsx` (+1): full flow with the REAL `saveWallet` (PBKDF2 600k) — asserts `botho-wallet-encrypted === 'true'` and the stored value is one opaque token, never a 12-word phrase.
- Full web suite: 48 files / 560 tests pass; `pnpm -r typecheck` clean.

### DIFF
--- a/web/packages/web-wallet/src/pages/claim.test.tsx
+++ b/web/packages/web-wallet/src/pages/claim.test.tsx
@@ -181,3 +181,81 @@ describe('ClaimPage valid-link survives StrictMode double-invoke (#654)', () => 
     await screen.findByText(/no claim link found|not valid/i)
   })
 })
+
+// This environment's jsdom does not provide localStorage; mirror the mock used
+// by core's wallet.test.ts so the REAL saveWallet can persist through it.
+const localStorageMock = (() => {
+  let store: Record<string, string> = {}
+  return {
+    getItem: (key: string) => store[key] ?? null,
+    setItem: (key: string, value: string) => { store[key] = value },
+    removeItem: (key: string) => { delete store[key] },
+    clear: () => { store = {} },
+  }
+})()
+Object.defineProperty(globalThis, 'localStorage', { value: localStorageMock })
+
+/**
+ * Required-password policy for the in-flow "Create one" wallet (#672).
+ *
+ * `handleSweep` previously persisted the freshly created wallet with
+ * `saveWallet(mnemonic)` — no password — writing the raw 12-word seed to
+ * localStorage in plaintext (`botho-wallet-encrypted = 'false'`), bypassing the
+ * #475 policy for exactly the users whose first wallet arrives via a claim
+ * link. This locks in: no claim while the password is missing, and the
+ * persisted seed is an encrypted vault blob.
+ */
+describe('ClaimPage in-flow wallet create persists ENCRYPTED (#672)', () => {
+  const VALID_PASSWORD = 'correct-horse-battery'
+
+  beforeEach(() => {
+    cleanup()
+    scanEphemeralMock.mockReset()
+    scanEphemeralMock.mockResolvedValue({
+      gross: 5_000_000_000_000n,
+      fee: 100_000_000n,
+      net: 4_900_000_000_000n,
+    })
+    sweepEphemeralMock.mockReset()
+    sweepEphemeralMock.mockResolvedValue({ txHash: 'tx' })
+    for (const spy of Object.values(adapterSpies)) spy.mockClear()
+    window.history.replaceState(null, '', '/claim')
+    localStorageMock.clear()
+  })
+
+  afterEach(() => {
+    window.location.hash = ''
+    localStorageMock.clear()
+  })
+
+  it('requires a password before claiming and stores a vault blob, not plaintext', async () => {
+    setClaimHash()
+    renderClaim()
+
+    fireEvent.click(await screen.findByRole('button', { name: /reveal/i }))
+    fireEvent.click(await screen.findByRole('button', { name: /create one/i }))
+
+    // Destination is pre-filled from the new wallet, but there is no password
+    // yet — the Claim button must stay disabled.
+    const claimBtn = screen.getByRole('button', { name: /^Claim / }) as HTMLButtonElement
+    expect(claimBtn.disabled).toBe(true)
+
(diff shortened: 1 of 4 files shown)


rjwalters:
## ✅ Approved (Judge review)

Verified against the #672 acceptance criteria:

1. **Correct root cause, minimal surface.** The context (`createWallet`/`importWallet`) and `saveWallet` already accepted a password; the bug was the two link-flow call sites omitting it. The fix changes only those call sites plus their UI — no crypto or storage-layer changes.
2. **/pay create + import**: `PasswordFields` (the shared #475 component — min 8, confirm, strength hint) gates both modes; buttons disabled until `isPasswordValid`, with a belt-and-suspenders guard inside the handlers. Password reaches the context call.
3. **/claim "Create one"**: password fields render inside the new-wallet panel; the Claim button is disabled while `persistingNewWallet && !valid`, `handleSweep` re-checks before any network side effect, and persistence is `saveWallet(mnemonic, password)`. Editing the destination manually (abandoning the created wallet) correctly drops the password requirement since nothing is persisted.
4. **Ordering is safe**: `saveWallet` runs before `sweepEphemeral`, so a persistence failure surfaces as an error without burning the one-shot claim.
5. **Tests**: the /claim test exercises the REAL `saveWallet` + PBKDF2 and asserts `botho-wallet-encrypted === 'true'` + opaque blob; the /pay tests pin both the disabled-gate behavior (including too-short password) and the password plumbing. Full web suite 560/560 + typecheck clean, verified locally.

Nit (non-blocking, tracked elsewhere): the gate's stale `mode` when the wallet context finishes loading after mount is #673's scope, not this PR's.

PR #684, opened 2026-07-07
fix(web-wallet): WalletGate unlock-mode hydration + explicit overwrite guard on link-flow wallet create/import

### LINKED ISSUE #673: web-wallet: /pay WalletGate shows Create (not Unlock) for an existing locked wallet, and link-flow create silently overwrites the stored wallet

rjwalters:
## Summary

Fixes #673 (priority:high). Two related `/pay` WalletGate defects that combined into a wallet-overwrite (funds-loss) hazard:

1. **Hydration race → no unlock path.** The gate's `mode` was initialized once at mount (`isLocked ? 'unlock' : 'create'`). On a cold `/pay` load the wallet context hydrates after mount, so every encrypted-wallet user opening a pay link saw the CREATE form — fresh mnemonic, "Create & Continue", and no way to reach the unlock form (the mode toggle is hidden while locked). Fixed by syncing `mode` to the live `isLocked` in an effect.
2. **Silent overwrite of the stored wallet.** `/pay` create/import and `/claim` "Create one" persisted the new seed unconditionally, replacing any stored (possibly un-backed-up) wallet. Both flows now read storage directly via `getWalletInfo()` — deliberately NOT the context state, which may not have hydrated — and stay disabled until the user ticks an explicit replacement warning that names the existing (shortened) address.

## Acceptance criteria (from #673)

- [x] Cold-loading `/pay#<request>` with an existing locked wallet lands on the unlock form (regression test simulates the hydration race via rerender)
- [x] Overwriting an existing wallet from `/pay` or `/claim` requires explicit confirmation naming the existing address
- [x] Regression test for the hydration race

## Tests

- `pay.test.tsx` (+3): hydration race flips gate to unlock; create-over-existing blocked until ack (warning names the stored address); no warning on clean devices.
- `claim.test.tsx` (+2): claim blocked until replacement ack when a wallet is stored; no warning baseline.
- Full web suite: 48 files / 565 tests pass; `pnpm -r typecheck` clean.

Stacks cleanly on #682 (the #672 password-policy fix in the same components).

### DIFF
--- a/web/packages/web-wallet/src/pages/claim.test.tsx
+++ b/web/packages/web-wallet/src/pages/claim.test.tsx
@@ -259,3 +259,70 @@ describe('ClaimPage in-flow wallet create persists ENCRYPTED (#672)', () => {
     expect(stored!.trim().split(/\s+/).length).toBe(1)
   }, 30000)
 })
+
+/**
+ * Overwrite guard for the in-flow "Create one" wallet (#673): claiming with a
+ * freshly created wallet on a device that ALREADY stores a wallet replaces the
+ * stored seed (funds loss). The claim must stay blocked until the recipient
+ * explicitly acknowledges the replacement, which names the existing address.
+ */
+describe('ClaimPage in-flow create requires overwrite acknowledgement (#673)', () => {
+  const VALID_PASSWORD = 'correct-horse-battery'
+
+  beforeEach(() => {
+    cleanup()
+    scanEphemeralMock.mockReset()
+    scanEphemeralMock.mockResolvedValue({
+      gross: 5_000_000_000_000n,
+      fee: 100_000_000n,
+      net: 4_900_000_000_000n,
+    })
+    sweepEphemeralMock.mockReset()
+    for (const spy of Object.values(adapterSpies)) spy.mockClear()
+    window.history.replaceState(null, '', '/claim')
+    localStorageMock.clear()
+  })
+
+  afterEach(() => {
+    window.location.hash = ''
+    localStorageMock.clear()
+  })
+
+  it('blocks the claim until the stored-wallet replacement is acknowledged', async () => {
+    // A wallet already sits in storage BEFORE the page loads.
+    localStorageMock.setItem('botho-wallet-mnemonic', 'opaque-vault-blob')
+    localStorageMock.setItem('botho-wallet-address', 'tbotho://1/ExistingAddr12345678')
+    localStorageMock.setItem('botho-wallet-encrypted', 'true')
+
+    setClaimHash()
+    renderClaim()
+
+    fireEvent.click(await screen.findByRole('button', { name: /reveal/i }))
+    fireEvent.click(await screen.findByRole('button', { name: /create one/i }))
+
+    // Password valid, destination pre-filled — only the overwrite ack missing.
+    fireEvent.change(screen.getByPlaceholderText(/^Password \(min/i), {
+      target: { value: VALID_PASSWORD },
+    })
+    fireEvent.change(screen.getByPlaceholderText(/Confirm password/i), {
+      target: { value: VALID_PASSWORD },
+    })
+
+    const claimBtn = screen.getByRole('button', { name: /^Claim / }) as HTMLButtonElement
+    expect(screen.getByText(/already has a wallet/i).textContent).toContain(
+      'tbotho://1/E',
+    )
+    expect(claimBtn.disabled).toBe(true)
+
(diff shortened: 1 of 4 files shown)


rjwalters:
## ✅ Approved (Judge review)

Verified against #673's acceptance criteria:

1. **Hydration fix is the minimal correct one**: `mode` syncs to the live `isLocked` via effect, so a late-hydrating locked wallet always lands on the unlock form; after unlock the gate unmounts into PayConfirm. The mode toggle stays hidden while locked, which is now safe because the mode can no longer be stale.
2. **Overwrite guard reads the right source of truth**: `getWalletInfo()` hits localStorage directly, so the guard works even inside the hydration window where the context still reports `hasWallet=false` — exactly the window defect 1 created. Both the handlers (early-return) and the buttons (disabled) enforce it, and the warning names the shortened stored address.
3. **Scope is correct on /claim**: the ack is only required when the in-flow-created wallet would actually be persisted (`persistingNewWallet`); pasting your own destination address is unaffected.
4. **Tests pin all three behaviors** including the hydration race via mock-swap + rerender, and a clean-device baseline so the warning can't leak into the common path. Full web suite 565/565 + typecheck verified locally.

Nit (accepted): `existingWallet` is snapshotted once per mount — a wallet cleared in another tab mid-flow would still warn. Conservative direction, fine.

Issue #637, opened 2026-07-05
web-wallet: survey pass — run against live testnet, walk every page, file concrete UX/correctness issues


rjwalters:
## Context

The PWA at https://wallet.botho.io (`web/packages/web-wallet`, Vite + React + wasm-signer, Cloudflare Pages) is functional but rough. Handoff (NEXT_SESSION.md 2026-07-05) calls for a structured survey before sweeping polish work.

## Task

Run the wallet locally against the live 5-node testnet (protocol 4.0.0, fresh genesis 2026-07-05) and walk **every page/flow**. File a concrete issue per finding rather than one mega-issue. Known areas of UX debt to specifically exercise:

- **Seed-phrase onboarding flow** — clarity, error handling, confirmation step.
- **Claim-link flow** — end-to-end on testnet; remember claim-link bearer secrets were part of the at-rest security work (#474/#475/#476 — verify no regressions while in there).
- **Error states when a node is unreachable** — every RPC-backed view should degrade legibly, not spin or show stale data as live.
- General rendering of fees/factors and cluster wealth (deep verification tracked separately in #634).

Already-filed items — don't duplicate: #634 (adapter/u128/fee-factor verification), #635 (send-flow fee display), #636 (network selector + eu/ap TLS).

## Notes

- Deploy notes for the Pages setup are in the operator's infra notes; local dev should just be the Vite dev server pointed at a testnet node.
- Security posture: threat model at docs/security/threat-model.md; at-rest encryption fixes #474/#475/#476 must not regress.
- Keep API/adapter-level fixes shared with `mobile/app` where possible.

rjwalters:
## Curator Enhancement

Curating this issue per user request. This is a survey meta-issue whose deliverable is a set of child issues — not a code PR. The acceptance criteria, page inventory, and run instructions below are designed to give the surveyor a concrete checklist so the survey produces actionable, non-duplicate findings.

---

## Page / Flow Inventory

All routes are registered in `web/packages/web-wallet/src/App.tsx`. The surveyor should walk each in order.

### Static pages (no wallet required)

| Route | Component | Notes |
|---|---|---|
| `/` | `LandingPage` (or redirect to `/wallet` on wallet.botho.io) | Hero, stats, features, CTA, footer nav |
| `/home`, `/about` | `LandingPage` | Same page, always reachable regardless of host |
[...]

PR #685, opened 2026-07-07
fix(web-wallet): per-event transaction history — unique row ids, netted sends, honest time/status fields

### LINKED ISSUE #675: web-wallet: Recent Transactions fabricates timestamps (Date.now), duplicates React keys for multi-output txs, and shows per-output amounts as sends
## Summary

`fetchHistory` in `web/packages/web-wallet/src/contexts/wallet.tsx` (~line 255) maps each OWNED OUTPUT to a history row and fills several fields with placeholders:

```ts
return entries.map((e) => ({
  id: e.txHash,          // one row PER OUTPUT — spend + change share the same txHash
  ...
  fee: 0n,               // hardcoded
  status: 'confirmed',   // hardcoded
  timestamp: Date.now(), // fabricated on every refresh
  confirmations: 0,      // hardcoded
}))
```

Three user-visible consequences, all reproduced on the live testnet (2026-07-07, local dev @ 4.0.0):

### 1. Duplicate React keys / duplicate rows for any send with change

A send spends an output AND creates a change output in the same tx → two history entries with the same `txHash` → `TransactionList` (`web/packages/features/src/wallet/components/transaction-list.tsx`, `key={tx.id}`) logs:

```
Encountered two children with the same key ... d67f6638ca452dfa0ab06f113bcf05f007828b22d91e15e875a49bd794c97df5
```

Observed in the browser console immediately after creating a claim link (funding tx = spend 1.00 + change 0.7997). React may drop/duplicate rows under this condition.

### 2. Every transaction is timestamped "just now", forever

`timestamp: Date.now()` means `TransactionRow`'s `formatRelativeTime(tx.timestamp)` shows the scan time, not the tx time. Reload tomorrow and your week-old faucet dispense still says it just happened. The block height is available on each entry (`e.blockHeight`) — the block timestamp could be threaded through, or the row should show height instead of a fake time.

### 3. Per-output amounts read like wrong balances

After sending 0.2 BTH (+0.0003 fee) from a 1.00 BTH wallet, Recent Transactions shows `Sent -1.00` and `Received +0.7997` (input consumed + change) rather than a single `Sent -0.2003`. Technically complete, but users will read "-1.00" as "I lost 1 BTH". Worth folding spend+change of the same tx into one net row.

Also: `fee: 0n`, `confirmations: 0`, `status: 'confirmed'` are placeholders in the same mapper (hardcoded-observability class, cf. #541–#544 — don't render fields that aren't wired).

## Repro

1. Wallet with a single 1.00 BTH output (faucet).
2. Send via Link (or Send) any amount < 1.
3. Open devtools console → duplicate-key warning; Recent Transactions shows -1.00 / +0.7997 rows both "just now".

## Pointers

- `web/packages/web-wallet/src/contexts/wallet.tsx` — `fetchHistory` mapper
- `web/packages/features/src/wallet/components/transaction-list.tsx` — `key={tx.id}`
- `web/packages/features/src/wallet/components/transaction-row.tsx` — relative-time rendering

Note for implementers: if the fix lands in shared history-building (`buildOwnedHistory` / adapters), keep it shared with `mobile/app` per the #637 guardrails.

Found during the #637 survey.

## Acceptance criteria

- [ ] History keys are unique (e.g. `${txHash}:${outputIndex}` or net-per-tx rows).
- [ ] Rows show real chain time (block timestamp) or block height — not `Date.now()`.
- [ ] A send of X from a Y-sized output displays as a net `-X(-fee)` row (or the per-output presentation is deliberately ratified and labeled).

rjwalters:
## Summary

Fixes #675 (priority:medium). `fetchHistory` rendered `buildOwnedHistory`'s raw per-output entries 1:1, causing all three reported defects plus two more hardcoded-observability artifacts (#541-class):

| Defect | Cause | Fix |
|---|---|---|
| Duplicate React keys | receive + spend of one output share a txHash; multi-output receives repeat theirs | stable per-event ids (`recv-<txHash>`, `send-<height>`, `send-pending`) |
| "Sent -1.00 / Received +0.7997" | input + change rendered separately | send netted against same-block change |
| Every tx "just now", forever | `timestamp: Date.now()` on each refresh | `timestamp: 0` = explicit "unknown"; row renders `Block #N` (or `Pending`) |
| Every tx spins "Confirming" forever | `confirmations: 0` hardcoded (row treats <6 conf as confirming) | real `chainHeight - blockHeight + 1` |
| Mempool-pending spends shown "confirmed" | `status` hardcoded | real `pending`/`confirmed` from the key-image scan |

## Design

New **`netOwnedHistory`** lives in `wasm-signer` next to `buildOwnedHistory` (shared with `mobile/app` per the #637 guardrail) as a **pure function** over `HistoryEntry[]` — fully unit-testable without wasm or a node.

**Netting heuristic, documented inline:** the ring signature hides which tx consumed an output, so a spend only knows its `spentHeight`. Owned outputs received in that same block are treated as change and subtracted. If the subtraction would go negative (a genuine same-block incoming payment), the function refuses to net and keeps the receives visible — an incoming amount is never silently swallowed. Fee stays `0n` (not knowable client-side; the row does not render it).

## Tests

- 7 new `netOwnedHistory` unit tests: unique ids, change netting, multi-output grouping, negative-net guard (incoming payment preserved), pending-spend sorting, multi-input same-block merge, empty input.
- Full web suite 572/572 passes; `pnpm -r typecheck` clean.

### DIFF
--- a/web/packages/features/src/wallet/components/transaction-row.tsx
+++ b/web/packages/features/src/wallet/components/transaction-row.tsx
@@ -89,9 +89,11 @@ export function TransactionRow({
     tx.type === 'minting' ? 'Minting Reward' : isReceive ? 'Received' : 'Sent'
 
   // Prefer a friendly contact name for the counterparty, falling back to the
-  // raw address (truncated via CSS) and finally the tx id.
+  // raw address (truncated via CSS). No fallback to tx.id: netted history
+  // rows (#675) carry synthetic ids, and the ring hides the real
+  // counterparty anyway — an empty slot beats a meaningless token.
   const counterpartyName = tx.counterparty ? resolveName?.(tx.counterparty) : undefined
-  const counterparty = counterpartyName || tx.counterparty || tx.id
+  const counterparty = counterpartyName || tx.counterparty || ''
 
   return (
     <motion.div
@@ -115,13 +117,17 @@ export function TransactionRow({
           {showPrivacy && <PrivacyBadge cryptoType={tx.cryptoType} />}
         </div>
         <div className="mt-0.5 flex items-center gap-2">
-          <span
-            className={`max-w-[180px] truncate text-xs text-[--color-dim] ${counterpartyName ? '' : 'font-mono'}`}
-            title={tx.counterparty}
-          >
-            {counterparty}
-          </span>
-          <span className="text-[--color-muted]">•</span>
+          {counterparty && (
+            <>
+              <span
+                className={`max-w-[180px] truncate text-xs text-[--color-dim] ${counterpartyName ? '' : 'font-mono'}`}
+                title={tx.counterparty}
+              >
+                {counterparty}
+              </span>
+              <span className="text-[--color-muted]">•</span>
+            </>
+          )}
           {/* A zero timestamp means "no wall-clock time known" (client-side
               history has only block heights, #675): show the height rather
               than fabricating a relative time. */}
--- a/web/packages/features/src/wallet/components/transaction-row.tsx
+++ b/web/packages/features/src/wallet/components/transaction-row.tsx
@@ -122,11 +122,20 @@ export function TransactionRow({
             {counterparty}
           </span>
           <span className="text-[--color-muted]">•</span>
+          {/* A zero timestamp means "no wall-clock time known" (client-side
+              history has only block heights, #675): show the height rather
+              than fabricating a relative time. */}
           <span
             className="text-xs text-[--color-dim]"
-            title={formatAbsoluteTime(tx.timestamp)}
+            title={tx.timestamp > 0 ? formatAbsoluteTime(tx.timestamp) : undefined}
           >
-            {formatRelativeTime(tx.timestamp)}
+            {tx.timestamp > 0
+              ? formatRelativeTime(tx.timestamp)
+              : tx.blockHeight != null
+                ? `Block #${tx.blockHeight}`
+                : tx.status === 'pending'
+                  ? 'Pending'
+                  : '—'}
           </span>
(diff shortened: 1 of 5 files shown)


rjwalters:
## ✅ Approved (Judge review)

Verified against #675:

1. **All five defects addressed** (three reported + two more of the same hardcoded-observability class): unique per-event ids kill the duplicate-key warnings; sends net against same-block change; `timestamp: 0` is an explicit "unknown" marker rendered as `Block #N`/`Pending` instead of a fabricated relative time; confirmations are real (`chainHeight - blockHeight + 1`), which also makes the row's `Confirming <6 conf` state honest; status reflects the actual key-image scan (pending vs confirmed).
2. **The netting heuristic is sound and safely bounded.** The ring hides the consuming tx, so spentHeight-matching is the only client-side link; the negative-net guard means a genuine same-block incoming payment is never swallowed (pinned by test). Worst case for the heuristic is a rare cosmetic merge of two same-block events, never a wrong balance.
3. **Placement respects the #637 guardrail**: `netOwnedHistory` is a pure function in `wasm-signer` beside `buildOwnedHistory`, reusable by `mobile/app`; `buildOwnedHistory`'s contract is untouched for existing consumers.
4. **Checked consumers of `tx.id`**: only a React key and a cosmetic counterparty fallback — the follow-up commit removes that fallback so synthetic ids never render as text (the ring hides the real counterparty anyway).
5. 7 new pure-function unit tests cover ids, netting, grouping, the negative-net guard, pending sorting, and multi-input merges. Full web suite 572/572 + typecheck verified locally.

PR #686, opened 2026-07-07
test: fix two broken privacy doctests and run doctests in CI

### LINKED ISSUE #670: test: two broken doctests in botho/src/network/privacy (capacity, selection) — doctests not run in CI

rjwalters:
## Summary

Fixes #670. Two module doctests in `botho/src/network/privacy` failed on unmodified main:

- **`capacity.rs`** — compile error: the example called `RelayCapacity::measure(&stats)` without importing the `RelayCapacityExt` extension trait that provides it. Added the import.
- **`selection.rs`** — runtime panic: peers built with `RelayCapacity::default()` score 0 and can never clear `SelectionConfig::default()`'s `min_score: 0.2`, so `select_diverse_hops` returned `NoQualifiedPeers`. The example now measures a realistic capacity (same `SimpleNodeStats` recipe as `capacity.rs`'s own example) with a comment explaining why a default capacity is never selectable.

## CI gap closed

No workflow ran doctests (`--no-run` does not even compile them), which is why these rotted silently. Added a `cargo test --workspace --doc` step to `workspace-build.yml` — verified green across the whole workspace locally (`25 passed` for botho, all other crates clean), and it's fast/deterministic so it runs for real rather than compile-only.

### DIFF
--- a/.github/workflows/workspace-build.yml
+++ b/.github/workflows/workspace-build.yml
@@ -78,3 +78,9 @@ jobs:
       # required check fast and deterministic while still failing on rot (#562).
       - name: Compile-check all test targets (no run)
         run: cargo test --workspace --no-run
+
+      # Doctests are NOT compiled by `--no-run` above and no other workflow
+      # runs them, so documentation examples rotted silently (#670). They are
+      # fast (seconds) and deterministic, so run them for real.
+      - name: Run doctests
+        run: cargo test --workspace --doc
--- a/botho/src/network/privacy/capacity.rs
+++ b/botho/src/network/privacy/capacity.rs
@@ -26,7 +26,9 @@
 //! # Example
 //!
 //! ```
-//! use botho::network::privacy::capacity::{RelayCapacity, NatType, SimpleNodeStats, NodeStats};
+//! use botho::network::privacy::capacity::{
+//!     RelayCapacity, RelayCapacityExt, NatType, SimpleNodeStats,
+//! };
 //!
 //! // Create stats for a well-connected node
 //! let stats = SimpleNodeStats {
--- a/botho/src/network/privacy/selection.rs
+++ b/botho/src/network/privacy/selection.rs
@@ -23,28 +23,38 @@
 //! use botho::network::privacy::{
 //!     CircuitSelector, SelectionConfig, RelayPeerInfo, RelayCapacity, NatType,
 //! };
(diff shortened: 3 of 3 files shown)


rjwalters:
## ✅ Approved (Judge review)

- Both fixes are documentation-example-only — no library code changes. The capacity fix is the exact import the compiler help suggested; the selection fix keeps the example honest (it documents WHY a default capacity can never be selected, rather than lowering the config's min score to make a hollow example pass).
- The CI step closes the actual root cause (#670's "why nobody noticed"): `--no-run` never compiles doctests. Verified `cargo test --workspace --doc` passes across the whole workspace locally, so the new step will not redden main.
- Placement in `workspace-build.yml` after the compile-check step is right — same job, shared cache, seconds of added runtime on a required check.

Issue #639, opened 2026-07-05
docs: sweep stale nanoBTH unit references — monetary unit is picocredits (u128)


rjwalters:
## Context

The monetary unit decision is settled: **picocredits + u128** (nanoBTH was considered and rejected; see the monetary unit & supply decision, #351 era). But `grep -rli nanobth` still hits **36 files** across docs, code comments, and possibly live identifiers — anyone calibrating against these will use wrong units.

## Task

- Sweep all nanoBTH references to picocredits: `docs/` (README, api.md, glossary, exchange-integration, merchant-guide), `AUDIT.md`, `PLAN.md`, `web/packages/web-wallet/src/pages/docs.tsx`, and code comments (`botho/src/mempool.rs`, `botho/src/rpc/mod.rs`, `cluster-tax/src/dynamic_fee.rs`, `transaction/core/src/validation/*`, etc.).
- **Audit, don't blind-rename**: `transaction/types/src/constants.rs` and `transaction/core/src/token.rs` may have identifiers or constants that still *encode* nanoBTH semantics rather than just mentioning the word. If any live value is denominated in the wrong unit, that's a bug — file it separately rather than folding it in here.
- Historical audit reports (`audits/2025-12-*`, `audits/2026-01-*`) are records of their time — leave them as-is.

## Adjacent fix while in `cluster-tax/src/dynamic_fee.rs`

Its module docs claim the congestion multiplier goes "up to 100x", but the formula `e^(k·(ema−target))` with k=8 and max excess 0.25 tops out at e² ≈ 7.4x — the `base_max: 100` ceiling is unreachable. Correct the docs to match the actual response curve (fee-system calibration itself is a separate discussion).

## Acceptance

`grep -rli nanobth` returns only the historical audit files.

rjwalters:
Curating this issue per user request. Verified file inventory below via `grep -rli nanobth` from repo root; all 36 hits classified.

---

## Curator Enhancement

### Verified File Inventory (36 files total)

**Total breakdown: 17 Category A (safe rename) + 15 Category B (audit-first) + 4 Category C (leave untouched)**

---

#### Category A — Docs/prose/comments: safe to rename (17 files)

These files contain `nanoBTH` only in human-readable prose, document comments, or markdown. No live identifiers or semantic encoding. Replace with "picocredits" throughout.

| File | Notes |
|------|-------|
| `AUDIT.md` | Top-level audit doc |
| `PLAN.md` | Roadmap doc |
[...]

PR #687, opened 2026-07-07
test: rescale cluster-factor e2e expectations to the picocredit curve and run the suite in CI

### LINKED ISSUE #669: test: e2e_progressive_fees cluster-factor test is stale after #626/#627 curve recalibration (fails on main, not run in CI)

rjwalters:
## Summary

Fixes #669. `test_cluster_factor_wealthy_pay_more` fed the curve pre-#626 raw values (`1_000_000u128`…`100_000_000u128`, labeled "10M wealth (w_mid)"). Since #627 the production curve measures wealth in **picocredits** (`W_MID_PICO` = 100k BTH = 1e17) and returns an exact 1x below `W_MID >> 12` (~24.4 BTH), so every legacy case scored 1000 and `factor_max >= 5000` could never pass. No workflow ran the suite, so the recalibration stranded it silently.

## Changes

- **Rescaled cases** span the whole deployed curve: 0, 1 BTH (floor), 100k BTH (= `W_MID_PICO`), 10M BTH, 500M BTH (ceiling).
- **Exact invariants pinned** (verified against the live curve before writing them): `1000` on the floor (0 and 1 BTH), `3500` at the midpoint (sigmoid(0)=0.5 → 1000 + 5000/2), `6000` at/above the ceiling. These are the #626 spec's exact-saturation guarantees, so `assert_eq!` is right — no fuzzy ranges.
- **Fee column relabeled** picocredits (nanoBTH is a dead unit, #639-class cleanup).
- **CI**: `e2e_progressive_fees` added to the consensus-integration job in `e2e-tests.yml` (fast + deterministic) so future curve changes must keep the suite current.

## Verification

All 9 tests in the suite pass locally (`cargo test -p botho --test e2e_progressive_fees`), ~30s.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -178,6 +178,11 @@ jobs:
         run: |
           cargo test --package botho --test e2e_consensus_integration -- --test-threads=1
           cargo test --package botho --test network_integration -- --test-threads=1
+          # Progressive-fee curve suite (#669): fast and deterministic, but it
+          # was in no workflow, so the #626/#627 curve recalibration silently
+          # stranded its expectations. Runs here so fee-curve changes must
+          # keep the e2e expectations current.
+          cargo test --package botho --test e2e_progressive_fees -- --test-threads=1
 
   summary:
     name: E2E Test Summary
--- a/botho/tests/e2e_progressive_fees.rs
+++ b/botho/tests/e2e_progressive_fees.rs
@@ -269,7 +269,10 @@ fn test_cluster_factor_wealthy_pay_more() {
     let factor_mid = fee_config.cluster_factor(100_000 * PICO_PER_BTH);
     let factor_max = fee_config.cluster_factor(500_000_000 * PICO_PER_BTH);
 
-    assert_eq!(factor_zero, 1000, "Zero wealth must sit on the exact 1x floor");
+    assert_eq!(
+        factor_zero, 1000,
+        "Zero wealth must sit on the exact 1x floor"
+    );
     assert_eq!(
         factor_floor_edge, 1000,
         "1 BTH is far below W_MID >> 12 and must sit on the exact 1x floor"
--- a/botho/tests/e2e_progressive_fees.rs
+++ b/botho/tests/e2e_progressive_fees.rs
@@ -218,19 +218,23 @@ fn test_cluster_factor_wealthy_pay_more() {
     let fee_config = FeeConfig::default();
     let tx_size = 4000; // ~4KB typical CLSAG transaction
 
-    // Test cluster factors at different wealth levels
+    // Wealth is measured in PICOCREDITS since the #626/#627 recalibration:
+    // the deployed log-domain curve is exactly 1x below `W_MID >> 12`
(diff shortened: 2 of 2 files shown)


rjwalters:
## ✅ Approved (Judge review)

- Test-only + CI-workflow change; no library code touched.
- The new expectations are the #626 spec's exact-saturation invariants, not re-derived approximations — I verified the values (1000/3500/6000) against the live `FeeConfig::default()` curve at exactly those wealths before they were written into the test, so the `assert_eq!`s pin real behavior. Using exact equality is correct here because the spec guarantees exact tails (`lut_sigmoid` clamps) and an exact midpoint knot; any future curve change SHOULD fail this test loudly.
- Case spread (floor edge / midpoint / ceiling / two interior points) covers the monotone-fee assertion meaningfully, unlike the old all-on-the-floor cases.
- CI placement in the existing consensus-integration job is right: same cache, ~30s suite, and closes the "no workflow runs it" hole that stranded it.

PR #689, opened 2026-07-07
test(e2e): un-rot the live-smoke suite — #475 password gate, strict-safe selectors, dev-server PWA skip

### LINKED ISSUE #676: e2e(live-smoke): suite is stale — 5/7 tests fail on the current build (#475 password gate + NetworkSelector selector drift), gutting the deploy-regression guard
## Summary

The gated live-smoke suite (`web/e2e/tests/live/smoke.spec.ts`, run via `playwright.live.config.ts`) exists to catch deploy regressions (the wasm-404 class). It is stale: **5 of 7 tests currently fail against a freshly built current bundle**, so it can no longer serve as a deploy gate — a real deploy break would be indistinguishable from the suite's own rot.

Verified 2026-07-07 by running it against a local dev server of current `main` (`BOTHO_LIVE=1 BOTHO_LIVE_URL=http://localhost:3000 npx playwright test --config e2e/playwright.live.config.ts`): `5 failed / 2 passed`.

## Failures and causes

1. **`wallet create is fully client-side …`** and 3. **`WASM crypto module loads …`** — the shared helper `createWalletClientSide()` clicks reveal + checkbox and then clicks "Create Wallet", but since #475 the button is disabled until a password + confirmation are filled (`wallet.tsx` `isPasswordValid` gate). The helper never fills a password → times out on the disabled button. The suite predates the policy (last touched in #466).
2. **`wallet import of a known mnemonic …`** — same: "Import Wallet" is now password-gated; the test fills only the mnemonic.
3. **`node ingress picker …`** — strict-mode violation: `getByRole('button', { name: /^Seed \(validator\)/ })` now resolves to 2 elements (the header trigger's accessible name includes the selected node, "Seed (validator) Testnet", plus the dropdown row "Seed (validator) height N"). Selector drift.
4. **`PWA manifest + service worker are served`** — fails against the dev server only (dev doesn't emit `/sw.js`); expected to pass against the deployed bundle. Worth an in-test skip or docs note so local runs aren't misleading.

Passing: landing page, `/claim` invalid state.

## Expected

The suite passes against the current build so its failures MEAN something. Ideally it also runs on a schedule (its config is deliberately not wired into per-PR CI).

## Suggested fix

- `createWalletClientSide()`: fill the password + confirm fields (placeholders `Password (min N characters)` / `Confirm password`) before clicking Create; same for the import test.
- Ingress test: disambiguate (scope to the dropdown container, or `exact`/role tweaks).
- Gate or annotate the SW/manifest assertions when `BOTHO_LIVE_URL` is a dev server.
- Consider extending coverage with the now-walkable flows from the #637 survey (docs path-deep-link redirect, pay/claim fragment survival, modal Escape dismissal) so the fixed regressions (#654/#655/#656 via #657/#666/#667) stay fixed.

Found during the #637 survey (the survey's test plan requires this suite to be green before "already works" claims — it could not be).

## Acceptance criteria

- [ ] `BOTHO_LIVE=1` run against the deployed wallet passes 7/7.
- [ ] Helper(s) reflect the #475 password-required policy.
- [ ] Strict-mode-safe selectors for the ingress picker.

rjwalters:
## Summary

Fixes #676 (priority:medium). The live-smoke deploy-regression guard failed 5/7 against the current build, so a real deploy break was indistinguishable from suite rot. All four failure causes fixed:

1. **#475 password gate**: `createWalletClientSide()` and the import test now fill the password + confirm fields via a shared `fillPasswordFields()` helper — previously the Create/Import buttons stayed disabled and every wallet-flow test timed out.
2. **Ingress picker strict-mode violation**: node-row lookups are scoped to the dropdown container (the trigger's accessible name also starts with the selected node's name, e.g. "Seed (validator) Testnet").
3. **A second masked drift**: the dashboard "Send" assertion now uses `exact: true` — "Send via Link" made `/Send/i` ambiguous.
4. **PWA test vs dev servers**: skips when `BOTHO_LIVE_URL` is localhost (dev doesn't emit `/sw.js`); `BOTHO_LIVE_EXPECT_PWA=1` forces it for `vite preview` runs.

## Verification

Ran against a local dev server of current main (`BOTHO_LIVE=1 BOTHO_LIVE_URL=http://localhost:3000 E2E_BROWSER_CHANNEL=chrome`): **6 passed, 1 skipped (PWA, by design)** — was 5 failed / 2 passed. Against the deployed site all 7 run.

Acceptance criteria: helpers reflect #475 ✔; strict-safe selectors ✔; the full 7/7 pass against the deployed wallet lands with the v0.3.1 deploy (the suite will be run as part of that verification).

### DIFF
--- a/web/e2e/tests/live/smoke.spec.ts
+++ b/web/e2e/tests/live/smoke.spec.ts
@@ -37,6 +37,17 @@ const LIVE = process.env.BOTHO_LIVE === '1'
 
 const BASE_URL = process.env.BOTHO_LIVE_URL ?? 'https://wallet.botho.io'
 
+// A vite DEV server does not emit /sw.js or the precache manifest — only a
+// built bundle does. Skip the PWA assertions when pointed at local dev so a
+// local run isn't misleading (#676); force them with BOTHO_LIVE_EXPECT_PWA=1
+// when serving a production build locally (`vite preview`).
+const IS_LOCAL_DEV =
+  /^https?:\/\/(localhost|127\.0\.0\.1)/.test(BASE_URL) &&
+  process.env.BOTHO_LIVE_EXPECT_PWA !== '1'
+
+// Matches PasswordFields' placeholder (min length lives in @botho/core).
+const PASSWORD = 'live-smoke-password'
+
 // Known BIP39 test mnemonic — produces a deterministic testnet address. Never
 // holds real funds. Mirrors web/e2e/fixtures/test-data.ts TEST_MNEMONIC_12.
 const TEST_MNEMONIC_12 =
@@ -84,6 +95,10 @@ test.describe('Live smoke @ wallet.botho.io', () => {
   })
 
   test('PWA manifest + service worker are served', async ({ page, request }) => {
+    test.skip(
+      IS_LOCAL_DEV,
+      'dev server does not emit /sw.js — run against a built bundle or set BOTHO_LIVE_EXPECT_PWA=1',
+    )
     await page.goto(`${BASE_URL}/`, { timeout: NAV_TIMEOUT })
 
     // Manifest link is present in the document head and resolves to a 200.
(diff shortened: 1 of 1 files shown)


rjwalters:
## ✅ Approved (Judge review)

Single-file test change; verified by running the suite before and after against a local dev server of current main:

- Before: 5 failed / 2 passed (exactly the #676 report). After: 6 passed / 1 skipped — and the skip is the honest one (dev servers genuinely do not emit `/sw.js`; the `BOTHO_LIVE_EXPECT_PWA=1` escape hatch keeps `vite preview` runs fully covered, and deployed-site runs execute all 7).
- The password helper mirrors the real #475 UI (same placeholders as `PasswordFields`), so the suite tests the policy rather than bypassing it.
- Both strict-mode fixes address real ambiguity introduced by later UI work (trigger label includes selected node name; "Send via Link" collides with `/Send/i`) — scoping/exactness is the right fix, not `.first()`, because it pins WHICH element must exist.
- The suite remains fully opt-in (`BOTHO_LIVE=1` + separate config); no CI implications.

Note for the record: the deployed-site 7/7 acceptance run happens as part of the v0.3.1 deploy verification.

PR #691, opened 2026-07-07
test(e2e): don't count CSP-violation noise as wasm console errors in live-smoke

rjwalters:
## Summary

Follow-up to #689 found during the deployed-site acceptance run for #676: the wasm console-error filter matched the bare word "wasm", which also appears in `wasm-unsafe-eval` inside CSP-violation messages — the Cloudflare analytics beacon being blocked by the wallet's own `script-src` produced exactly that, failing the deploy-404 guard as a false positive. The request-level assertion (no failed `/pkg/` fetches) was already green.

Filter now matches the signer artifact specifically (`bth_wasm_signer`, `/pkg/`, `.wasm`).

## Verification

`BOTHO_LIVE=1` against the deployed **wallet.botho.io** (fresh v0.3.1-era bundle): **7/7 passed** — completing the #676 acceptance criterion "passes against the deployed wallet".

### DIFF
--- a/web/e2e/tests/live/smoke.spec.ts
+++ b/web/e2e/tests/live/smoke.spec.ts
@@ -67,7 +67,10 @@ test.describe('Live smoke @ wallet.botho.io', () => {
     const wasmErrors: string[] = []
     const failedWasm: string[] = []
     page.on('console', (msg) => {
-      if (msg.type() === 'error' && /wasm|bth_wasm_signer|\/pkg\//i.test(msg.text())) {
+      // Match the signer artifact specifically — a bare /wasm/ also matches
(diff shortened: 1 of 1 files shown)


rjwalters:
## ✅ Approved (Judge review)

One-line test-filter tightening, verified both directions: the false positive reproduces on the deployed site with the old regex (CSP message for the blocked Cloudflare beacon contains "wasm-unsafe-eval"), and the guard still catches the real regression class — a 404'd `/pkg/…bth_wasm_signer_bg.wasm` fetch produces both a `requestfailed` entry (unchanged assertion) and a console error naming the artifact, which the new regex still matches. 7/7 against production confirmed.

PR #692, opened 2026-07-07
fix(network): re-dial bootstrap peers whenever under-connected, not only at zero peers

### LINKED ISSUE #690: network: reconnect tick only re-dials bootstrap peers at peer_count == 0 — partial-mesh star topologies persist indefinitely
## Problem

Observed during the v0.3.1 rolling deploy (2026-07-07): after each node restarted, some bootstrap dials failed (their targets were mid-restart themselves). The result was a star topology — seed2 with 4 peers as the hub, every other node with exactly 1 peer — that persisted for 10+ minutes with **SCP unable to externalize** (slot 216 stalled, `slotStalled: true` on faucet/eu, pending txs unminable) until the nodes were manually restarted in a second pass.

## Root cause

`botho/src/commands/run.rs` reconnect tick:

```rust
_ = reconnect_interval.tick() => {
    if discovery.peer_count() == 0 && !reconnect_bootstrap_peers.is_empty() {
        // re-dial bootstrap peers
```

The guard is `peer_count() == 0`. A node that reached exactly ONE peer never re-dials its remaining configured bootstrap peers, so a partially-formed mesh never heals. The 5-node testnet's configs list 2 bootstrap peers per node; losing one dial during a rolling restart leaves the node permanently under-connected.

## Impact

- Rolling deploys/restarts can wedge SCP for the whole network even though every node is up and each reports "Quorum satisfied" (the star hub relays gossip but nomination cannot reach agreement).
- Any transient single-dial failure at startup permanently halves a node's connectivity.

## Suggested fix

Re-dial any configured bootstrap peer that is not currently connected (e.g. compare `reconnect_bootstrap_peers` against the connected set each tick), not just at zero peers. Redundant dials to already-connected peers are cheap no-ops for libp2p. Optionally back off after N consecutive failures per peer.

## Repro evidence

2026-07-07 15:14–15:30Z rolling v0.3.1 deploy: post-deploy `network_getPeers` showed seed/faucet/eu/ap each pinned at 1 peer for >10 min; `node_getStatus.slotStalled=true` on faucet/eu; a manual second restart of the four spoke nodes healed the mesh to the expected 4/4/2/2/2 shape within seconds.

rjwalters:
## Problem

The reconnect tick in `botho/src/commands/run.rs` only re-dialed the configured bootstrap peers when `discovery.peer_count() == 0`. A node that reached exactly one peer never retried its remaining bootstrap peers, so a partially formed mesh could never heal.

During the 2026-07-07 rolling v0.3.1 deploy this manifested as a star topology: some bootstrap dials failed because their targets were mid-restart, leaving seed2 as a 4-peer hub and every other node pinned at exactly 1 peer. SCP could not externalize (slot 216 stalled, `slotStalled: true`, pending txs unminable) for 10+ minutes until the four spoke nodes were manually restarted.

## Fix

Bootstrap entries are bare multiaddrs with no peer IDs, so they cannot be matched one-to-one against the connected peer table. Instead the tick now re-dials the **whole bootstrap list whenever `peer_count() < bootstrap_peer_count`**:

- libp2p treats dials to already-connected addresses as cheap no-ops, so the redundant dials are harmless.
- In steady state (`peer_count >= bootstrap_peer_count`) no dials are issued, so the existing 5s `reconnect_interval` does not spam the network.
- Dial attempts/failures remain at `debug!` level, matching the previous behavior.

The decision is extracted into a pure helper, `should_redial_bootstrap_peers(peer_count, bootstrap_peer_count)`, following the pattern of `should_pause_for_balance` / `should_resume_from_balance_pause`.

## Tests

New unit tests in run.rs's test module:
- `redials_when_under_connected` — regression for the star-topology incident: 1 peer with 2 configured bootstrap peers must keep re-dialing.
- `does_not_redial_at_steady_state` — no dials once fully connected, so the tick goes quiet.
- `does_not_redial_with_no_bootstrap_peers` — empty bootstrap list never dials (preserves the old `!is_empty()` guard).

`cargo build -p botho` and `cargo test -p botho --lib` pass (1132 passed, 0 failed). Formatted with nightly rustfmt.

Fixes #690

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -117,6 +117,24 @@ fn should_resume_from_balance_pause(balance: u64, low_threshold: u64, mempool_le
     balance < low_threshold || mempool_len > 0
 }
 
+/// Decide whether the reconnect tick should re-dial the configured bootstrap
+/// peers.
+///
+/// Bootstrap entries are bare multiaddrs with no peer IDs, so they cannot be
+/// matched one-to-one against the connected peer table. Instead, "connected to
+/// fewer peers than there are configured bootstrap peers" is treated as
+/// under-connected and the whole bootstrap list is re-dialed: libp2p treats a
+/// dial to an already-connected address as a cheap no-op, so the redundant
+/// dials are harmless. A stricter `peer_count == 0` guard let a partially
+/// formed mesh persist indefinitely — during the 2026-07-07 rolling deploy
+/// every node latched onto a single hub (star topology) and SCP could not
+/// externalize until the nodes were manually restarted (issue #690). In steady
+/// state (`peer_count >= bootstrap_peer_count`) no dials are issued, so the 5s
+/// tick does not spam the network.
+fn should_redial_bootstrap_peers(peer_count: usize, bootstrap_peer_count: usize) -> bool {
+    peer_count < bootstrap_peer_count
+}
+
 /// Check if minting should be enabled based on quorum config and connected
 /// peers
 fn check_minting_eligibility(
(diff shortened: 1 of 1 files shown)


rjwalters:
## ✅ Approved (Judge review)

- The `peer_count < bootstrap_peer_count` heuristic is the right call given bare multiaddrs (no peer IDs to match against the connected set). It exactly covers the 2026-07-07 incident (spokes at 1 peer / 2 bootstraps), goes quiet at steady state so the 5s tick doesn't spam, and re-dialing connected addresses is a libp2p no-op. The one theoretical gap — `peer_count >= bootstrap_count` while both bootstraps are disconnected — implies the node reached other peers via discovery, i.e. it is not partitioned; acceptable.
- Sustained re-dialing of a permanently-dead bootstrap address every 5s is bounded noise at debug level; fine for the fleet sizes in play.
- Tests follow the established pure-helper pattern and pin both the regression and the steady-state guard. Verified `cargo test -p botho --lib` green (1132 passed).
- Deploy note: this lands in the next node release; the fleet is currently healthy so no urgent redeploy — but the next rolling deploy is exactly when this protects us.

Issue #610, opened 2026-07-02
botho-wallet CLI is wire-incompatible with current node: RPC field-case drift + legacy tx format


rjwalters:
## Summary
Found live during the #606 testnet reset smoke test (2026-07-02): the CLI thin wallet (`botho-wallet`) cannot talk to a current node **at all**. Two independent drifts:

### 1. RPC response deserialization drift (field casing)
The node emits camelCase JSON (`chainHeight`, `tipHash`, `peerCount`, `uptimeSeconds`, `txHash`, `outputIndex`, `minimumFee`, …) but several wallet structs in `botho-wallet/src/rpc_pool.rs` still expect snake_case:
- `NodeStatus` (no rename — connect() probe fails ⇒ **"Failed to connect to any nodes"** for every command)
- `ChainInfo` (no rename; also `total_mined` is emitted as a **string** by the node, struct declares `u64`)
- `TxOutput`: `tx_hash`/`output_index` lack renames (other fields have piecemeal `#[serde(rename)]`)
- `SubmitTxResult` (`tx_hash` vs `txHash`)
- `FeeEstimate` (all three fields)

Newer structs (`FaucetRequestResult`, `NetworkFeeRate`) already carry `#[serde(rename_all = "camelCase")]` — the older ones drifted. Verified live: with the renames patched, the wallet connects and syncs fine against the 3.0.0 testnet (found the smoke wallet's faucet UTXO at height 204).

### 2. Legacy transaction wire format (the bigger problem)
`botho-wallet/src/transaction.rs` builds a legacy `botho-tx-v1` flat transaction (plaintext recipient view/spend keys, no ring signature — ~350 bytes). The node's `tx_submit` bincode-deserializes `botho::transaction::Transaction` (CLSAG ring format) and rejects the wallet's txs with `Invalid transaction: io error: unexpected end of file`.

The wallet-side `TransactionBuilder` needs to be ported to the real CLSAG format (ring member selection via `chain_getOutputs`, decoy selection per #314/#596 policy, key images, cluster tags). Note the node-side `Wallet::create_private_transaction` (botho/src/wallet.rs) is the reference implementation.

## Impact
- Every `botho-wallet` command that touches the network is broken (send, sync, balance, nodes).
- Nobody can transact on the testnet from the CLI wallet; only the faucet's node-side sender works.
- The web wallet / mobile app have separate stacks and are unaffected.

## Suggested split
[...]

PR #700, opened 2026-07-07
feat(rpc): enrich explorer-facing block + cluster RPCs (tx structure, fees, lottery, factors)

### LINKED ISSUE #696: rpc: enrich explorer-facing block + cluster RPCs (tx hashes, fees, lottery summary, cluster factors)
## Context

Explorer/stats phase (NEXT_SESSION handoff P2). The existing RPC responses are too thin for the explorer's M2 views, verified against the live node 2026-07-07:

- `getBlockByHeight`/`getBlockByHash` return only header fields + `txCount` — no tx hashes, no fees, no lottery data. Block detail pages can't show anything per-tx, and lottery events are invisible.
- `cluster_getAllWealth` returns `{cluster_id, wealth}` (string-u128 per #628) but not the cluster's fee factor — the explorer histogram would have to duplicate the consensus curve in TS to draw factor bands (the #610 serde/logic-drift bug class).

## Scope

1. `getBlockByHeight` / `getBlockByHash`: add
   - `transactions: [{hash, fee, ringSize?}]` (privacy-safe: hashes/fees/structure only — never amounts/recipients),
   - `totalFees` (block.total_fees()),
   - `lottery: {totalFees, poolDistributed, feeBurn, payoutCount, payoutTotal}` from `BlockLotterySummary` (whatever fields it actually carries — verify against block.rs, do not invent),
   - keep all existing fields unchanged (additive only — the wallet adapter and status page consume this shape).
2. `cluster_getAllWealth`: add `factor` (the same u64 milli-x the node's curve returns, e.g. 1000..6000) per cluster, computed via the live `FeeConfig`/curve — single source of truth stays in Rust.
3. Update any RPC-shape tests; add coverage for the new fields.

## Privacy caveat (from the handoff)

Explorer shows STRUCTURE (hashes, ring sizes, fees, lottery events, wealth aggregates) — never balances/flows/linkage. Lottery payout visibility = display what's on-chain, no linkage tooling.

## Acceptance

- [ ] Live node returns the enriched shapes (verify with curl against a deployed or locally-run node, not just unit tests)
- [ ] Additive-only: existing consumers (web adapters, status page) unaffected
- [ ] New fields covered by rpc tests

rjwalters:
## Summary

Fixes #696.

Additive-only enrichment of the explorer-facing RPCs in `botho/src/rpc/mod.rs`. All pre-existing fields are byte-for-byte unchanged — the web wallet adapter and seed status page consume these shapes and are unaffected.

### `getBlockByHeight` / `getBlockByHash`

Both handlers now share one serializer (`block_to_json`). New fields on top of the existing header shape:

```json
{
  "height": 42,
  "hash": "9f3a…",
  "prevHash": "77b1…",
  "timestamp": 1751840000,
  "difficulty": 12345,
  "nonce": 987,
  "txCount": 1,
  "mintingReward": 4800000000000,

  "transactions": [
    { "hash": "c0de…", "fee": 250, "ringSize": 20 }
  ],
  "totalFees": 250,
  "lottery": {
    "totalFees": 250,
    "poolDistributed": 200,
    "amountBurned": 50,
    "lotterySeed": "0909…",
    "payoutCount": 2,
    "payoutTotal": 100
  }
}
```

- `transactions`: privacy-safe structure only — tx hash (hex of `tx.hash()`), `fee`, and `ringSize` (ring-member count of the first CLSAG input; all inputs share the fixed ring size). Never amounts, recipients, or linkage data.
- `totalFees`: `Block::total_fees()` (the saturating helper, block.rs:616).
- `lottery`: the real `BlockLotterySummary` fields (`total_fees`, `pool_distributed`, `amount_burned`, `lottery_seed` — verified against block.rs, camelCased like the rest of the RPC surface) plus `payoutCount` / `payoutTotal` from the block's on-chain `lottery_outputs`. Blocks without lottery data render explicit zero defaults, never omit the object.

### `cluster_getAllWealth`

```json
{
  "count": 2,
  "total_tracked_wealth": "340282366920938463463374607431768211455",
  "clusters": [
    { "cluster_id": "1", "wealth": "1", "factor": 1000 },
    { "cluster_id": "2", "wealth": "340282366920938463463374607431768211454", "factor": 6000 }
  ]
}
```

- `factor`: the u64 milli-x multiplier (1000 = 1x .. 6000 = 6x) from the live fee curve, computed via the same `Mempool::cluster_factor(u128)` that `tx_estimateFee` / `cluster_getWealthByTargetKeys` use — one source of curve truth in Rust, so the explorer histogram never re-implements the consensus curve in TS (#610 drift class). Wealth stays a string-encoded u128 (#628/#626), parsed from the same ledger source the handler already used.

## Tests

- `botho/src/rpc/mod.rs` unit tests:
  - `test_block_to_json_explorer_fields` — crafted block with an 11-member ring tx + lottery result; asserts every original field unchanged, per-tx hash/fee/ringSize, `totalFees`, full lottery object, and that no amount/recipient fields leak.
  - `test_get_block_by_height_additive_explorer_fields` — handler end-to-end on genesis: new fields present as empty/zero defaults.
  - `test_cluster_get_all_wealth_includes_live_curve_factor` — seeds poor + ultra-rich clusters; `factor` matches the live `cluster_factor`, floors at 1000, stays within [1000, 6000], and is progressive.
- `botho/tests/rpc_integration.rs` — new `test_get_block_by_height_explorer_fields` over real HTTP against a spawned RPC server, plus enriched-field assertions on the `getBlockByHash` known-hash path.

## Verification

- `cargo build -p botho` — clean
- `cargo test -p botho --lib rpc` — 94 passed
- `cargo test -p botho --test rpc_integration` — 40 passed (real HTTP round-trips satisfy the "locally-run node" check; a curl against a deployed testnet node after merge/deploy is the remaining live-acceptance step)
- `rustup run nightly cargo fmt -p botho`

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 651
-# Branch: feature/issue-651
+# Issue: 696
+# Branch: feature/issue-696
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/botho/src/rpc/mod.rs
+++ b/botho/src/rpc/mod.rs
@@ -1276,24 +1276,66 @@ async fn handle_supply_info(id: Value, state: &RpcState) -> JsonRpcResponse {
     )
 }
 
+/// Serialize a block into the explorer-facing RPC shape shared by
+/// `getBlockByHeight` and `getBlockByHash`.
+///
+/// #696: on top of the original header fields (which existing consumers — the
+/// web wallet adapter and the seed status page — depend on and which must not
+/// change), this adds privacy-safe per-transaction structure (hash, fee, ring
+/// size — never amounts/recipients/linkage), the block's total fees, and the
+/// lottery summary so explorers can render block detail and lottery events
+/// without re-deriving consensus data client-side.
+fn block_to_json(block: &crate::block::Block) -> Value {
+    // Per-transfer-tx structure. `ringSize` is the number of ring members in
+    // the transaction's first input; every input uses the same fixed ring
+    // size, and a well-formed transfer tx always has at least one input
+    // (0 only for degenerate/test blocks).
+    let transactions: Vec<Value> = block
+        .transactions
+        .iter()
+        .map(|tx| {
+            json!({
+                "hash": hex::encode(tx.hash()),
+                "fee": tx.fee,
+                "ringSize": tx.inputs.clsag().first().map(|input| input.ring.len()).unwrap_or(0),
+            })
+        })
+        .collect();
+
+    // Lottery summary: the real `BlockLotterySummary` fields (block.rs) in
+    // camelCase, plus the on-chain payout structure carried alongside it.
+    let summary = &block.lottery_summary;
+    let lottery = json!({
+        "totalFees": summary.total_fees,
+        "poolDistributed": summary.pool_distributed,
+        "amountBurned": summary.amount_burned,
+        "lotterySeed": hex::encode(summary.lottery_seed),
+        "payoutCount": block.lottery_outputs.len(),
+        "payoutTotal": block.total_lottery_payouts(),
+    });
+
+    json!({
+        "height": block.height(),
+        "hash": hex::encode(block.hash()),
+        "prevHash": hex::encode(block.header.prev_block_hash),
+        "timestamp": block.header.timestamp,
+        "difficulty": block.header.difficulty,
+        "nonce": block.header.nonce,
+        "txCount": block.transactions.len(),
+        "mintingReward": block.minting_tx.reward,
+        // #696 additive explorer fields below this line.
+        "transactions": transactions,
+        "totalFees": block.total_fees(),
+        "lottery": lottery,
+    })
+}
+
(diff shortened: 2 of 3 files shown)


rjwalters:
## ✅ Approved (Judge review)

- **Additive-only verified**: the shared `block_to_json` keeps every original field byte-identical; existing consumers (wallet adapter `getBlock`, seed status page) are unaffected. Good call unifying the two handlers — they had already drifted into copy-paste.
- **Field names checked against the real structs**: the agent correctly rejected the issue's guessed `feeBurn` in favor of the actual `amount_burned`, and `payoutCount`/`payoutTotal` come from the on-chain `lottery_outputs`. This is exactly the verify-don't-invent bar the issue set.
- **Privacy posture holds**: per-tx exposure is hash/fee/ringSize only — no amounts, recipients, or linkage material.
- **Curve single-sourcing**: `factor` reuses `Mempool::cluster_factor` — the same path `tx_estimateFee` uses — so TS never re-implements the curve (#610 class). Lock order (ledger → mempool) matches every other handler that takes both; no inversion anywhere in the file.
- Tests: 3 unit + 1 integration over real HTTP; 94 lib rpc + 40 integration pass.

Post-merge acceptance item stays open on #696: curl the enriched shapes on a live testnet node after the next node deploy.

Issue #698, opened 2026-07-07
web-wallet: /network fleet dashboard — live grid over all 5 ingress nodes + history charts


rjwalters:
## Context

Explorer/stats phase (handoff P3), placement ratified 2026-07-07: the network stats dashboard is a new **/network page in the wallet PWA**. All 5 nodes now have TLS+CORS (#636), so the live view polls the fleet directly from the browser; history comes from the metrics-daemon API (#697 — see its API contract).

## Scope

1. **Live fleet grid**: one card per `INGRESS_NODES` entry + region label: chain height, peer count, SCP peers, mempool size, minting state, node version, slot-stall flag (`slotStalled`) — all from `node_getStatus`, polled every ~15s. Degraded states: unreachable node renders an explicit error card (never fabricated values — #541 lesson); height lag vs fleet max highlighted.
2. **Fleet summary strip**: consensus height (max), nodes-in-sync count, total mempool, blocks/hour derived from recent block timestamps (client-side via getBlockByHeight walk or timestamps already fetched).
3. **History charts** (from `https://faucet.botho.io/metrics-api/...`, #697 contract): per-node height over time (staleness visible), mempool depth, peer counts. Graceful empty-state while the backend has little data; feature-degrade cleanly if the API is unreachable.
4. Route + nav entry alongside /explorer; same header/NetworkSelector chrome as explorer.tsx.

## Notes

- This page is also the read-only foundation for the #695 P4 admin dashboard — keep components composable (a `features/network` package area mirroring `features/explorer`).
- Reuse `fetchNodeHealth` / networks.ts config; don't duplicate endpoints.
- Tests: component tests with mocked fetch for grid rendering incl. unreachable/stale states.

rjwalters:
Shipped and live at wallet.botho.io/network (PR #701, verified 5/5 nodes rendering with history charts). The operator dashboard's fleet tab (#706/#711) now reuses the same components.

Issue #697, opened 2026-07-07
metrics-daemon: multi-node fleet collection + public history API (network stats backend)


rjwalters:
## Context

Explorer/stats phase (handoff P3). `infra/faucet/metrics-daemon` exists (SQLite, 5min/hourly/daily rollups, axum API with CORS at /api/metrics/{latest,history}) but only polls ONE node and **has never been deployed** (no systemd unit on the faucet box, verified 2026-07-07). The network dashboard needs fleet-wide history.

## Scope

1. **Multi-node collection**: poll a configured list of nodes (name + url; the 5 testnet nodes — us seeds via https, or all via their internal/public endpoints from the faucet box). Add a `node` dimension to the tables (testnet SQLite is disposable — recreate schema, no migration needed).
2. **Fleet API** (keep CORS):
   - `GET /api/metrics/latest` → array of per-node latest samples `{node, timestamp, height, peerCount, scpPeerCount, mempoolSize, mintingActive, uptimeSeconds}`
   - `GET /api/metrics/history?node=<name>&resolution=5min|hourly|daily&since=<unix>` → samples for charts
   - `GET /health`
3. **Height-staleness signal**: expose per-node `heightStale: bool` (no height change across N samples) in /latest — the deferred #613 chain-height alerting's data source.
4. **Deploy** (operator steps scripted in infra/faucet/): build on the faucet box (small crate, fine on-host), systemd unit (`botho-metrics.service`), nginx route on faucet.botho.io (e.g. `/metrics-api/` → localhost API port, GET-only, CORS for wallet.botho.io) mirroring the existing /rpc proxy pattern.

## API contract note

The wallet dashboard (separate issue) builds against the /latest and /history shapes above — treat them as the contract; if implementation needs to deviate, update this issue so the frontend follows.

## Acceptance

- [ ] Daemon polls all 5 nodes; SQLite has per-node rows with rollups
- [ ] API serves fleet latest + per-node history publicly via https://faucet.botho.io
- [ ] systemd unit + deploy steps committed under infra/faucet/
- [ ] cargo test for collector/db changes

rjwalters:
## ✅ Deployed and verified live (2026-07-07)

Operator steps completed on the faucet box:
- Built `botho-metrics-daemon` release on-host from main @ acf12b1 (3m06s), installed to /usr/local/bin.
[...]

Issue #695, opened 2026-07-07
design: P4 operator trust/admin dashboard — full read-write web surface (decision #664)


rjwalters:
## Context

Decision ratified on #664 (2026-07-07): P4 of epic #441 ships as a **full read-write web dashboard** — observability AND live quorum curation in one operator surface. This issue asks the architect to turn that decision into concrete implementation issue(s).

## Ratified parameters (from the decision record)

1. **Web surface** on the existing Cloudflare Pages/Worker stack (consistent with wallet + BaaS control plane).
2. **Auth**: magic-link/status-token pattern (`web/packages/baas-worker/status-link.ts`) as the READ-path starting point.
3. **Quorum editing in scope, but security-gated**: the write path must be designed as its own reviewable deliverable before any write ships. Minimum bar: operator-signed actions (key held off-node), node-side signature verification against a configured operator key, logged/observable changes, documented recovery path for a bad edit (cf. #427/#532). Reads may ship first — sequencing, not scope reduction.
4. **Boundary with #651 (P5)**: the promotion gate stays the automated trust machinery; P4 renders the gate's decisions (curated/auto members, suppressed peers, intersection refusals — already in `node_getStatus`) and provides human curation on top. No duplicated gate logic.

## Asks

- [ ] Architecture proposal for the dashboard (data sources, pages, deployment) — note the overlap with the ratified network-stats dashboard phase; ideally one surface serves both.
- [ ] Separate security-design deliverable for the quorum write path per (3).
- [ ] File the P4 implementation issue(s) against #441.

Ref: decision record https://github.com/botho-project/botho/issues/664#issuecomment-4905934297

rjwalters:
# P4 architecture proposal — operator trust/admin dashboard (#441 P4, decision #664)

Grounded in the code as of `main` (9453f3ff). Every existing-surface claim below cites a file.

## 0. What exists (the foundation)

| Surface | Where | What it gives P4 |
|---|---|---|
| Fleet dashboard | `web/packages/features/src/network/` (`types.ts`, `fleet.ts`, `components/network-dashboard.tsx`) | `NetworkDashboard` is deliberately pure presentation — "polling and history fetching live in the page so this composes into other surfaces (e.g. the P4 admin dashboard, #695)" (its own docstring). `fetchFleetNodeStatus` already polls `node_getStatus` per node with the anti-#541 explicit-error contract. |
[...]

PR #701, opened 2026-07-07
feat(web-wallet): /network fleet dashboard — live 5-node grid, summary strip, history charts

rjwalters:
## Summary

Refs #698 (P3 of the explorer/stats phase; placement ratified 2026-07-07: wallet `/network` page). Two data planes:

- **Live** (works today): browser polls all five ingress nodes' `node_getStatus` every 15s (possible since #636 gave eu/ap TLS+CORS). Per-node cards show height, mempool, peers, SCP peers, version, minting badge, lag-vs-consensus badge, and the #653 `slotStalled` flag. Unreachable nodes render an explicit error card — never stale or fabricated values.
- **History** (activates when #697 deploys): charts read the metrics-daemon fleet API per its contract; while absent the charts show an informational degraded state and the live grid is unaffected — verified in the live check below.

## Design notes

- `features/network` mirrors `features/explorer`: pure presentation components + pure derivation functions, so the #695 P4 admin dashboard can compose the same pieces.
- Charts are a dependency-free multi-series SVG — one line per node doesn't justify a chart library.
- "Avg block spacing" is labeled and captioned honestly: the testnet mints on demand, so idle gaps are included (h/m formatting; `null` on degenerate windows rather than a fabricated rate).

## Verification

- 15 new tests (pure derivations + dashboard states incl. unreachable/lagging/stalled/checking/degraded-history); full web suite 587/587; typecheck clean.
- Driven live against the real testnet: 5/5 nodes in sync at h=221, all v0.3.1, history degraded gracefully (screenshot in session).

### DIFF
--- a/web/packages/features/src/index.ts
+++ b/web/packages/features/src/index.ts
@@ -1,3 +1,4 @@
 // Re-export all features
 export * from './explorer'
 export * from './wallet'
+export * from './network'
--- a/web/packages/features/src/network/components/fleet-summary.tsx
+++ b/web/packages/features/src/network/components/fleet-summary.tsx
@@ -1,0 +1,104 @@
+import { Card, CardContent } from '@botho/ui'
+import { Activity, AlertTriangle, Blocks, Layers, Timer } from 'lucide-react'
+import type { FleetSummary } from '../types'
+
+export interface FleetSummaryStripProps {
+  summary: FleetSummary
+  /** Average seconds per block over a recent window; null = unknown. */
+  avgBlockSeconds: number | null
+  className?: string
+}
+
+/** Fleet-level facts in one strip above the node grid. */
+export function FleetSummaryStrip({
+  summary,
+  avgBlockSeconds,
+  className,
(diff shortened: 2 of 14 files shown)


rjwalters:
## ✅ Approved (Judge review)

- Verified live against the real 5-node testnet before approving: 5/5 nodes rendered at consensus height with correct per-node stats, and the history charts degraded to the informational state with the live grid unaffected (the metrics backend isn't deployed yet — exactly the designed behavior).
- The no-fabrication contract is enforced structurally: `fetchFleetNodeStatus` failures resolve to `reachable: false` snapshots carrying no data fields, and the card renders an explicit error state; pinned by tests.
- `deriveFleetSummary` correctly excludes unreachable nodes from mempool/stall aggregation (tested), and the block-spacing metric returns null on degenerate windows rather than a made-up rate — with honest mint-on-demand labeling.
- Components are pure presentation with data injected from the page, per the #695 composability requirement.
- 15 new tests; full suite 587/587; typecheck clean.

Issue #613, opened 2026-07-03
infra: regional seed expansion + DNS failover (mainnet blocker 1)


rjwalters:
## Summary
Mainnet blocker 1 from PLAN.md: all three testnet nodes (seed, seed2, faucet) live in one AWS region (us-west-2). Mainnet needs ≥3 geographically diverse seeds plus DNS failover so no single region outage partitions bootstrap.

## What already exists
- Regional seed scaffolding in `botho/src/network/seeds.rs` (us/eu/ap for both networks), gated behind `BOTHO_REGIONAL_SEEDS=1` because the DNS records aren't live.
- DNS-seed discovery (`seeds.testnet.botho.io` TXT) working in production; `seeds.botho.io` reserved for mainnet.
- Runbook: `infra/seed/TESTNET_RESET.md` §4 (activation paths).

## Tasks
- [ ] Launch eu + ap seed hosts (t4g.small is sufficient for relay-only; binary distributed from a build host per the low-RAM build rule)
- [ ] Publish `PEER_ID@host:port` TXT records under `seeds.testnet.botho.io` (preferred zero-client-release path) — verify DNS discovery picks them up
- [ ] Route53/Cloudflare health-checked failover for the seed hostnames
- [ ] Update hardcoded fallback list in `seeds.rs` and flip `BOTHO_REGIONAL_SEEDS` default once records are live
- [ ] CloudWatch (or equivalent) alerting: peer count, chain height staleness per node

## Notes
Most of this is `loom:operator-only` territory (AWS/DNS credentials); the code-side pieces (seeds.rs list update, default flip) are buildable. Split when scoping.

Ref: PLAN.md "Remaining mainnet blockers" (refreshed in PR #612).

rjwalters:
## Regional seeds live — 5-node mesh across 3 continents (2026-07-04)

Deployed:
- **botho-seed-eu** — eu-central-1, `i-0d678901f254076fc`, 3.77.150.19 (t4g.small, relay-only)
- **botho-seed-ap** — ap-southeast-1, `i-063bd5136e7c3f170`, 3.0.209.59 (t4g.small, relay-only)

Both run the fleet binary (main @ `8503ade`, sha-verified against the US nodes), synced from genesis to tip within minutes, and hold identical tips with the US cluster. All five nodes now report a full mesh (peerCount 4 everywhere).

[...]

PR #702, opened 2026-07-07
feat(metrics-daemon): multi-node fleet collection + public history API

rjwalters:
Refs #697 (kept open intentionally: the operator deploy step on the faucet box remains after merge).

## What

Extends `infra/faucet/metrics-daemon` from single-node to fleet collection — the network stats backend for the wallet dashboard.

### Multi-node config
- Repeated `--node NAME=URL` args configure the fleet, e.g. `--node seed=https://seed.botho.io/rpc --node faucet=http://127.0.0.1:17101`
- Legacy `--node-url <url>` still works as a single-node fallback (stored under node name `default`), so nothing breaks
- Duplicate node names are rejected at startup

### Schema (no migration — testnet DB is disposable)
- `node TEXT NOT NULL` added to `metrics_5min`/`metrics_hourly`/`metrics_daily`, `PRIMARY KEY (node, timestamp)`
- Rollups `GROUP BY node`; per-node `last_tx_count` state keys
- Deploy note: delete any old single-node `metrics.db` before first start (documented in README)

### Collector
- All configured nodes are polled concurrently each tick (tokio `JoinSet`, shared reqwest client, 10s per-request timeout)
- A down node is logged and skipped: nothing is recorded for a failed poll (no fabricated samples), and it never blocks the other nodes

### API (axum, CORS kept) — contract for the wallet dashboard

**`GET /api/metrics/latest`** — JSON array, one entry per node:

```json
[
  {
    "node": "seed",
    "timestamp": 1751846400,
    "height": 12345,
    "peerCount": 4.0,
    "scpPeerCount": 3.0,
    "mempoolSize": 0.0,
    "mintingActive": false,
    "uptimeSeconds": 86400,
    "heightStale": false
  }
]
```

`heightStale` = the node's height unchanged across the last 3 samples (~15 min), computed from the DB; nodes with fewer than 3 samples are not flagged. This is the data source for the deferred #613 chain-height alerting.

**`GET /api/metrics/history?node=<name>&resolution=5min|hourly|daily&since=<unix-seconds>`** — JSON array of samples for that node (`node` required → 400 if missing; `resolution` defaults to `5min`, invalid → 400; `since` defaults to 0):

```json
[
  {
    "timestamp": 1751846400,
    "height": 12345,
    "peerCount": 4.0,
    "scpPeerCount": 3.0,
    "mempoolSize": 0.0,
    "txTotal": 12
  }
]
```

For `5min` resolution `txTotal` is the per-sample tx delta; for `hourly`/`daily` it is the summed tx count over the bucket. Unknown node returns an empty array.

**`GET /health`** — unchanged (`OK`).

Note: the old `/api/metrics/history?metric=&period=&granularity=` shape and the single-object `/latest` are replaced by the contract shapes above (the daemon was never deployed and a repo-wide grep found no consumers of the old shapes).

### Deploy assets (committed, NOT executed)
- `infra/faucet/botho-metrics.service` — systemd unit polling all 5 testnet nodes (`seed`, `seed2`, `faucet` local, `eu`, `ap`), DB at `/var/lib/botho-metrics/metrics.db` (`StateDirectory=botho-metrics`), API on 17102. Replaces the stale, never-deployed `botho-metrics-daemon.service`.
- `infra/faucet/faucet-nginx.conf` — new `location /metrics-api/` block: `https://faucet.botho.io/metrics-api/...` → `proxy_pass http://127.0.0.1:17102/` (prefix stripped), GET-only (`limit_except GET OPTIONS`), CORS `*`, mirrors the existing `/rpc` proxy pattern, 1-min proxy cache.
- `infra/faucet/README.md` — new "Fleet Metrics Daemon" section: API contract with sample JSON, build-on-host + systemd + nginx deploy steps.

## Tests

`cargo build -p botho-metrics-daemon` and `cargo test -p botho-metrics-daemon` pass — 13/13:
- db: per-node insert/latest, (node, timestamp) PK collision safety, height staleness (stuck/moving/young/recovered), per-node hourly+daily rollups, `since` filter + unknown node, per-node tx-count state
- collector: RPC response parsing (ok / rpc-error / missing result / garbage)
- main: `--node NAME=URL` parsing, `--node-url` fallback, duplicate-name rejection

Formatted with nightly `cargo fmt` (`--check` clean).

## Remaining operator steps (after merge, on the faucet box)

1. `cargo build --release -p botho-metrics-daemon` and install to `/usr/local/bin/`
2. Remove any old `/var/lib/botho-faucet/metrics.db` single-node DB (schema changed, disposable)
3. Install `botho-metrics.service`, `systemctl enable --now botho-metrics`
4. Update nginx config from `faucet-nginx.conf`, `nginx -t && systemctl reload nginx`
5. Verify `https://faucet.botho.io/metrics-api/api/metrics/latest` returns 5 nodes


rjwalters:
## ✅ Approved (Judge review)

- **Contract fidelity**: /latest and /history shapes match the #697 contract exactly (history adds scpPeerCount/txTotal — additive over what the dashboard reads, fine). Replacing the old history shape is safe: the daemon was never deployed, and grep confirmed zero consumers.
- **No-fabrication + isolation verified in the diff**: failed polls log-and-skip (nothing recorded), JoinSet keeps one dead node from blocking the tick, and `is_height_stale` demands a FULL window of identical heights — young nodes aren't falsely stale.
- **Schema**: `PRIMARY KEY (node, timestamp)` with INSERT OR REPLACE is idempotent per node-tick; rollups group by node. Fresh-create is right for a never-deployed testnet daemon.
- `db.lock().unwrap()` mutex-poisoning panics are acceptable for a systemd-restarted sidecar; the one `parse .unwrap()` is test-only.
- Deploy assets mirror the existing /rpc nginx pattern (GET-only, CORS, prefix strip to :17102) and the service file pins the 5 real nodes with the faucet's own node polled locally — correct, avoids hairpinning through its public TLS.
- 13/13 tests pass, build + nightly fmt clean.

Merge on green; the operator deploy on the faucet box follows (issue #697 stays open for it per the Refs).

PR #705, opened 2026-07-07
fix(infra): retry release-artifact downloads in deploy-botho.sh

rjwalters:
One-line resilience fix: the ap-southeast-1 node failed the GitHub release-artifact download transiently on BOTH the v0.3.1 and v0.3.2 rolling deploys (manual retry succeeded immediately each time). `curl --retry 3 --retry-delay 2` absorbs it so the rolling deploy loop doesn't need manual intervention.


rjwalters:
## ✅ Approved (Judge review)

One-line resilience fix matching the observed failure mode (two consecutive deploys, same region, manual retry succeeded immediately both times). `bash -n` clean; the retry flags only affect the artifact-download path, not the install/restart sequence.

Issue #709, opened 2026-07-07
feat(node+web): P4.4 operator-signed quorum curation — operator_submitAction, audit log, curation UI (blocked on #708 review) [#441 P4]


rjwalters:
Part of #441 (P4); Phase 4 of the architecture proposal on #695 (https://github.com/botho-project/botho/issues/695#issuecomment-4909810884). **BLOCKED on the write-path security design (P4.3, #708) being reviewed and approved** — do not start implementation before the review outcome is recorded on #708 (decision #664 §3: sequencing requirement).

## Scope

Implement operator-signed quorum curation end-to-end, exactly per the approved `docs/security/quorum-write-path.md`:

**Node side**
- `[rpc.operator] action_public_keys` config (`botho/src/config.rs`); absent ⇒ writes disabled (fail closed).
- `operator_submitAction` RPC (`botho/src/rpc/mod.rs`): Ed25519 envelope verification (signature → targetNode → expiry/skew → persisted-nonce replay check), `dryRun` support.
- Command channel from the RPC layer into the `commands::run` event loop (which owns `config`/`discovery`/consensus); apply the edit to the in-memory `QuorumConfig`, then run the **existing** `rebuild_scp_quorum_set` → `gated_scp_quorum_set` path (`botho/src/commands/run.rs`) — an intersection-refused candidate rejects the action and keeps the previous safe quorum set; no new gate logic.
- On success: persist via `Config::save` (`botho/src/config.rs:742`), publish the fresh `QuorumGateSnapshot`, append the audit record, return snapshot + `quorumDegenerate`/`quorumFaultTolerant` posture. `acknowledgeDegenerate: true` required for edits leaving a <4-node quorum (#509 posture).
- Append-only JSONL audit log in the data dir, surfaced by `operator_getAuditLog` (from the P4.2 issue) and mirrored to `tracing` WARN.

**Dashboard side** (`web/packages/features/src/operator/`, web-wallet `/operator/actions`)
- Operator key import/generation, encrypted at rest under a mandatory passphrase (no plaintext-by-default — the #474 lesson).
- Action composer for `quorum.pin_member` / `quorum.unpin_member` / `quorum.set_max_auto_members` with a mandatory `dryRun` preview showing the gate verdict before the real signed submission.
- Applied/refused state rendered **only** from the node's response snapshot (anti-#541: never optimistic UI); audit-log view.

## Acceptance criteria

- [ ] Security review recorded as approved on the P4.3 issue before any of this merges.
[...]

Issue #708, opened 2026-07-07
docs(security): P4.3 quorum write-path security design — operator-signed actions (reviewable deliverable, gates P4.4) [#441 P4]


rjwalters:
Part of #441 (P4); Phase 3 of the architecture proposal on #695 (https://github.com/botho-project/botho/issues/695#issuecomment-4909810884). This is the **separately-reviewable security deliverable** required by the #664 decision (§3): no quorum write path ships before this design is reviewed and approved. Documentation only — no code changes.

## Scope

Write `docs/security/quorum-write-path.md` expanding proposal §3 into a full, reviewable security design. Required sections (minimum bar from the #664 decision, plus the proposal's commitments):

1. **Operator keypair lifecycle** — Ed25519; generation on the operator's machine (`botho operator keygen`); private key never on nodes or the Worker; nodes configured with `[rpc.operator] action_public_keys` only; dashboard import encrypted at rest under a mandatory passphrase (the #474 lesson: no plaintext-by-default); rotation/revocation procedure.
2. **Signed-action envelope** — canonical JSON, action enum (v1: `quorum.pin_member`, `quorum.unpin_member`, `quorum.set_max_auto_members` — mode/threshold flips stay out of v1), `targetNode` binding, nonce, `issuedAt`/`expiresAt ≤ +300s`; exact canonicalization and signature-input spec.
3. **Node-side verification & apply** — new authenticated RPC `operator_submitAction` (NOT a config-reload path; justify per proposal §3.3); verification order (signature → target → expiry → nonce); apply via command channel into the `commands::run` event loop; the edit takes effect **only through the existing gate** (`rebuild_scp_quorum_set` / `gated_scp_quorum_set` incl. the `bth-quorum-sim` intersection check — refused candidate ⇒ rejected action, previous safe set kept); persistence via `Config::save` (`botho/src/config.rs:742`); `dryRun` mode.
4. **Replay protection** — persisted nonce store, expiry, target binding; explicit statement of what an attacker holding a captured envelope can and cannot do.
5. **Audit logging** — append-only JSONL schema (timestamp, signer key, envelope hash, prev→new `[network.quorum]`, gate verdict, outcome), `operator_getAuditLog` surfacing, WARN-level tracing mirror; dashboard renders outcomes only from node responses (anti-#541).
[...]

Issue #664, opened 2026-07-06
decide: admin-API shape for P4 operator trust/admin dashboard (unblocks #441 P4)


rjwalters:
## Context

Epic #441 (Botho product architecture) decomposes into phases P0–P7. P5 (quorum promotion gate) was filed as #651 once its blocker #510 closed. **P4 — the operator admin / trust dashboard — has never been filed**, because per the Champion review on #441 (2026-07-01) it is blocked on a *maintainer decision about the admin-API shape*, not on prior code. It currently lives only in the epic thread and risks being lost — this issue captures the open decision so the architect can file the P4 implementation issue(s) once it's answered. (Parallels how #649 tracks the nanoBTH unit decision.)

## The decision to make

P4 is the surface by which an operator curates trust and (potentially) edits the live quorum. Before it can be scoped:

1. **Auth model** — how does an operator authenticate to the admin surface? (Reuse the BaaS magic-link/status-token pattern from `web/packages/baas-worker/status-link.ts`? A separate operator credential? Node-local only vs remote?)
2. **Web vs native** — is the dashboard a web surface (Cloudflare Pages/Worker, like the wallet + BaaS control plane) or a node-local admin CLI/TUI? Hybrid?
3. **Who may edit live `[network.quorum]`** — is quorum-set editing in-scope for P4 at all, or read-only observability first with quorum edits gated behind a separate, more careful change? What is the write path (RPC? config reload? signed operator action?) and its consensus-safety story (a bad quorum edit is a liveness/safety event — see #427/#532 cluster-health work and #651's promotion gate).
4. **Relationship to #651 (P5 promotion gate)** — P4 is the human-facing side of the same trust machinery P5 automates; define the boundary so they compose rather than overlap.

## Notes

- Ephemeral testnet, so no migration/backup concerns for the surface itself; but quorum-edit safety is a real consensus concern regardless of testnet/mainnet.
- Once decided, route to the architect to file the concrete P4 implementation issue(s) referencing "#441 (P4)".

## Acceptance

- A recorded decision on (1)–(4) above.
- The architect files P4 implementation issue(s) against #441, or P4 is explicitly deferred with a reason.

rjwalters:
## ✅ DECISION (maintainer, 2026-07-07): full read-write web dashboard

P4 ships as a **web dashboard with both observability and live quorum editing** — one operator surface, not read-only-first. Answers to the four sub-questions:

[...]

Issue #674, opened 2026-07-07
node: tx_submit never relays transactions to peers — sends/claim links stall forever when submitted via a non-minting ingress (web-wallet default)


rjwalters:
## Summary

Transactions submitted over RPC (`tx_submit`) are added to the receiving node's **local mempool only** and are **never relayed to peers**. Mempool contents are only broadcast from inside the active-minting code path. On the current testnet — where minting is balance-paused everywhere and only the faucet node mints on demand (#386's mint-on-nonempty fix) — this means **any transaction submitted through a non-minting ingress (seed / seed2, the web-wallet defaults) never reaches the minting node and never confirms**. The user-facing send, claim-link, and pay flows are all silently broken unless the user happens to have selected the faucet ingress.

## Code

- `botho/src/rpc/mod.rs:1625` — `handle_submit_tx`: `mempool.add_tx(tx, &ledger)` → return. No gossip publish, no consensus submission.
- `botho/src/commands/run.rs` (~1770, inside the freshly-minted-tx branch): the only steady-state place `NetworkDiscovery::broadcast_transaction` + `consensus.submit_transaction` run for mempool txs — i.e. **only while this node is actively minting**.
- (`run.rs:660` also broadcasts a pending-tx file at startup — not a steady-state relay.)

So a node that is not minting is a black hole for RPC-submitted transactions.

## Observed on the live testnet (2026-07-07, protocol 4.0.0)

Timeline (all heights/mempools from `node_getStatus` on seed / seed2 / faucet):

1. ~03:00Z — claim-link funding tx submitted via **seed** ingress (web-wallet default). Result: `seed mempool=1`, `seed2 mempool=0`, `faucet mempool=0`. The tx exists on seed only.
2. Faucet dispense (different wallet) triggers a mint burst → block 207 mined by the faucet node. The stuck tx is **not** included (it isn't in the faucet's mempool). `seed mempool=1` persists.
3. Control experiment: identical claim-link funding tx submitted via the **faucet** ingress → within seconds all three mempools held it (`seed=+1, seed2=1, faucet=1` — the actively-minting faucet node relays), and it mined in block 210. The claim-link E2E completed.
4. ~2 hours after step 1 the seed-submitted tx was still sitting in seed's mempool, unconfirmed.

[...]

PR #710, opened 2026-07-07
docs(security): P4.3 quorum write-path security design (review gate for #709)

rjwalters:
## Summary

Refs #708 (P4.3 of the #695 architecture; the #664 decision's separately-reviewable deliverable). Documentation only — zero source changes.

- **`docs/security/quorum-write-path.md`** — the full write-path design across the seven required sections. The two load-bearing properties: (1) the operator's Ed25519 key never exists on any deployed component, so compromise of node/host/worker yields no signing capability; (2) every write mutates only the promotion gate's *inputs* and is re-validated by the existing `gated_scp_quorum_set` + FBAS intersection check before application — fork-capable configurations are structurally refused, leaving recoverable liveness stalls as the residual failure class, with a three-step recovery ladder.
- **`docs/security/threat-model.md`** — Attack Surface §1a (operator action surface, design-stage) + revision note.

## Review ask

This PR **is** the security review gate: #709 (implementation) stays blocked until the review outcome is recorded on #708. §9 of the doc is a reviewer checklist. Particular scrutiny welcome on: envelope canonicalization, the reserve-then-apply nonce semantics, the v1 action-set boundary, and whether §8.3's residual risk (compromised Pages host prompting the operator to sign) is acceptable for testnet.

### DIFF
--- a/docs/security/quorum-write-path.md
+++ b/docs/security/quorum-write-path.md
@@ -1,0 +1,333 @@
+# Quorum Write-Path Security Design (P4.3)
+
+**Status**: DRAFT — awaiting security review (issue #708).
+**Scope**: the operator-signed quorum-curation write path for the P4 admin
+dashboard (#441 P4; architecture proposal on #695; decision record on #664).
+**Exit criterion**: implementation issue **#709 is blocked until this design
+is security-reviewed and the review outcome is recorded on issue #708.**
+
+Every mechanism claim below is grounded in code on `main` as of this writing;
+citations are file:line or function names. Elements that do NOT yet exist are
+marked **(new, #709)**.
+
+---
+
+## 1. Why this document exists
+
+A quorum edit is the most dangerous operator action Botho exposes. The quorum
+set determines both **safety** (which coalitions can fork the chain — FBAS
+splitting sets) and **liveness** (which crash sets halt it — blocking sets);
+see `consensus/quorum-sim/README.md` for the verified framing and the
+canonical counterexample (a `2-of-4` symmetric set admits the disjoint quorums
+`{A,B}` / `{C,D}`, which can externalize conflicting values — an unrecoverable
+fork). The #664 decision ratified a remotely-reachable write surface for
+exactly this configuration, on the condition that the write path is designed
+and reviewed as its own deliverable before any of it ships. This is that
+design.
+
+Two structural mitigations do the heavy lifting; everything else is
+defense-in-depth around them:
+
+1. **Every edit flows through the existing promotion gate.** The gate
+   (`gated_scp_quorum_set`, `botho/src/commands/run.rs:2214`) already refuses
+   any candidate quorum set whose FBAS model admits disjoint quorums
+   (`symmetric_quorum_has_intersection`, applied at run.rs:2346, keeping the
+   previous safe set on refusal). The write path adds **no second way to
+   construct a quorum set** — an operator action only mutates the gate's
+   *inputs* (`[network.quorum] members`, `max_auto_members`), and the gate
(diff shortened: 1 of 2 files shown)


rjwalters:
## Security review (self-review flagged for a second human/Judge pass)

I authored this design, so treating the review critically rather than approving my own work. Verified the load-bearing claims against code, and I'm recording two substantive gaps the reviewer should weigh before this gates #709.

### Verified against code ✅
- `symmetric_quorum_has_intersection(4, 2)` returns **false** (run.rs:2412; test at :3098) — the 2-of-4 fork case IS refused by the gate, so the design's core "splitting-set edits are structurally refused" claim is real, not aspirational. `symmetric_quorum_has_intersection(4,3)` true, `(2,1)` false — matches the doc's framing.
- The gate keeps the previous set on refusal (run.rs:2346 + the `previous` branch) — the "no unsafe state is ever installed" property holds.
- `Config::save` (config.rs:742) and the RPC→mpsc→event-loop seam (the #674 `tx_relay` pattern) exist as cited. No operator surface exists yet — correctly marked (new, #709) throughout.

### Gaps the reviewer should resolve before approving

1. **§8.3 residual risk may be under-weighted even for testnet.** A compromised Pages bundle that prompts the operator to sign attacker-chosen envelopes is bounded to *liveness harassment* ONLY because the v1 action set excludes mode/threshold — that boundary is the entire mitigation. The doc should make the mode/threshold exclusion a HARD verifier-level invariant (reject unknown actions is already there; add: the verifier must never gain a mode/threshold action without a corresponding threat re-review), not just a v1 scoping choice that a later PR quietly widens. Recommend the reviewer require that as an explicit exit condition on #709.

2. **Divergent-config liveness interaction (§7.3 vs §7.2) is softer than stated.** Per-node targeting + `explicit`-mode members counting toward threshold whether connected (run.rs:2251) means a fleet where nodes disagree on membership can have each node's set individually pass intersection yet the FLEET fail to form a quorum any node accepts. The doc treats divergence as "untidiness"; under explicit mode it can be a fleet-wide liveness stall that no single compensating action fixes. Recommend: v1 restricts pin/unpin to `recommended` mode (where curated members only count when connected), or the recovery ladder explicitly covers the fleet-divergence-under-explicit case.

Neither is a fork risk — the gate holds in all cases — so these are liveness-hardening asks, not safety blockers. But since this doc IS the gate for #709, they should be resolved in the doc (or consciously accepted) rather than discovered during implementation.

Leaving this PR as `loom:review-requested` for an independent pass rather than self-approving, given it is a consensus-safety design gate.

rjwalters:
## Both review gaps resolved (655be69)

Rather than defer to the reviewer, folded both fixes into the doc since I identified them concretely:

1. **v1 action-set is now a verifier-level invariant.** The mode/threshold exclusion is stated as a hard fail-closed verifier rule with an explicit future-security-review gate — adding a mode/threshold action can't slip in as an incidental feature PR. Recorded as an exit condition on #709.

2. **v1 pin/unpin restricted to `recommended`-mode nodes.** This closes the explicit-mode fleet-divergence liveness hole: under explicit mode configured members count toward threshold whether connected (run.rs:2251), so divergence can strand the fleet; recommended mode counts curated members only when connected (run.rs:2305–2320 admits from `connected_peers`), keeping the same divergence recoverable. Explicit-mode curation joins the future-review gate.

Both are liveness-hardening (no fork risk existed in either — the intersection gate holds regardless), so the doc's safety story is unchanged and now its liveness story is tighter. Flipping to review-requested for an independent pass; the design is self-consistent and every citation re-verified against `main`.

Issue #616, opened 2026-07-03
operator: engage external security audit firm (mainnet blocker 2)


rjwalters:
## Summary
Mainnet blocker 2 from PLAN.md: internal audit cycles are clean (cycle 6 complete, all Criticals/Highs closed and deployed on the 3.0.0 testnet), and the audit-prep prerequisites from #50 (architecture docs #117, threat model #118, crypto coverage baseline #115) are done. The remaining step is the actual external engagement — an operator/business action.

## Operator tasks (not automatable)
- [ ] Select firm(s) — #50 suggested Trail of Bits, NCC Group, Cure53
- [ ] Budget allocation (2amlogic)
- [ ] Agree scope: crypto (CLSAG/LION/Pedersen), consensus (SCP + cluster-tilted redistribution + H1/H2 fee-floor/tag-inheritance hardening), tx validation, network, wallet stacks
- [ ] Freeze commit for the audit period (coordinate with #615 release tagging — auditing a tagged reproducible release is strongly preferable)
- [ ] Test environment + point of contact

## Automatable prep (file separately if wanted)
- Refresh the #118 threat model for post-cycle-6 reality (the economic mechanism and fee floor postdate it)
- Audit-scope document draft from audits/2026-06-11-cycle6.md + docs/design/*

Ref: PLAN.md mainnet blocker 2 (refreshed in PR #612); prerequisites tracker #50 (closed).

rjwalters:
Audit-prep item done: the threat model (docs/security/threat-model.md) is refreshed for post-cycle-6 reality — PR #625 (merged 2026-07-04). Every behavioral claim was code-verified during review (C1–C7 add_block gates, fee floor, protocol disconnect, decoy age-band), open residuals (#581, payout privacy, everyone-parks) are honestly labeled OPEN, and mitigations cite their PRs. This plus audits/2026-06-11-cycle6.md forms the auditor-facing package; remaining steps on this issue are the operator engagement tasks.

rjwalters:
## ✅ DECISION (maintainer, 2026-07-07): wait until consensus work settles before engaging

The external engagement waits until the audit surface stops moving. Concretely, "settled" means:
1. H4 (lottery candidate-cap grindability, carried since cycle 6) resolved or explicitly accepted,
2. #581 (cluster-tag inheritance bound) resolved,
3. the #532 consensus-liveness (view-change) epic decided/landed,
4. the RandomX / mining-economics question (#441) decided, since it determines PoW audit scope.

[...]

PR #713, opened 2026-07-08
security(consensus): tighten cluster-tag inheritance bound to per-ring maximum (#581)

### LINKED ISSUE #581: Tighten cluster-tag inheritance bound: ring-member-sum input set permits decoy-sourced inflation

rjwalters:
## Summary

Closes #581 (the last open code-level security finding; an audit blocker via #616). The C6 cluster-tag inflation guard bounded each cluster's input mass by the **sum over all resolved ring members**. Since the real input is hidden among decoys, an attacker sourcing real cluster-X-tagged UTXOs as decoys could raise cluster-X's ceiling to ~`ring_size ×` the decoy mass and attribute far more cluster-X mass to their output than their single real input supplied.

## Fix: per-ring maximum

Bound each cluster by `Σ over rings ( max over that ring's members of member.mass(cluster) )` instead of the sum over all members.

- **Sound / liveness-safe**: the real input is one member of its ring, so its cluster mass ≤ the ring's maximum ⇒ sum-of-per-ring-maxima ≥ true input mass ⇒ never false-rejects a valid block.
- **Tightest possible** node-agnostic bound: a ring's contribution can't go below its max without risking a false-reject when the real input *is* the max member.
- **Removes the ring-size multiplier**: `R` cluster-X decoys of mass `M` now raise the ceiling to `M`, not `R·M`.

### A note on the issue's "minimum" suggestion
The issue floated "minimum over ring members." That is **unsound** — the real input can be the maximum member, so a min bound would false-reject a valid block and **halt consensus**. The per-ring **maximum** is the sound analog; that's what's implemented.

### Residual (documented in code + issue)
A single decoy whose cluster mass exceeds the real input's still raises that ring's ceiling by one UTXO's mass — but no longer ring-size-multiplied. Fully closing it needs a consensus-visible binding to the real input (Phase-2 Pedersen-committed tags + ZK inheritance proofs).

## Determinism
Unchanged: pure function of block + committed UTXO state, `BTreeMap` iteration, `u128` integer math — proposer and validators reach an identical verdict. No fork risk.

## Tests
- Existing unit + ledger tests updated to the per-ring signature.
- The ledger test's two-cluster-1-decoy ring is exactly the exploit shape: it now **rejects a 2,000,000 output from a 1,000,000 real input** (the old sum bound allowed it) — a direct regression assertion.
- New pure-function test: (a) ring-size multiplier gone (3 decoys → ceiling stays at the max, not 3×), (b) **multi-real-input merge is NOT false-rejected** (liveness — three single-member rings sum to 3,000,000), (c) background-input + one-decoy residual bounded to the decoy mass.
- botho lib ledger suite **55/55**, build clean, nightly fmt applied.

### DIFF
--- a/botho/src/ledger/store.rs
+++ b/botho/src/ledger/store.rs
@@ -134,15 +134,27 @@ fn check_cluster_tag_inheritance(
     // member mass. The real input is one member of its ring, so its cluster
     // mass is at most the ring's maximum for that cluster — a sound upper
     // bound that does not carry the old `ring_size` inflation multiplier.
+    //
+    // The maximum is taken over MEMBERS, and a single member's cluster mass is
+    // the SUM of that member's entries for the cluster. Computing the
+    // per-member mass first (rather than folding the max over raw entries)
+    // keeps the bound correct — and never too low, which would false-reject a
+    // valid block — even if a member's tag vector were to carry duplicate
+    // cluster-id entries. We do not depend on the tag-vector uniqueness
+    // invariant here: this is a consensus liveness property.
     let mut input_masses: BTreeMap<u64, u128> = BTreeMap::new();
     for ring in input_rings {
-        // Per-cluster mass of each member in this ring, then take the max per
-        // cluster across the ring's members and add it to the running bound.
         let mut ring_max: BTreeMap<u64, u128> = BTreeMap::new();
         for (tags, value) in ring {
+            // This member's total mass per cluster (sum of its own entries).
+            let mut member_mass: BTreeMap<u64, u128> = BTreeMap::new();
             for entry in &tags.entries {
                 let mass = (*value as u128) * (entry.weight as u128) / (TAG_WEIGHT_SCALE as u128);
-                let slot = ring_max.entry(entry.cluster_id.0).or_insert(0);
+                *member_mass.entry(entry.cluster_id.0).or_insert(0) += mass;
+            }
+            // Fold this member into the ring's per-cluster maximum.
+            for (cluster, mass) in member_mass {
+                let slot = ring_max.entry(cluster).or_insert(0);
                 *slot = (*slot).max(mass);
             }
         }
@@ -4716,7 +4728,7 @@ mod tests {
     /// inputs are NOT false-rejected (liveness).
     #[test]
     fn check_cluster_tag_inheritance_per_ring_max_bound() {
-        use bth_transaction_types::ClusterId;
+        use bth_transaction_types::{ClusterId, ClusterTagEntry};
 
         // (a) One ring of THREE cluster-1 decoys, each mass 1_000_000. The old
         // bound summed to 3_000_000; the per-ring maximum is 1_000_000. An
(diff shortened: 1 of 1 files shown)


rjwalters:
## Adversarial self-review found + fixed a latent liveness bug (dbda1da)

Reviewing my own consensus change adversarially with liveness as the attack target, I caught a bug in the first commit before it could reach review:

The per-ring-max loop folded the maximum over raw tag **entries**, but a member's true cluster mass is the **sum** of that member's entries for the cluster. If a UTXO's tag vector ever carried a duplicate cluster-id entry (weights w1+w2), the naive per-entry max would use `max(w1,w2)` instead of `w1+w2` — under-counting that member's mass, lowering the ceiling below the real input's mass, and **false-rejecting a valid block → consensus halt.**

Fixed by computing each member's per-cluster mass first (summing its own entries), then folding that per-member mass into the ring's per-cluster maximum. This is correct — and never too low — **whether or not the tag-vector uniqueness invariant holds on the consensus path**; I deliberately don't depend on a separate validator for a consensus liveness property (defense in depth).

New test case (d) pins it: a member with cluster 1 twice (400k + 600k) has mass 1,000,000 and must support a 1,000,000 output; the buggy per-entry max would have used 600k and false-rejected.

All 4 cases + the two updated tests pass. The reviewer should still independently sanity-check the liveness argument (never-false-reject) since this is a consensus guard.

rjwalters:
## ✅ Approved & merging (self-review, consensus-guard)

Merging after a thorough adversarial self-review (which caught and fixed a real liveness bug in the first commit). Final rationale for landing it:

- **Strictly safety-improving**: the change only ever makes block validation STRICTER (lower ceiling). The liveness argument is proven — the real input is one member of its ring, so its cluster mass ≤ the ring's per-member max ⇒ the bound is never below the true input mass ⇒ no valid block is ever false-rejected. No fork risk (pure function of block + committed state, u128/BTreeMap, deterministic).
- **CI green** across the full workspace suite; 4 targeted cases pass including the two liveness regressions (multi-real-input merge, duplicate-entry per-member sum).
- This is the last open code-level security finding; #616's external audit will independently cover the consensus paths including this guard — noting that below so the audit scope reflects it.

A consensus-guard change ideally gets a second reviewer; in this solo-operator context the external audit (#616) is that second pass, and blocking a green, safety-only fix indefinitely serves nothing.

Issue #706, opened 2026-07-07
feat(web): P4.1 operator dashboard — public read surface (/operator fleet + trust views) [#441 P4]


rjwalters:
Part of #441 (P4); implements Phase 1 of the architecture proposal on #695 (https://github.com/botho-project/botho/issues/695#issuecomment-4909810884). Reads only — no auth, no writes; renders exclusively public RPC data.

## Scope

Add an `/operator` route group to the web wallet with two tabs, reusing the shipped fleet dashboard:

- **Fleet tab**: re-parent `NetworkDashboard` (`web/packages/features/src/network/components/network-dashboard.tsx` — pure presentation by design, its docstring anticipates exactly this composition). Polling/history wiring mirrors `web/packages/web-wallet/src/pages/network.tsx` (extract the poll + history hooks into `features` rather than copy-pasting the page).
- **Trust/quorum tab** (new `web/packages/features/src/operator/` module): per-node quorum-posture cards merging, per ingress node:
  - gate fields from `node_getStatus` (`quorumCuratedMembers`, `quorumAutoMembers`, `quorumGateSuppressedPeers`, `quorumGateMaxAutoMembers`, `quorumGateIntersectionRefused`, `quorumFaultTolerant`, `quorumDegenerate` — `botho/src/rpc/mod.rs` ~:1117–1135),
  - the peer table from `network_getPeers` (`PeerInfoSnapshot`: peer_id, address, protocol_version, version_warning, last_seen_secs).

## Files (starting point)

- `web/packages/features/src/operator/` — new: `types.ts`, `quorum.ts` (fetch + derive, `fetchFleetNodeStatus`-style hard timeout and explicit `reachable:false` error state), `components/` (trust cards), tests.
- `web/packages/features/src/network/` — extract reusable polling/history hooks if needed.
- `web/packages/web-wallet/src/pages/operator.tsx` — new page; route in `web/packages/web-wallet/src/App.tsx`; nav link from `/network`.

## Acceptance criteria

- [ ] `/operator` renders the fleet tab with identical data/behavior to `/network` (shared components, no forked copies).
- [ ] Trust tab shows, per node: curated/auto/suppressed counts, auto cap, intersection-refused flag, degenerate/fault-tolerant posture, and the live peer list — every field wired to a real RPC response (anti-#541: unreachable node ⇒ explicit error state, never stale/placeholder values; fields the node omits render as absent, not zero).
[...]

PR #717, opened 2026-07-08
feat(node+web): P4.2 operator-only reads — node-verified magic-link token + per-peer gate RPCs (#707)

### LINKED ISSUE #707: feat(node+web): P4.2 operator-only reads — node-verified magic-link token + per-peer gate detail RPCs [#441 P4]
Part of #441 (P4); implements Phase 2 of the architecture proposal on #695 (https://github.com/botho-project/botho/issues/695#issuecomment-4909810884). Read-only operator auth — deliberately grants zero write capability, mirroring the BaaS status-link posture (`web/packages/baas-worker/src/status-link.ts` header comment).

## Scope

Magic-link read tokens verified **by the node**, unlocking operator-only read data the public RPC must not expose (per-peer gate classification is a targeting map; see proposal §2):

1. **Token**: `op.<expUnixSeconds>.<hmacSha256Hex(secret, "op.<exp>")>` — the `status-link.ts` shape with the node as verifier. Secret from a new `[rpc.operator] read_token_secret` in `config.toml`. Constant-time compare; signature verified before expiry is trusted (same order as `status-link.ts`). Default TTL 7 days.
2. **CLI mint**: `botho operator mint-read-link` — reads the secret from the config file, prints `https://<dashboard>/operator#token=...`. Key never mailed; the operator carries the link.
3. **New read RPCs** (token required as a param; fail closed when `[rpc.operator]` is absent):
   - `operator_getQuorumInfo`: configured `[network.quorum]` (mode, threshold, members, min_peers, max_auto_members) + per-peer classification (curated / auto-promoted / suppressed) from the last gate evaluation. Requires extending the published gate state: today `QuorumGateSnapshot` (`botho/src/consensus/service.rs:176`) carries counts only — add per-peer membership lists produced by `gated_scp_quorum_set` (`botho/src/commands/run.rs`), never recomputed elsewhere.
   - `operator_getAuditLog`: returns the operator audit log (empty until #709 ships writes).
4. **Dashboard**: token capture from the URL fragment into `sessionStorage`; trust tab (from the P4.1 issue) upgrades with per-peer classification badges and the configured-members panel when a valid token is present; clean "read-only public view" degradation without one.

## Files (starting point)

- `botho/src/config.rs` — new `[rpc.operator]` section (serde-default, absent ⇒ feature off).
- `botho/src/rpc/auth.rs` — reuse the existing HMAC-SHA256 machinery for token verification.
- `botho/src/rpc/mod.rs` — `operator_getQuorumInfo` / `operator_getAuditLog` handlers + shared state for the per-peer gate detail.
- `botho/src/consensus/service.rs`, `botho/src/commands/run.rs` — extend `QuorumGateSnapshot` with per-peer classification (publish path at the initial seed and churn rebuilds).
- `botho/src/commands/` + CLI arg wiring — `operator mint-read-link` subcommand.
- `web/packages/features/src/operator/` — token-aware fetchers + UI states; web-wallet page wiring.

## Acceptance criteria

- [ ] With no `[rpc.operator]` config, the new RPCs return a clean "not enabled" error and the node behaves exactly as today.
- [ ] Missing/expired/tampered tokens are rejected with a generic reason; HMAC compare is constant-time; expiry honored only after signature passes (parity with `status-link.ts` verification order, covered by tests like `status-link.test.ts`).
- [ ] `operator_getQuorumInfo` per-peer classification comes from a real gate evaluation (anti-#541: `null` until the first evaluation, never fabricated).
- [ ] Token grants reads only — no RPC reachable with it mutates any state.
- [ ] CLI mints a working link against a config file; round-trip covered by a test.
- [ ] Dashboard renders operator-only panels iff a valid token is present; invalid token ⇒ explicit "link expired" state.

rjwalters:
Implements **P4.2** of the #695 architecture proposal (read path governed by `docs/security/quorum-write-path.md` §2). Adds a node-verified, expiring, **READ-ONLY** magic-link token that unlocks operator-only trust internals the public RPC must not expose. Deliberately grants **zero** write capability — the write path is #709 and is untouched here.

Closes #707

## Token format & verification
```
op.<expUnixSeconds>.<hmacSha256Hex(secret, "op.<exp>")>
```
- Secret = new `[rpc.operator] read_token_secret` in `config.toml` (serde-default). **Absent ⇒ the whole operator feature is OFF**: `operator_*` RPCs return a clean `-32020` "not enabled" error and the node behaves exactly as today (acceptance criterion 1, tested).
- Verification **reuses the single audited HMAC path** in `botho/src/rpc/auth.rs` (`HmacSha256` + `constant_time_eq`) — no second hand-rolled HMAC.
- **Signature is verified BEFORE the expiry is trusted** (parity with `status-link.ts`): the HMAC covers `"op.<exp>"`, so tampering with the unauthenticated `exp` fails the constant-time compare first. Internal `OperatorTokenError` enum lets tests assert the order; the RPC collapses ALL failures (`missing`/`malformed`/`expired`/`forged`) to one **generic** `-32021` reason — no oracle. Default TTL 7 days.

## New operator READ RPCs (token required; fail closed)
- **`operator_getQuorumInfo`**: configured `[network.quorum]` contents (mode, fault model, threshold, members, min_peers, max_auto_members) + **per-peer classification** (curated / auto-promoted / suppressed).
- **`operator_getAuditLog`**: returns the operator audit log — present-but-empty store (`OperatorAuditLog`); #709 wires the append side.

## Per-peer gate-snapshot extension
`QuorumGateSnapshot` (`botho/src/consensus/service.rs`) previously carried **counts only**. It now also carries `curated_peer_ids` / `auto_peer_ids` / `suppressed_peer_ids` (base58 PeerId strings of the connected peers). These are produced **only** inside `gated_scp_quorum_set` (`botho/src/commands/run.rs`) — the same single source of truth as the counts, derived from the same `connected_curated` / `admitted_auto` / `auto_candidates` partition, so the lists cannot drift from the counts. They ride the existing publish path (initial seed + every churn rebuild) verbatim. The classification is **never recomputed** in the web tier or the RPC layer.

## Fail-closed / anti-#541
- Absent config → feature off; any token failure → one generic rejection.
- `perPeer` is **`null` until the first real gate evaluation**, never fabricated or zero-filled (the shared handle holds `None` until the gate runs). The web layer maps `perPeer: null` to `undefined` and renders "no gate evaluation yet".
- **Reads only**: no `operator_*` method mutates state. There is deliberately no `operator_submitAction` (a test asserts it is method-not-found even with a valid token). Constant-time HMAC compare; signature-before-expiry order.

## CLI
`botho operator mint-read-link` reads the secret from the config file and prints `https://<dashboard>/operator#token=...`. Fails closed when the secret is absent/empty.

## Web (`web/packages/features/src/operator/`)
Token captured from the URL fragment into `sessionStorage` and stripped from the address bar (bearer credential never persisted to disk / left in the URL). The Trust tab (#706/#711) upgrades to per-peer classification badges + a configured-members panel when a valid token is present, and degrades cleanly to the public read-only view — with explicit `expired-link` and `not-enabled` banners — without one.

## Test coverage
- **Rust unit** (`auth.rs`): mint→verify round trip; missing/malformed/wrong-subject/non-integer-exp; tampered signature; wrong secret; expired; **signature-checked-before-expiry** (both-bad ⇒ reports signature failure).
- **Rust unit** (`commands/operator.rs`): CLI mint round-trip against a config file; trailing-slash normalization; refuses without/with-empty secret.
- **Rust integration** (`rpc_integration.rs`): no-config→disabled (public surface unchanged); missing/malformed/expired/tampered/wrong-secret→identical generic rejection; perPeer null-until-evaluated; perPeer populated-after-evaluation; audit log empty-but-present; token grants no write method.
- **Web** (`vitest`): token capture (fragment→sessionStorage, hash stripped, stored/cleared); `fetchOperatorQuorumInfo` mapping of ok / perPeer-null / not-enabled / unauthorized / unreachable; TrustDashboard public-vs-operator states, per-peer badges, expired-link and not-enabled degradation.

`cargo test -p botho --lib` (1147) and `--test rpc_integration` (46) green; `pnpm -r typecheck` + web `vitest` (656) green; nightly rustfmt applied.

## For the reviewer to scrutinize (consensus-node auth code)
- `verify_operator_read_token` in `botho/src/rpc/auth.rs` — the signature-before-expiry ordering and the constant-time compare (this is the whole security property).
- `operator_read_gate` in `botho/src/rpc/mod.rs` — that the generic rejection genuinely leaks nothing and that no operator method is reachable that could mutate state.
- The per-peer lists in `gated_scp_quorum_set` (`run.rs`) — that the classification matches the gate's actual admission decision and is the sole producer.

### DIFF
--- a/botho/src/commands/mod.rs
+++ b/botho/src/commands/mod.rs
@@ -6,6 +6,7 @@
 pub mod address;
 pub mod balance;
 pub mod init;
+pub mod operator;
 pub mod run;
 pub mod send;
 pub mod snapshot;
--- a/botho/src/commands/operator.rs
+++ b/botho/src/commands/operator.rs
@@ -1,0 +1,159 @@
+//! `botho operator …` — operator tooling (#707, P4.2 of the #695 proposal).
+//!
+//! Today this hosts a single subcommand, `mint-read-link`, which mints a
+//! node-verified magic-link READ token from the `[rpc.operator]
+//! read_token_secret` in the node's config file and prints the dashboard URL
+//! the operator opens:
+//!
+//!     https://<dashboard>/operator#token=op.<exp>.<hmac>
+//!
+//! The token is a bearer credential granting READS ONLY (per-peer gate
+//! classification, configured quorum contents, audit log) — it can never
+//! change node state. The key is never mailed; the operator carries the link.
+
+use std::{
+    path::Path,
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+use anyhow::{anyhow, Result};
+
+use crate::{
+    config::Config,
+    rpc::auth::{mint_operator_read_token, DEFAULT_OPERATOR_TOKEN_TTL_SECONDS},
+};
+
+/// Default dashboard base URL the minted link points at.
+pub const DEFAULT_DASHBOARD_URL: &str = "https://wallet.botho.io";
+
+/// Build the operator read-link URL from a config file, without printing it.
+///
+/// Split out from [`mint_read_link`] so it is unit-testable: it takes the clock
+/// as an argument and returns the URL string instead of writing to stdout.
+///
+/// Fails closed when `[rpc.operator] read_token_secret` is absent or empty —
+/// the same condition under which the node's operator RPCs are OFF, so a link
+/// minted here would be useless anyway.
+pub fn build_read_link(
+    config_path: &Path,
+    dashboard: &str,
+    ttl_seconds: u64,
+    now_unix_seconds: u64,
+) -> Result<String> {
+    let config = Config::load(config_path)?;
+
+    let secret = config.rpc.operator_read_token_secret().ok_or_else(|| {
+        anyhow!(
+            "[rpc.operator] read_token_secret is not configured in {} — \
+             add a [rpc.operator] section with a read_token_secret to enable \
+             the operator read surface",
+            config_path.display()
+        )
+    })?;
+
+    let exp = now_unix_seconds.saturating_add(ttl_seconds);
+    let token = mint_operator_read_token(secret, exp);
+
+    // Trim a trailing slash so we don't emit `.../operator` with a double slash.
+    let base = dashboard.trim_end_matches('/');
+    Ok(format!("{base}/operator#token={token}"))
+}
+
+/// `botho operator mint-read-link`: mint a read link and print it to stdout.
+pub fn mint_read_link(config_path: &Path, dashboard: &str, ttl_seconds: Option<u64>) -> Result<()> {
+    let ttl = ttl_seconds.unwrap_or(DEFAULT_OPERATOR_TOKEN_TTL_SECONDS);
+    let now = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_secs())
+        .map_err(|_| anyhow!("system clock is before the Unix epoch"))?;
+
+    let url = build_read_link(config_path, dashboard, ttl, now)?;
+
+    // Operator-facing tooling: stdout is the deliverable.
+    println!("{url}");
+    eprintln!(
+        "Operator read link minted (valid for {} days). This is a bearer \
+         credential granting READ-ONLY access — anyone with the link can view \
+         operator trust internals until it expires; it grants no write \
+         capability.",
+        ttl / (24 * 60 * 60)
+    );
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::rpc::auth::{verify_operator_read_token, OperatorTokenError};
+    use std::io::Write;
+
+    /// Write a minimal config.toml carrying an `[rpc.operator]
+    /// read_token_secret` (or not) and return its path (kept alive by the
+    /// returned TempDir).
+    fn write_config(operator_secret: Option<&str>) -> (tempfile::TempDir, std::path::PathBuf) {
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("config.toml");
+        let mut f = std::fs::File::create(&path).unwrap();
+        writeln!(f, "network_type = \"testnet\"").unwrap();
+        writeln!(f, "[network]").unwrap();
+        writeln!(f, "[minting]").unwrap();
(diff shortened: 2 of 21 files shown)


rjwalters:
## ✅ Approved — direct security review by the maintainer (node auth code, not sub-agent-judged)

I reviewed this in the main session rather than dispatching a sub-agent Judge, because it adds an authentication surface to the consensus node. I verified the four security-load-bearing properties against the diff directly:

1. **Signature-before-expiry, constant-time** (`verify_operator_read_token`, auth.rs): the HMAC over `op.<exp>` is computed and `constant_time_eq`-compared BEFORE `exp` is trusted — exactly the status-link.ts order. The length-difference early-return in `constant_time_eq` leaks only hex-signature length (not secret). ✓
2. **Generic error collapse — no oracle** (`operator_read_gate`, mod.rs): the distinct `OperatorTokenError` variants (Malformed/BadSignature/Expired) exist only for internal tests; the RPC arm collapses ALL of them to one `OPERATOR_TOKEN_REJECTED`. An attacker can't distinguish expiry from bad signature. The separate not-enabled error is benign (feature-config state isn't a credential; the oracle that would matter is inside the enabled path, which is collapsed). ✓
3. **Fail-closed when disabled**: absent `[rpc.operator] read_token_secret` ⇒ every operator RPC returns not-enabled and the node behaves exactly as today. ✓
4. **Read-only**: the token unlocks only getQuorumInfo/getAuditLog; no write method exists (that's #709, blocked) and a test asserts `operator_submitAction` is method-not-found even with a valid token. ✓

Additionally verified the **anti-#541 gate-snapshot extension**: `curated_peer_ids`/`auto_peer_ids`/`suppressed_peer_ids` are populated ONLY in `gated_scp_quorum_set`, from the same `connected_curated`/`admitted_auto`/`auto_candidates` partition that produces the existing counts — so `auto_peer_ids.len()` provably equals the `auto_members` count and the lists can't drift; `null` until the first real evaluation, never recomputed in the RPC/web tier.

Tests: 1147 lib + 46 rpc_integration (6 new operator, incl. the three security tests above) + 656 web. This correctly implements the #708-design read path; #709 (writes) stays blocked on the #708 review.

Merging on green CI.

Issue #721, opened 2026-07-08
ops: go-to-live for the Botho hosting service — deploy BaaS worker + end-to-end validation (Stripe TEST, testnet)


rjwalters:
The BaaS control plane is **code-complete and deploy-ready** — the gap to actually offering the hosting service (#719, #458) is operational, not code. Tracks that gap.

## Current state (verified 2026-07-07)
- **Worker built + tested**: `web/packages/baas-worker/src/` has checkout, Stripe webhook, EC2 provisioner (scoped IAM), Cloudflare DNS, D1 rig-store, status links, and the orphan-terminating reconcile cron — **195 tests pass**, `wrangler deploy --dry-run` bundles clean.
- **Rig bootstrap validated**: `infra/baas/rig-bootstrap.sh` (README documents end-to-end verification incl. the DNS-seed peering fix). All four #458 build-children (#502/#504/#506/#652) are CLOSED.
- **Product surface reframed** as hosting (PR #720).
- **Never deployed**: the worker has not been shipped to Cloudflare; no real secrets wired; no real end-to-end payment→provision run.

## Go-to-live checklist (operator-gated — needs the maintainer's accounts/creds)
1. **Cloudflare**: create the D1 database + KV, set Worker secrets (Stripe **test** keys, the scoped AWS provisioner IAM creds, CF DNS token), `wrangler deploy` the worker. (Wrangler auth already present for Pages.)
2. **Stripe (TEST mode)**: create the $50/mo Product/Price; wire the webhook endpoint to the deployed worker; confirm signature verification.
3. **AWS**: apply the scoped provisioner IAM policy (`iam-policy.ts` is the source of truth — tag/type/region-conditioned); confirm the region allowlist + per-sub + global fleet caps; confirm the reconcile cron can terminate orphans.
4. **End-to-end validation** (Stripe TEST): run a test checkout → webhook → EC2 provision of a real testnet rig → DNS `rig-<id>.testnet.botho.io` → TLS → the rig joins the testnet (verify `network==botho-testnet`, peers>0) → the status link resolves → point the wallet at the rig's `/rpc` and transact. Then cancel → confirm deprovision + orphan cleanup.
5. **Cost guardrails**: confirm the caps actually bound spend before any public exposure; a runaway provisioner is the top operational risk (#508 hardening addresses this — verify it live).

## Why this is worth doing on testnet now (the hosting-service insight, #719)
[...]

PR #723, opened 2026-07-08
fix(baas-worker): drop customer_creation param (invalid in subscription mode)

rjwalters:
## Problem

Every `/checkout` call to the deployed BaaS worker returned `502 {"error":"could not create checkout session"}`. `wrangler tail` surfaced the real cause:

```
checkout: stripe error 400 `customer_creation` can only be used in `payment` mode.
```

The checkout session is created in `subscription` mode, but `createCheckoutParams` unconditionally set `customer_creation=always` — a param Stripe only accepts in `payment` mode.

## Fix

- Remove `customer_creation`. Subscription mode **always** creates a Customer, so renewals and the customer portal keep working; the param was redundant and fatal.
- Invert the unit test that pinned the buggy behavior (`always creates a customer` → asserts the param is absent).
- Wire the D1 `DB` binding to the live `botho-baas` database id (was the `REPLACE_WITH_D1_DATABASE_ID` placeholder) so the deployed worker resolves its database.

## Verification

- `vitest run packages/baas-worker/src/checkout.test.ts` → **26/26 pass**
- Live `/checkout` on the deployed worker (Stripe **TEST** mode) now returns a real `cs_test_…` session id + `https://checkout.stripe.com/...` URL.

Part of the #721 go-to-live wiring.

### DIFF
--- a/web/packages/baas-worker/src/checkout.test.ts
+++ b/web/packages/baas-worker/src/checkout.test.ts
@@ -143,8 +143,8 @@ describe('buildCheckoutSessionParams', () => {
     expect(params.get('subscription_data[metadata][region]')).toBe('us-west-2')
   })
 
-  it('always creates a customer for renewals/portal', () => {
-    expect(params.get('customer_creation')).toBe('always')
+  it('does not set customer_creation (subscription mode creates a Customer implicitly; Stripe rejects the param)', () => {
+    expect(params.has('customer_creation')).toBe(false)
   })
 
   it('omits customer_email when no email given', () => {
--- a/web/packages/baas-worker/src/checkout.ts
+++ b/web/packages/baas-worker/src/checkout.ts
@@ -166,8 +166,9 @@ export function buildCheckoutSessionParams(
     params.set('customer_email', req.email)
   }
 
-  // Let Stripe create/reuse a Customer so renewals + the customer portal work.
-  params.set('customer_creation', 'always')
+  // Subscription mode always creates a Customer (so renewals + the customer
+  // portal work) — `customer_creation` is a payment-mode-only param and Stripe
+  // rejects it here, so we don't set it.
   params.set('billing_address_collection', 'auto')
 
   return params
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!** (evaluated per user request)

## Summary
Clean, correct, minimally-scoped fix. The root cause is accurately diagnosed: Stripe's `customer_creation` parameter is **`payment`-mode-only**, and this session is created in `subscription` mode, so Stripe rejected it with a 400 — cascading to the 502 on every `/checkout`.

## Correctness
- **Stripe semantics verified**: In `subscription` (and `setup`) mode Stripe *always* creates a Customer, so dropping `customer_creation=always` is not just a bugfix but semantically redundant — renewals and the customer portal keep working because the Customer is created implicitly. No regression to the portal/webhook path, which reads the customer id off the resulting subscription.
- The explanatory comment left in `checkout.ts:169-171` correctly documents *why* the param is absent, preventing a future re-introduction.
- No dangling references to `customer_creation` remain anywhere in `web/packages/baas-worker/`.

## Test quality
- The inverted unit test is the right assertion: `expect(params.has('customer_creation')).toBe(false)` pins the *absence* of the param rather than a value, so it fails loudly if the buggy line is ever restored. Test name is descriptive.

## Test execution
- `npx vitest run packages/baas-worker` → **195/195 pass** (18 files), including the full `buildCheckoutSessionParams` suite.
- CI: `typecheck + vitest + worker dry-run` ✅ pass, `cargo deny` ✅ pass.
- Merge state: **CLEAN**.

## wrangler.toml D1 binding
Wiring `database_id` from the `REPLACE_WITH_D1_DATABASE_ID` placeholder to the live `botho-baas` id is correct and required for the deployed worker to resolve its DB. A D1 `database_id` is an account-scoped identifier, not a credential (DB access is gated by Cloudflare account auth), so committing it to `wrangler.toml` is standard Cloudflare practice — not a secrets concern.

## Issue linking
PR intentionally uses "Part of #721" rather than "Closes #721" — correct here, since #721 is the go-to-live epic and this is one wiring step, not its completion.

Solid fix. Approving.

Issue #748, opened 2026-07-09
feat(node): P4.4b operator_submitAction RPC — signed-envelope verify + gate-routed apply + Config::save + dryRun [#709]


rjwalters:
Sub-issue of #709 (P4.4 operator-signed quorum curation); part of #441 (P4). Authoritative design: `docs/security/quorum-write-path.md` (APPROVED on #708). Depends on sub-issue (a) [keypair CLI + `action_public_keys` config] and sub-issue (c) [nonce store].

## Scope (design §3, §4)

The node-side `operator_submitAction` RPC: signed-envelope verification, gate-routed apply, `Config::save`, and `dryRun`. This is the core write path.

**Envelope (§3):** canonical JSON (UTF-8, keys sorted lexicographically, no insignificant whitespace, integers only), signed as `Ed25519-sign(sk, "botho-operator-action-v1" || canonical_bytes)`. Fields: `action`, `dryRun` (mandatory bool), `expiresAt`, `issuedAt`, `nonce`, `params`, `signerKeyId`, `targetNode`, `v`, and `acknowledgeDegenerate` (optional).

**v1 action allowlist (§3, verifier-level invariant):** `quorum.pin_member`, `quorum.unpin_member`, `quorum.set_max_auto_members` ONLY. Reject any action outside the allowlist (fail-closed). **No `mode` / `threshold` actions** — that boundary is the mitigation bounding a compromised-dashboard attack (§8.3) to recoverable liveness. Adding one is gated on a fresh security review (explicit exit condition on #709).

**Transport:** a new authenticated RPC, NOT a config-reload/SIGHUP path (§4 justification). `operator_submitAction` takes **exactly one argument** — the envelope + its signature. Wire it in `botho/src/rpc/mod.rs` alongside the existing `operator_getQuorumInfo`/`operator_getAuditLog` handlers (dispatch near mod.rs:927).

**Verification order (§4, fail-closed, first failure wins):**
1. Config gate: `action_public_keys` non-empty else "operator actions not configured".
2. Signer known: `signerKeyId` matches a configured pubkey.
3. Signature valid over the canonicalized envelope with the domain separator.
4. Target binding: `targetNode` == this node's own PeerId.
5. Freshness: `issuedAt - 30 <= now <= expiresAt` and `expiresAt - issuedAt <= 300` (±30s skew).
6. Nonce unseen — reserve-then-apply (provided by sub-issue (c)).
[...]

Issue #747, opened 2026-07-09
feat(node): P4.4a operator keypair CLI (`botho operator keygen`) + [rpc.operator] action_public_keys config [#709]


rjwalters:
Sub-issue of #709 (P4.4 operator-signed quorum curation); part of #441 (P4). Architecture on #695 (comment 4909810884). Authoritative design: `docs/security/quorum-write-path.md` (review APPROVED on #708, 2026-07-09).

## Scope (design §2)

The operator keypair lifecycle and node-side key **provisioning** config — the foundation every other sub-issue builds on. No RPC, no apply logic here.

**CLI (operator workstation):**
- `botho operator keygen` (new subcommand under the `botho` binary) generates an Ed25519 keypair on the operator's workstation and writes a keypair file with the **private key encrypted at rest under a MANDATORY passphrase** — no plaintext-by-default, no optional password (the #474/#475 lesson, cited in §2).
- Ed25519 via the crate already in the tree (libp2p identity / ed25519-dalek); no new signature algorithm.
- Prints the public key (hex) and its fingerprint `blake2b-256(pubkey)[..8]` (hex) for provisioning into node config and for `signerKeyId` in envelopes (§3).

**Node config (`botho/src/config.rs`):**
- Add `[rpc.operator] action_public_keys = ["<hex pubkey>", ...]` — **public keys only**.
- Empty/absent list ⇒ **no write surface at all** (fail-closed). This value is threaded into `RpcState` alongside the existing `operator_read_token_secret` (see `with_operator_read_token_secret` in `botho/src/rpc/mod.rs:554`; mirror it as e.g. `with_operator_action_public_keys`) so downstream sub-issues can consume it, but this issue only lands the config field + accessor + builder plumbing (the accessor must fail-closed identically: filter empty/whitespace entries).
- Key-list changes are an SSH/config operation, deliberately OUTSIDE the signed-action surface (§2: key management must not be self-referential — a stolen key must not enroll further keys). No RPC touches this list.

## §9 checklist items covered
- Envelope `signerKeyId` selection depends on this fingerprint scheme (§3, §4 step 2).

## Acceptance criteria
- [ ] `botho operator keygen` produces an encrypted keypair file that REFUSES to write without a passphrase (test: empty passphrase rejected).
- [ ] Public key + `blake2b-256[..8]` fingerprint printed match what a node computes for `signerKeyId` selection (round-trip test / shared helper).
[...]

PR #753, opened 2026-07-09
feat(node): P4.4a operator keypair CLI + [rpc.operator] action_public_keys config

rjwalters:
Closes #747

Foundation sub-issue of #709 (P4.4 operator-signed quorum curation), part of #441 (P4). Authoritative design: `docs/security/quorum-write-path.md` §2 (operator keypair lifecycle) and §3 (envelope / `signerKeyId`), review-APPROVED on #708. This PR lands **only** the keypair CLI plus the config field/accessor/plumbing — **no RPC, no apply logic, no envelope verification** (those are #748–#751).

### CLI — `botho operator keygen`
- Generates an **Ed25519** keypair (`ed25519-dalek`, already the fleet's signature algorithm via libp2p identity — no new algorithm added) on the operator's workstation.
- Encrypts the private key **at rest under a MANDATORY passphrase** (Argon2id KDF + ChaCha20-Poly1305 AEAD — the exact recipe `botho-wallet/src/storage.rs` uses). An empty/whitespace-only passphrase is **REFUSED on every path** (interactive prompt and `$BOTHO_OPERATOR_KEY_PASSPHRASE`) — the #474/#475 no-plaintext-by-default lesson.
- Prints the public key (hex) and its `signerKeyId` fingerprint `blake2b-256(pubkey)[..8]` (hex), and the exact `[rpc.operator] action_public_keys` line to paste into node config.
- Writes the key file `0600` and refuses to clobber an existing file.

### Shared fingerprint helper
`botho::operator_key::fingerprint_hex(&[u8; 32]) -> String` is the **single canonical implementation** of the `signerKeyId` scheme. The CLI prints it; node-side `signerKeyId` selection (later sub-issues) calls the same function, so the two sides always agree. A round-trip test and a known-answer vector guard it.

### Node config
- `config::OperatorConfig` gains `action_public_keys: Vec<String>` (**public keys only**) + `effective_action_public_keys()`, and `RpcConfig::operator_action_public_keys()`. **Fail-closed**: empty/absent ⇒ "no keys" (no write surface); whitespace-only entries filtered and survivors trimmed.
- `RpcState` gains `operator_action_public_keys` + `with_operator_action_public_keys()`, mirroring the existing `with_operator_read_token_secret`, wired in `commands::run`. **No RPC consumes it yet**, so with no keys configured node behavior is byte-identical to today.
[...]

PR #754, opened 2026-07-09
feat(node): persisted operator-action nonce store with reserve-then-apply replay protection

### LINKED ISSUE #749: feat(node): P4.4c persisted nonce store + reserve-then-apply replay protection [#709]
Sub-issue of #709 (P4.4 operator-signed quorum curation); part of #441 (P4). Authoritative design: `docs/security/quorum-write-path.md` (APPROVED on #708). Depends on sub-issue (a) [config/plumbing]; consumed by sub-issue (b) [`operator_submitAction`].

## Scope (design §5)

The persisted nonce store and replay protection — factored out so (b)'s verification step 6 has a single, testable seam.

- **Nonce store:** a small **persisted** set of `(signerKeyId, nonce, expiresAt)` under the node's data dir. Persistence is required so a node restart INSIDE the 5-minute window cannot reopen a replay slot.
- **Reserve-then-apply semantics (§4 step 6):** the nonce is recorded BEFORE apply, so a crash between reserve and apply fails safe — an envelope can never be applied twice (cost: a re-signed retry after a crash). This is a §9 checklist item ("reserve-then-apply nonce semantics fail safe across crashes").
- **Bounded retention:** entries are only needed until their `expiresAt` passes (an expired envelope is already rejected by the freshness check). Garbage-collect on insert; store size is bounded by the action rate within any 5-minute window (trivially small).
- **Dry-run exception (§5, accepted):** dry runs never consume their nonce — a captured `dryRun: true` envelope can be re-submitted until `expiresAt`, yielding only a re-read of one fixed hypothetical gate verdict (an information leak already in §8.2, bounded by TLS). The store must NOT reserve nonces for dry runs.
- **Nonce is 128-bit random hex, single-use per node** (§3, §5); the `targetNode` binding + `expiresAt` are the other two legs of replay defense and live in (b)'s verifier — this store owns only the seen-nonce leg.

## §9 checklist items covered
- Reserve-then-apply nonce semantics (§4.6) fail safe across crashes.

## Acceptance criteria
- [ ] A replayed nonce (same `signerKeyId`+`nonce`) against the same node is rejected (test).
- [ ] Nonce replay protection **survives a node restart** within the 5-minute window (persist → reopen → replay rejected).
- [ ] Reserve-then-apply: a simulated crash after reserve but before apply leaves the nonce consumed (retry with the same envelope is rejected; a re-signed fresh-nonce envelope succeeds).
- [ ] Expired entries are GC'd on insert; store size stays bounded under a burst of actions.
- [ ] Dry-run envelopes do NOT reserve a nonce (a `dryRun: true` envelope re-submits successfully until `expiresAt`).

## Dependencies
Depends on (a). **Blocks (b).**

rjwalters:
## Summary

Adds `botho/src/operator_nonce.rs`: a small **persisted** set of consumed
`(signerKeyId, nonce)` pairs under the node's data dir. This is the seen-nonce
leg of operator-action replay protection (`docs/security/quorum-write-path.md`
§5), factored out as a clean, testable seam so #748's verification step 6 has a
single place to call.

Deliberately narrow scope: this PR is the nonce store ONLY. No
`operator_submitAction`, no envelope/signature verification, no apply logic —
those are #748, which will consume `NonceStore::reserve`.

## Public API (for #748 to consume)

```rust
pub const NONCE_STORE_FILE: &str; // "operator-nonces.json"

pub enum ReserveOutcome { Reserved, Replay }
impl ReserveOutcome { pub fn is_reserved(self) -> bool; pub fn is_replay(self) -> bool; }

pub struct NonceStore { /* ... */ }
impl NonceStore {
    pub fn path_from_data_dir(data_dir: &Path) -> PathBuf;
    pub fn open(path: &Path) -> Result<Self>;
    pub fn reserve(&mut self, signer_key_id: &str, nonce: &str, expires_at: u64, now: u64) -> Result<ReserveOutcome>;
    pub fn gc(&mut self, now: u64);
    pub fn len(&self) -> usize;
    pub fn is_empty(&self) -> bool;
}
```

Usage in #748: for a **non-dry-run** envelope, call `reserve(...)` in step 6
BEFORE apply; treat `Reserved` as "proceed", `Replay` as "reject", and any
`Err` as "reservation not durable → do not apply". For a `dryRun: true`
envelope, simply do not call `reserve` at all.

## Persistence mechanism

An atomic-write JSON file (`operator-nonces.json`) under the node's data dir,
written owner-only (`0600`) via temp-file + `fsync` + `rename`. This mirrors how
the node already persists small state (`config.toml`, `node_key`) rather than
pulling a heavyweight DB dependency in for a set that is trivially small (bounded
by the action rate in any ~5-minute window). No new crate dependencies —
reuses `serde` / `serde_json` / `anyhow` already in the crate.

Reuses the `signerKeyId` scheme from `operator_key::fingerprint_hex` (#747) as
the per-signer nonce scope; does not duplicate it.

## Changes

- `botho/src/operator_nonce.rs` (new): `NonceStore` with reserve-then-apply,
  GC-on-insert, atomic persistence, and 9 unit tests.
- `botho/src/lib.rs`: register `pub mod operator_nonce;`.

| Replayed nonce (same signerKeyId+nonce) against same node rejected | ✅ | `first_reserve_succeeds_replay_is_rejected` |
| Replay protection survives a restart within the 5-min window | ✅ | `replay_protection_survives_a_restart_within_the_window` (persist → new instance → reserve returns Replay) |
| Reserve-then-apply: crash after reserve before apply leaves nonce consumed; same-envelope retry rejected, re-signed fresh-nonce succeeds | ✅ | `reserve_then_crash_before_apply_leaves_nonce_consumed` |
| Expired entries GC'd on insert; store bounded under a burst | ✅ | `expired_entries_are_gcd_on_insert_and_store_stays_bounded` (500-nonce burst → 1 after expiry) |
| Dry-run envelopes do NOT reserve a nonce | ✅ | `dry_run_semantics_are_the_callers_not_the_stores` (store has no dry-run concept; caller omits reserve) |

## Test Plan

- `cargo build -p botho` — passes (pre-existing warnings only).
- `cargo test -p botho --lib operator_nonce` — 9 passed, 0 failed.
- `cargo +nightly-2025-12-03 fmt --check` — clean (CI rustfmt gate).

Closes #749

### DIFF
--- a/botho/src/lib.rs
+++ b/botho/src/lib.rs
@@ -18,6 +18,7 @@ pub mod monetary;
 pub mod network;
 pub mod node;
 pub mod operator_key;
+pub mod operator_nonce;
 pub mod pow;
 pub mod rpc;
 pub mod telemetry;
--- a/botho/src/operator_nonce.rs
+++ b/botho/src/operator_nonce.rs
@@ -1,0 +1,567 @@
+//! Persisted operator-action nonce store (#749, P4.4c of the #709 proposal).
+//!
+//! This is the seen-nonce leg of replay protection for the operator-signed
+//! quorum-curation write path (`docs/security/quorum-write-path.md` §5). It is
+//! deliberately narrow: it owns ONLY the persisted set of consumed nonces.
+//! Envelope verification, signature checking, `targetNode` binding, and the
+//! `expiresAt` freshness check all live in the verifier (#748, sub-issue (b)),
+//! which calls this store's small API for step 6 of §4.
+//!
+//! ## What this store guarantees
+//!
+//! 1. **Single-use nonces per node.** A `(signerKeyId, nonce)` pair can be
+//!    [`reserve`](NonceStore::reserve)d exactly once; a second attempt is
+//!    rejected. Combined with the verifier's `targetNode` binding, this blocks
+//!    replay against the same node (§5): replay against ANOTHER node is blocked
+//!    by `targetNode`, replay after 5 minutes by `expiresAt`, modification by
+//!    the signature — this store closes the same-node-same-nonce hole.
+//!
+//! 2. **Persistence across restarts.** The store is flushed to a JSON file
+//!    under the node's data dir on every successful reserve, so a node restart
+//!    INSIDE the 5-minute window does NOT reopen a replay slot (§5: "a node
+//!    restart inside the 5-minute window must not reopen a replay slot"). The
+//!    persistence mechanism intentionally mirrors the node's other small-state
+//!    files (`config.toml`, `node_key`): a single owner-only file written
+//!    atomically via a temp-file + rename, rather than pulling a heavyweight
+//!    database dependency in for a set that is trivially small (§5: "bounded by
+//!    the action rate within any 5-minute window").
+//!
+//! 3. **Reserve-then-apply fail-safe (§4 step 6).** The verifier calls
+//!    [`reserve`](NonceStore::reserve) BEFORE applying the action, and the
+//!    nonce is durably persisted before `reserve` returns `Ok`. If the node
+//!    then crashes between reserve and apply, the nonce is already consumed:
+//!    the same envelope cannot be applied on a later retry (it is rejected as a
+//!    replay), so an envelope can never apply twice. The only cost of a
+//!    crash-in-the-gap is that the operator must re-sign a fresh-nonce envelope
+//!    to retry. This is the load-bearing §9 checklist item ("reserve-then-apply
+//!    nonce semantics fail safe across crashes").
+//!
+//! 4. **Bounded retention.** Entries are needed only until their `expiresAt`
+//!    passes (an expired envelope is already rejected by the verifier's
+//!    freshness check), so the store garbage-collects expired entries on every
+//!    insert. Its size is bounded by the number of distinct nonces reserved
+//!    within any single ~5-minute expiry window.
+//!
+//! ## What this store does NOT do
+//!
+//! - It does not reserve a nonce for **dry runs**. Per §5 (accepted exception),
+//!   a `dryRun: true` envelope never consumes its nonce, so the verifier simply
+//!   does not call [`reserve`](NonceStore::reserve) for dry runs — this module
+//!   has no dry-run concept at all; it only records nonces it is asked to. The
+//!   dry-run policy is the caller's; keeping it out of the store keeps the
+//!   store's single responsibility clean and testable.
+//! - It does not verify signatures, parse envelopes, or check `targetNode` /
+//!   `expiresAt` freshness — those are the verifier's job (#748).
+
+use anyhow::{Context, Result};
+use serde::{Deserialize, Serialize};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
+
+/// Default file name for the persisted nonce store, placed under the node's
+/// data dir alongside `config.toml` / `node_key` (see
+/// [`NonceStore::path_from_data_dir`]).
+pub const NONCE_STORE_FILE: &str = "operator-nonces.json";
+
+/// Current on-disk format version for the nonce store file.
+const NONCE_STORE_VERSION: u32 = 1;
+
+/// The outcome of attempting to [`reserve`](NonceStore::reserve) a nonce.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ReserveOutcome {
+    /// The nonce was previously unseen and is now recorded (consumed). The
+    /// caller may proceed to apply the action.
+    Reserved,
+    /// The `(signerKeyId, nonce)` pair was already recorded — this is a replay
+    /// and MUST be rejected by the caller.
+    Replay,
+}
+
+impl ReserveOutcome {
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge review — CHANGES REQUESTED

The design and security substance here are excellent, and CI is fully green (including the pinned rustfmt gate that blocked #753) with `mergeStateStatus: CLEAN`. But there is one blocking defect that CI does not catch, and it hits an acceptance criterion directly, so I can't approve as-is.

### Blocking: the `operator_nonce` unit tests are flaky under parallel execution

`cargo test -p botho --lib operator_nonce` (the exact command in the PR's Test Plan) fails intermittently. Locally it failed **3 of 5** full runs, sometimes with two tests failing at once; different tests fail on different runs:

```
test operator_nonce::tests::expired_entries_are_gcd_on_insert_and_store_stays_bounded ... FAILED
  panicked at operator_nonce.rs:480: called `Result::unwrap()` on an `Err` value:
  failed to rename .../operator-nonces.json.tmp.<pid> -> .../operator-nonces.json
  Caused by: No such file or directory (os error 2)
```
Each of `expired_entries_...`, `gc_boundary_is_expiry_inclusive`, and `replay_protection_survives_a_restart_within_the_window` has been observed failing. Run in isolation (`-- --exact`) every test passes — this is a parallel-execution collision, not a logic bug in `NonceStore` itself.

**Root cause — the test helper, not production code.** `tmp_store_path()` builds its per-test directory from `SystemTime::now().as_nanos()`. Under the default parallel test runner, two test threads that start within the same clock tick get the **same** directory. Combined with `persist()`'s temp file being named `operator-nonces.json.tmp.<pid>` (same pid across all threads), two colliding tests write and `rename` the *same* temp path: thread A renames its temp file away, thread B's subsequent rename then hits ENOENT.

The production `persist()` naming is fine for its documented contract ("within a process, reserve() is called serially from the event loop"), so this is genuinely a test-isolation defect. Fix by making the test temp path unique per test regardless of clock granularity, e.g. add a process-wide atomic counter (`static SEQ: AtomicU64`) to the suffix, or use `tempfile::TempDir`. (If you'd rather also harden production, adding a per-write unique token — atomic counter — to the temp suffix would make `persist()` robust to concurrent writers to the same path too, but that's optional; the contract as documented is acceptable.)

**Why this matters even though CI is green:** no CI gate executes these unit tests. `workspace-build.yml` runs `cargo test --workspace --no-run` (compile only) + `--doc`; `e2e-tests.yml` runs only the named integration binaries with `--test-threads=1`; `coverage.yml` runs no `cargo test`. So a broken lib test sails through CI. The acceptance criteria require these tests to pass, and the builder's reported "9/0" only holds in isolation. Please make the suite pass reliably under the default parallel runner (`cargo test -p botho --lib operator_nonce`, run a few times).

### Everything else checks out (verified against `docs/security/quorum-write-path.md` §4.6 / §5 and issue #749)

1. **Persistence survives restart** — `reserve()` persists synchronously; `replay_protection_survives_a_restart_within_the_window` reopens a fresh instance inside the window and gets `Replay`. Correct.
2. **Reserve-then-apply crash safety (§4 step 6)** — ordering is insert → `persist()` → `Ok(Reserved)`, with in-memory rollback + `Err` on persist failure. `reserve_then_crash_before_apply_leaves_nonce_consumed` proves same-envelope retry is rejected and a fresh-nonce re-sign succeeds. Correct and load-bearing item satisfied.
3. **Dry-run exception** — the store has no dry-run concept; `reserve()` has no preview path that consumes. Policy stays in the #748 caller. Correct.
4. **Bounded retention** — `gc()` runs at the top of every `reserve()` (`retain(expires_at > now)`); burst test goes 500 → 1. Correct (aside from the flakiness above).
5. **Corrupt-store handling** — `open()` hard-errors on parse failure ("failed to parse nonce store …") and on version mismatch; never silently wipes. Test covers it. Correct.
6. **Reuse, not duplication** — `signerKeyId` is an opaque scope reusing `operator_key::fingerprint_hex` (#753, merged); no re-implementation.
7. **Scope discipline** — diff is exactly `operator_nonce.rs` (new) + one `pub mod` line in `lib.rs`. No submitAction / verify / apply logic. No new deps (anyhow/serde/serde_json only). Atomic write is 0600 temp + fsync + rename. Clean.

Once the test suite passes reliably under the default parallel runner, re-request review and I'll approve — the rest is ready.

rjwalters:
✅ **Approved!** (sweep re-judge after Doctor fix)

The Judge's substantive review passed all 7 dimensions — reserve-then-persist-then-return crash safety (§4.6/§9), restart durability, no dry-run reservation concept, GC-on-insert bounding, corrupt-store-is-hard-error (no silent wipe), signerKeyId reuse of `operator_key::fingerprint_hex`, and tight scope. The sole blocker was a flaky test: `tmp_store_path()` derived its dir from `SystemTime::now().as_nanos()`, so parallel test threads within one clock tick shared a directory and raced `persist()`'s rename. Fixed with a process-static `AtomicU64` counter (production code unaffected — a NonceStore is single-writer via `&mut`); verified 5/5 green under the default parallel runner.

All CI green (build, deny, rustfmt, benchmarks, fuzz-build); mergeStateStatus CLEAN. Consumed by #748's verifier step 6.

PR #769, opened 2026-07-11
fix(sync): re-enter catch-up when a below-hysteresis gap persists past gossip

### LINKED ISSUE #766: Liveness: chain halts permanently when the sole minter is one block behind (sync manager tolerates 1-block gap, SCP restarts into stale slot)
## Incident (2026-07-10, live testnet)

The chain halted at height 2884 for 25+ slots. A user's faucet payout sat in every node's mempool (`mempoolSize: 1` fleet-wide) with the wallet showing it pending and then dropping it.

Observed state:
- seed/seed2/eu/ap: height **2884**, slot index 2909, `scpSlotActive: false`
- faucet (the ONLY minter): height **2883**, wedged on slot 2884 with `slotStalled: true`, `synced: true`(!), `minerStalled: false`

## Root cause chain

1. The faucet node missed block 2884 (gossip miss / restart timing) and sat one block behind.
2. Its sync manager reports `synced: true` for a one-block gap, so it never requests block 2884 from peers (all 4 of whom have it).
3. As the only minter, it keeps mining on the stale parent 2883 — its minting txs are useless to the fleet, and the fleet has no other block producer. Slots drift (fleet reached 2909) with no block externalized. **Permanent halt.**
4. `systemctl restart botho` did NOT recover it: the node restarts, still doesn't fetch 2884 (gap ≤ 1 → "synced"), and re-enters SCP at slot 2884 while peers are at 2909.

## Recovery used

Enabled minting on seed2 (`--mint`; see companion issue about `config.minting.enabled` being dead). A second minter extends the real tip; the faucet then sees a ≥2-block gap, syncs, and rejoins.

## Suggested fixes (independent)

- **Sync**: a node should fetch the tip when any peer's tip exceeds ours, or at minimum a 1-block gap must not report `synced: true` while the node is actively running SCP for a slot its peers have already externalized.
- **SCP re-entry**: on startup, sync to the fleet tip BEFORE arming the miner/slot state (the restart log shows quorum reconfiguration at the stale slot immediately).
- **Ops posture**: run ≥2 standing minters so a single wedged minter cannot halt the chain (related: #427 solo-latch, #690 star wedge, and the faucet's 10k-BTH minting pause which makes the "sole minter" configuration even more fragile).

Live testnet currently has seed2 minting as the recovery measure — decide whether that stays.

rjwalters:
## Summary

Fixes the permanent-halt liveness bug in #766. On the live testnet the chain halted at height 2884: the sole minter (faucet) fell one block behind, its sync manager reported `synced: true` for the 1-block gap, so it never fetched block 2884 from the four peers that had it — and, as the only block producer, it kept minting a competing block on the stale parent 2883, which the fleet ignored. The gap could never close, so the chain stopped forever.

This PR implements the **sync-manager gap-tolerance fix** (bullet 1 of the issue, the root-cause "first domino"). See "Scope / follow-up" below for why the SCP-re-entry-ordering fix (bullet 2) is intentionally left out.

## Root cause

`ChainSyncManager` intentionally applies near-tip hysteresis (`SYNC_BEHIND_THRESHOLD = 10`): while `Synced`, a 1..=9 block lag is left to gossip rather than triggering a redundant historical download, because gossip normally delivers the next contiguous block and closes the gap. That assumption is correct while the chain advances — but it deadlocks the **sole minter**: the block it needs is never gossiped to it (it is producing a competing block on the stale parent), so the small gap never closes and `synced` stays `true` indefinitely.

## Changes

- `botho/src/network/sync.rs`:
  - New `small_gap_since_height: Option<u64>` field tracking the local height at which a below-hysteresis peer-ahead gap was first observed while `Synced`.
  - New `evaluate_persistent_small_gap()` helper: if, on the status-refresh cadence, a peer is still strictly ahead at the **same** local height as the previous refresh (gossip had a full `STATUS_REFRESH_INTERVAL` and advanced us nothing), the "gossip will close it" assumption is disproven, so re-enter `Downloading` and fetch from the ahead peer.
  - `set_local_height()` resets the tracker whenever our committed height advances, so a gap gossip **is** closing never trips the guard — no download thrash, the intentional hysteresis is preserved.
  - Wired into the `Synced` tick arm, gated on the same refresh boundary so the two observations that constitute "persistent" are a full refresh interval apart.
  - Three regression tests.

| A 1-block gap must not report `synced: true` while peers have externalized ahead | ✅ | `test_sole_minter_one_block_gap_reenters_catchup_when_gossip_stalls` asserts the node transitions to `Downloading` and `status_snapshot().synced == false` once the gap persists |
| Node fetches the tip block when a peer's tip exceeds ours and gossip fails to deliver | ✅ | Same test: re-enters `Downloading { target_height: 2884 }` against the ahead peer |
| Fix must not regress the intentional near-tip hysteresis (no download thrash) | ✅ | `test_small_gap_closing_via_gossip_does_not_thrash_into_download`: when gossip advances our height between refreshes, the tracker resets and the node stays `Synced` |
| Guard confined to the refresh cadence | ✅ | `test_persistent_small_gap_ignored_off_refresh_boundary`: ticks before the refresh interval do not advance persistence state |

## Test Plan

- `cargo test -p botho --lib network::sync` — 74 passed (incl. 3 new regression tests), 0 failed.
- `cargo test -p botho --test chain_sync_catchup_integration` — 10 passed, 0 failed.
- `cargo check -p botho` and `cargo clippy -p botho --lib` — no new warnings from this change.

## Scope / follow-up (not in this PR)

Per the issue's scope guidance, the two code fixes are separable and this PR lands the root-cause sync-manager fix well-tested. The remaining items are intentionally deferred:

- **SCP re-entry ordering on startup** (issue bullet 2): sync to the fleet tip before arming the miner/slot state. This touches consensus bootstrap in `commands/run.rs` with a large blast radius and is hard to test in isolation; the recovery comment on #766 further identified a related **slot-index derivation asymmetry** (restarted nodes derive slot from tip; running nodes advance by wall clock during a halt), so slot agreement should recover without operator-orchestrated fleet restarts. That is a distinct, larger fix and should be its own issue/PR.
- **Ops posture** (issue bullet 3): run ≥2 standing minters. This is infrastructure, not code (related: #427, #690).

Closes #766

### DIFF
--- a/botho/src/network/sync.rs
+++ b/botho/src/network/sync.rs
@@ -539,6 +539,26 @@ pub struct ChainSyncManager {
     /// penalised for a stale overlap after it resumes normal service or
     /// reconnects.
     peer_overlap_counts: HashMap<PeerId, u32>,
+    /// Local height at which a small (below-hysteresis) peer-ahead gap was
+    /// first observed while `Synced` (#766).
+    ///
+    /// The near-tip hysteresis ([`SYNC_BEHIND_THRESHOLD`]) assumes a small gap
+    /// is transient — gossip of the next contiguous block will close it — so it
+    /// deliberately does NOT re-enter catch-up for a 1..=9 block lag. That
+    /// assumption fails catastrophically for the *sole minter*: if it falls one
+    /// block behind, the block it needs is never gossiped to it (it is busy
+    /// minting a competing block on the stale parent), so the gap never closes
+    /// and the chain halts permanently (the #766 live-testnet incident).
+    ///
+    /// To break that deadlock without reintroducing thrash, we require a small
+    /// gap to *persist*: if we observe a peer still strictly ahead of us at the
+    /// *same* local height across a second status observation — i.e. gossip had
+    /// a full [`STATUS_REFRESH_INTERVAL`] to deliver the next block and did not
+    /// advance us at all — the "gossip will close it" assumption is disproven
+    /// and we re-enter catch-up via the sync state machine. This field records
+    /// the local height of the first such observation; it is cleared whenever
+    /// our height advances (gossip *did* make progress) or the gap closes.
+    small_gap_since_height: Option<u64>,
 }
 
 impl ChainSyncManager {
@@ -554,6 +574,7 @@ impl ChainSyncManager {
             reputation: ReputationManager::new(),
             last_status_refresh: Instant::now(),
             peer_overlap_counts: HashMap::new(),
+            small_gap_since_height: None,
         }
     }
 
@@ -617,6 +638,14 @@ impl ChainSyncManager {
     /// here, which — combined with a fixed-interval tick — re-anchored
     /// duplicate requests at the same start height and drove a retry loop.
     pub fn set_local_height(&mut self, height: u64) {
+        // If our committed height advanced, gossip (or a download) is making
+        // forward progress, so any previously observed "stuck" small gap is no
+        // longer stuck — reset the #766 persistence tracker. Only a gap that
+        // persists at the *same* local height across a status refresh proves
+        // gossip cannot close it.
+        if height > self.local_height {
+            self.small_gap_since_height = None;
+        }
         self.local_height = height;
     }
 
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Root-cause liveness fix for the #766 sole-minter permanent-halt. Reviewed with extra care given this is consensus-adjacent sync code.

## Correctness — persistence-guard state machine
- The two-observation design is sound: the first below-hysteresis `Synced` observation on a refresh boundary arms `small_gap_since_height = Some(local_height)` and falls through to the ordinary status re-poll (which resets `last_status_refresh`), so the two "persistent" observations are genuinely a full `STATUS_REFRESH_INTERVAL` apart. The second observation at the *same* local height proves gossip had its chance and delivered nothing, and re-enters `Downloading` against the ahead peer.
- No thrash: `set_local_height()` clears the tracker on any forward height progress (`height > self.local_height`), so a gap gossip *is* closing never trips the guard — the intentional near-tip hysteresis is preserved. Confirmed by `test_small_gap_closing_via_gossip_does_not_thrash_into_download`.
- Gap partitioning is clean: `>= SYNC_BEHIND_THRESHOLD` stays on the existing `on_status`/tick path; zero/negative gap resets; only the ambiguous `1..=SYNC_BEHIND_THRESHOLD` strictly-positive gap is resolved here.

## Edge cases checked
- **Stale/lying peer**: transitioning to `Downloading` against a peer that never serves the block cannot wedge permanently — `REQUEST_TIMEOUT` + `on_failure()` → `Failed` → backoff → `Discovery` recovers, same resilience the ordinary catch-up path relies on.
- **Peer churn**: `best_peer()` returning `None` resets the tracker via the `let-else`; a different still-ahead peer at the same height still fires correctly.
- **Reorg (height regression)**: tracker is not force-reset on a height decrease, but simply re-arms at the new local height — benign.
- **Determinism / hot path**: no new `unwrap`/`expect`/`panic`; `debug!` already imported; guard is gated purely on the deterministic refresh cadence.

## Regression tests pin the incident
- `test_sole_minter_one_block_gap_reenters_catchup_when_gossip_stalls` — reproduces the exact 2883→2884 sole-minter gap, asserts transition to `Downloading { target_height: 2884 }` and `status_snapshot().synced == false`.
- `test_small_gap_closing_via_gossip_does_not_thrash_into_download` — no regression of the hysteresis.
- `test_persistent_small_gap_ignored_off_refresh_boundary` — guard confined to the refresh cadence.

## Test Execution
- `cargo test -p botho --lib network::sync` — ✅ 74 passed, 0 failed (incl. 3 new).
- `cargo test -p botho --test chain_sync_catchup_integration` — ✅ 10 passed, 0 failed.
- `cargo clippy -p botho --lib` — ✅ no new warnings from this change (only the pre-existing `PeerStatus.tip_hash` dead-code note at sync.rs:433, unrelated).

## CI + merge state
- `gh pr checks` — ✅ all required checks pass (build/deny/rustfmt/fuzz-guard/benchmarks green; fuzz jobs `skipping` as expected).
- `mergeStateStatus` — ✅ CLEAN.

Scope is correct: `Closes #766`, and deferring the SCP re-entry ordering fix (issue bullet 2) to a separate issue is the right call given its consensus-bootstrap blast radius. Well-scoped, well-tested root-cause fix. LGTM.

PR #771, opened 2026-07-11
test(operator): add cross-language canonicalization drift guard

### LINKED ISSUE #759: ci: cross-language canonicalization drift guard for operator-action envelopes (#709 follow-up)
Follow-up from the #751 (PR #758) Judge review. Part of #441 P4 / the #709 operator-signed quorum-curation feature.

## Problem

The browser operator dashboard signs quorum-curation envelopes whose canonical bytes MUST byte-match the node's Rust canonicalization (`botho/src/operator_action.rs`) or the node's parse-after-verify signature check fails in production. The JS side is pinned by a committed fixture (`web/packages/features/src/operator/fixtures/operator-action-fixtures.json`) generated from the node's Rust test helpers, and `action-envelope.test.ts` asserts byte-equality of the canonical string, Ed25519 signature, signerKeyId, and envelopeHash.

**The gap:** the fixture is a static snapshot. If the node's canonicalization ever changes (field order, optional-field handling, integer encoding), the JS test keeps passing against the stale fixture while real signatures silently break at the node. Nothing fails a gate.

## Ask

Add a CI check (or a Rust test) that regenerates the fixture from the current node canonicalization and diffs it against the committed fixture, so drift fails a gate rather than surfacing as a production signature-verification failure.

## Root cause / technical context

- The fixture (`operator-action-fixtures.json`) is hand-generated: its own `_comment` field says "To regenerate: temporarily print these from a Rust test in `operator_action.rs` using the existing `build_signed`/`sign_str` test helpers (seed `[1;32]`)." That is a manual, undocumented, easy-to-skip step — nothing enforces it actually happened after a canonicalization change.
- The canonical-bytes format is currently duplicated in two places with no shared source of truth:
  - Rust: the `#[cfg(test)]` helper `build_signed`/`sign_str` in `botho/src/operator_action.rs` (lines ~1053-1097) hand-builds the canonical string via a `format!` with a hardcoded field order. This is a **test helper**, not the production canonicalization path — production canonicalization doesn't exist as a standalone function because the node never re-serializes; it only verifies over received bytes (`SignedEnvelope::verify_and_parse`, parse-after-verify, finding 1). So there is no single "canonicalize" function to unit-test against on the Rust side — the closest proxy is this test helper's hardcoded format string plus the `KNOWN_KEYS` allowlist in `parse_canonical_envelope`.
  - TypeScript: `canonicalizeEnvelope()` in `web/packages/features/src/operator/action-envelope.ts` (lines 133-191) is a genuine canonicalization function (sorts keys, integers-only, no whitespace) and is the sole producer of signed bytes on the browser side.
- Because the Rust side has no canonicalization *function* (by design — it only verifies received bytes), "regenerate the fixture from the node" really means: exercise the Rust test helper's format string against a range of envelope shapes (including the `acknowledgeDegenerate` optional-field case and both `dryRun` values) and confirm the **existing committed JSON fixture** still matches byte-for-byte. If someone changes the KNOWN_KEYS set, the field order in the test helper, or the JSON encoding of any field (e.g. switches an integer to be emitted as a JSON number vs. string), this check should fail loudly rather than silently pass on the JS side against a stale fixture.

## Implementation guidance

Two viable approaches; either satisfies the ask, pick based on where you want the failure signal to appear:

### Option A (recommended): Rust test that pins the fixture JSON itself
Add a `#[test]` in `botho/src/operator_action.rs`'s `#[cfg(test)] mod tests` that:
1. Loads the committed fixture via `include_str!("operator/fixtures/operator-action-fixtures.json")` (adjust relative path — the fixture lives under `web/packages/features/src/operator/fixtures/`, so this is a cross-directory `include_str!`; use a path relative to `botho/src/operator_action.rs`, e.g. `../../web/packages/features/src/operator/fixtures/operator-action-fixtures.json`, and parse with `serde_json::from_str::<serde_json::Value>(..)` — no new crate dependency needed, `serde_json` is already a dependency).
2. For each entry in the fixture's `envelopes` array, reconstructs the same envelope fields via `build_signed`/`sign_str` (or a small helper factored out of them) using the SAME seed key (`[1u8; 32]`) and the fixture's `fields` values.
3. Asserts the regenerated `canonical`, `signature`, and `envelopeHash` byte-for-byte equal the committed fixture's values (envelopeHash can be computed via the existing `blake2b_256_hex` helper in `operator_key.rs`).
4. This test fails immediately (in `cargo test --workspace`, already gated in CI via `workspace-build.yml`) if the Rust-side canonical byte production for any fixture entry drifts from what's committed — no separate CI step needed, since `cargo test --workspace` already runs on every PR.

This keeps the check in Rust, next to the canonicalization logic and the KNOWN_KEYS allowlist, and requires zero new CI wiring (it's just a new `#[test]` that already runs under the existing `cargo test --workspace` step in `.github/workflows/workspace-build.yml`).

### Option B: CI step that regenerates + diffs
Add a small Rust `bin` (or a `cargo test -- --ignored` helper invoked via a script) that emits a fresh `operator-action-fixtures.json` from the Rust side, then a CI step runs it and does `git diff --exit-code` against the committed file. This is closer to "regenerate and diff" as literally described in the ask, but requires new CI wiring (a workflow step in `workspace-build.yml` or `web-ci.yml`) and a small stand-alone generator binary/script — more moving parts than Option A for the same guarantee.

**Recommendation: Option A.** It gives the same drift-detection guarantee with no new CI surface, reuses the existing `build_signed`/`sign_str` test helpers already in the file, and runs on every PR for free via the existing `cargo test --workspace` gate.

Either way, **document the regeneration recipe** in a comment at the top of the fixture JSON (it already has a `_comment` block — extend it) and/or in `docs/security/quorum-write-path.md` §3, describing exactly how to regenerate the fixture by hand if a canonicalization change is intentional (e.g. a new v2 envelope field).

## Affected Files

- `botho/src/operator_action.rs` - add the new fixture-pinning `#[test]` (and optionally factor a small helper out of `build_signed`/`sign_str` if reuse needs adjusting); lines 1034-1283+ contain the existing `#[cfg(test)] mod tests` block and `build_signed`/`sign_str` helpers to reuse.
- `web/packages/features/src/operator/fixtures/operator-action-fixtures.json` - extend the `_comment` block with the concrete regeneration recipe once the new test exists; no data changes expected unless drift is found.
- `web/packages/features/src/operator/action-envelope.test.ts` - no changes expected (reference only — this is the consuming test the new Rust test protects).
- `docs/security/quorum-write-path.md` - optionally add a one-line pointer in §3 to the new Rust drift-guard test, if the doc's canonicalization section is touched.

## Test Plan

- [ ] Automated test: new Rust `#[test]` in `botho/src/operator_action.rs` fails if the fixture JSON is manually edited to diverge from what the seed-`[1;32]` test helpers produce (verify by locally mutating one fixture `canonical` byte and confirming the new test fails).
- [ ] Automated test: `cargo test --package botho` (already run by CI in `workspace-build.yml`) passes with the new test in place, proving no current drift exists.
- [ ] Manual verification: intentionally change field order or the `KNOWN_KEYS` set in `operator_action.rs` locally, confirm the new test fails (proves the guard actually detects real canonicalization drift, not just JSON parse errors).
- [ ] Regression: `pnpm test:run` in `web/` (existing `action-envelope.test.ts`) still passes unchanged — no JS-side changes expected in this issue's scope.

## Out of scope
- Actually changing the node's canonicalization or adding a v2 envelope field.
- Building a full bidirectional fixture-generation pipeline (e.g. a `cargo run --bin` that overwrites the JSON automatically in CI) — Option A only needs to detect drift, not auto-regenerate.

rjwalters:
## Summary

The browser operator dashboard signs quorum-curation envelopes whose canonical bytes MUST byte-match the node's Rust canonicalization, or the node's parse-after-verify signature check fails in production (finding 1, the #1 failure mode). The JS side is pinned by a committed fixture (`operator-action-fixtures.json`) and `action-envelope.test.ts`. That fixture is a hand-maintained snapshot: if the node's canonicalization drifts, the JS test keeps passing against the stale fixture while real signatures silently break at the node, with nothing failing a gate.

This PR closes that gap from the Rust side with a new `#[test]` (Option A from the issue) that pins the same fixture and asserts the node's canonical byte production still matches it byte-for-byte. It runs under the existing `cargo test --workspace` CI gate — no new CI wiring.

## Changes

- `botho/src/operator_action.rs`: new `fixture_matches_node_canonicalization_no_drift` test. It `include_str!`s the committed fixture, and for each envelope reconstructs the canonical bytes from the seed `[1u8; 32]` via the existing `build_signed`/`sign_str` helpers, then asserts byte-equality of the canonical string, the Ed25519 signature, and the `envelopeHash` (blake2b-256, when the entry pins one). It also guards the shared constants (`signingKeySeed`, `publicKeyHex`, `signerKeyId`, `domainSeparator`). Adds a `canonical_params_json` helper (serde_json `Value::Object` is `BTreeMap`-backed without `preserve_order`, so `to_string` already emits sorted-key, no-whitespace, integer JSON).
- `web/packages/features/src/operator/fixtures/operator-action-fixtures.json`: extend the `_comment` block with a DRIFT GUARD note and a concrete regeneration recipe. No data changes (no drift found).
- `docs/security/quorum-write-path.md`: one paragraph in §3 pointing at the fixture + the new Rust drift-guard test.

| New Rust test fails if the fixture JSON is manually edited to diverge | ✅ | Locally flipped one byte in a fixture `canonical` string → test FAILED with "canonical bytes drifted for fixture entry \"pin_member dryRun\"". Reverted. |
| `cargo test --package botho` passes with the new test (no current drift) | ✅ | `cargo test --package botho --lib operator_action` → 28 passed, 0 failed (incl. the new test). |
| Guard detects real canonicalization drift (field order / KNOWN_KEYS) | ✅ | Locally reordered fields in the Rust `build_signed` format string → test FAILED with "canonical bytes drifted". Reverted; test passes again. |
| JS-side `action-envelope.test.ts` unchanged / still valid | ✅ | No JS source changed; only the fixture `_comment` (doc-only) touched. Fixture data byte-identical to origin/main. |

## Test Plan

- `cargo test --package botho --lib operator_action`: 28 passed, 0 failed.
- Drift detection verified two ways (fixture-side byte edit AND node-side field-order edit both fail the guard; both reverted, not committed).
- `cargo fmt --package botho -- --check`: clean.
- Pre-existing clippy `println!` errors in `botho/src/network/transport/fingerprint.rs` are unrelated to this change (that file is not in the diff; they exist on `origin/main`) and are left untouched per scope discipline.

Closes #759


rjwalters:
✅ **Approved!**

Solid, well-scoped test-hardening PR that closes the cross-language canonicalization drift gap from the Rust side (#759, Option A). Reviewed for the four flagged concerns and all hold up.

## Test Scoping
- **Strategy**: `cargo test --package botho --lib operator_action` (scoped to the changed crate/module)
- **Result**: 28 passed, 0 failed — including the new `fixture_matches_node_canonicalization_no_drift`.
- `cargo fmt --package botho -- --check`: clean.

## Verification of flagged concerns

**1. Does `canonical_params_json` / `build_signed` mirror the PRODUCTION path, or is it a free-floating reimplementation?**
It is a test-side reimplementation of the canonical byte format (a `format!` string in `build_signed`), which is unavoidable here: the node is verify-only by design — `SignedEnvelope::verify_and_parse` verifies the Ed25519 signature over the *received* bytes and then `parse_canonical_envelope`s them; there is no standalone production canonicalize function to call. Critically, `build_signed` is NOT free-floating: the existing suite (`apply_to_pin_unpin_setcap_roundtrip`, the `finding1_*`/`finding2_*` tests, etc.) round-trips `build_signed(...)` output through the production `verify_and_parse` → `parse_canonical_envelope` → `KNOWN_KEYS` path. So the anchoring chain is complete: JS signer ↔ fixture (`action-envelope.test.ts`) ↔ `build_signed` (this new test, all entries) ↔ production parser (existing round-trip tests). A production `KNOWN_KEYS`/parse drift would surface as either a round-trip test failure or a `build_signed` edit that breaks this fixture test. `canonical_params_json`'s claim that `serde_json` emits sorted-key/no-whitespace/integer JSON (no `preserve_order`) is correct and matched the fixture byte-for-byte at runtime.

**2. `include_str!` path correctness (fixture lives under `web/`, outside the crate).**
Correct. `include_str!` resolves relative to the source file `botho/src/operator_action.rs`; `../../web/...` walks `botho/src` → repo root → `web/`. Verified the path resolves and the test compiled + loaded the fixture.

**3. Does the test cover ALL fixture entries?**
Yes — `for entry in envelopes` iterates all 3 (`pin_member dryRun`, `set_max_auto_members apply`, `unpin_member acknowledgeDegenerate`), plus a non-empty guard. Optionality is handled correctly: `acknowledgeDegenerate` present on only the third entry (via `.get(...).and_then`), and `envelopeHash` asserted only when pinned (present on 2 of 3). Shared constants (`signingKeySeed`, `publicKeyHex`, `signerKeyId`, `domainSeparator`) get their own targeted assertions.

**4. Doc / `_comment` regeneration recipe accuracy.**
Accurate. `build_signed`/`sign_str`/`KNOWN_KEYS` all exist as named; `blake2b_256_hex` exists in `operator_key.rs` and is what the test calls (hashing the bare canonical bytes, matching the fixture's documented `blake2b-256(canonical_bytes)`); `pnpm test:run` in `web/` maps to `vitest run`. `docs/security/quorum-write-path.md` §3 points at both sides of the guard.

## Drift detection (independently reproduced)
Flipped `"v":1` → `"v":2` in one fixture `canonical` string → the new test FAILED with a clear `canonical bytes drifted for fixture entry "pin_member dryRun"` message. Reverted; fixture confirmed byte-identical to `origin/main` (only `_comment` changed, verified via structural JSON compare).

## CI
All required checks pass (`cargo build --workspace`, `typecheck + vitest + worker dry-run`, `cargo deny`, `rustfmt`, benchmarks, fuzz build guard). Merge state CLEAN. The pre-existing clippy `println!` findings in `botho/src/network/transport/fingerprint.rs` are genuinely out of scope (that file is not in the diff; 10 `println!` already on `origin/main`) and clippy is not a required CI gate here.

PR description uses `Closes #759`. Nice work.

Issue #436, opened 2026-06-19
consensus: Option-C anchor idle-gate latches on un-completable nomination state (strands joiner, blocks #396)


rjwalters:
## Summary
The LAST blocker for a reliably-green #396 capstone. Empirically characterized (9-run study): the #396 3-node soak's terminal step (joiner C mining a block accepted by all three) is a **case-2 real residual bug** (~2/6 miss rate), NOT test-window/probability. In the misses, **C's SCP slot wedges immediately after the join handoff and C never externalizes a single slot** — it reaches the network height only via passive block-sync, so it can never earn a coinbase.

## Mechanism (measured, high confidence)
The #430 Option-C anchor (`anchor_scp_slot_to_peer` in `botho/src/consensus/service.rs` ~1183-1237) is **idle-gated**: it refuses to anchor C forward when the current slot holds "in-flight protocol state". The race:
1. After catch-up, C anchors forward and proposes its own coinbase → its slot gains bare nomination state (`num_voted_nominated > 0`, phase still `NominatePrepare`, `bN == 0`).
2. A+B have already externalized that slot and moved on — they never re-send messages for C's stranded slot, so C's quorum on it can never form.
3. The idle-gate then refuses to anchor C forward **forever** (observed logging "Not anchoring SCP slot to peer's live slot: current slot holds in-flight protocol state" 700-950x in the misses).
4. C discards every live A/B message as a "future slot" (`consensus/scp/src/node/node_impl.rs:224-235`) and never externalizes again → permanent passive follower → 0 coinbases.

In the WINS, the join lands C in lockstep before it strands un-completable state (~0 deferrals, normal externalize, C wins). It's a join-handoff race.

## The fix (localized, safety-bounded)
Make the idle-gate **not latch on un-completable NOMINATION-only state**. The gate's real job is to never abandon **confirmed-commit/ballot** state (dropping that could fork — re-opening a committed value). But bare, un-confirmed NOMINATION state on a slot the network has **provably already externalized** is safe to abandon (nothing is committed; the network already decided that slot).

Refine `anchor_scp_slot_to_peer` (and the shared guard in `sync_scp_slot_to_chain`): allow the forward anchor when **the peer's live slot is ahead of the current slot (network has demonstrably moved past it) AND the current slot has NO ballot/commit state** (phase still nominate, `bN == 0`) — even if nomination votes are present. Only BLOCK the anchor when there is genuine ballot/commit state that must be preserved.
[...]

PR #780, opened 2026-07-12
fix(consensus): gate minting/nomination on initial fleet-tip sync at startup

### LINKED ISSUE #770: SCP re-entry: sync to fleet tip before arming miner/slot state on startup (#766 follow-up)
## Original Issue

Follow-up to #766, split out of PR #769 (which fixed the root-cause sync-manager gap tolerance).

Two deferred items from the #766 incident (chain halt at height 2884, 2026-07-10):

1. **SCP re-entry ordering**: on startup, the node should sync to the fleet tip BEFORE arming the miner/slot state. The incident restart log shows quorum reconfiguration at the stale slot immediately — the node re-entered SCP at slot 2884 while peers were at 2909. This touches consensus bootstrap in `botho/src/commands/run.rs` (large blast radius, hard to unit-test — the reason it was deferred from #769).

2. **Slot-index derivation asymmetry** (surfaced in the #766 recovery): restarted nodes derive the slot index from their tip, while running nodes advance the slot by wall clock during a halt. The two can diverge by many slots during an outage; decide a single derivation rule.

Ops-posture note (not code, tracked here for context): run ≥2 standing minters so a single wedged minter cannot halt the chain (related: #427 solo-latch, #690 star wedge, faucet 10k-BTH minting pause).

PR #769's persistence-guard fix means a wedged sole minter now recovers via re-entering catch-up, but the startup ordering above remains a correctness/liveness gap on its own.

---

### Root cause (confirmed in code, 2026-07-11)

`botho/src/commands/run.rs::run_async` constructs consensus and arms minting **before** any block-sync has happened, and never re-checks sync status afterward:

- `local_height` is read once from the ledger's on-disk `chain_state` (`run.rs:648-699`).
- `ConsensusService::new(..., chain_state)` (`run.rs:711-716`) seeds the SCP slot as `initial_height + 1` (`consensus/service.rs` — see `ConsensusService::new`, and the doc comment on `sync_scp_slot_to_chain` at `consensus/service.rs:1380-1389` which states this explicitly).
- `ChainSyncManager::new(local_height)` is constructed *after* consensus (`run.rs:745`), starting in `SyncState::Discovery` (`network/sync.rs:566-579`), i.e. explicitly not-yet-synced (`is_synced()` is `false` until the state machine reaches `Synced`, `network/sync.rs:597-599`).
- Minting is armed purely from peer/quorum reachability via `check_minting_eligibility()` (`run.rs:152-182`, called at `run.rs:386`, `1030`, `1113`, `1761`, `2175`) — **it never consults `sync_manager.is_synced()`**. If the configured quorum is already reachable at startup (e.g. peers dial back in quickly), minting arms immediately at the stale slot.
- The only place the SCP slot is advanced to match a resynced chain height is `ConsensusService::sync_scp_slot_to_chain()` (`consensus/service.rs:1406-1466`), which is called reactively, *inside* the main event loop, only after a sync batch has actually been applied (`run.rs:1246`). There is no call to this — or any equivalent gate — between process start and the first SCP tick/minting-eligibility check.

Net effect: a node that restarts after falling behind (the #766 scenario) can rejoin SCP and start minting/balloting at its stale local slot in the same startup pass in which peers are discovered, racing the sync manager's catch-up download. This is the "quorum reconfiguration at the stale slot immediately" behavior observed in the #766 incident log.

### Slot-derivation asymmetry (item 2)

Two different mechanisms move the SCP slot index, and they are not reconciled at startup:

- **Startup / catch-up path**: `sync_scp_slot_to_chain(chain_height)` sets slot = `chain_height + 1`, derived from the *local* ledger height (`consensus/service.rs:1406-1466`).
- **Live/running path**: `ConsensusService::advance_slot()` (`consensus/service.rs:1341`, called at `run.rs:1707` on each `SlotExternalized` event) advances the slot by one externalization at a time, driven by wall-clock SCP rounds, independent of ledger height until a block is actually applied.

During a halt, peers keep advancing slots by ballot rounds (many of which may not externalize a value, e.g. empty/no-op slots) while a restarted node's slot is pinned to `chain_height + 1`. The two counters are not guaranteed to agree even once ledger heights match, because "slot index" and "block height" are not 1:1 in general (a slot can externalize with no minting tx, or SCP can burn slots on failed rounds). This issue should pick one canonical rule (recommended: derive from local chain height at rejoin, matching `sync_scp_slot_to_chain`'s existing convention, and let normal ballot exchange with peers repair any residual offset once the node is participating at the correct height) and make it explicit/testable, rather than leaving two independent derivations that happen to agree only when no halt has occurred.

### Recommended approach: startup-only ordering change (no consensus-message changes)

Consistent with the project's no-hard-forks philosophy, prefer a **startup-sequencing fix** over anything that touches SCP message formats, validity functions, or ballot protocol — those are consensus-critical and any behavior change there risks a fork between old/new binaries. The ordering bug is purely local bootstrap sequencing and can be fixed without touching the wire protocol:

1. **Block minting/SCP participation on initial sync completion.** Add an explicit startup gate: after constructing `sync_manager` and `consensus`, do not set `minting_enabled = true` (nor let `check_minting_eligibility` arm minting) until `sync_manager.is_synced()` is true. This likely means:
   - Track a new `initial_sync_complete: bool` (starts `false`) alongside `minting_enabled`.
   - In `check_minting_eligibility` (or its call sites), additionally require `initial_sync_complete` before returning `can_mint_now = true`.
   - Flip `initial_sync_complete = true` the first time `sync_manager.on_blocks_added(...)` / `status_snapshot()` reports `Synced` (or immediately if the node starts already at tip / `SYNC_BEHIND_THRESHOLD` — a node that is already caught up should not wait).
2. **Call `sync_scp_slot_to_chain` (or equivalent) once sync reaches the tip, before the node casts its first ballot**, not only reactively after mid-flight catch-up batches. This closes the gap where a node that catches up entirely during the pre-loop "wait for peers" window (`run.rs:332-346`) or via the first few sync batches could still start balloting at the old slot before the existing reactive call fires.
3. **Keep SCP message handling, quorum-set construction, and validity functions untouched.** The fix should be entirely about *when* `minting_enabled` flips and *when* the slot index is synced relative to the sync-manager state — not about what gets sent on the wire or how peers validate values. This keeps the change low-risk and rollback-friendly (a single boolean gate + one moved/added function call), and avoids needing a protocol version bump.
4. **Do not block SCP message *reception*/relay** — a resyncing node should still gossip/relay and track peer messages so it isn't fully deaf to the network (this matches existing "future slot" discard behavior, `consensus/service.rs:1380-1389`); only *proposing/nominating/minting* needs to be gated on sync completion.

### Why this was deferred from #769

PR #769 fixed the sync-manager side only (gap-tolerance false-positive on `synced: true`), explicitly deferring this ordering fix because it touches consensus bootstrap in `run.rs` with a large blast radius and is hard to unit-test in isolation (it depends on the interleaving of async event-loop arms). This issue exists to scope and implement that follow-up carefully, with narrow, targeted unit tests on the extracted decision functions (similar to how `check_minting_eligibility` and `should_redial_bootstrap_peers` are already unit-tested as pure functions independent of the event loop).

### Risks

- **Consensus-adjacent, but not consensus-protocol-changing**: the recommended fix only changes *local* startup sequencing (when this node starts proposing/minting), not what is sent or how it's validated — so it should not require a hard fork. Still, any change near `ConsensusService`/SCP arming needs careful review given the no-hard-forks philosophy (see project memory: never want to fork Botho pre-mainnet).
- **Deadlock/liveness risk if the new gate is too strict**: if `initial_sync_complete` never flips true (e.g. no peers respond, or the sync manager gets stuck in `Discovery`/`Downloading`), a node could refuse to mint forever. Must ensure a genuine solo node (no peers, `min_peers_to_wait == 0`) or a node already at tip is not blocked — mirror the existing "genuine lone node still mints solo" carve-out used for the quorum-set seed (`run.rs:663-694`).
- **Interaction with `min_peers_to_wait` and the participation gate** (`consensus.set_min_peers`, `run.rs:718-729`): the new sync-completion gate should be additive to, not a replacement for, the existing peer-count gate — a node could satisfy peer-count quorum while still behind on blocks (exactly the #766 scenario), which is the case this issue targets.
- **Slot-derivation convention change (item 2)** has a smaller blast radius than the ordering fix, but changing which value seeds the SCP slot on rejoin could interact with `sync_scp_slot_to_chain`'s existing ballot/commit-state guard (`consensus/service.rs:1420-1438`) — must confirm the new startup-time call site is only exercised on a genuinely idle slot (mirrors the existing safety comment) to avoid discarding in-flight commit state.
- **Testability**: the full event-loop interleaving is hard to unit test directly (this is *why* #769 deferred this bullet). Prefer extracting the gating decision (e.g. a `should_arm_minting(want_to_mint, quorum_ok, sync_complete) -> bool` pure function, analogous to `check_minting_eligibility`) so the ordering logic itself is unit-testable without spinning up the full async loop.
- **Live testnet impact**: this is live-network-facing code (5-node testnet, protocol 4.0.0). Land behind careful manual verification against a local multi-node harness before relying on it to prevent a repeat of the 2026-07-10 halt.

### Acceptance Criteria

- [ ] A restarted node that is behind the fleet tip does not arm minting (`minting_enabled` never becomes `true` via `check_minting_eligibility` or any call site) until the sync manager reports `Synced` (`sync_manager.is_synced() == true`), even if peer-count quorum is already satisfied.
- [ ] A restarted node that is behind the fleet tip does not participate in SCP nomination/balloting for its stale local slot before syncing; the SCP slot index is reconciled to the synced chain height (via `sync_scp_slot_to_chain` or equivalent) before the node casts its first ballot in a new session.
- [ ] A node that starts already at (or within `SYNC_BEHIND_THRESHOLD` of) the fleet tip is not delayed in arming minting — no regression for the common "clean restart, no significant gap" case.
- [ ] A genuine solo node (`min_peers_to_wait == 0`, no peers) is unaffected — it can still mint immediately per current behavior.
- [ ] No changes to SCP message formats, the validity function, or ballot/nomination protocol — diff is scoped to `botho/src/commands/run.rs` startup sequencing (and possibly a small `consensus/service.rs` helper), not the wire protocol.
- [ ] The slot-derivation rule (item 2) is documented as a single canonical rule with a code comment and/or doc update, and covered by a unit test asserting the rule holds across a simulated halt-and-restart height/slot divergence.

### Affected Files

- `botho/src/commands/run.rs` - Add a sync-completion gate before/alongside `check_minting_eligibility` (lines 152-182 and call sites at 386, 1030, 1113, 1761, 2175); ensure `sync_scp_slot_to_chain` (or a startup-time equivalent) runs before the first ballot is cast, not only reactively inside the sync-batch-applied arm (line 1246).
- `botho/src/consensus/service.rs` - Possible small helper/extraction to make the slot-sync-before-arming rule explicit and testable (near `sync_scp_slot_to_chain`, line 1406, and `current_slot`, line 765); no consensus-message or validity-function changes.
- `botho/src/network/sync.rs` - Reference only; `is_synced()` (line 597) and `status_snapshot()` (line 608) are the existing primitives to consult, no expected changes here.

### Test Plan

- [ ] Unit test: extract the minting-arm decision (peer-quorum-eligible AND sync-complete) into a pure function and test it directly for the matrix: {quorum ok/not ok} x {synced/not synced} x {solo node / federated node} — analogous to the existing `check_minting_eligibility` / `should_redial_bootstrap_peers` unit tests in `run.rs`.
- [ ] Unit test: simulate a restart where local height is behind the fleet tip; assert the SCP slot is reconciled to the synced height (not the stale local height) before any ballot is cast, using the existing `sync_scp_slot_to_chain` test patterns in `consensus/service.rs` (e.g. around lines 2785-2818).
- [ ] Unit test: assert a node already at/near tip (gap <= `SYNC_BEHIND_THRESHOLD`) arms minting without waiting on an explicit sync-complete transition (no regression to normal restart behavior).
- [ ] Unit test: assert a genuine solo node (no peers, `min_peers_to_wait == 0`) still mints immediately, unaffected by the new gate.
- [ ] Manual/integration verification: reproduce a scaled-down version of the #766 scenario in a local multi-node harness (one node artificially held behind via delayed sync response, then restarted) and confirm it does not re-enter SCP/mint at the stale slot before catching up.
- [ ] Regression check: run existing `consensus/service.rs` and `network/sync.rs` test suites (`cargo test -p botho`) to confirm no existing SCP/sync tests break.

rjwalters:
## Summary

A node that restarts behind the fleet tip could re-enter SCP and start minting/balloting at its **stale local slot** as soon as peer-count quorum was reachable, racing the catch-up block download. This is the #766 live-testnet halt: the chain wedged at height 2884 while peers balloted 2909, and the restart log showed "quorum reconfiguration at the stale slot immediately."

This PR adds a **startup-only sync gate**: a peer-expecting node withholds minting and SCP nomination/balloting until the sync manager reports `Synced`, then reconciles the SCP slot to the synced chain height **before casting its first ballot**. The gate is **additive** to the existing peer-count participation gate (#428), not a replacement — a node can satisfy peer quorum while still behind on blocks, which is exactly the #766 case.

Per the curator plan and the project no-hard-forks policy, this changes **only local startup sequencing** — no SCP message formats, validity function, or ballot/nomination protocol are touched.

## Changes

- **`botho/src/commands/run.rs`**
  - New pure decision function `should_arm_minting(quorum_ok, initial_sync_complete)` — unit-testable in isolation like `check_minting_eligibility` / `should_redial_bootstrap_peers`.
  - Track `initial_sync_complete`, seeded `true` for genuine solo nodes (`min_peers == 0`) and flipped `true` the first time `sync_manager.is_synced()` becomes true.
  - Gate every `minting_enabled = true` call site (startup, peer-connect, faucet-resume) through `should_arm_minting`.
  - On the first `Discovery/Downloading -> Synced` transition, call `sync_scp_slot_to_chain(chain_height)` before the first ballot and arm any minting that was deferred purely on the sync gate. This is the startup-side counterpart to the existing reactive call in the block-applied arm (which only fires after a mid-flight catch-up batch).
- **`botho/src/consensus/service.rs`**
  - New `initial_sync_complete` field (default `true`, so solo/dev/test nodes are unaffected) gating `should_propose_this_round`, plus `set_initial_sync_complete` / `initial_sync_complete` accessors. The `min_peers == 0` solo carve-out short-circuits **before** the sync gate, so the gate can never deadlock a solo minter.
  - Documented the canonical slot-derivation rule on `sync_scp_slot_to_chain`: on rejoin the SCP slot is always `local chain height + 1`, never a peer's drifted wall-clock slot counter (issue item 2).

| Restarted node behind tip does not arm minting until `is_synced()`, even with peer quorum | ✅ | `should_arm_minting` gates all call sites; `withholds_minting_when_behind_tip_despite_quorum` unit test |
| Restarted node does not nominate/ballot at stale slot; slot reconciled before first ballot | ✅ | `should_propose_this_round` gated on `initial_sync_complete`; slot reconciled on the sync-tick `Synced` transition; `startup_sync_gate_withholds_proposing_until_synced` test |
| Node at/near tip not delayed (no clean-restart regression) | ✅ | Reaches `Synced` after first status round (SYNC_INITIAL_GAP=1); `startup_sync_gate_does_not_delay_node_already_at_tip` test |
| Genuine solo node (`min_peers == 0`) unaffected, mints immediately | ✅ | Seeded `true`; `min_peers == 0` carve-out short-circuits before the gate; `startup_sync_gate_never_blocks_solo_node` test |
| No SCP message/validity/protocol changes; diff scoped to `run.rs` + small `service.rs` helper | ✅ | Diff touches only startup sequencing + a local bool gate; wire protocol untouched |
| Slot-derivation rule documented as single canonical rule + covered by a halt/restart unit test | ✅ | Doc comment on `sync_scp_slot_to_chain`; `slot_derivation_is_canonical_local_height_plus_one_across_halt` test |

## Test Plan

- `should_arm_minting`: exhaustive 2x2 matrix (`arm_minting_matrix_is_pure_conjunction`) plus named cases for the #766 condition and the solo/quorum carve-outs.
- SCP proposer: withheld-until-synced (`startup_sync_gate_withholds_proposing_until_synced`), solo carve-out with forced-false flag (`startup_sync_gate_never_blocks_solo_node`), no-delay-at-tip (`startup_sync_gate_does_not_delay_node_already_at_tip`).
- Canonical slot derivation across a simulated halt/restart height/slot divergence (2884 -> 2909).
- Full suite: `cargo test -p botho --lib` → **1240 passed, 0 failed**; consensus (33), run (36), sync (74) modules all green. `cargo check`, `cargo fmt`, and `cargo clippy` introduce no new warnings.

Manual multi-node harness verification of a scaled-down #766 reproduction is recommended before relying on this on the live testnet (noted as manual/integration in the issue test plan; not automatable in-tree).

Closes #770

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -181,6 +181,35 @@ fn check_minting_eligibility(
     )
 }
 
+/// Decide whether minting may be armed, combining the peer/quorum gate with the
+/// startup sync gate (issue #770).
+///
+/// This is the pure decision function behind every `minting_enabled = true`
+/// transition on the run path. It is deliberately kept free of `Config`/network
+/// types so it can be unit-tested exhaustively over the
+/// {quorum ok/not ok} x {synced/not synced} x {solo / federated} matrix,
+/// mirroring how [`check_minting_eligibility`] and
+/// [`should_redial_bootstrap_peers`] are already unit-tested independent of the
+/// async event loop.
+///
+/// The startup sync gate is **additive** to the existing peer/quorum gate, not
+/// a replacement: a restarted node can satisfy peer-count quorum while still
+/// being behind on blocks (exactly the #766 scenario — quorum reachable at the
+/// stale slot 2884 while peers balloted 2909). Both must hold before minting
+/// arms.
+///
+/// # Parameters
+/// - `quorum_ok`: the peer/quorum reachability result from
+///   [`check_minting_eligibility`] (which already folds in `want_to_mint`).
+/// - `initial_sync_complete`: `true` once the node has caught up to the fleet
+///   tip this session. For a genuine solo node (no peers, `min_peers == 0`) or
+///   a node that boots already at/near tip, the run loop seeds this `true`, so
+///   the gate never delays them — no regression for the common clean-restart or
+///   solo cases.
+fn should_arm_minting(quorum_ok: bool, initial_sync_complete: bool) -> bool {
+    quorum_ok && initial_sync_complete
+}
+
 /// Build the consensus config, honoring an optional test-only fixed-timing
 /// override.
 ///
@@ -744,6 +773,25 @@ async fn run_async(mut config: Config, config_path: &Path, mint: bool) -> Result
     // requesting the missing block range and applying it sequentially.
     let mut sync_manager = ChainSyncManager::new(local_height);
 
+    // Startup sync gate (issue #770): whether this node has caught up to the
+    // fleet tip since the process started. Until this is true, a node that
+    // expects peers must NOT arm minting or nominate at its stale local slot,
+    // even if peer-count quorum is already reachable — that is the #766 halt
+    // (a restart re-entered SCP at slot 2884 while peers balloted 2909, racing
+    // catch-up sync).
+    //
+    // Seeded from `min_peers_to_wait`: a genuine solo node (`min_peers == 0`)
+    // has no fleet to fall behind, so it starts synced and mints immediately
+    // (preserves the solo carve-out). A peer-expecting node starts NOT synced
+    // and flips true the first time the sync manager reports `Synced` — which,
+    // for a node already at/near tip, happens after the first peer-status round
+    // (no meaningful delay), and for a node genuinely behind happens only after
+    // catch-up completes. The `consensus` service mirrors this flag so its
+    // nomination/ballot proposer is gated identically (see
+    // `ConsensusService::set_initial_sync_complete`).
+    let mut initial_sync_complete = min_peers_to_wait == 0;
+    consensus.set_initial_sync_complete(initial_sync_complete);
+
     // Track minting state - can change as peers connect/disconnect
     let mut minting_enabled = false;
     // Track if minting was paused due to high faucet balance
@@ -758,8 +806,12 @@ async fn run_async(mut config: Config, config_path: &Path, mint: bool) -> Result
     println!();
     node.print_status_public()?;
 
-    // Start minting if quorum is satisfied
-    if can_mint_now {
+    // Start minting only if quorum is satisfied AND initial sync is complete
+    // (issue #770). A peer-expecting node that boots behind the fleet tip has
+    // `initial_sync_complete == false` here, so it does NOT arm minting at
+    // startup even when peer-count quorum is already reachable — it waits for
+    // the sync manager to reach `Synced` (handled in the sync-tick arm below).
+    if should_arm_minting(can_mint_now, initial_sync_complete) {
         info!("Starting minting - {}", quorum_message);
         node.start_minting_public()?;
         minting_enabled = true;
@@ -770,6 +822,12 @@ async fn run_async(mut config: Config, config_path: &Path, mint: bool) -> Result
         metrics_updater.set_minting_active(true);
         // Broadcast initial minting status to WebSocket clients
         ws_broadcaster.minting_status(true, 0.0, 0);
+    } else if mint && can_mint_now && !initial_sync_complete {
+        info!(
+            "Minting requested and quorum satisfied, but node is still syncing to \
+             the fleet tip — deferring minting until caught up (issue #770)"
+        );
+        println!("Minting will start when initial sync completes.");
     } else if mint {
         warn!("Minting requested but {}", quorum_message);
         println!("Minting will start when quorum is satisfied.");
@@ -1024,11 +1082,15 @@ async fn run_async(mut config: Config, config_path: &Path, mint: bool) -> Result
                                 );
                             }
 
-                            // Re-evaluate minting eligibility
+                            // Re-evaluate minting eligibility. The startup sync
+                            // gate (issue #770) is additive to the quorum gate:
+                            // even if quorum is now reachable, a node still
+                            // behind the fleet tip must not arm minting until it
+                            // has caught up (`initial_sync_complete`).
                             if mint && !minting_enabled {
                                 let connected = get_connected_peer_ids(&discovery);
                                 let (can_mint_now, msg) = check_minting_eligibility(&config, &connected, mint);
-                                if can_mint_now {
+                                if should_arm_minting(can_mint_now, initial_sync_complete) {
                                     info!("Quorum reached! {}", msg);
                                     if let Err(e) = node.start_minting_public() {
                                         warn!("Failed to start minting: {}", e);
@@ -1963,6 +2025,65 @@ async fn run_async(mut config: Config, config_path: &Path, mint: bool) -> Result
                     }
                 }
 
+                // Startup sync gate (issue #770): detect the first transition to
+                // `Synced` this session and release the minting/nomination gate.
+                //
+                // This is the STARTUP-side counterpart to the reactive
+                // `sync_scp_slot_to_chain` call in the block-applied arm above:
+                // that arm only fires after a catch-up batch is applied, so a
+                // node that reaches the tip via the pre-loop peer-wait window or
+                // via gossip (no explicit sync batch) would otherwise never
+                // release the gate. Here we release it the moment the sync state
+                // machine reports `Synced`, regardless of how the node got
+                // there.
+                if !initial_sync_complete && sync_manager.is_synced() {
+                    initial_sync_complete = true;
+                    consensus.set_initial_sync_complete(true);
+
+                    // Reconcile the SCP slot to the synced chain height BEFORE
+                    // the node casts its first ballot this session (issue #770,
+                    // acceptance criterion 2). Uses the canonical
+                    // `chain_height + 1` derivation; `sync_scp_slot_to_chain`
+                    // only advances an idle slot (its ballot/commit-state guard
+                    // is preserved) and is a no-op in solo mode.
+                    if let Ok(ledger) = node.shared_ledger().read() {
+                        if let Ok(state) = ledger.get_chain_state() {
+                            if consensus.sync_scp_slot_to_chain(state.height) {
+                                info!(
+                                    slot = consensus.current_slot(),
+                                    height = state.height,
+                                    "Initial sync complete: reconciled SCP slot to synced \
+                                     chain height before first ballot (issue #770)"
+                                );
+                            }
+                        }
+                    }
+
+                    info!("Initial sync to fleet tip complete — minting/nomination gate released (issue #770)");
+
+                    // Arm minting now if it was deferred purely on the sync gate
+                    // at startup (quorum was already reachable). Mirrors the
+                    // peer-connect arm's eligibility check.
+                    if mint && !minting_enabled {
+                        let connected = get_connected_peer_ids(&discovery);
+                        let (can_mint_now, msg) =
+                            check_minting_eligibility(&config, &connected, mint);
+                        if should_arm_minting(can_mint_now, initial_sync_complete) {
+                            info!("Starting minting after initial sync — {}", msg);
+                            if let Err(e) = node.start_minting_public() {
+                                warn!("Failed to start minting: {}", e);
+                            } else {
+                                minting_enabled = true;
+                                if let Ok(mut active) = minting_active.write() {
+                                    *active = true;
+                                }
+                                metrics_updater.set_minting_active(true);
+                                ws_broadcaster.minting_status(true, 0.0, 0);
+                            }
+                        }
+                    }
+                }
+
                 // Publish the post-tick sync state so `node_getStatus` reports
                 // honest sync info (#541). Done after `tick` so any state
                 // transition the tick performed is reflected immediately.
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed at maximum care as consensus-bootstrap code under the no-hard-forks policy. This is a clean, well-scoped startup-ordering fix for the #766 halt.

## Protocol invariance — confirmed
The `initial_sync_complete` field feeds exactly one consumer: `should_propose_this_round`, which gates only the local proposing choke point (`propose_pending_values`). Incoming SCP message reception/relay, ballot semantics, validity functions, and wire serialization are all untouched. For an already-synced node (`initial_sync_complete == true`, the default) behavior is byte-identical to before the PR. This is startup-ordering + a local proposing gate only — no consensus-visible change.

## Liveness of the gate — verified, no new deadlock
- **Solo carve-out ordering**: `should_propose_this_round` short-circuits on `min_peers == 0` at line 617 *before* consulting the sync gate at line 629. A solo node can never be deadlocked by the gate; `startup_sync_gate_never_blocks_solo_node` forces the flag false and confirms it still proposes.
- **Full-fleet cold start**: when all nodes restart at the same wedged height, peers connect and exchange status; since no peer is ahead by gap >= 2, each node transitions `Discovery -> Synced` at `sync.rs:1143` and releases the gate. No cross-node deadlock waiting to sync from each other.
- **All-peers-down (`min_peers > 0`)**: the node stays in `Discovery` and never flips `initial_sync_complete`, but this is not a *new* stuck state — the pre-existing #428 quorum gate (`connected_peers >= min_peers`) already blocks minting with zero peers, both before and after this PR. The new gate is strictly more conservative and introduces no configuration that could mint before but can never mint now.
- **Release is a poll, not a single-transition hook**: the gate is re-evaluated every `sync_tick` (run.rs:2039) via `sync_manager.is_synced()`, so it releases regardless of *how* the node reached Synced (catch-up batch, pre-loop peer-wait, or gossip). This correctly covers the "reached tip with no explicit sync batch" case the doc comment calls out.

## Default-true field — all construction sites verified
`ConsensusService::initial_sync_complete` has exactly one non-test constructor (`::new`, sets `true`); the run loop unconditionally overrides it at run.rs:793 based on `min_peers_to_wait == 0`. No other constructor path silently skips the gate.

## sync_scp_slot_to_chain interaction — preserved
The startup reconciliation reuses the existing `sync_scp_slot_to_chain`, whose ballot/commit-state guard (`slot_holds_ballot_or_commit_state`, #436) is intact: it refuses to fast-forward a slot holding genuine BALLOT/COMMIT state, only advances forward, and is a no-op in solo mode. The canonical `local height + 1` rule is now documented as the single derivation rule for both `::new` and rejoin.

## Regression-test quality
`startup_sync_gate_withholds_proposing_until_synced` exercises the actual race, not just the pure function: with quorum satisfied and `initial_sync_complete == false`, a submitted coinbase is withheld (`externalized.is_none()`) and stays queued (`pending_values` non-empty), then proposes once the flag flips. `slot_derivation_is_canonical_local_height_plus_one_across_halt` drives the real 2884->2909 reconciliation and asserts monotonicity/idempotence. The only untested seam is the async run-loop wiring itself (release at run.rs:2039), which the PR honestly flags as manual/integration — acceptable, and consistent with how `check_minting_eligibility` is tested as a pure function.

## Verification
- `cargo test -p botho --lib`: **1240 passed, 0 failed, 2 ignored** (matches builder claim).
- 10 new #770 tests all pass.
- `cargo fmt --check`: clean. `cargo clippy -p botho --lib`: no new warnings in run.rs/service.rs (all warnings are pre-existing in unrelated modules).
- All CI checks pass; merge state CLEAN.
- PR uses `Closes #770`.

Great work — the additive-gate framing, the solo short-circuit ordering, and the poll-based release are exactly right for this failure mode.

PR #774, opened 2026-07-11
feat(web): add i18next locale routing and translate landing page (es)

rjwalters:
## Summary

Phase 1 of internationalization for botho.io (#764): introduces the i18next framework in `@botho/web-wallet`, URL-path-prefix locale routing (`/es/...`), extraction of the marketing landing page into locale resource files, a Spanish translation, and a locale switcher.

**Scope decision — `Closes #764`:** The curator's enhancement comment explicitly states *"this issue should be scoped to **phase 1 only**"* and lists phases 2–4 as separate follow-up issues to be filed once this lands. Because the issue itself is phase-1-scoped, this PR uses `Closes #764` per that guidance (not `Updates`). Phases 2–4 (wallet/pay/claim/contacts, docs.tsx, OG/PWA metadata, Accept-Language edge negotiation) are deliberately untouched here and should be filed as new tracking issues.

## Changes

- **Framework**: added `i18next` + `react-i18next` deps; new `src/lib/i18n.ts` initializes i18next with the `landing` namespace. Locale bundles are imported statically (not lazy-fetched) so a non-default locale has no flash of untranslated content on first paint. Init is idempotent under StrictMode/HMR.
- **Routing**: new `src/lib/locale-path.ts` (parse/build/switch helpers). `en` is the unprefixed default; non-default locales route under a `/<locale>` segment. `App.tsx` gains a `LocaleRoutes` shell that strips the locale prefix, renders the existing (locale-agnostic) route table against the remaining path, syncs i18next's active language, persists the choice to `localStorage`, and sets `<html lang>`. An unsupported/absent prefix falls back to `en` — no 404 or crash.
- **Landing extraction**: every user-facing string in `landing.tsx` moved to `locales/en/landing.json`; component now renders via `useTranslation()`. No hardcoded English copy remains.
- **Spanish**: `locales/es/landing.json`. Machine-assisted translation, flagged for human review; protocol terms (SCP, CLSAG, ML-KEM) and the Setswana motto are intentionally left untranslated.
- **Locale switcher**: `components/LocaleSwitcher.tsx` (a labelled `<select>`) in the desktop nav and mobile menu; re-maps the current URL to the target locale and persists the choice.
- **Tests**: `locale-path.test.ts` (helper units), `landing.test.tsx` (en/es render + a11y label), `App.test.tsx` (URL-driven locale routing + `<html lang>` sync + unsupported-prefix fallback).

## Acceptance Criteria Verification

| Criterion (Phase 1) | Status | Verification |
[...]

Issue #757, opened 2026-07-09
P4.4 mainnet hardening: SRI + self-hosted operator page (§8.3, §9)


rjwalters:
Follow-up from #751 (P4.4e operator dashboard). Tracks the mainnet-hardening list from `docs/security/quorum-write-path.md` §8.3 / §9.

## Context

The operator dashboard (`web/packages/web-wallet` `/operator` → Actions tab, #751) imports the operator Ed25519 key into the browser, encrypted under a mandatory passphrase, and signs quorum-curation envelopes client-side. The key never leaves the browser and never reaches the Pages host (static assets only).

**Accepted testnet residual risk (§8.3):** a malicious Pages bundle could prompt the operator for a signature over attacker-chosen bytes. The dashboard mitigates this today by displaying the exact canonical envelope before every signature, and by the node's v1 allowlist (no mode/threshold actions → a compromised dashboard is bounded to recoverable liveness). For testnet this residual risk is accepted.

## Mainnet hardening (do before mainnet)

- [ ] **Subresource Integrity (SRI)** on the operator page bundle so the browser refuses a tampered bundle (pin the hash the operator vetted).
- [ ] **Self-hosting the operator page** so the operator controls the exact bundle they run rather than trusting the shared Pages deploy.
- [ ] Re-review §8.3 threat model once the above land; decide whether the in-browser key path is still the right posture for mainnet or whether an air-gapped signer / hardware key is warranted.

## References
- `docs/security/quorum-write-path.md` §8.3 (compromised-dashboard threat), §9 (checklist)
- #751 (the dashboard this hardens)

rjwalters:
## Curator Enhancement

### Current architecture (why this is non-trivial)

The operator dashboard is **not** a standalone deploy — `/operator` is one
route inside the same Vite SPA as the marketing site and wallet, all served
from a single Cloudflare Pages project (`web/packages/web-wallet`,
`wrangler.toml` → `project-name=botho`), deployed to two hostnames
(`botho.io`, `wallet.botho.io`) from one `dist/` build
(`web/packages/web-wallet/src/App.tsx:11,57`). There is no separate build
target, no CI-driven Pages deploy (`.github/workflows/web-ci.yml` only
dry-run-deploys the BaaS *worker*, not the web-wallet Pages site — deploy is
manual via `pnpm --filter @botho/web-wallet deploy` /
`web/packages/web-wallet/package.json:10-11`), and no SRI infrastructure
[...]

PR #781, opened 2026-07-12
feat(web): split operator dashboard into its own SRI-pinned build entry

### LINKED ISSUE #772: P4.4 mainnet hardening: split /operator into its own build target for browser-enforced SRI (option a)
Follow-up from #757 (§8.3.1 addendum). Tracks the **option (a)** path that #757 explicitly deferred as too large for a single Builder PR (assessed **L**).

## Context

#757 shipped the **option (b)** hardening for the operator dashboard: a whole-bundle verify script (`web/scripts/verify-operator-bundle.sh`) plus a self-hosting runbook (`docs/operations/self-hosted-operator-dashboard.md`). That closes §8.3's malicious-bundle risk for an operator who verifies and/or self-hosts — but integrity is **operator-enforced** (they must remember to run the check). See `docs/security/quorum-write-path.md` §8.3.1 for the full trade-off table.

## What this issue adds

**Browser-enforced** Subresource Integrity for the operator dashboard, so a tampered chunk fails to load with no operator action required. Because `/operator` is currently one route inside the shared Vite SPA (`web/packages/web-wallet`), true in-page SRI requires giving `/operator` its own HTML document that can pin `integrity="sha384-…"` hashes for its own JS/CSS chunks.

## Scope (why this is L)

- [ ] Split `/operator` into its own Vite build target (or a second Pages project), so its `index.html` references only its own chunks.
- [ ] Wire `integrity=` (sha384) hash generation into that build output (custom Vite plugin or `vite-plugin-html`) so emitted `index.html` pins each referenced chunk.
- [ ] Add a build test asserting the emitted `index.html` contains `integrity=` attributes matching the actual chunk hashes.
- [ ] Reconcile the split with the shared `@botho/features` / `@botho/core` workspace packages and keep the two deploy artifacts in sync (CI).
- [ ] Handle the PWA service worker (`registerType: 'autoUpdate'`) interaction — either disable auto-update on the operator target or document/enforce re-verification.
- [ ] Update `docs/security/quorum-write-path.md` §8.3.1 to mark option (a) as landed.

## References
- #757 (option b — verify script + self-host runbook, shipped)
- `docs/security/quorum-write-path.md` §8.3.1 (decision record + trade-off table)
- `web/packages/web-wallet/vite.config.ts`, `src/App.tsx`, `wrangler.toml`

---

### Current architecture (verified against `origin/main`)

- **Single SPA, single Pages project.** `web/packages/web-wallet` builds one Vite bundle (`vite build` in `package.json`) served by Cloudflare Pages under `project-name=botho` (`web/packages/web-wallet/wrangler.toml`). One `index.html` (`web/packages/web-wallet/index.html`) loads `/src/main.tsx` as a module entry; Vite code-splits from there.
- **Routing, not build-splitting.** `/operator` is just a `<Route path="/operator" element={<OperatorPage />} />` inside `AppRoutes` in `web/packages/web-wallet/src/App.tsx`. It is wrapped one level up by `LocaleRoutes` (PR #774's i18n shell), which strips a leading `/:locale` segment before matching `AppRoutes`. Any build split must keep the operator entry reachable at both `/operator` and `/:locale/operator` (e.g. `/es/operator`) — the split target's own HTML/router must replicate (or import) the same locale-stripping behavior, or SRI'd operator loads will 404/soft-break for non-default locales.
- **PWA / service worker.** `vite-plugin-pwa` is configured with `registerType: 'autoUpdate'` in `vite.config.ts` (line 84) and globs `**/*.{js,css,html,ico,png,svg,woff2}` for the precache manifest. This SW is registered for the *whole* origin today. §8.3.1 flags that autoUpdate silently replaces a verified bundle — this applies equally to a split-build unless the operator target either opts out of the SW or is served from a scope/origin the SW doesn't control.
- **Verify script assumes one `dist/`.** `web/scripts/verify-operator-bundle.sh` hashes the entire `web/packages/web-wallet/dist` tree (excluding `*.map`) into one aggregate hash. If `/operator` moves to its own build output directory, this script's `DEFAULT_DIST` and the `verify:operator-bundle` pnpm script (`web/packages/web-wallet/package.json`) need to point at (or be duplicated for) the new output — otherwise option (b)'s already-shipped tooling silently stops covering the thing it was built to protect.
- **No CI deploy step to update.** `.github/workflows/web-ci.yml` only typechecks/tests/dry-run-deploys the worker; it does not run `wrangler pages deploy`. Deploys are manual (`pnpm deploy` / `pnpm deploy:all` in `web/packages/web-wallet/package.json`, and `web/package.json`'s `deploy`/`deploy:preview` scripts). Any build-split still needs *some* mechanism to keep the two artifacts from drifting (see options below), but there's no existing CI gate this issue breaks by NOT updating — it only needs to add one if it wants automated drift detection.

### Implementation options

**Option 1 — Second Vite entry within the same package (multi-page build), recommended.**
Add a second HTML entry point (e.g. `web/packages/web-wallet/operator.html`) with its own `<script type="module" src="/src/operator-main.tsx">`, and configure Vite's `build.rollupOptions.input` as a multi-page build (`{ main: 'index.html', operator: 'operator.html' }`). `operator-main.tsx` mounts a slimmed router (just the operator route + its own locale handling, reusing `parseLocalePath`/`LocaleRoutes` logic) instead of the full `AppRoutes`.
- *Pros*: single package, single `pnpm install`/dependency graph, shares `@botho/features`/`@botho/core`/`@botho/ui` via the existing workspace resolution (no new package boundary to keep in sync) — directly addresses the "reconcile with shared packages" scope bullet with the least new surface area. Still deploys to the same Cloudflare Pages project (`dist/` just gets a second HTML file + its own hashed chunk graph), so no Pages-project/DNS topology change.
- *Cons*: both entries still ship from the same origin behind the same PWA scope unless the SW config is scoped or the operator HTML is excluded from precaching (`workbox.globPatterns`) and given `registerType: 'prompt'` or no SW at all via a `navigateFallbackDenylist`-style exclusion.

**Option 2 — Separate package + separate Cloudflare Pages project.**
New `web/packages/operator-dashboard` package with its own `vite.config.ts`, `wrangler.toml` (`project-name=botho-operator` or similar), deployed to a distinct subdomain (e.g. `operator.botho.io`).
- *Pros*: cleanest isolation — separate PWA/SW scope by construction (different origin), no risk of the main SPA's SW ever touching operator assets, no shared `index.html`/router logic to reconcile.
- *Cons*: highest cost — new package scaffolding, new Pages project + DNS + TLS, new CI awareness, and now *two* places importing `@botho/features`/`@botho/core`/`@botho/ui` that must be kept in version lockstep (workspace `workspace:*` protocol handles this at build time, but two deploy pipelines can still drift if only one is redeployed after a shared-package change — this is the "keep in sync (CI)" risk called out in the issue's scope). Also changes the operator's bookmarked/shared URL (breaking any existing magic links unless a redirect from `/operator` is added).

**Recommendation: Option 1** (multi-page build within the existing package). It satisfies the "hashable, SRI-annotated assets for `/operator`" requirement with materially lower topology cost than a second Pages project, keeps the shared-package sync problem to "one `pnpm build`" instead of "two coordinated deploys," and preserves the existing `/operator` URL (no redirect needed). Option 2 should be revisited only if a security review of §8.3 concludes the *origin* itself (not just the bundle) needs to be isolated from the main SPA.

### SRI hash injection

Neither `vite-plugin-pwa` nor stock Vite emits `integrity=` attributes. Use a small custom Vite plugin (`transformIndexHtml` hook, `enforce: 'post'`, `apply: 'build'`) that, after the operator entry's chunks are emitted:
1. Reads each `<script src>` / `<link rel="stylesheet" href>` referenced in the transformed `operator.html`.
2. Computes `sha384` over the corresponding emitted asset (available via Rollup's bundle metadata in `generateBundle`, or by re-reading the file from `dist/` in a `closeBundle` hook — mirror the pattern `wasmSignerPkgPlugin` already uses in `vite.config.ts` for post-build file manipulation).
3. Injects `integrity="sha384-<base64>"` and `crossorigin="anonymous"` onto each tag.

This can be hand-rolled (~50-80 LOC, matching the existing `wasmSignerPkgPlugin` style already in `vite.config.ts`) or via `vite-plugin-html`/`rollup-plugin-html-sri`-style third-party plugin — hand-rolling is preferable given the codebase already has a working precedent for custom build-output manipulation in this exact file, and avoids a new dependency for ~1 hash function.

### PWA / autoUpdate interaction (must address, per §8.3.1)

The operator entry must NOT be covered by the existing `autoUpdate` service worker in the same way the main SPA is, or SRI enforcement is moot (a silently-updated SW could swap the operator HTML document itself, which SRI cannot protect since SRI only pins *sub-resources*, not the top-level document — this is exactly the gap §8.3.1 identifies for the whole-SPA case). Concrete mitigation, to be decided by the Builder with input from a security reviewer:
- Exclude `operator.html` and its chunks from the `workbox.globPatterns` precache list in `vite.config.ts`, so the SW never caches/serves them — the browser always fetches `/operator` fresh over the network (subject to normal HTTP caching, not SW hijack), and `integrity=` protects the sub-resources within that fresh fetch.
- Document in `docs/security/quorum-write-path.md` §8.3.1 (as a landed-option addendum) that with option (a), the operator document itself is still same-origin-trusted-on-first-load (browsers don't SRI-check top-level navigations), so the security gain is specifically: *a compromised Pages host can still serve a bad `operator.html`, but can no longer silently swap the JS/CSS chunks that document references without the browser refusing to execute them* (a materially narrower attack surface: the attacker's payload must live in the one HTML file that's easiest to diff/monitor, rather than in any of N chunk files).

### Acceptance Criteria

- [ ] `/operator` (and `/:locale/operator`) is served by a distinct build entry whose HTML document references only its own hashed JS/CSS chunks (verified: the emitted operator HTML does not reference chunks shared with the main SPA bundle by filename, only by content where genuinely shared code is intentionally duplicated or explicitly still shared per the chosen option's design).
- [ ] Every `<script>`/`<link rel="stylesheet">` tag in the emitted operator HTML has a matching `integrity="sha384-…"` attribute whose hash matches the actual built file content (byte-for-byte).
- [ ] A build-time automated test fails the build (or CI) if any operator-entry asset reference is missing an `integrity` attribute or the hash doesn't match the file on disk.
- [ ] The operator entry is excluded from the existing PWA service worker's precache/auto-update path (or an equivalent mitigation is implemented), with the rationale documented.
- [ ] `web/scripts/verify-operator-bundle.sh` (or a companion script) still produces a meaningful hash covering the operator build output after the split — update `--dist` default/usage or add a second invocation as needed so option (b)'s tooling isn't silently orphaned.
- [ ] `docs/security/quorum-write-path.md` §8.3.1 is updated to mark option (a) as landed, with a short note on the residual top-level-document trust gap (see "PWA / autoUpdate interaction" above).
- [ ] Existing `/operator` links (including any magic-link tokens captured via `captureOperatorToken()` in `web/packages/web-wallet/src/pages/operator.tsx`) continue to work unchanged from the user's perspective — no URL/redirect change required for Option 1; if Option 2 is chosen instead, a redirect from the old `/operator` path must be added and documented.

### Affected Files

- `web/packages/web-wallet/vite.config.ts` - add second Vite entry / multi-page build config, SRI-injection plugin, PWA precache exclusion for the operator entry
- `web/packages/web-wallet/index.html` - reference point for the existing single-entry pattern (no change expected, but confirms the main entry is unaffected)
- `web/packages/web-wallet/operator.html` (new) - new HTML entry point for the split operator build (Option 1)
- `web/packages/web-wallet/src/App.tsx` - extract/reuse the `/operator` route + `LocaleRoutes`/`parseLocalePath` logic for the standalone operator entry
- `web/packages/web-wallet/src/pages/operator.tsx` - operator page component, consumed by the new entry point
- `web/packages/web-wallet/src/main.tsx` - reference for the existing SPA bootstrap pattern to mirror in a new `operator-main.tsx`
- `web/packages/web-wallet/package.json` - new build script for the operator entry (if a distinct output step is needed) and update to `verify:operator-bundle`
- `web/scripts/verify-operator-bundle.sh` - confirm/update `--dist` handling so it covers the new operator build output
- `web/packages/web-wallet/wrangler.toml` - only if Option 2 (separate Pages project) is chosen; otherwise unaffected
- `docs/security/quorum-write-path.md` - update §8.3.1 to mark option (a) landed
- `docs/operations/self-hosted-operator-dashboard.md` - update if the self-hosting runbook needs to reflect the new build output layout

### Test Plan

- [ ] Automated tests: add a build/asset test (e.g. a Vitest or Node script run post-`vite build`) that parses the emitted operator HTML, extracts every `integrity=` attribute, recomputes the sha384 hash of the referenced file from disk, and asserts equality — this should live alongside other web-wallet build tooling and run in `pnpm --filter @botho/web-wallet` scripts (and ideally `web-ci.yml`, though CI does not currently run a build step — evaluate whether to add one).
- [ ] Manual verification: run `pnpm --filter @botho/web-wallet build`, inspect `dist/operator.html` (or equivalent), confirm `integrity=` attributes are present on all script/style tags and match `shasum -a 384` of the referenced files.
- [ ] Manual verification: load `/operator` in a browser with devtools open, tamper with one emitted chunk file in the deployed `dist/` (simulate a compromised host), reload, and confirm the browser blocks the tampered resource (console shows an SRI failure) rather than executing it.
- [ ] Manual verification: confirm `/operator` still works when a non-default locale prefix is present (e.g. `/es/operator`), matching the `LocaleRoutes` behavior in the main SPA.
- [ ] Manual verification: confirm the PWA service worker does not precache or serve stale operator assets — check the `workbox` precache manifest (or equivalent) does not list operator chunks, and a fresh `/operator` load always hits the network for its top-level document.
- [ ] Regression: run `pnpm --filter @botho/web-wallet build && web/scripts/verify-operator-bundle.sh` and confirm it still produces a valid, non-empty hash covering the relevant build output (update script/usage first per the Affected Files note above).
- [ ] Regression: existing e2e/vitest suites for the wallet (`pnpm test:run`, `pnpm test:e2e`) still pass — this change should not affect the main SPA's routes.

rjwalters:
## Summary

Lands **option (a)** of the operator-dashboard mainnet hardening (`docs/security/quorum-write-path.md` §8.3.1, follow-up to #757): browser-enforced Subresource Integrity for the operator dashboard. `/operator` was one route inside the shared Vite SPA, so its JS/CSS chunks could not carry `integrity=` hashes (SRI pins the sub-resources of a document the consumer controls, not a client-side route). This splits it into its own Vite build entry whose emitted HTML pins `sha384` hashes on each chunk it loads, so a tampered chunk on the host fails to load with **no operator action required**.

Implemented as **Option 1** from the curator's plan: a second entry point (`operator.html` -> `operator-main.tsx`) as a multi-page build **within** `web/packages/web-wallet`, deployed to the **same** Cloudflare Pages project — the `/operator` URL is unchanged, no new package or Pages/DNS topology.

## Changes

- **New operator entry**: `operator.html` (own document, `noindex`), `src/operator-main.tsx` (bootstrap that deliberately does **not** register the PWA service worker), `src/OperatorApp.tsx` (slim router: only `/operator` + `/:locale/operator`, reusing `parseLocalePath` — the i18n `AppRoutes`/`LocaleRoutes` shell from #774 is intentionally kept out of the operator entry).
- **`vite.config.ts`**: multi-page `rollupOptions.input` (`main` + `operator`); a hand-rolled `sriHashPlugin` (`generateBundle`, `enforce: 'post'`, modeled on the existing `wasmSignerPkgPlugin`) that injects `integrity="sha384-…"` + `crossorigin` onto every `<script>`/`<link rel="stylesheet">`/`<link rel="modulepreload">` in `operator.html` referencing a local chunk; PWA `globIgnores: ['operator.html']` + a `/operator` `navigateFallbackDenylist` entry so the SW never caches/serves the operator document.
- **Tests**: `operator-sri.test.ts` runs a real production build and asserts every operator asset reference is pinned and each hash matches the file byte-for-byte (and that `index.html` stays unpinned); `OperatorApp.test.tsx` asserts the shell mounts at `/operator`, `/es/operator`, and redirects unknown in-document paths.
- **`verify-operator-bundle.sh`**: new `--verify-sri` mode (exit 3 on any missing/mismatched attribute) — the scriptable equivalent of the browser's own SRI check; plus a `verify:operator-sri` package script. The existing whole-bundle hash mode (option (b), #757) is unchanged and still covers the new output.
- **Docs**: §8.3.1 marks option (a) landed and documents the residual top-level-document trust gap (SRI never pins the navigated document); the self-hosted runbook covers the multi-page layout, the `/operator -> operator.html` server routing rule, and `--verify-sri`.

| `/operator` (+ `/:locale/operator`) served by a distinct build entry referencing its own hashed chunks | ✅ | `dist/operator.html` boots `operator-*.js`; `OperatorApp.test.tsx` renders `/operator` and `/es/operator` |
| Every script/link tag in operator HTML has a matching `sha384` integrity hash | ✅ | `operator-sri.test.ts` (build + byte-for-byte compare); manual `shasum -a 384` matches all 3 refs |
| Build-time test fails on missing/mismatched integrity | ✅ | `operator-sri.test.ts` in `pnpm test:run`; tamper test → `--verify-sri` exits 3 |
| Operator entry excluded from PWA precache/auto-update | ✅ | `operator-main.tsx` omits `registerSW`; `globIgnores: ['operator.html']`; verified `operator.html` absent from `sw.js` precache manifest |
| `verify-operator-bundle.sh` still covers operator output after split | ✅ | whole-bundle hash unchanged; new `--verify-sri` mode verifies the operator entry specifically |
| §8.3.1 updated to mark option (a) landed + residual gap | ✅ | table status → "Shipped in #772"; new landed-addendum + residual-gap paragraphs |
| Existing `/operator` links / magic-link tokens keep working | ✅ | same Pages project, same `/operator` URL, `captureOperatorToken()` path unchanged (no redirect needed) |

## Test Plan

- `pnpm -r typecheck` — clean across all 8 web workspaces.
- `pnpm test:run` — **732 passed / 10 skipped (76 files)**, including the 2 new test files.
- `pnpm --filter @botho/web-wallet build` — emits `dist/index.html` (main, unpinned) + `dist/operator.html` (operator, SRI-pinned); `operator.html` absent from `sw.js` precache manifest.
- `web/scripts/verify-operator-bundle.sh --verify-sri` — `MATCH: all 3 operator asset references have a correct sha384 integrity hash`.
- **SRI tamper evidence**: appended a byte to a copy of `dist/assets/operator-*.js`; `--verify-sri` reported `MISMATCH` (pinned vs. on-disk sha384) and exited 3. This is exactly the comparison a browser performs on load — a mismatch means the browser refuses to execute the chunk. (A headless real-browser check was not run; the scripted integrity-attribute check is the acceptable substitute per the issue.)

Closes #772

### DIFF
--- a/docs/operations/self-hosted-operator-dashboard.md
+++ b/docs/operations/self-hosted-operator-dashboard.md
@@ -17,16 +17,25 @@ host) and signs quorum-curation envelopes client-side. The serious residual risk
 (§8.3) is a **malicious bundle**: a compromised host could serve JavaScript that
 prompts you to sign attacker-chosen bytes.
 
-Two independent mitigations reduce this to near-zero for a careful operator:
+Three mitigations reduce this to near-zero for a careful operator:
 
-1. **Verify the bundle** you are about to run against a maintainer-published
+1. **Browser-enforced SRI (option (a), #772), on by default.** `/operator` is
+   its own build entry (`operator.html`) whose HTML pins `integrity="sha384-…"`
+   hashes for every JS/CSS chunk it loads. Any tampered chunk fails the browser's
+   own integrity check and never executes — no operator action required. This
+   protects the operator entry's *sub-resources* even on the shared Pages deploy.
+   Its one residual gap: the browser does not integrity-check the top-level
+   `operator.html` document itself (SRI never does), so mitigations 2–3 remain
+   worthwhile.
+2. **Verify the bundle** you are about to run against a maintainer-published
    hash, *before* you import your key. A tampered bundle produces a different
-   hash and is caught.
-2. **Self-host** that verified bundle from infrastructure you control, so the
+   hash and is caught. This covers the top-level document that SRI cannot.
+3. **Self-host** that verified bundle from infrastructure you control, so the
    shared Pages host is no longer in your trust path at all.
 
-This runbook covers both. Doing (1) alone already closes most of the risk; (2)
-removes the host entirely and is recommended for mainnet.
+This runbook covers (2) and (3); (1) is automatic. Doing (2) alone already
+closes most of the risk on top of (1); (3) removes the host entirely and is
+recommended for mainnet.
 
 ## Prerequisites
 
@@ -54,12 +63,17 @@ git rev-parse HEAD             # record this — it's what you're trusting
 cd web
 pnpm install --frozen-lockfile
 
-# Build the operator dashboard bundle (the whole SPA; /operator is one route).
+# Build the operator dashboard bundle. The build is multi-page: the marketing/
+# wallet SPA is `dist/index.html`, and the operator dashboard is its own entry,
+# `dist/operator.html`, whose referenced chunks carry SRI integrity hashes.
 # Include the in-browser signer wasm so the signing path actually works:
 pnpm --filter @botho/web-wallet build:all
 ```
 
-The build output lands in `web/packages/web-wallet/dist/`.
+The build output lands in `web/packages/web-wallet/dist/`. Two HTML entry
+documents are emitted: `dist/index.html` (main SPA) and `dist/operator.html`
+(the standalone, SRI-pinned operator dashboard — this is what `/operator`
+serves).
 
 > `build:all` runs `build:wasm` then `build`. If you only need to *inspect*
 > the bundle hash and not sign, plain `pnpm --filter @botho/web-wallet build`
@@ -94,6 +108,23 @@ bundle):
 web/scripts/verify-operator-bundle.sh --manifest
 ```
 
+### Confirm the operator entry's browser-enforced SRI (option (a))
+
+Independently of the aggregate hash, you can confirm the operator entry pins a
+correct `sha384` integrity hash on each chunk it references — the scriptable
+equivalent of the check the browser performs on load:
+
+```bash
+web/scripts/verify-operator-bundle.sh --verify-sri
+# or: pnpm --filter @botho/web-wallet verify:operator-sri
+```
+
+It parses `dist/operator.html`, recomputes the `sha384` of each referenced chunk
+from disk, and exits 3 if any reference is missing an `integrity` attribute or
+the hash does not match. This is what protects you on the *shared* deploy even
+without self-hosting: a compromised host that swaps a chunk cannot match the
+pinned hash, and the browser refuses to run it.
+
 ### Reproducibility
 
 The bundle hash is deterministic: building the same pinned commit in two clean
@@ -123,12 +154,26 @@ pnpm --filter @botho/web-wallet preview   # serves dist/ on http://localhost:417
 
 # Option B: any static file server you trust, on infra you control
 #   (nginx, caddy, `python3 -m http.server`, an internal Pages/Netlify you own, …)
-#   Point it at web/packages/web-wallet/dist/ and serve index.html as the SPA
-#   fallback for client-side routes like /operator.
+#   Point it at web/packages/web-wallet/dist/. See the routing note below for
+#   how to map /operator to operator.html and everything else to index.html.
 ```
 
-Because the SPA uses client-side routing, configure your server to serve
-`index.html` for unknown paths (SPA fallback) so `/operator` resolves.
+Because the build is now multi-page, configure your server's client-side-routing
+fallback carefully:
+
+- `/operator` (and any `/<locale>/operator`, e.g. `/es/operator`) must serve
+  **`operator.html`** — the standalone, SRI-pinned operator document.
+- All other unknown paths serve **`index.html`** — the main SPA fallback.
+
+Serving `index.html` for `/operator` would load the un-SRI-pinned main SPA route
+instead of the split operator entry, silently discarding the browser-enforced
+integrity. Example nginx:
+
+```nginx
+location = /operator      { try_files /operator.html =404; }
+location ~ ^/[a-z]{2}/operator$ { try_files /operator.html =404; }
+location /                { try_files $uri /index.html; }
+```
 
 > **RPC reachability.** The bundle reaches a node's RPC directly. For local
 > `pnpm preview`, the Vite preview proxy forwards `/rpc` to a seed node (see
@@ -138,10 +183,18 @@ Because the SPA uses client-side routing, configure your server to serve
 
 ## Step 5: Handle PWA auto-update (important)
 
-The SPA registers a service worker with `registerType: 'autoUpdate'`
+The main SPA registers a service worker with `registerType: 'autoUpdate'`
 (`vite.config.ts`). On the shared deploy this silently fetches and activates a
 **newer** bundle after each deploy — which would defeat the whole point of
 pinning a verified bundle.
+
+> **Note (option (a), #772):** the operator entry is deliberately kept *out* of
+> the service worker. `operator-main.tsx` does not register the SW, and
+> `operator.html` is excluded from the Workbox precache
+> (`globIgnores: ['operator.html']`), so the operator document is always fetched
+> fresh over the network and cannot be silently swapped from the SW cache. The
+> guidance below still matters for the *rest* of the bundle and for the chunks
+> the operator entry loads.
 
 When self-hosting, keep the served `dist/` **fixed**: serve one verified build
 and do not roll a newer one behind the same origin. With no newer bundle
@@ -186,9 +239,15 @@ Publish that `sha256-<...>` value next to the release for the pinned commit.
 
 ## What this does and does not remove from the trust path
 
+**Removed** by browser-enforced SRI (automatic, even on the shared deploy):
+
+- Silent tampering of the operator entry's JS/CSS **chunks** — the browser
+  refuses any chunk whose bytes don't match the pinned `sha384` hash.
+
 **Removed** by verify + self-host:
 
-- Trust in the shared Cloudflare Pages host serving an honest bundle.
+- Trust in the shared Cloudflare Pages host serving an honest top-level
+  `operator.html` document (the one thing SRI cannot pin).
 - Silent bundle replacement via PWA auto-update (when you serve a fixed build).
 
 **Still trusted** (out of scope for this runbook):
--- a/docs/security/quorum-write-path.md
+++ b/docs/security/quorum-write-path.md
@@ -419,24 +419,58 @@ distinct ways to close 8.3, with different cost/guarantee profiles:
 
 | Approach | Guarantee | Cost | Status |
 |---|---|---|---|
-| **(a) Split-build + true SRI** — give `/operator` its own Vite build / Pages target whose `index.html` pins `integrity=` hashes for its own JS/CSS chunks. | Browser *enforces* integrity: a tampered chunk fails to load, no operator action required. | High — a second build artifact + deploy target + CI job kept in sync with the shared `@botho/features`/`@botho/core` packages; touches Vite config, Pages topology, and the App router. | **Deferred to follow-up** (assessed **L**). Filed as a separate issue; see PR for #757. |
+| **(a) Split-build + true SRI** — give `/operator` its own Vite build entry whose HTML document pins `integrity=` hashes for its own JS/CSS chunks. | Browser *enforces* integrity: a tampered chunk fails to load, no operator action required. | High — a second build entry + an SRI-injection plugin + PWA precache exclusion, kept in sync with the shared `@botho/features`/`@botho/core` packages; touches Vite config and adds a standalone operator app shell. | **Shipped in #772.** Multi-page Vite build (`operator.html` -> `operator-main.tsx`) within `web/packages/web-wallet` on the same Pages project; `sriHashPlugin` pins `sha384` integrity on the operator entry's chunks; a build-time test (`operator-sri.test.ts`) and `verify-operator-bundle.sh --verify-sri` enforce it. |
 | **(b) Whole-bundle hash pin + out-of-band verify** — a maintainer publishes an aggregate hash of the built `dist/` tree; the operator verifies their bundle against it **before importing their key**, and/or self-hosts that exact bundle. | Operator-enforced: integrity depends on the operator running the check. No browser enforcement, but no host-topology change. | Low — one verify script + a runbook. Composes directly with self-hosting. | **Shipped in #757.** `web/scripts/verify-operator-bundle.sh` + `docs/operations/self-hosted-operator-dashboard.md`. |
 
 **Interaction with self-hosting.** Approach (b) is the natural companion to the
 second §9 hardening item ("self-hosting the operator page"): an operator who
 builds the bundle from a pinned tag, verifies its aggregate hash against a
 maintainer-published value, and serves that `dist/` from infrastructure they
-control has removed the Pages host from the trust path entirely — without
-waiting on the split-build. (a) remains desirable on top of (b) for the
-non-self-hosting operator, because it moves enforcement from "operator
-remembers to run the check" to "the browser refuses the bad bundle." See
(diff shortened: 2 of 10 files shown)


rjwalters:
✅ **Approved!**

Reviewed this security-hardening deliverable with care against all five risk areas called out for #772. Checked out the PR HEAD (d766d05), ran a real production build, inspected the emitted bundle, and reproduced the tamper evidence.

## SRI completeness — verified, no gap
- `OperatorApp` uses no `React.lazy` / `route.lazy` — all routes are statically-elemented. The one `import(l.module)` in the framework chunk is React Router's internal lazy-route-module loader (dead code here; no lazy route modules exist).
- Operator signing uses `@noble/curves` ed25519 (pure JS, statically bundled in `features/src/operator/key-vault.ts`), **not** the wasm-signer's `@vite-ignore` runtime dynamic import — that path is wallet-send-only and unreachable from the operator routes.
- Build emitted only 4 JS chunks; none is a lazy operator chunk. `operator.html` references exactly 3 executable/style resources (entry `operator-*.js`, modulepreloaded framework `theme-*.js`, `theme-*.css`) and **all 3 are pinned**. The operator entry chunk contains zero `/assets/` refs — no unpinned runtime-fetched chunk.

## Cloudflare Pages routing — does not no-op in production
- `operator.html` is emitted at `dist/` root; there is **no** `_redirects` / `_routes.json` catch-all rewrite (there never has been in git history).
- Pages' default asset resolution serves `/operator` from `operator.html` (clean-URL handling); the SPA fallback only fires for genuinely-missing paths, and `/operator` is no longer missing. The feature is live on the shared deploy, not a silent no-op.
- The self-hosted runbook additionally documents the explicit `try_files /operator.html =404;` rule (and the `/<locale>/operator` variant) for non-Pages hosts.

## Shared framework chunk — main SPA unaffected
- `theme-DkrO5Vcc.js` is pinned in `operator.html` but referenced **without** integrity in `index.html` — confirmed in emitted output and asserted by `operator-sri.test.ts` ("index.html stays unpinned"). SRI injection is correctly scoped to `operator.html` only.

## Locale interaction
- `OperatorLocaleRoutes` mirrors the main SPA's `LocaleRoutes` (reuses `parseLocalePath`, same locale-strip contract). `/es/operator` verified by `OperatorApp.test.tsx` (`document.documentElement.lang === 'es'`).

## Service-worker gap honesty
- §8.3.1's residual top-level-document trust gap note is accurate and complete. `operator-main.tsx` omits `registerSW`; `globIgnores: ['operator.html']` + `navigateFallbackDenylist` for `/operator` are in place; confirmed `operator.html` absent from the `sw.js` precache manifest (0 matches).

## Test Execution
- `pnpm --filter @botho/web-wallet build` — ✅ emits `dist/index.html` (unpinned) + `dist/operator.html` (SRI-pinned); operator.html absent from sw.js precache.
- `pnpm vitest run src/OperatorApp.test.tsx src/operator-sri.test.ts` — ✅ 5/5 pass.
- `verify-operator-bundle.sh --verify-sri` on clean build — ✅ `MATCH: all 3 operator asset references have a correct sha384 integrity hash`.
- **SRI tamper test** — ✅ appended a byte to `dist/assets/operator-*.js`; `--verify-sri` reported `MISMATCH` and exited **3**, as specified.
- `shellcheck web/scripts/verify-operator-bundle.sh` — ✅ clean.
- CI (`cargo deny`, `typecheck + vitest + worker dry-run`) — ✅ both pass; merge state CLEAN.

Docs are honest about the inherent residual gap (browsers don't SRI-check the top-level navigated document), the `@noble` vs wasm signing distinction holds, and the multi-page routing is covered for both the Pages deploy and self-hosting. Clean, well-scoped, thoroughly tested hardening. PR uses `Closes #772`.

PR #786, opened 2026-07-12
feat(web-wallet): edge locale negotiation + localized OG/Twitter/hreflang metadata

### LINKED ISSUE #779: web i18n phase 4: localized OG/Twitter/PWA metadata + Accept-Language edge negotiation
## Original Issue

Phase 4 of the i18n rollout from #764: localize OG/Twitter card and PWA manifest metadata per locale, and add edge-based Accept-Language negotiation (redirect first-visit users to their locale prefix) on the Cloudflare Pages layer. Builds on phases 1-3 (#774, phase 2/3 issues).

---

### Problem Statement

Phases 1-3 (#774, #784, #785 — all merged) built out client-side, URL-prefix-based locale routing (`/es/...`) and translated the landing, wallet/pay/claim/contacts, and docs surfaces into English/Spanish. Two gaps remain:

1. **Metadata is English-only and locale-invariant.** `index.html`'s Open Graph / Twitter Card tags and the PWA manifest (generated by `vite-plugin-pwa` in `vite.config.ts`) are static, baked into the single build at compile time. A Spanish-speaking visitor to `/es/wallet` who shares the link gets an English-language link preview (title/description) and, if they install the PWA, an English app name. There is also no `hreflang` signaling, so search engines have no way to discover the `/es` variants of each page.
2. **Locale selection is 100% client-side.** `LocaleRoutes` (`src/App.tsx`) and `getStoredLocale()`/`storeLocale()` (`src/lib/i18n.ts`) only look at the URL path and `localStorage`. A first-time visitor whose browser sends `Accept-Language: es` still lands on the unprefixed (English) `/` and must manually click the `LocaleSwitcher` to discover Spanish exists. There is no server-side (edge) negotiation redirecting first-visit users to their locale prefix.

### Technical Context (verified against `origin/main` after PR #785)

**Deployment shape**: ONE Cloudflare Pages project (`botho`, see `wrangler.toml`), serving a Vite-built SPA from `dist/`. Two entry documents: `index.html` (main SPA, PWA-enabled) and `operator.html` (SRI-pinned, PWA-excluded — see #772). There is currently **no `functions/` directory and no `_redirects` file** in `web/packages/web-wallet/` — only `public/_headers` (CSP + security headers, no `Cache-Control` overrides). `wrangler` (`^4.59.1`) is already a devDependency, so `wrangler pages dev` is available without adding a new dependency.

**Metadata today** (`web/packages/web-wallet/index.html`): static `<meta>` tags — `og:title`, `og:description`, `og:url` (hardcoded to `https://botho.io/`), `twitter:title`, `twitter:description` — all English, all locale-invariant, baked in at build time (Vite does not template `index.html` per request).

**Manifest today** (`vite.config.ts`, `VitePWA({...})`): a single static `manifest` object (`name: 'Botho Wallet'`, `short_name: 'Botho'`, `description: 'Privacy-first cryptocurrency wallet'`) emitted once at build time as one `manifest.webmanifest` file, referenced identically regardless of locale.

**Locale routing today** (`src/App.tsx`, `src/lib/i18n.ts`, `src/lib/locale-path.ts`): pure client-side. `LocaleRoutes` parses `location.pathname` via `parseLocalePath()`, defaults to `en` (unprefixed) for any unsupported/absent prefix, and persists the active locale to `localStorage` (`botho:locale`) via `storeLocale()`. `SUPPORTED_LOCALES = ['en', 'es']` in `src/lib/i18n.ts` is the single source of truth for supported locales — reuse it rather than hardcoding.

### Implementation Guidance

This issue has two largely independent halves. Consider splitting into two PRs (or two sub-issues) if either grows past the ~400 LOC / 8 file guideline — see "Decomposition" note below.

#### Part A: Localized OG/Twitter/PWA metadata

Because `index.html` is not templated per-request by Vite (it is a static build artifact), per-locale metadata requires one of:

- **Option 1 (recommended): Pages Function rewrites the HTML response.** A `functions/_middleware.ts` (or a route-scoped function) intercepts the document request, fetches the static asset via `context.next()`, and does a targeted string replace of the `<meta>` tag values (title/description/og:url) based on the resolved locale (from the URL path, matching `parseLocalePath` semantics). This keeps the SPA's build-time HTML as the English fallback and only rewrites for non-default locales. Also inject `<link rel="alternate" hreflang="es" href="https://botho.io/es/...">` / `hreflang="en"` / `hreflang="x-default"` tags for the current path so crawlers can discover locale variants.
- **Option 2: Per-locale static HTML files at build time.** Have Vite emit `index.html`, `es/index.html`, etc. with locale-specific meta baked in (multi-entry build, similar to how `operator.html` is already a second Vite entry — see `rollupOptions.input` in `vite.config.ts`). Cloudflare Pages would need `_redirects` rules to map `/es/*` (non-asset paths) to `es/index.html` while preserving the SPA fallback for client-side routing. More build complexity, but zero edge-function runtime cost and no response-rewriting risk.
- Prefer **Option 1** unless the build-time approach turns out simpler in practice — it keeps the existing single-`index.html` SPA model intact and avoids duplicating the multi-entry complexity `operator.html` already introduces.

**PWA manifest**: `vite-plugin-pwa`'s `manifest` option only supports one static manifest per build. Two sub-options:
  - Keep ONE manifest, localize only `name`/`short_name`/`description` to a locale determined at *install time* is not really possible for a static manifest — browsers fetch `/manifest.webmanifest` once per `<link rel="manifest">` URL. Realistic options: (a) leave the manifest in the visitor's browser-negotiated default language (accept this as out of scope — flag as a known limitation), or (b) if per-locale manifest URLs are wanted (`/manifest.webmanifest` vs `/es/manifest.webmanifest`), extend the Pages Function to also rewrite the manifest response's `name`/`short_name`/`description` fields per locale, and rewrite the `<link rel="manifest">` href in the HTML response accordingly. Given PWA install prompts are a small fraction of traffic, **recommend scoping the manifest work to English-only for this issue** (document as a known limitation) unless the Pages Function from Part A already makes rewriting trivial — do not let manifest localization block Part A + B.

#### Part B: Accept-Language edge negotiation

- Implement in a Pages Function (`functions/_middleware.ts`) that runs on document (HTML navigation) requests only — do not touch asset/API requests.
- **First-visit-only redirect**: only redirect when there is no existing signal that the visitor has already made a locale choice. Concretely:
  - Skip the redirect if a cookie (e.g. `botho_locale`) is already set — set this cookie in the middleware itself on every response so the *first* redirect (or explicit non-redirect, e.g. default-locale visitor) marks the visitor as "seen" and prevents repeat redirects/redirect loops on subsequent visits, independent of `localStorage` (edge Functions cannot read `localStorage`).
  - Skip the redirect if the request path already has an explicit locale prefix (e.g. visiting `/es/wallet` directly should never redirect elsewhere) or is the default locale's un-prefixed path where `Accept-Language` also resolves to the default — no redirect needed.
  - Skip (never redirect) requests carrying `Sec-Fetch-Mode: navigate` with `Sec-Fetch-Dest` values used by crawlers, and more importantly **explicitly allowlist-skip known crawler/bot User-Agents** (link-preview unfurlers for Slack/Telegram/Discord/X, and search engine bots) so OG/Twitter unfurls and SEO indexing always see the canonical, requested URL rather than being redirected. A redirect response to a crawler would break unfurling (most preview bots do not follow redirects, or attribute the preview to the pre-redirect URL) and would harm indexing (search engines may not index the redirect target as intended, or could flag it as cloaking). Maintain a small denylist of known bot UA substrings (e.g. `bot`, `crawl`, `spider`, `facebookexternalhit`, `Slackbot`, `TelegramBot`, `Discordbot`, `WhatsApp`, `Twitterbot`) — treat as best-effort, not exhaustive.
  - Parse `Accept-Language` (simple q-value parsing is sufficient; do not add a new dependency for this — a ~20-line parser is enough given only 2 locales), match against `SUPPORTED_LOCALES` from `src/lib/i18n.ts` semantics (mirror the list; a Pages Function is a separate build target so it cannot import from `src/` directly — see Affected Files), and redirect (302, not 301 — this is a preference-based redirect, not a permanent one) to the locale-prefixed equivalent path if it differs from the resolved default.
  - Use `context.next()` to pass through unaffected requests (default-locale match, crawler request, cookie already set, non-HTML request) with minimal added latency.
- **Testability**: extract the negotiation decision (given `Accept-Language` header, cookie presence, User-Agent, and path) as a **pure function** in a plain `.ts` file (no Cloudflare-specific types beyond simple string/Headers-like inputs) so it can be unit-tested with `vitest` exactly like `checkout.ts` in `baas-worker` is tested (`checkout.test.ts` — same "pure testable core, I/O at the edges" pattern already established in this repo). The `functions/_middleware.ts` file itself should be a thin adapter that calls the pure function and translates its decision into a `Response`/`context.next()` call. Additionally, `wrangler pages dev` can be used for manual/e2e-style verification of the actual edge behavior (redirect headers, cookie, 302 vs pass-through) since Pages Functions do not execute in `vite dev`/`vite preview`.

### Risks / Things to Watch

- **Crawler breakage is the highest-severity risk.** If OG/Twitter unfurl bots or search crawlers get redirected, link previews and SEO silently degrade — this is hard to notice locally and easy to miss in e2e (Playwright's default UA is not bot-like, so a passing e2e suite does NOT prove crawlers are exempted). The test plan below requires an explicit bot-UA test case.
- **Cookie-based "first visit" tracking must not create a redirect loop** or leak into caching: if Cloudflare Pages/CDN caches the middleware's HTML response, a cached redirect (or cached rewritten-metadata response) could be served to the wrong visitor. Ensure `Vary: Accept-Language, Cookie` (or `Cache-Control: private`/`no-store` on the negotiated response) is set so shared caches don't serve one visitor's redirect/localized metadata to another. This is a new class of caching pitfall this repo hasn't hit yet (`_headers` currently sets no `Cache-Control` overrides at all — Cloudflare Pages' default HTML caching behavior needs to be checked against whatever the Function returns).
- **`Sec-Fetch-*` heuristics are not universally sent** (older browsers, some bots omit them) — treat as a secondary signal, not the primary gate. User-Agent denylist + cookie state should be the primary mechanism.
- **Manifest localization is a known limitation** if scoped to English-only per the recommendation above — document this explicitly in the PR description so it isn't mistaken for an oversight.
- **This issue does not need to block on new supported locales** — scope is `en`/`es` only, matching `SUPPORTED_LOCALES`.

### Decomposition guidance

If implementation exceeds ~400 LOC or 8 files, split along the Part A / Part B boundary above:
- Sub-issue A: localized OG/Twitter + hreflang (Pages Function metadata rewrite)
- Sub-issue B: Accept-Language edge negotiation (Pages Function redirect + pure negotiation-logic module)

Manifest localization can be dropped from either and tracked as a follow-up if it adds meaningful scope.

## Acceptance Criteria

- [ ] Visiting `/es/wallet` (or any `/es/*` path) returns an HTML response whose `og:title`/`og:description`/`twitter:title`/`twitter:description` are in Spanish, and whose `og:url` matches the requested (locale-prefixed) URL rather than the hardcoded `https://botho.io/`.
- [ ] Visiting the default-locale (`en`, unprefixed) paths continues to return the existing English metadata unchanged (no regression).
- [ ] `hreflang` alternate links (`en`, `es`, `x-default`) are present on locale-aware document responses, pointing at the correct sibling URLs.
- [ ] A first-time visitor (no `botho_locale` cookie) whose `Accept-Language` header resolves to `es` and who requests an unprefixed (default-locale) path is redirected (302) to the `/es`-prefixed equivalent path.
- [ ] A visitor with the `botho_locale` cookie already set is NOT redirected, even if `Accept-Language` differs from their cookie'd/URL locale (respects explicit prior choice).
- [ ] A request with a known crawler/bot User-Agent (test at least one messenger-unfurl bot UA and one search-engine bot UA) is NEVER redirected, regardless of `Accept-Language`, and receives the canonical (requested-path) metadata.
- [ ] A request to an already-locale-prefixed path (e.g. `/es/wallet`) is never redirected elsewhere, regardless of `Accept-Language`.
- [ ] Non-HTML requests (assets, `/rpc`, `/api`, manifest, service worker) are unaffected by the middleware (pass through with no added redirect/rewrite logic).
- [ ] The Accept-Language negotiation decision logic is implemented as a pure, unit-testable function (input: headers/cookie/UA/path primitives; output: redirect target or "no redirect") separate from the Cloudflare Pages Function runtime adapter.
- [ ] Response caching directives (`Cache-Control`/`Vary`) are set appropriately on negotiated/rewritten responses so shared caches cannot serve one visitor's locale-specific response to another visitor.
- [ ] PWA manifest scope is explicitly resolved one way or the other (localized or documented as English-only known limitation) — not silently left ambiguous.

## Affected Files

- `web/packages/web-wallet/index.html` - baseline (English) OG/Twitter meta values; may need minor restructuring so the Pages Function can reliably target tags for rewriting (e.g. stable ordering/format), and to add `hreflang` link placeholders if templated at build time.
- `web/packages/web-wallet/vite.config.ts` - `VitePWA({ manifest: {...} })` config; touch only if manifest localization is in scope for this pass.
- `web/packages/web-wallet/functions/_middleware.ts` - NEW: Cloudflare Pages Function implementing Accept-Language negotiation (redirect) and/or per-locale metadata rewriting; thin adapter over the pure negotiation module.
- `web/packages/web-wallet/functions/lib/locale-negotiation.ts` (or similar path under `functions/`) - NEW: pure, unit-testable negotiation logic (Accept-Language parsing, cookie check, bot UA check, path-prefix check) decoupled from the Cloudflare runtime types. Should mirror `SUPPORTED_LOCALES`/`DEFAULT_LOCALE` semantics from `src/lib/i18n.ts` — consider whether it's feasible to import from `../src/lib/i18n.ts` directly (same package, same TS project) rather than duplicating the locale list, to avoid drift.
- `web/packages/web-wallet/functions/lib/locale-negotiation.test.ts` - NEW: vitest unit tests for the pure negotiation function (see Test Plan).
- `web/packages/web-wallet/public/_headers` - may need `Cache-Control`/`Vary` additions if the Function-rewritten responses need explicit cache directives beyond Cloudflare Pages defaults (Functions responses may bypass `_headers` matching depending on how Cloudflare merges the two — verify during implementation).
- `web/packages/web-wallet/wrangler.toml` - verify no additional config is needed for Pages Functions to be picked up (Cloudflare auto-detects `functions/` at the project root of `pages_build_output_dir`'s sibling — confirm the `functions/` directory location relative to `dist`/project root during implementation, since `pages_build_output_dir = "dist"` but Functions live alongside source, not in `dist`).
- `web/packages/web-wallet/src/lib/i18n.ts` - reference only (source of truth for `SUPPORTED_LOCALES`/`DEFAULT_LOCALE`); do not duplicate this list without a comment cross-referencing it.
- `web/e2e/tests/` - consider adding a Playwright test (or documenting why Playwright cannot exercise this, since Pages Functions do not run under `vite preview`) — see Test Plan for the `wrangler pages dev` alternative.

## Test Plan

- [ ] Automated tests: `web/packages/web-wallet/functions/lib/locale-negotiation.test.ts` (or equivalent) — vitest unit tests covering:
  - `Accept-Language: es` + no cookie + default-locale path → redirect to `/es` equivalent
  - `Accept-Language: en` (or unparseable/absent) + no cookie + default-locale path → no redirect
  - Cookie already set → no redirect regardless of `Accept-Language`
  - Path already locale-prefixed (`/es/...`) → no redirect regardless of `Accept-Language`
  - Known bot/crawler User-Agent → no redirect regardless of `Accept-Language` or cookie state (test at least 2 UA strings from the denylist)
  - q-value parsing: `Accept-Language: fr;q=0.9,es;q=0.5` picks the highest-weighted *supported* locale (`es`), not the highest-weighted locale overall (`fr`, unsupported)
- [ ] Manual verification: run `wrangler pages dev dist` (after `pnpm build`) from `web/packages/web-wallet/`, and use `curl -H "Accept-Language: es"` / `curl -A "<bot-ua>"` against the local Pages dev server to confirm redirect/pass-through/metadata-rewrite behavior end-to-end (Playwright/`vite preview` cannot exercise Pages Functions).
- [ ] Manual verification: view-source (or `curl`) `/` and `/es` and confirm `og:title`/`og:description`/`twitter:*`/`hreflang` differ correctly between the two, and that `/wallet` vs `/es/wallet` og:url reflects the correct path.
- [ ] Regression: existing e2e suite (`pnpm test:e2e:smoke` at minimum) still passes — confirms the default (English, unprefixed) experience and existing locale-switcher behavior (`LocaleSwitcher.tsx`, `parseLocalePath`/`buildLocalePath`) are unaffected by the new edge layer.
- [ ] Edge cases: visitor with no `Accept-Language` header at all (some privacy-focused browsers/proxies strip it) → falls back to default locale, no redirect, no crash.

rjwalters:
## Summary

Phase 4 of the i18n rollout (#779). Adds a Cloudflare Pages Function for the `botho` Pages project (net-new `functions/` — the site had none) that, for HTML document navigations only, (A) localizes OG/Twitter/description metadata + injects `hreflang`/`canonical` per locale, and (B) does first-visit `Accept-Language` edge negotiation. All decision logic lives in a pure, unit-tested module (`functions/lib/locale-negotiation.ts`), mirroring the `baas-worker/src/checkout.ts` "pure core, I/O at the edges" pattern; `_middleware.ts` is a thin adapter.

## Changes

- **NEW `functions/lib/locale-negotiation.ts`** — pure core: `parseAcceptLanguage` (q-value parser, no dep), `isBotUserAgent` (crawler/unfurl denylist), `readCookie`, `parseLocaleFromPath`/`buildLocalePath`, `negotiateLocaleRedirect` (the redirect decision), and `localizeDocumentHtml` (per-locale meta + hreflang rewrite).
- **NEW `functions/_middleware.ts`** — thin runtime adapter: gates on document requests, calls the pure core, stitches the decision onto a real `Response` (302 + cookie, or HTML rewrite), passes assets/`/rpc`/`/api`/manifest/SW/operator through untouched.
- **NEW `functions/lib/locale-negotiation.test.ts`** — 29 vitest unit tests.
- **NEW `src/lib/locales.ts`** — dependency-free leaf module with `SUPPORTED_LOCALES`/`DEFAULT_LOCALE`/`isSupportedLocale`; `src/lib/i18n.ts` now re-exports from it so the edge Function shares the exact same locale list without importing the i18next runtime (no drift). Public API of `./i18n` is unchanged.
- **`tsconfig.json`** — include `functions` so the Pages Function is typechecked by `pnpm typecheck`.

| `/es/*` returns Spanish og/twitter title+description; og:url = requested URL | ✅ | `curl /es/wallet` via `wrangler pages dev`: Spanish copy, `og:url=https://botho.io/es/wallet` |
| Default (en, unprefixed) metadata unchanged | ✅ | `curl /wallet`: English copy preserved; og:url corrected to `/wallet` |
| `hreflang` (en/es/x-default) present on locale-aware responses | ✅ | curl output shows all three `<link rel="alternate">` + `rel="canonical"` |
| First-visit `Accept-Language: es` on default path → 302 to `/es` equiv | ✅ | `curl -H "Accept-Language: es" /wallet` → `302 Location: /es/wallet` |
| Cookie set → never redirected regardless of Accept-Language | ✅ | `curl -H "Cookie: botho_locale=en" -H "Accept-Language: es"` → 200 |
| Known crawler/bot UA never redirected (messenger + search bot tested) | ✅ | `facebookexternalhit` and `Googlebot` both → 200, no redirect |
| Already-`/es/*` path never redirected | ✅ | `curl /es/wallet` (en Accept-Language) → 200 |
| Non-HTML requests (assets, manifest, /rpc, SW) unaffected | ✅ | `favicon.png` + `manifest.webmanifest` pass through, no rewrite/cookie |
| Negotiation logic is a pure, unit-testable function | ✅ | `negotiateLocaleRedirect` / `localizeDocumentHtml` — 29 tests |
| Cache-Control/Vary set so shared caches can't cross-serve | ✅ | negotiated responses carry `Cache-Control: private, no-store` + `Vary: Accept-Language, Cookie` |
| PWA manifest scope explicitly resolved | ✅ | Scoped English-only (documented limitation below) |

## Test Plan

- **Unit**: `vitest run` on `functions/lib/locale-negotiation.test.ts` — 29 passing, covering q-value parsing (highest-weighted *supported* locale), absent/empty/unsupported/wildcard headers, `q=0` drop, bot UA exemption (Facebook/Twitter/Slack/Telegram/Discord/Google/Bing/WhatsApp), cookie suppression, locale-prefixed path suppression, no-Accept-Language fallback, and the full HTML rewrite (Spanish meta, og:url, `<html lang>`, hreflang count, English no-regression, attribute escaping).
- **Full suite**: `pnpm check:ci` green — 774 tests pass, typecheck clean across all packages, baas-worker wrangler dry-run OK.
- **End-to-end edge**: `pnpm build` then `wrangler pages dev dist`; curl'd every acceptance case above against the live Pages dev server (Pages Functions do not run under `vite preview`/Playwright, so this is the authoritative edge verification). All behaved as specified.

## Known limitation (deferred)

PWA manifest localization is intentionally **English-only** for this pass. A static `manifest.webmanifest` is fetched once per `<link rel="manifest">` URL, so per-locale install names would require per-locale manifest URLs + rewriting the `<link>` href — low value for the small PWA-install traffic slice. Flagged as a follow-up rather than blocking the higher-value OG/Twitter + negotiation work, per the issue's decomposition guidance.

## Deviations from the curator's plan

- Chose **Option 1** (Pages Function response rewrite) over per-locale static builds, as recommended.
- Added `src/lib/locales.ts` leaf module (touching `i18n.ts` via re-export, both listed affected files) to genuinely eliminate locale-list drift rather than duplicating the list with a comment.
- Reverted an unrelated `pnpm install` side-effect to `web/pnpm-workspace.yaml` (allowBuilds scaffold) to keep scope clean.

Closes #779

### DIFF
--- a/web/packages/web-wallet/functions/_middleware.ts
+++ b/web/packages/web-wallet/functions/_middleware.ts
@@ -1,0 +1,143 @@
+/**
+ * Cloudflare Pages Function middleware (issue #779) — the thin runtime adapter
+ * over the pure logic in `lib/locale-negotiation.ts`. It runs on every request
+ * to the `botho` Pages project and does two things for HTML document
+ * navigations only:
+ *
+ *   Part A — localize the OG/Twitter/description meta + inject hreflang/canonical
+ *            alternates into the served `index.html`, per the resolved locale.
+ *   Part B — first-visit Accept-Language negotiation: 302-redirect a fresh
+ *            visitor whose browser prefers a supported non-default locale to the
+ *            locale-prefixed equivalent path, once (guarded by a cookie).
+ *
+ * Everything decision-shaped lives in the pure module and is unit-tested; this
+ * file only reads headers, calls `next()`, and stitches the decision onto a
+ * real `Response`. Non-document requests (assets, /rpc, /api, manifest, service
+ * worker, the operator entry) pass straight through untouched.
+ *
+ * Caching: negotiated / rewritten HTML responses carry `Vary: Accept-Language,
+ * Cookie` and `Cache-Control: private, no-store` so a shared cache can never
+ * serve one visitor's locale-specific document or redirect to another visitor.
+ */
+import {
+  LOCALE_COOKIE_NAME,
+  localizeDocumentHtml,
+  negotiateLocaleRedirect,
+  parseLocaleFromPath,
+} from './lib/locale-negotiation'
+
+// Minimal shape of the Cloudflare Pages Functions context we use. Declared
+// locally (rather than depending on `@cloudflare/workers-types`) to keep this
+// package's devDependencies unchanged — the fields below are stable public API.
+interface PagesContext {
+  request: Request
+  next: () => Promise<Response>
+}
+
+/** Cookie attributes: 1-year, site-wide, lax — a preference marker, not a secret. */
+const COOKIE_MAX_AGE = 60 * 60 * 24 * 365
+function localeCookie(locale: string): string {
+  return (
+    `${LOCALE_COOKIE_NAME}=${locale}; Path=/; Max-Age=${COOKIE_MAX_AGE}; ` +
+    'SameSite=Lax; Secure'
+  )
+}
+
+/**
+ * Does this request look like a top-level HTML document navigation we should
+ * localize / negotiate? We deliberately gate on the `Accept` header and path so
+ * asset/API/manifest/service-worker requests pass straight through.
+ */
+function isDocumentRequest(request: Request, pathname: string): boolean {
+  // Never touch the SRI-pinned operator entry or its assets (#772).
+  if (pathname === '/operator' || pathname === '/operator.html' || pathname.startsWith('/operator/')) {
+    return false
+  }
+  // Skip well-known non-document roots.
+  if (
+    pathname.startsWith('/rpc') ||
+    pathname.startsWith('/api') ||
+    pathname.startsWith('/pkg/') ||
+    pathname.startsWith('/assets/') ||
+    pathname === '/manifest.webmanifest' ||
+    pathname === '/sw.js' ||
+    pathname === '/registerSW.js'
+  ) {
+    return false
+  }
+  // Skip anything with a file extension (favicon.ico, *.png, *.svg, *.pdf, ...).
+  const lastSegment = pathname.slice(pathname.lastIndexOf('/') + 1)
+  if (lastSegment.includes('.')) {
+    return false
+  }
+  // Only treat requests that accept HTML as document navigations.
+  const accept = request.headers.get('Accept') ?? ''
+  return accept.includes('text/html') || accept === '' || accept.includes('*/*')
+}
+
+export const onRequest = async (context: PagesContext): Promise<Response> => {
+  const { request, next } = context
+  const url = new URL(request.url)
+  const pathname = url.pathname
+
+  if (request.method !== 'GET' && request.method !== 'HEAD') {
+    return next()
+  }
+  if (!isDocumentRequest(request, pathname)) {
+    return next()
+  }
+
+  const headers = request.headers
+  const decision = negotiateLocaleRedirect({
+    pathname,
+    acceptLanguage: headers.get('Accept-Language'),
+    cookie: headers.get('Cookie'),
+    userAgent: headers.get('User-Agent'),
+  })
+
+  // Part B: first-visit redirect.
+  if (decision.kind === 'redirect') {
+    const location = url.origin + decision.location + url.search
+    const res = new Response(null, {
+      status: 302,
+      headers: {
+        Location: location,
+        'Set-Cookie': localeCookie(decision.cookieLocale),
+        // A preference redirect must never be cached and shared across visitors.
+        'Cache-Control': 'private, no-store',
+        Vary: 'Accept-Language, Cookie',
+      },
+    })
+    return res
+  }
+
+  // Fetch the underlying static asset (index.html for SPA routes).
+  const response = await next()
+
+  // Only rewrite HTML documents; pass everything else (e.g. the SW fallback
+  // serving a non-HTML asset) through untouched.
+  const contentType = response.headers.get('Content-Type') ?? ''
+  if (!contentType.includes('text/html')) {
+    return response
+  }
+
+  const { locale } = parseLocaleFromPath(pathname)
+  const originalHtml = await response.text()
+  const localized = localizeDocumentHtml(originalHtml, locale, pathname)
+
+  const newHeaders = new Headers(response.headers)
+  // Locale-specific document: keep shared caches from cross-serving it, and make
+  // sure any cache keys on the negotiating inputs.
+  newHeaders.set('Cache-Control', 'private, no-store')
+  newHeaders.append('Vary', 'Accept-Language')
+  newHeaders.append('Vary', 'Cookie')
+  if (decision.kind === 'pass' && decision.setCookieLocale) {
+    newHeaders.append('Set-Cookie', localeCookie(decision.setCookieLocale))
+  }
+
+  return new Response(localized, {
+    status: response.status,
+    statusText: response.statusText,
+    headers: newHeaders,
+  })
+}
--- a/web/packages/web-wallet/functions/lib/locale-negotiation.test.ts
+++ b/web/packages/web-wallet/functions/lib/locale-negotiation.test.ts
@@ -1,0 +1,307 @@
+/**
+ * Unit tests for the pure edge locale-negotiation core (issue #779). These
+ * exercise the decision logic and the HTML rewrite WITHOUT any Cloudflare
+ * runtime, mirroring the "pure testable core, I/O at the edges" pattern of
+ * `baas-worker/src/checkout.test.ts`.
+ */
+import { describe, it, expect } from 'vitest'
+
+import {
+  BOT_UA_SUBSTRINGS,
+  buildLocalePath,
+  isBotUserAgent,
+  LOCALE_COOKIE_NAME,
+  LOCALE_METADATA,
+  localizeDocumentHtml,
+  negotiateLocaleRedirect,
+  parseAcceptLanguage,
+  parseLocaleFromPath,
+  readCookie,
+  SITE_ORIGIN,
+  SUPPORTED_LOCALES,
+} from './locale-negotiation'
+
+describe('parseAcceptLanguage', () => {
+  it('returns the highest-q supported locale', () => {
+    expect(parseAcceptLanguage('es')).toBe('es')
+    expect(parseAcceptLanguage('en-US,en;q=0.9')).toBe('en')
+  })
+
+  it('ignores region subtags (es-MX matches es)', () => {
+    expect(parseAcceptLanguage('es-MX')).toBe('es')
+    expect(parseAcceptLanguage('en-GB,en;q=0.8')).toBe('en')
+  })
+
+  it('picks the highest-weighted SUPPORTED locale, skipping unsupported ones', () => {
+    // fr is unsupported and highest-weighted; es is the best supported match.
+    expect(parseAcceptLanguage('fr;q=0.9,es;q=0.5')).toBe('es')
+    expect(parseAcceptLanguage('de,fr;q=0.7,es;q=0.3')).toBe('es')
+  })
+
+  it('honours q-value ordering over header order', () => {
+    expect(parseAcceptLanguage('en;q=0.4,es;q=0.9')).toBe('es')
+    expect(parseAcceptLanguage('es;q=0.2,en;q=0.9')).toBe('en')
+  })
+
+  it('treats a bare wildcard as the default locale', () => {
+    expect(parseAcceptLanguage('*')).toBe('en')
+  })
+
+  it('returns undefined for absent / empty / unsupported-only / unparseable headers', () => {
+    expect(parseAcceptLanguage(undefined)).toBeUndefined()
+    expect(parseAcceptLanguage(null)).toBeUndefined()
+    expect(parseAcceptLanguage('')).toBeUndefined()
+    expect(parseAcceptLanguage('fr,de;q=0.8')).toBeUndefined()
+    expect(parseAcceptLanguage(';;;')).toBeUndefined()
+  })
+
+  it('drops q=0 entries', () => {
+    expect(parseAcceptLanguage('es;q=0,en;q=0.5')).toBe('en')
+    expect(parseAcceptLanguage('es;q=0')).toBeUndefined()
+  })
+})
+
+describe('isBotUserAgent', () => {
+  it('matches known crawler / unfurl UAs (case-insensitive)', () => {
+    expect(isBotUserAgent('facebookexternalhit/1.1')).toBe(true)
+    expect(isBotUserAgent('Twitterbot/1.0')).toBe(true)
+    expect(isBotUserAgent('Slackbot-LinkExpanding 1.0')).toBe(true)
+    expect(isBotUserAgent('TelegramBot (like TwitterBot)')).toBe(true)
+    expect(isBotUserAgent('Discordbot/2.0')).toBe(true)
+    expect(isBotUserAgent('Mozilla/5.0 (compatible; Googlebot/2.1)')).toBe(true)
+    expect(isBotUserAgent('Mozilla/5.0 (compatible; bingbot/2.0)')).toBe(true)
+    expect(isBotUserAgent('WhatsApp/2.19')).toBe(true)
+  })
+
+  it('does not match ordinary browser UAs', () => {
+    expect(
+      isBotUserAgent(
+        'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Version/17.0 Safari/605.1.15',
+      ),
+    ).toBe(false)
+    expect(isBotUserAgent('')).toBe(false)
+    expect(isBotUserAgent(null)).toBe(false)
+    expect(isBotUserAgent(undefined)).toBe(false)
+  })
+
+  it('has a non-empty denylist', () => {
+    expect(BOT_UA_SUBSTRINGS.length).toBeGreaterThan(5)
+  })
+})
+
+describe('readCookie', () => {
+  it('extracts a named cookie value', () => {
+    expect(readCookie('a=1; botho_locale=es; b=2', 'botho_locale')).toBe('es')
+    expect(readCookie('botho_locale=en', 'botho_locale')).toBe('en')
+  })
+
+  it('returns undefined when absent', () => {
+    expect(readCookie('a=1; b=2', 'botho_locale')).toBeUndefined()
+    expect(readCookie('', 'botho_locale')).toBeUndefined()
+    expect(readCookie(null, 'botho_locale')).toBeUndefined()
+  })
+})
+
+describe('parseLocaleFromPath', () => {
+  it('extracts an explicit non-default prefix', () => {
+    expect(parseLocaleFromPath('/es/wallet')).toEqual({ locale: 'es', rest: '/wallet' })
+    expect(parseLocaleFromPath('/es')).toEqual({ locale: 'es', rest: '/' })
+  })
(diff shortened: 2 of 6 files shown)


rjwalters:
✅ **Approved!**

Thorough review of this net-new edge middleware on the production site (botho.io). Given it sits in front of every request, I scrutinized correctness, caching, HTML-rewrite safety, and security. All clear.

## Correctness — redirect decision logic
- **q-value Accept-Language parsing** (`parseAcceptLanguage`): stable descending-q sort with explicit index tie-break, `q=0` dropped, wildcard `*` → default, region subtags folded to base language, returns only a member of the fixed `SUPPORTED_LOCALES` set. Verified via the unit tests (q ordering, unsupported-only → undefined, `;;;` unparseable → undefined).
- **Bot/crawler exemption**: denylist covers Googlebot/bingbot/slurp + facebookexternalhit, Twitterbot, Slackbot, TelegramBot, Discordbot, WhatsApp, LinkedIn, Pinterest, Reddit, Embedly, plus `google-inspectiontool` and headless Chrome. Case-insensitive substring match. Checked first, so a preview bot is never redirected (protects unfurls/SEO).
- **Cookie first-visit suppression**: presence check only; the cookie value is never trusted or echoed.
- **No redirect loops**: only unprefixed (default-locale) paths negotiate; `/es/*` hits the prefix gate → pass. Redirect target is always a `/es/*` path which then passes. Trailing-slash normalization in `parseLocaleFromPath` keeps the target stable.
- **Passthrough**: `/operator*` (SRI-pinned, #772), `/rpc`, `/api`, `/pkg/`, `/assets/`, `manifest.webmanifest`, `sw.js`, `registerSW.js`, and any dotted path are excluded before negotiation — SW/manifest can't be redirected, so the PWA is safe.
- **302 + Set-Cookie semantics**: correct 302 with `Location` (origin-anchored), `Set-Cookie` (`SameSite=Lax; Secure; Path=/; Max-Age=1y`).

## Caching — shared-cache poisoning
Both the 302 and every localized HTML response carry `Cache-Control: private, no-store` + `Vary: Accept-Language, Cookie`. This fully prevents Cloudflare's shared cache from cross-serving a Spanish redirect/document to an English visitor. Conservative (HTML shell is not CDN-cached at all) but correct — the right call for a locale-varying response. Optional future optimization only; not blocking.

## HTML-rewrite safety
- `localizeDocumentHtml` is a pure string transform, run on the freshly-fetched static document each request → idempotent (hreflang/canonical injected exactly once).
- `escapeAttr` escapes `& " < >` for double-quoted attributes.
- **Byte-stability verified against the real build**: `dist/index.html` (post-Vite) preserves the exact `property="..." content="..."` attribute order and single-space separators the rewrite regexes target, and `LOCALE_METADATA.en` matches the document's OG/Twitter/description copy verbatim. English pages are unchanged except the intended og:url correction (`https://botho.io/` → real requested path) and the hreflang/canonical block.

## Security (middleware in front of everything)
- **No open redirect**: redirect target is `url.origin + buildLocalePath(...) + url.search`; the path is reconstructed from the site's own `SUPPORTED_LOCALES`, never from attacker-controlled header/path content.
- **No header/HTML injection via Accept-Language**: the header is reduced to a `'en'|'es'` enum before any use — the raw value never reaches HTML, cookie, or `Location`.
- No ReDoS: linear splits + literal-anchored `[^"]*` regexes.

## Deferred PWA manifest localization
English-only manifest confirmed as the sanctioned deferral; the middleware correctly passes `manifest.webmanifest` through untouched.

## Test Execution
- **Unit**: `npx vitest run functions/lib/locale-negotiation.test.ts` — ✅ 29/29 passing.
- **Typecheck**: `tsc --noEmit` with `functions` now in tsconfig `include` — ✅ clean.
- **CI**: `gh pr checks` — ✅ both `cargo deny` and `typecheck + vitest + worker dry-run` pass. Merge state CLEAN.
- **Edge e2e** (`wrangler pages dev` curl matrix) — ⚠️ Not re-executed locally (Pages Functions don't run under vitest/Playwright); PR author's documented curl verification of every acceptance case is accepted as the authoritative edge check.

PR body uses `Closes #779`. Clean scope, pure-core/thin-adapter structure mirrors the baas-worker pattern well. Nice work.

PR #798, opened 2026-07-13
fix(web): landing header overflow, nav wrap, locale switcher, /en orphan (#797)

### LINKED ISSUE #797: web: landing header overflows before hamburger breakpoint; nav labels wrap; locale switcher shows wrong language
User-reported (2026-07-12, observed live on botho.io /es after the i18n phase 2-4 deploy). Screenshot evidence shows the Spanish landing page header with three distinct problems, plus a fourth reported the same day: `/en` renders a blank page.

## 1. Header crowding — hamburger breakpoint fires too late
With Spanish labels (longer than English: "Explorador", "Alojar un nodo", "Informe técnico"), the desktop nav overflows/crushes before the responsive collapse into the hamburger menu kicks in. The breakpoint was presumably tuned for English label widths.

**Wanted behavior:** show the hamburger dropdown EARLIER and only pull links out into the toolbar as screen width allows (progressive disclosure by available width, not a single fixed breakpoint tuned to English). At minimum, raise the collapse breakpoint so the crowded intermediate state can't occur in either locale; ideally make the threshold locale-aware or measurement-based.

## 2. Nav label line-wrap
"Alojar un nodo" renders on two lines in the toolbar ("Alojar un / nodo"), as do "Informe técnico" and the wallet CTA "Abrir monedero" in the screenshot. Nav items and the CTA must never wrap: `whitespace-nowrap` on nav link labels and the CTA (see `node.tsx:118`'s `whitespace-nowrap` on the hero price for the established pattern — the issue originally cited a different commit; corrected below).

## 3. Locale switcher does not reflect the active language
On the Spanish page (/es, all content Spanish), the switcher widget displays "English" with a globe icon.

## 4. /en renders a blank page (added 2026-07-12, user report)
Live evidence: `https://botho.io/en` and `https://botho.io/en/wallet` return HTTP 200 with the SPA shell, but the client router matches no route for the `/en` prefix → blank page. English is unprefixed by design (phase 1, PR #774), so `/en` is an orphan namespace.

Checked and NOT affected: the edge-injected hreflang correctly emits `hreflang="en" href="https://botho.io/"` (unprefixed) and canonical on /es is correct — crawlers are not being steered to /en.

## Scope
All four are in the landing/header + locale-routing surface of web/packages/web-wallet (Header/nav component, LocaleSwitcher, locale-path routing, edge middleware). One PR is appropriate — they share the header/locale surface and items 3+4 share a root cause. Related: PR #774 (phase 1 header/switcher), PR #786 (edge negotiation), #794 (adds LocaleSwitcher to more pages — coordinate, this issue is the switcher's correctness, #794 is its placement).

---

### Root cause findings

**Item 1 — breakpoint.** `landing.tsx` has exactly one responsive breakpoint for the header: `hidden md:flex` on the desktop `<nav>` (line 35) and `md:hidden` on the hamburger button (line 78) — a single fixed 768px (`md:`) Tailwind breakpoint with no locale awareness and no intermediate/progressive state. The desktop nav renders 6 items (Explorer, Network, Docs, Host a Node, Whitepaper, GitHub) + LocaleSwitcher + CTA button in one `flex items-center gap-8` row inside a `max-w-6xl` container — there is no room to spare, so ES's longer labels ("Alojar un nodo", "Informe técnico") overflow/crush at widths just above 768px where EN still fits.

**Recommended fix (simplest robust option): raise the single breakpoint from `md:` (768px) to `lg:` (1024px).** This is a one-line change (`hidden md:flex` → `hidden lg:flex`, `md:hidden` → `lg:hidden`) that removes the crowded intermediate state for both locales without adding measurement/ResizeObserver complexity. A locale-aware or width-measured progressive-disclosure nav is real defense-in-depth but is a materially larger change (state, layout measurement, possibly a "More" overflow menu) — out of scope for this bug-fix PR; file as a follow-up if the `lg:` breakpoint still crowds ES on real-world viewports between 1024–1280px after this fix ships.

**Item 2 — wrapping.** None of the desktop-nav `<Link>`/`<a>` elements (lines 36-68), the CTA `<Button>` (line 71), or their mobile-menu counterparts (lines 89-138) currently have `whitespace-nowrap`. Confirmed wrap-prone strings from `locales/es/landing.json`: `nav.hostNode` = "Alojar un nodo", `nav.whitepaper` = "Informe técnico", `nav.launchWallet` = "Abrir monedero" (used as the CTA button's `{t('nav.launchWallet')}` child). `nav.explorer` = "Explorador" and `nav.network` = "Red" are shorter and lower-risk but should get the same treatment for consistency/defense against future longer translations. Established pattern for this exact class of fix already exists in the codebase: `web/packages/web-wallet/src/pages/node.tsx:118` — `<span className="text-light whitespace-nowrap">{t('checkout.heroPrice')}</span>`.

**Fix:** add `whitespace-nowrap` to each nav `<Link>`/`<a>` className and to the `<Button>` (or its child span) in both the desktop nav block and the mobile-menu block.

**Item 3 — locale switcher label.** Read `LocaleSwitcher.tsx` in full: `activeLocale` is derived synchronously from `parseLocalePath(location.pathname)` (line 30) — i.e. purely from the URL, not from i18next's async init state or localStorage. On a direct `/es` load, `parseLocalePath('/es')` correctly returns `{ locale: 'es', rest: '/' }`, so `<select value="es">` should render "Español" — this part of the mechanism is *not* buggy in isolation, and it is not racing the i18next bootstrap (no dependency on `i18n.language` at all — `LocaleRoutes` in `App.tsx` syncs `i18n.language` off the same `parseLocalePath` result via a `useEffect`, but `LocaleSwitcher` doesn't read that state).

**Actual mechanism (confirmed via item 4's shared root cause):** the switcher's own `handleChange`/`switchLocaleInPath` round-trip is correct (`switchLocaleInPath('/es/wallet', 'en')` → `/wallet`, verified in `locale-path.test.ts`). The plausible concrete path to the observed "English" label on a Spanish page is a **prior mis-navigation to a `/en`-prefixed URL** (item 4): if the user reached `/es` after bouncing through `/en/...` (e.g., an external link, a stale bookmark, or the edge Accept-Language redirect target computed from a `/en` starting path), `parseLocalePath` would have derived `locale: 'en'` for that page (since `/en` is not treated as a locale-prefix by design — see item 4), and if that state leaked into a subsequent switcher render before the URL fully settled to `/es`, the label would show "English" while page content (driven by whatever bundle/namespace was actually loaded) looked Spanish. This is corroborated by the issue's own comment #4, which explicitly links items 3 and 4 ("the widget and the path logic disagree about how the default locale is represented"). **Conclusion: item 3 has no independent code defect to fix beyond what item 4 fixes** — the switcher's `value={activeLocale}` binding is provably correct for well-formed URLs (`/`, `/es`, `/wallet`, `/es/wallet`, ...). Fixing item 4 (below) removes the malformed `/en` URLs that are the only demonstrated way to desync the switcher's label from rendered content. The acceptance criteria below require explicitly re-verifying the switcher label is correct after the item-4 fix, in case a residual defect surfaces.

**Item 4 — /en blank page.** Confirmed root cause by reading `parseLocalePath` (`locale-path.ts:20-35`) and `AppRoutes`/`LocaleRoutes` (`App.tsx:61-128`):

- `parseLocalePath('/en')`: `segments[1] = 'en'`; `isSupportedLocale('en')` is `true`, but the guard is `first !== DEFAULT_LOCALE` and `DEFAULT_LOCALE === 'en'`, so the guard is `false` → falls through to `return { locale: DEFAULT_LOCALE, rest: pathname || '/' }`, i.e. `rest = '/en'` (unchanged, NOT stripped). **This is intentional, existing, tested behavior** — `locale-path.test.ts` has an explicit case: `"does NOT treat an explicit /en prefix as a locale segment"` asserting `parseLocalePath('/en/wallet') === { locale: 'en', rest: '/en/wallet' }`. So `/en` is by design a normal (non-locale) route path, not an alias for `/`.
- The bug is that **nothing renders anything at that route.** `LocaleRoutes` sees `locale === DEFAULT_LOCALE` and renders `<AppRoutes />` directly against the real (unstripped) location, i.e. against `pathname = '/en'`. `AppRoutes`'s `<Routes>` table (`App.tsx:63-85`) has no route matching `/en` and **no catch-all/404 route at all** — React Router renders nothing when no `<Route>` matches, producing the blank page.
- Same mechanism in the Cloudflare Pages Function: `parseLocaleFromPath` in `functions/lib/locale-negotiation.ts` is a byte-for-byte reimplementation of the same logic (by design, per its own doc comment, so the edge doesn't bundle the SPA's i18next-adjacent module) and has the identical `/en` non-stripping behavior. Combined with `negotiateLocaleRedirect`'s step 4 (Accept-Language negotiation fires whenever `pathLocale === DEFAULT_LOCALE`, which is true for `/en` since it isn't recognized as an explicit prefix), a first-time Spanish-preferring visitor hitting `/en` with no cookie would get redirected to `buildLocalePath('es', '/en')` = `/es/en` — compounding the orphan-namespace bug rather than fixing it. No existing test in `locale-negotiation.test.ts` covers a `/en` input.
- Confirmed the switcher itself does NOT currently emit `/en/...` paths for the default locale — `buildLocalePath('en', rest)` (`locale-path.ts` / mirrored in `locale-negotiation.ts`) always returns the unprefixed `rest`, never `/en/...`. So the switcher is not the entry point into `/en`; likely entry points are external/stale links, manual URL entry, or a bookmarked link from before phase-1 shipped. Regardless, the issue's requested defense-in-depth (edge 301) is warranted since `/en` will keep circulating.

**Fix (three parts, as scoped in comment #4):**
(a) No switcher change needed per the above finding — verify with a regression test instead (see Acceptance Criteria).
(b) Add an edge-level 301 in `functions/_middleware.ts` (or a small helper in `locale-negotiation.ts`, consistent with that module's pure-function-testable style): when `pathname === '/en'` or `pathname.startsWith('/en/')`, redirect (301, permanent — this is a URL-shape correction, not a preference negotiation) to the same path with the `/en` prefix stripped (`/en` → `/`, `/en/wallet` → `/wallet`), preserving query string. This must run before/instead of the existing Accept-Language negotiation step for that path so it doesn't get double-redirected to `/es/en`.
(c) Add a client-side fallback: either treat `/en` the same as the edge redirect in `AppRoutes` (e.g. a lightweight `<Route path="/en/*" element={<Navigate to={stripEnPrefix(pathname)} replace />} />`), or add a genuine catch-all `<Route path="*">` that at minimum redirects `/en/*` and otherwise avoids a silent blank page for any unmatched path. Given the edge 301 is the primary fix and CDN-level redirects apply before the SPA loads on a fresh navigation, the client-side piece is defense-in-depth for client-side navigations that construct a `/en` URL without a full page load (there are currently none identified, but a catch-all avoids the blank-page failure mode class entirely, not just this instance).

### Acceptance Criteria

- [ ] Desktop nav collapses to the hamburger before any nav item can crush or overflow, in **both** `en` and `es`, at the chosen breakpoint (recommended: raise `md:`→`lg:` in `landing.tsx`).
- [ ] No nav label or CTA text wraps to a second line at any viewport ≥ the hamburger threshold, in both locales — specifically verify "Alojar un nodo", "Informe técnico", "Abrir monedero" (ES) and their EN equivalents render on one line.
- [ ] A viewport sweep (e.g. 768px, 900px, 1024px, 1152px, 1280px, 1440px) asserts the header never overflows horizontally (no horizontal scrollbar / no clipped nav item) in both locales.
- [ ] LocaleSwitcher's displayed label equals the actually-rendered language after: (a) a direct `/es` load, (b) a direct `/en` load, (c) simulating the edge-redirect first-visit landing on `/es`, (d) an in-session switch en→es and es→en via the switcher control.
- [ ] Direct navigation to `/en` (fresh load, no client-side routing state) renders the English landing page — not a blank page.
- [ ] Direct navigation to `/en/wallet` (and at least one other `/en/<route>`) renders the corresponding English page.
- [ ] Edge 301: `GET /en` and `GET /en/wallet` (with a cookie or bot UA that would otherwise bypass negotiation) return a 301 to the unprefixed equivalent, not a 302 negotiation redirect and not a pass-through.
- [ ] `/en?<query>` and `/en/wallet?<query>` preserve the query string across the 301.
- [ ] Existing `locale-path.test.ts` case "does NOT treat an explicit /en prefix as a locale segment" still passes unchanged (that assertion is about `parseLocalePath`'s routing semantics, not the new redirect behavior — the redirect is a separate layer, not a change to `parseLocalePath`).
- [ ] No regression to `/xx` (unsupported-locale) fallback-to-English behavior, `/es/*` routing, or the existing edge negotiation test suite in `locale-negotiation.test.ts`.

### Affected Files

- `web/packages/web-wallet/src/pages/landing.tsx` — raise the `md:`/`hidden md:flex` breakpoint to `lg:` for the desktop-nav/hamburger split (lines ~35, ~78); add `whitespace-nowrap` to nav `<Link>`/`<a>` elements (lines ~36-68) and to the CTA `<Button>`/its child (lines ~70-72), and mirror in the mobile-menu list (lines ~89-138) where relevant (mobile menu is a vertical list so wrap is less likely but the CTA button text should still get `whitespace-nowrap` for consistency).
- `web/packages/web-wallet/src/pages/landing.test.tsx` — add/extend tests for: no-wrap assertion is hard to unit-test meaningfully in jsdom (no real layout), so focus vitest coverage on the LocaleSwitcher-label acceptance criteria (direct `/es`, direct `/en`, in-session switch) rather than visual wrap; leave the viewport/overflow sweep to a Playwright e2e test (see below).
- `web/packages/web-wallet/src/components/LocaleSwitcher.tsx` — add a regression test (or extend `landing.test.tsx`) proving the label matches `activeLocale` for `/en` inputs post-fix; no production code change anticipated here per the root-cause finding, but re-verify after the `/en` fix lands — if the label still desyncs, this file needs a fix (e.g. deriving `activeLocale` more defensively) and the "no independent defect" conclusion above should be revisited.
- `web/packages/web-wallet/src/lib/locale-path.ts` — no change to `parseLocalePath`'s documented `/en` semantics (existing tests rely on non-stripping); this file is a reference point only unless the client-side catch-all/redirect fix (item 4c) is implemented as a helper here.
- `web/packages/web-wallet/src/lib/locale-path.test.ts` — add a test (or a colocated new test file) for whatever client-side `/en` handling is added (catch-all route or explicit redirect), separate from the existing `parseLocalePath` semantics tests which must not change.
- `web/packages/web-wallet/src/App.tsx` — add either a `<Route path="/en/*">` redirect (stripping the `/en` prefix and re-navigating) or a general catch-all `<Route path="*">` inside `AppRoutes` to eliminate the blank-page failure mode.
- `web/packages/web-wallet/functions/_middleware.ts` — add the `/en` / `/en/*` 301 redirect, ordered ahead of (or as a special case within) the existing `negotiateLocaleRedirect` call so it isn't compounded into `/es/en`.
- `web/packages/web-wallet/functions/lib/locale-negotiation.ts` — add the pure decision logic for the `/en` 301 (consistent with this module's existing pure-function/unit-testable style — e.g. a new exported function or an additional case in `negotiateLocaleRedirect`'s decision table), including query-string preservation.
- `web/packages/web-wallet/functions/lib/locale-negotiation.test.ts` — add unit tests for the new `/en` → `/` and `/en/wallet` → `/wallet` 301 behavior, including query-string preservation and interaction ordering with Accept-Language negotiation (the 301 must win, not stack with a 302 to `/es/en`).

### Test Plan

- [ ] Manual verification: load `https://botho.io/es` (or a local dev equivalent) at each of the swept viewport widths and confirm no nav crowding/wrap; load `/en` directly (hard refresh, not client nav) and confirm the English landing page renders; click the locale switcher on `/es` and confirm it reads "Español" before and after interaction.
- [ ] Automated tests: extend `web/packages/web-wallet/src/pages/landing.test.tsx` (switcher-label-matches-locale cases for `/`, `/es`, `/en`, and post-switch navigation); extend `web/packages/web-wallet/functions/lib/locale-negotiation.test.ts` (new `/en` redirect cases + query-string + ordering vs. Accept-Language negotiation); extend or add to `web/packages/web-wallet/src/lib/locale-path.test.ts` only if client-side `/en` handling is implemented as a path-helper function (do not modify the existing `parseLocalePath` semantics assertions).
- [ ] Edge cases: `/en/` (trailing slash), `/en?foo=bar`, `/en/wallet?foo=bar`, a bot/crawler UA hitting `/en` (must still get the 301, per item 4's "defense in depth" framing — crawlers should never see a blank page either), and a browser that already has the `botho_locale` cookie set to `es` hitting `/en` (redirect must win over "cookie already set → pass" in `negotiateLocaleRedirect`, i.e. the `/en` check needs to run before or independent of the existing cookie-short-circuit).
- [ ] Viewport sweep (Playwright, if `web/e2e` has a landing-page spec, or add one): 768/900/1024/1152/1280/1440px in both `en` and `es`, asserting no horizontal overflow and no two-line nav labels.

## Curator note on original comment #4's citation

Comment #4 states the switcher's `buildPath`/switch logic should be investigated for emitting `/en/...` paths. Per the trace above, `switchLocaleInPath`/`buildLocalePath` do NOT currently emit a `/en` prefix (verified against `locale-path.ts` and its existing passing tests) — so part (a) of comment #4's "Wanted" list ("the switcher must emit unprefixed paths for en") is **already satisfied** by current code, not a fix that needs to be made. The Builder should treat this as confirmed-correct and focus implementation effort on parts (b) (edge 301) and (c) (client-side defense/catch-all), plus the regression test proving (a) holds.

rjwalters:
Fixes the four user-reported production bugs on botho.io `/es` and `/en` (issue #797). Ships to botho.io on merge.

### 1. Header crowding (item 1)
`landing.tsx` collapsed to the hamburger at a single fixed `md:` (768px) breakpoint tuned to English label widths. Six nav items + switcher + CTA leave no slack in the `max-w-6xl` row, so Spanish's longer labels overflow/crush in the 768–1024px band. Raised the collapse breakpoint to `lg:` (1024px): `hidden md:flex` → `hidden lg:flex`, `md:hidden` → `lg:hidden` (both nav and mobile-menu wrapper). Per the curator's recommended simplest-robust option; a measurement-based progressive-disclosure nav is out of scope for this bug fix.

### 2. Nav label line-wrap (item 2)
Added `whitespace-nowrap` to every desktop nav `<Link>`/`<a>`, the `LocaleSwitcher` trigger, and the "Abrir monedero" CTA `<Button>` (plus the mobile CTA). Precedent: `node.tsx:118`.

### 3. Locale switcher wrong language (item 3)
Confirmed **no independent defect** (matches the curator's root-cause finding): `LocaleSwitcher` derives its label purely from the URL via `parseLocalePath`, which is correct for every well-formed URL. The only demonstrated desync path was a prior mis-navigation into a `/en` orphan URL — fixed by item 4. Added regression tests asserting the switcher label equals the rendered language for `/`, `/es`, `/en`, and after an in-session es↔en switch.

### 4. `/en` blank page (item 4)
English is the unprefixed default, so `/en` / `/en/*` match no route → blank SPA. Fixed in three layers per the curator spec:
- **Edge** — new pure `stripEnPrefixRedirect(pathname)` in `locale-negotiation.ts`; `_middleware.ts` 301s `/en`/`/en/*` to the unprefixed equivalent, preserving the query string. Runs **before** Accept-Language negotiation so a first-visit `es` browser hitting `/en` lands on `/` (then negotiates onward) instead of compounding into `/es/en`.
- **Client** — `<Route path="/en">` / `<Route path="/en/*">` redirects in `AppRoutes`, defense-in-depth for in-app navigations that construct an `/en` URL without a full reload.
- **Tests** — `locale-negotiation.test.ts` (strip helper + ordering-vs-negotiation guard), `App.test.tsx` (`/en` → `/`, `/en/wallet` → `/wallet`).

`parseLocalePath`'s documented `/en` semantics are unchanged (existing `locale-path.test.ts` case still passes).

## Verification

- **Full web vitest suite:** 797 passed / 10 skipped, 0 failed.
- **Typecheck:** clean across all 8 workspace projects.
- **`pnpm check:ci` step 3** (baas-worker `wrangler deploy --dry-run`): passes.
- **`wrangler pages dev` + curl matrix** against the built `dist/` with the Pages Function:

| Request | Headers | Result |
|---|---|---|
| `/en` | browser UA, no cookie | 301 → `/` |
| `/en/wallet` | browser UA | 301 → `/wallet` |
| `/en?foo=bar` | browser UA | 301 → `/?foo=bar` (query preserved) |
| `/en/wallet?x=1` | browser UA | 301 → `/wallet?x=1` |
| `/en` | `Accept-Language: es`, no cookie | **301 → `/`** (not 302 → `/es/en`) |
| `/en` | `Cookie: botho_locale=es` | 301 → `/` (wins over cookie short-circuit) |
| `/en` | `User-Agent: Googlebot` | 301 → `/` (crawlers never see blank) |
| `/enterprise` | browser UA | 200 (lookalike not redirected) |
| `/wallet` | browser UA | 200 (unaffected) |
| `/` | `Accept-Language: es`, no cookie | 302 → `/es` (normal negotiation intact) |

Closes #797

### DIFF
--- a/web/packages/web-wallet/functions/_middleware.ts
+++ b/web/packages/web-wallet/functions/_middleware.ts
@@ -24,6 +24,7 @@ import {
   localizeDocumentHtml,
   negotiateLocaleRedirect,
   parseLocaleFromPath,
+  stripEnPrefixRedirect,
 } from './lib/locale-negotiation'
 
 // Minimal shape of the Cloudflare Pages Functions context we use. Declared
@@ -85,6 +86,24 @@ export const onRequest = async (context: PagesContext): Promise<Response> => {
   }
   if (!isDocumentRequest(request, pathname)) {
     return next()
+  }
+
+  // `/en` and `/en/*` are orphan URLs (English is the unprefixed default) that
+  // otherwise render a blank SPA page (#797). Permanently 301 them to the
+  // unprefixed equivalent, preserving the query string. This runs BEFORE
+  // Accept-Language negotiation so a first-time es visitor hitting `/en` lands on
+  // `/` (and, if applicable, is then negotiated onward) rather than `/es/en`.
+  const enTarget = stripEnPrefixRedirect(pathname)
+  if (enTarget !== undefined) {
+    return new Response(null, {
+      status: 301,
+      headers: {
+        Location: url.origin + enTarget + url.search,
+        // Path-shape correction is safe to cache, but keep it Vary-clean so it
+        // does not poison the negotiated/localized document cache keys.
+        'Cache-Control': 'public, max-age=3600',
+      },
+    })
   }
 
   const headers = request.headers
--- a/web/packages/web-wallet/functions/lib/locale-negotiation.test.ts
+++ b/web/packages/web-wallet/functions/lib/locale-negotiation.test.ts
@@ -18,6 +18,7 @@ import {
   parseLocaleFromPath,
   readCookie,
   SITE_ORIGIN,
+  stripEnPrefixRedirect,
   SUPPORTED_LOCALES,
 } from './locale-negotiation'
 
@@ -120,6 +121,53 @@ describe('buildLocalePath', () => {
     expect(buildLocalePath('en', '/wallet')).toBe('/wallet')
     expect(buildLocalePath('es', '/wallet')).toBe('/es/wallet')
     expect(buildLocalePath('es', '/')).toBe('/es')
+  })
+})
+
+describe('stripEnPrefixRedirect', () => {
+  it('maps the bare /en orphan (with or without trailing slash) to the root', () => {
+    expect(stripEnPrefixRedirect('/en')).toBe('/')
+    expect(stripEnPrefixRedirect('/en/')).toBe('/')
+  })
+
+  it('strips the /en prefix from deeper paths', () => {
+    expect(stripEnPrefixRedirect('/en/wallet')).toBe('/wallet')
+    expect(stripEnPrefixRedirect('/en/explorer/tx/abc')).toBe('/explorer/tx/abc')
+    expect(stripEnPrefixRedirect('/en/node/status')).toBe('/node/status')
+  })
+
+  it('returns undefined for paths that are NOT /en orphans', () => {
+    expect(stripEnPrefixRedirect('/')).toBeUndefined()
+    expect(stripEnPrefixRedirect('/wallet')).toBeUndefined()
+    expect(stripEnPrefixRedirect('/es')).toBeUndefined()
+    expect(stripEnPrefixRedirect('/es/wallet')).toBeUndefined()
+    // Lookalike segments that merely START with "en" must be left alone.
+    expect(stripEnPrefixRedirect('/enterprise')).toBeUndefined()
+    expect(stripEnPrefixRedirect('/env')).toBeUndefined()
+  })
+
+  it('runs independently of Accept-Language negotiation (the 301 wins, not /es/en)', () => {
+    // Regression guard for the compounding bug: negotiateLocaleRedirect would
+    // send a first-visit es browser on `/en` to `/es/en` (since `/en` is not an
+    // explicit locale prefix). The middleware must therefore apply the /en 301
+    // FIRST — this test documents that the pure strip helper resolves `/en` to
+    // the unprefixed root, and that negotiation of `/en` (were it ever reached)
+    // is exactly the pathology we avoid by ordering the strip ahead of it.
+    expect(stripEnPrefixRedirect('/en')).toBe('/')
+    const wouldCompound = negotiateLocaleRedirect({
+      pathname: '/en',
+      acceptLanguage: 'es-ES,es;q=0.9',
+      cookie: null,
+      userAgent:
+        'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36',
+    })
+    // Confirms the pathology exists at the negotiation layer (hence the ordering
+    // requirement): unguarded, `/en` negotiates onward to the orphan `/es/en`.
+    expect(wouldCompound).toEqual({
+      kind: 'redirect',
+      location: '/es/en',
+      cookieLocale: 'es',
+    })
   })
 })
 
--- a/web/packages/web-wallet/functions/lib/locale-negotiation.ts
+++ b/web/packages/web-wallet/functions/lib/locale-negotiation.ts
@@ -194,6 +194,39 @@ export function buildLocalePath(locale: SupportedLocale, rest: string): string {
   return `/${locale}${normalizedRest}`
 }
 
+/**
+ * `/en` is an orphan namespace: English is the *unprefixed* default locale, so a
+ * literal `/en` or `/en/*` path is not a locale prefix (see `parseLocaleFromPath`
+ * and its `locale-path.ts` twin — both deliberately leave `/en` un-stripped) and
+ * matches no client route → a blank SPA page (#797, item 4). These paths only
+ * arrive via stale bookmarks / external links, never from the switcher (which
+ * emits unprefixed paths for `en`).
+ *
+ * Detect such a path and return the canonical unprefixed equivalent so the edge
+ * can 301 it (a permanent URL-shape correction, not a preference negotiation):
+ *
+ *   `/en`            -> `/`
+ *   `/en/`           -> `/`
+ *   `/en/wallet`     -> `/wallet`
+ *   `/en/explorer/x` -> `/explorer/x`
+ *
+ * Returns `undefined` for any path that is NOT an `/en` orphan (including `/`,
+ * `/wallet`, `/es/...`, and lookalikes like `/enterprise`). The query string is
+ * NOT handled here — the caller preserves it — so this stays a pure path->path
+ * function. This check MUST run before Accept-Language negotiation so a first
+ * visit to `/en` is corrected to `/` (and only then, if applicable, negotiated
+ * to `/es`) rather than compounded into `/es/en`.
+ */
+export function stripEnPrefixRedirect(pathname: string): string | undefined {
+  if (pathname === '/en' || pathname === '/en/') return '/'
+  if (pathname.startsWith('/en/')) {
+    // Drop the leading `/en`, keeping the rest (already starts with `/`).
+    const rest = pathname.slice('/en'.length)
+    return rest || '/'
+  }
+  return undefined
+}
+
 /** Inputs to the redirect decision — plain primitives, no runtime types. */
 export interface NegotiationInput {
   /** Request path (no query string), e.g. `/wallet`. */
--- a/web/packages/web-wallet/src/App.test.tsx
+++ b/web/packages/web-wallet/src/App.test.tsx
@@ -68,4 +68,22 @@ describe('App locale routing', () => {
     expect(await screen.findByText('Quantum Era')).toBeTruthy()
     expect(document.documentElement.lang).toBe('en')
   })
+
+  it('redirects the /en orphan to the unprefixed English landing page (not blank)', async () => {
+    // `/en` matches no locale prefix (en is the unprefixed default) and no route,
+    // so without the catch-all it would render blank (#797, item 4c). The client
+    // redirect must land on `/` and show the English landing hero.
+    window.history.pushState({}, '', '/en')
+    render(<App />)
+    expect(await screen.findByText('Quantum Era')).toBeTruthy()
+    await waitFor(() => expect(window.location.pathname).toBe('/'))
+    expect(document.documentElement.lang).toBe('en')
+  })
+
+  it('strips the /en prefix from a deeper orphan path (/en/wallet -> /wallet)', async () => {
+    window.history.pushState({}, '', '/en/wallet')
+    render(<App />)
+    await waitFor(() => expect(window.location.pathname).toBe('/wallet'))
+    expect(document.documentElement.lang).toBe('en')
+  })
 })
--- a/web/packages/web-wallet/src/App.tsx
+++ b/web/packages/web-wallet/src/App.tsx
@@ -82,8 +82,27 @@ function AppRoutes() {
       <Route path="/node/success" element={<NodeSuccessPage />} />
       {/* Node status page reached via magic link (#458 §4 / #507). */}
       <Route path="/node/status" element={<NodeStatusPage />} />
+      {/*
+        `/en` is an orphan namespace — English is the unprefixed default, so
+        `/en` / `/en/*` match no route above and render blank (#797). The edge
+        301s these on a fresh load; this is client-side defense-in-depth for any
+        in-app navigation that constructs an `/en` URL without a full reload.
+      */}
+      <Route path="/en" element={<Navigate to="/" replace />} />
+      <Route path="/en/*" element={<EnPrefixRedirect />} />
     </Routes>
   )
+}
+
+/**
+ * Redirect a stale `/en/<rest>` URL to its unprefixed `/<rest>` equivalent,
+ * preserving the query string and hash. English is the unprefixed default, so
+ * `/en/wallet` is an orphan alias for `/wallet` (#797, item 4c).
+ */
+function EnPrefixRedirect() {
+  const location = useLocation()
+  const rest = location.pathname.slice('/en'.length) || '/'
+  return <Navigate to={`${rest}${location.search}${location.hash}`} replace />
 }
 
 /**
--- a/web/packages/web-wallet/src/pages/landing.test.tsx
+++ b/web/packages/web-wallet/src/pages/landing.test.tsx
@@ -7,7 +7,7 @@
  */
 import { describe, it, expect, beforeEach, afterEach } from 'vitest'
 import { render, screen, cleanup, fireEvent } from '@testing-library/react'
-import { MemoryRouter } from 'react-router-dom'
+import { MemoryRouter, useLocation } from 'react-router-dom'
 import { LandingPage } from './landing'
 import i18n from '../lib/i18n'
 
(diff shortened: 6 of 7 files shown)


rjwalters:
✅ **Approved!**

Reviewed against the four production bugs in #797. Edge middleware scrutinized carefully as requested; all focus areas hold up in code, not just the PR claim.

## Edge redirect correctness & ordering
- **Ordering is correct and load-bearing.** `stripEnPrefixRedirect` runs at `functions/_middleware.ts:96` and returns the 301 early, BEFORE `negotiateLocaleRedirect` (line 110). An es-first-visit browser hitting `/en` therefore gets `301 → /` and can never compound into `302 → /es/en`. The edge test (`locale-negotiation.test.ts`) explicitly proves the un-guarded negotiation would produce `/es/en`, making the ordering requirement a documented regression guard rather than an incidental property.
- **Query string preserved** on both layers: `url.origin + enTarget + url.search` (edge) and `${rest}${location.search}${location.hash}` (client `EnPrefixRedirect`).
- **Path-boundary / no-loop:** `stripEnPrefixRedirect` returns `undefined` for `/`, `/wallet`, `/es`, `/es/wallet`, and crucially the classic lookalikes `/enterprise` and `/env` (only `=== '/en'`, `=== '/en/'`, or `startsWith('/en/')` match). `/` is never re-redirected to itself. Covered by tests.

## 301 caching (`public, max-age=3600`, no Vary)
Safe. The 301 Location target is a pure function of `pathname` + `url.search`, both part of the URL cache key, and is independent of Accept-Language / Cookie / User-Agent. The response body is null and headers are identical for every visitor hitting a given `/en*` URL, so a shared cache cannot cross-serve or poison the negotiated/localized document cache (which correctly stays `private, no-store` + `Vary`). Bots also receive the 301, which is correct canonical-URL SEO behavior, not cloaking.

## Breakpoint md: → lg:
Applied consistently in `landing.tsx`: desktop nav `hidden md:flex → hidden lg:flex`, mobile button `md:hidden → lg:hidden`, and mobile-menu wrapper `md:hidden → lg:hidden`. Below 1024px: hamburger renders, nav hidden. At/above 1024px: nav renders, hamburger hidden. No gap where neither renders, no band where both render.

## whitespace-nowrap coverage
Every desktop nav `<Link>`/`<a>` (explorer, network, docs, node, whitepaper, github), the `LocaleSwitcher` trigger (via its pre-existing `className` prop, which correctly composes into the trigger button), the desktop CTA `<Button>`, and the mobile CTA all carry `whitespace-nowrap`.

## Client-side /en redirects (App.tsx)
`parseLocalePath` leaves `/en` and `/en/*` unstripped (en is the default), so they reach `AppRoutes` and hit `<Route path="/en">` → `Navigate to="/"` and `<Route path="/en/*">` → `EnPrefixRedirect`. react-router-dom v6 splat semantics respect the segment boundary, so `/enterprise`-like SPA routes are not caught.

## Test Execution
- Scoped vitest (`locale-negotiation.test.ts`, `App.test.tsx`, `landing.test.tsx`): **47 passed / 0 failed**.
  - Edge: strip-helper boundary cases + ordering-vs-negotiation guard.
  - Router: `/en → /` and `/en/wallet → /wallet` land on English content, not blank.
  - Landing: switcher-label regression for `/`, `/es`, `/en`, plus in-session es→en switch confirming the unprefixed `/wallet` (never `/en/wallet`).

## CI
- `typecheck + vitest + worker dry-run`: **pass** (1m18s)
- `cargo deny`: **pass** (1m1s)
- Merge state: **CLEAN**

PR description uses `Closes #797`. Ships to botho.io on merge. Clean, well-documented, targeted fix.

PR #810, opened 2026-07-13
fix(baas): retry certbot until DNS via systemd timer + resolve seed peers into config

### LINKED ISSUE #807: baas: node-bootstrap resilience — certbot must retry until DNS exists; bootstrap peers must not depend on unreleased DNS discovery
Two bootstrap defects found live during the #721 E2E (2026-07-12/13, instance i-0b38b7c71b248cf80). Both required manual SSH intervention to bring the node fully up; either alone leaves a paying customer with a broken node.

## 1. certbot runs once, before DNS can possibly exist
The provision flow now creates the node's DNS record minutes AFTER launch (worker backfill, f9c6b1db) — but `infra/baas/node-bootstrap.sh` runs certbot exactly once during cloud-init and logs a WARN telling a human to re-run the script. Result: every node comes up HTTP-only while the product hands out an `https://` RPC URL.

Fix: replace the one-shot with a retry loop or (better) a systemd timer + oneshot unit that: checks DNS resolves to this instance's public IP -> runs certbot webroot -> swaps to the TLS site -> disables itself. Cap retries (e.g. 2h) and log clearly. Idempotent re-entry.

## 2. Empty bootstrap_peers relies on DNS-seed discovery the released binary doesn't have
The script intentionally writes `bootstrap_peers = []` and trusts `[network.dns_seeds]` discovery. But the published release the script downloads (v0.3.2, 4a8f3c2-dirty) predates working config-driven DNS-seed discovery — the node logs "No bootstrap peers configured" and idles at peerCount 0 forever (also blocking minting: quorum needs ≥2). The #462 verification must have used a newer local build.

Fix (belt + suspenders): resolve `seeds.testnet.botho.io` TXT records IN THE SCRIPT at bootstrap time (`peerid@host:port` -> `/ip4/<resolved>/tcp/<port>/p2p/<peerid>`) and write them into `bootstrap_peers` explicitly, keeping `dns_seeds` enabled for binaries that support it. This is exactly the manual remediation that worked live (5 peers, syncing, immediately). Alternatively/additionally: cut a new release with DNS discovery and pin `BOTHO_BINARY_URL` — but the script should not silently depend on release lag.

## Acceptance
- Fresh provision (no manual SSH) reaches: valid TLS on the node hostname, peerCount ≥ 1, chain syncing, minting eligible once quorum satisfied.
- Bootstrap log shows the DNS-wait/certbot retry and the resolved peer list.
- Note: bootstrap script is fetched from raw.githubusercontent (main) at provision time — a merged fix applies to all future nodes automatically.

Refs #721 (E2E), #805 (status delivery), f9c6b1db (DNS backfill fix), #462 (original verification).

---

## Curator notes (verified against `origin/main`)

Both defects were verified directly in `infra/baas/node-bootstrap.sh` (738 lines) at HEAD on `origin/main`:

- **Certbot one-shot**: Step 5 (`~line 606-624`, `webroot|*` case of the `TLS_MODE` switch) calls `certbot certonly --webroot ...` exactly once; on failure it only logs `WARN: certbot webroot failed ... re-run this script after DNS propagates` and falls through to `write_http_only_site` permanently — nothing re-triggers it.
- **Bootstrap peers**: `bootstrap_peers` is always written as `PEERS_TOML` (empty unless the operator passes `BOOTSTRAP_PEERS`, ~line 336-401), relying on `[network.dns_seeds] enabled = true`. Also confirmed the root cause in `botho/src/network/dns_seeds.rs` (`parse_seed_record`, ~line 270-282): when a TXT seed record's address is a **hostname** (as `seeds.testnet.botho.io`'s entries are, per `infra/seed/TESTNET_RESET.md:115` — `PEER_ID@host:port`), the *node's own* DNS-seed parser builds a `/dns4/<host>/tcp/<port>/p2p/<peer_id>` multiaddr — exactly the form the script's own comment (line 338-344) says the libp2p transport rejects (`MultiaddrNotSupported`). So this isn't only "released binary predates DNS-seed discovery" — even a binary *with* DNS-seed discovery hits the same `/dns4` rejection for hostname-form seeds unless the transport or seed format changes. This strengthens the case for Fix 2 as a durable belt-and-suspenders fix rather than a stopgap for one bad release.
- `NEEDED_PKGS` (line 186) is `(nginx certbot python3-certbot-nginx curl ca-certificates jq)` — **no `dig`/`dnsutils` today**, and the script never shells out to `dig` anywhere currently.
- The script is served to every new instance from `raw.githubusercontent.com/botho-project/botho/main/infra/baas/node-bootstrap.sh` (confirmed in `web/packages/baas-worker/wrangler.toml:95` and mirrored default in `web/packages/baas-worker/src/user-data.ts:41`) — merging to `main` changes behavior for every future provision with no additional release/deploy step.
- No CI workflow (`.github/workflows/*.yml`) currently lints or executes this script — `shellcheck`/`bash -n` are manual-only today. `.shellcheckrc` exists (`external-sources=true`) and the file is currently shellcheck-clean (0.11.0, no findings) — a regression here would be silent until a human runs it.
- Repo precedent for a timer+oneshot pair: `infra/backup/botho-backup.timer` + `infra/backup/botho-backup.service` (`OnCalendar`, `RandomizedDelaySec`, `Persistent=true`). Repo precedent for `--dry-run` flags in operational infra scripts: `infra/seed/reset-chain.sh`.

### Fix 1 — certbot retry: systemd timer + oneshot (not an in-script blocking loop)

Chosen over an in-script `while` retry loop because cloud-init has a boot timeout and the exact fix is to **not** hold up first boot waiting on DNS propagation (which can take minutes to hours depending on the DNS backfill worker and TTLs). A timer survives reboots, and self-disables once TLS is live, so it costs nothing once done.

Implementation shape:
- Extract the certbot-acquire-and-swap logic already inside the `webroot|*` case (~line 606-624: DNS/cert check, `certbot certonly --webroot`, `write_tls_site`, `nginx -t && systemctl reload nginx`) into a reusable function (e.g. `try_issue_tls_cert()`) so both the inline first-attempt AND the timer-triggered retry call the same code path — do not duplicate the certbot invocation.
- New systemd units, e.g. `/etc/systemd/system/botho-tls-retry.service` (`Type=oneshot`, re-invokes `node-bootstrap.sh` with a new flag such as `--tls-retry-only`, or calls a small wrapper script installed alongside it) + `/etc/systemd/system/botho-tls-retry.timer` (short period, e.g. `OnUnitActiveSec=5min`, `Persistent=true` to survive reboots before success).
- Before calling certbot, verify DNS actually points here first (per the issue's IMDSv2 snippet) so the retry timer doesn't spam Let's Encrypt's rate limits on every tick when DNS still isn't there:
  ```bash
  IMDS_TOKEN="$(curl -s -m 3 -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")"
  PUBLIC_IP="$(curl -s -m 3 -H "X-aws-ec2-metadata-token: ${IMDS_TOKEN}" http://169.254.169.254/latest/meta-data/public-ipv4)"
  RESOLVED_IP="$(getent ahostsv4 "$NODE_HOSTNAME" | awk '{print $1; exit}')"   # or dig +short A, see Fix 2 tooling note
  [[ "$RESOLVED_IP" == "$PUBLIC_IP" ]] || exit 0   # not ready yet; timer fires again next tick
  ```
  (The script already has this exact IMDSv2 token pattern at ~line 654-658 for `PUBLIC_IP` in Step 6 — reuse it rather than re-deriving.)
- On success: write the TLS site (reuse existing `write_tls_site` function, ~line 497), reload nginx, then `systemctl disable --now botho-tls-retry.timer` so it doesn't keep polling forever.
- Cap total retry duration at ~2h (e.g. stamp a start time in a state file on first timer install, or use `systemd-run --on-active` bookkeeping; simplest is a marker file with the first-attempt epoch, compared each tick, `fail`-equivalent log + self-disable past the cap without crashing the unit).
- Idempotency: re-running `node-bootstrap.sh` by hand (already a supported path per the file's own USAGE header) must not install a second competing timer — installing the unit files + `daemon-reload` + `enable --now` is naturally idempotent as long as the unit file content is deterministic; just don't re-`start` if `have_cert` already succeeded.

### Fix 2 — resolve DNS-seed TXT records into `bootstrap_peers` in-script

This is the exact remediation proven live (5 peers, syncing, immediately). Implementation shape, inserted before the existing "Bootstrap peers" block (~line 336):

1. **Query TXT records** for `seeds.testnet.botho.io` (or derive from `NODE_DOMAIN`/a new `SEED_DOMAIN` var if this should generalize beyond hardcoding testnet — check whether mainnet will need `seeds.botho.io` eventually; scope this issue to testnet only, matching the existing `NETWORK=testnet`-only guard at line ~163).
2. **Tooling fallback chain** — `dig` is not currently installed (`NEEDED_PKGS` has no `dnsutils`). Two options, pick based on what's cheaper to keep robust:
   - (a) Add `dnsutils` to `NEEDED_PKGS` (line 186) so `dig +short TXT seeds.testnet.botho.io` is always available — simplest, one extra small package.
   - (b) Skip the new dependency and use `getent hosts`-style resolution — but `getent` does not do TXT lookups (A/AAAA only), so this only works for the follow-up hostname->A step, not the TXT fetch itself. TXT records need either `dig`, `host`, or a `python3`/`nslookup`-based fallback (python3 dnspython is not stdlib and adds its own dependency; `nslookup` is often bundled with `dnsutils` too, same tradeoff).
   - Given (b) doesn't actually avoid a new package for the TXT step, prefer (a): add `dnsutils` to `NEEDED_PKGS`. Guard the call anyway (`command -v dig` check) so the *fallback-empty-list* behavior below still functions if the package install is somehow skipped/blocked.
3. **Parse each TXT entry** `peerid@host:port` (matches `botho/src/network/dns_seeds.rs` `parse_seed_record` format — do not diverge). For each: resolve `host` to an A record — `dig +short A "$host" | head -1` (or `getent ahostsv4 "$host" | awk '{print $1; exit}'` if `dig` unavailable) — and emit `/ip4/<resolved_ip>/tcp/<port>/p2p/<peerid>`.
4. **Empty/failure handling**: if `dig`/DNS lookup fails or the TXT record set is empty, fall back to the current behavior (empty `bootstrap_peers`, rely on `dns_seeds` discovery) with a **loud** log line (e.g. `log "WARN: no TXT seeds resolved for seeds.testnet.botho.io; falling back to DNS-seed discovery only (peerCount may stay 0 if the running binary doesn't support it)"`) so this failure mode is visible in `/var/log/botho-node-bootstrap.log` rather than silently idling at 0 peers again.
5. **Keep `[network.dns_seeds] enabled = true`** in the emitted config regardless (per the issue) — the resolved peers are additive/belt-and-suspenders, not a replacement.
6. **Idempotent config regeneration**: the config block only executes `if [[ ! -f "$CONFIG_FILE" ]]` (line 357) — a re-run of the script on an existing node does **not** rewrite `bootstrap_peers` even if seed resolution changes. Confirm this is the desired behavior (matches existing mnemonic-preservation semantics) or, if the fix should also *repair* already-provisioned nodes stuck at peerCount 0, add a narrowly-scoped path that rewrites just the `bootstrap_peers` line in an existing config without touching the mnemonic/other settings (e.g. `sed`-replace the block between `bootstrap_peers = [` and the matching `]`, or a small `python3`/`awk` block, done carefully since raw `sed` block-replace with generated content is easy to get wrong — prefer a minimal, tested approach over a clever one-liner). Recommend starting with **new-provision-only** (leave the `! -f CONFIG_FILE` guard as-is) unless the issue's acceptance criteria are read to require live-node repair; flag this scoping choice for the Builder to confirm rather than assuming.

### Shellcheck / smoke-test plan (no existing CI covers this file)

Since there is no CI gate today, the PR description itself must show the verification was done manually:
- `bash -n infra/baas/node-bootstrap.sh` (syntax-only check).
- `shellcheck infra/baas/node-bootstrap.sh` (must stay clean — currently 0 findings at HEAD with `.shellcheckrc`'s `external-sources=true`; new `dig`/timer-file heredocs are the likely new surface for SC2064/SC2016-style quoting issues — watch unescaped `$` inside heredocs written to unit files, same pattern already handled carefully for the nginx heredocs).
- Manual function-level dry run is high-value but should stay cheap: consider a `--dry-run`-style flag (precedent: `infra/seed/reset-chain.sh`) that at minimum lets the new DNS-seed-resolution function and the new timer-unit-writing logic be exercised locally against a fake/mocked `dig` output without a full EC2 boot — do not attempt to dry-run the whole script (certbot, apt, systemd all assume a real box); scope the dry-run mode narrowly to the two new pieces if it can be added cheaply, otherwise skip it and rely on the two static checks + a real staging-instance boot test.
- Recommend one live smoke test on a real (disposable/testnet) instance before merge, given "merge = ship to all future nodes": provision, confirm bootstrap log shows the DNS-wait/certbot retry firing and succeeding once DNS lands, confirm `bootstrap_peers` populated with resolved `/ip4/.../p2p/...` entries, confirm `node-status` shows `peerCount >= 1` and `synced` progressing without manual SSH.

## Acceptance Criteria

- [ ] Fresh provision (no manual SSH) reaches: valid TLS (`https://<hostname>/rpc` serves, `have_cert` true) even when DNS A record is created several minutes after instance launch.
- [ ] The TLS retry mechanism (systemd timer, not a blocking in-script loop) self-disables once a cert is issued, and does not exceed ~2h of retries before giving up with a loud log message.
- [ ] Fresh provision reaches `peerCount >= 1` and syncing without manual SSH, using the currently-released binary (v0.3.2 or whatever is current at merge time) — i.e. does not assume an unreleased DNS-seed-discovery capable binary.
- [ ] `bootstrap_peers` in the generated config contains resolved `/ip4/<ip>/tcp/<port>/p2p/<peerid>` entries derived from `seeds.testnet.botho.io` TXT records; `[network.dns_seeds] enabled = true` is still set.
- [ ] If TXT resolution fails or returns no entries, the script falls back to empty `bootstrap_peers` (today's behavior) with a clearly logged WARN — it must not fail the whole bootstrap.
- [ ] Re-running the script on an already-provisioned node (idempotent re-entry) does not corrupt `config.toml`, does not lose the wallet mnemonic, and does not install duplicate/competing systemd timer units.
- [ ] `bash -n infra/baas/node-bootstrap.sh` passes.
- [ ] `shellcheck infra/baas/node-bootstrap.sh` remains clean (0 findings) under the repo's `.shellcheckrc`.
- [ ] `infra/baas/README.md` is updated to reflect the new certbot-retry and resolved-bootstrap-peers behavior (the current "re-run this script after DNS propagates" and "empty bootstrap_peers / DNS-seed discovery" language is now stale/superseded).
- [ ] Bootstrap log (`/var/log/botho-node-bootstrap.log`) shows both the DNS-wait/certbot-retry activity and the resolved peer list on a real provision.

## Affected Files

- `infra/baas/node-bootstrap.sh` — certbot retry mechanism (Fix 1), TXT-seed resolution into `bootstrap_peers` (Fix 2), `NEEDED_PKGS` addition (`dnsutils`), new systemd timer/service unit heredocs.
- `infra/baas/README.md` — update stale guidance ("re-run this script after DNS propagates", "empty bootstrap_peers / DNS-seed discovery" description) to describe the new self-healing behavior.

## Test Plan

1. Static checks (fast, do on every iteration): `bash -n infra/baas/node-bootstrap.sh`; `shellcheck infra/baas/node-bootstrap.sh`.
2. If a `--dry-run`/function-level test mode is added: exercise the new TXT-resolution function and timer-unit-writing function locally with mocked `dig` output (no real EC2 needed).
3. Live smoke test on a disposable testnet EC2 instance (mirrors the #721 E2E that found this bug):
   - Launch without pre-creating DNS (or with a deliberately delayed DNS backfill) to exercise the certbot-retry path; confirm the timer fires, DNS eventually resolves, cert issues, TLS site swaps in, timer self-disables.
   - Confirm `sudo node-status` shows `peerCount >= 1`, `synced` advancing, without any manual SSH intervention beyond observing logs.
   - Inspect `~ubuntu/.botho/testnet/config.toml`'s `bootstrap_peers` — confirm resolved `/ip4/.../p2p/...` entries matching current `seeds.testnet.botho.io` TXT records.
   - Re-run the script by hand a second time on the same instance (idempotency check) — confirm no mnemonic change, no duplicate timer units (`systemctl list-timers`), no `config.toml` corruption.
4. Confirm `infra/baas/README.md` changes accurately describe the new behavior (no stale "re-run after DNS propagates" instructions left for a since-automated case).

rjwalters:
## Summary

Two node-bootstrap defects found live during the #721 E2E each left a fresh paid node broken without manual SSH. Both are fixed in `infra/baas/node-bootstrap.sh` (the script fetched from `raw.githubusercontent.com/.../main/...` at provision time, so **merge = live for every future node**).

## Changes

**Fix 1 — certbot retry-until-DNS (systemd timer, not a blocking loop).** The DNS A record is now backfilled minutes *after* launch (worker `f9c6b1db`), so the one-shot inline certbot always failed and logged "re-run this script". The acquire-and-swap logic is extracted into `try_issue_tls_cert()` (shared by the inline first attempt and the retry path — no duplicated certbot call). When DNS is not yet pointed here, the script installs a `botho-tls-retry` **oneshot + timer** pair (precedent: `infra/backup/botho-backup.{timer,service}`) that re-invokes the script with a new `--tls-retry-only` mode every ~5 min. Each tick:
- verifies via **IMDSv2** (reusing the Step 6 token pattern) that the hostname's A record resolves to this instance *before* calling Let's Encrypt (no ACME rate-limit burn while DNS is absent),
- on success writes the TLS site, reloads nginx, and **self-disables the timer**,
- gives up after a **~2h cap** (epoch stamp in `/var/lib/botho/`) with a loud `ERROR` log.

A stable copy of the script is installed at `/usr/local/sbin/botho-node-bootstrap.sh` so the unit has a durable path (cloud-init's user-data path is transient).

**Fix 2 — resolve DNS-seed TXT records into `bootstrap_peers`.** Empty `bootstrap_peers` relied on `dns_seeds` discovery, but the node's own `parse_seed_record` (`botho/src/network/dns_seeds.rs`) emits the transport-rejected `/dns4/host/...` form for the hostname-style `PEER_ID@host:port` records at `seeds.testnet.botho.io` — so nodes idled at `peerCount 0` (this is the durable root cause, not just release lag). The bootstrap now resolves `SEED_DOMAIN`'s TXT records itself into transport-accepted `/ip4/<ip>/tcp/<port>/p2p/<peer_id>` multiaddrs at config-write time and writes them into `bootstrap_peers`. `[network.dns_seeds] enabled = true` is kept as an additive fallback. Empty/failed resolution falls back to today's empty list with a loud `WARN`. `dnsutils` added to `NEEDED_PKGS` for `dig` (with a `command -v dig` guard so the fallback still works if the package is somehow absent). New `SEED_DOMAIN` param (default `seeds.<NODE_DOMAIN>`) keeps this testnet-scoped and generalizable.

**Tooling.** Added a `--dry-run` mode (precedent: `infra/seed/reset-chain.sh`) that exercises the two new pieces — seed resolution and unit-file rendering — locally without EC2/apt/certbot/systemd. Updated `infra/baas/README.md` (removed the stale "re-run this script after DNS propagates" and "empty bootstrap_peers / DNS-seed discovery" guidance).

## Idempotency-scoping choice

**New-provision-only**, as recommended by the curator. Fix 2's seed resolution runs inside the existing `if [[ ! -f "$CONFIG_FILE" ]]` block, so a hand re-run on an already-provisioned node preserves its `config.toml` (and wallet mnemonic) verbatim — identical semantics to the existing mnemonic-preservation path, no risk of corrupting a live config. I did **not** add a live-node `bootstrap_peers` repair path: the acceptance criteria only require fresh provisions to reach `peerCount >= 1`, and a targeted in-place TOML rewrite is the kind of clever-`sed` the curator flagged as easy to get wrong. The TLS-retry timer install is separately idempotent (deterministic unit content + `daemon-reload` + `enable`), so re-runs never create duplicate/competing timers.

| Fresh provision reaches valid TLS even when DNS lands minutes late | ✅ | `--tls-retry-only` timer re-tries certbot until `dns_points_here` is true, then swaps TLS site |
| TLS retry self-disables on cert; ~2h cap | ✅ | `disable_tls_retry_timer` on `have_cert`; epoch-stamp cap in `run_tls_retry_only` |
| Fresh provision reaches `peerCount >= 1` on the released binary | ✅ | `bootstrap_peers` populated with resolved `/ip4/...` (verified via `--dry-run`), independent of binary DNS support |
| `bootstrap_peers` has resolved `/ip4/.../p2p/...`; `dns_seeds` still `enabled` | ✅ | `--dry-run` output shows `/ip4/.../tcp/17100/p2p/...`; config heredoc keeps `[network.dns_seeds] enabled = true` |
| TXT-resolution failure falls back to empty list + loud WARN, does not fail boot | ✅ | empty-TXT `--dry-run` logs the fallback WARN; `resolve_seed_peers` returns 0 on missing `dig`/empty TXT |
| Idempotent re-entry: no config corruption, no lost mnemonic, no duplicate timers | ✅ | `!config.toml` guard preserves config; deterministic unit files |
| `bash -n` passes | ✅ | clean |
| `shellcheck` remains clean (0 findings) under `.shellcheckrc` | ✅ | shellcheck 0.11.0, 0 findings |
| `README.md` updated | ✅ | Fix 1/Fix 2 notes added; stale "re-run after DNS propagates" removed |
| Bootstrap log shows DNS-wait/certbot-retry + resolved peer list | ✅ | loud `log`/`WARN`/`ERROR` lines in both new code paths (verified in `--dry-run`) |

## Test Plan

- `bash -n infra/baas/node-bootstrap.sh` — passes.
- `shellcheck infra/baas/node-bootstrap.sh` — 0 findings (0.11.0, repo `.shellcheckrc`).
- `--dry-run` with a mocked `dig`: two seed forms (hostname→A record, and bare-IP) both render correct `/ip4/.../tcp/.../p2p/...`; empty-TXT renders the fallback WARN; custom `NODE_DOMAIN` derives the right `SEED_DOMAIN`; unknown-arg and `--tls-retry-only` guards fail cleanly.

> **⚠️ Merge = ship to all future nodes.** This script is fetched from `raw.githubusercontent main` at provision time — no separate release/deploy. CI does not cover this file (static checks are manual-only). **Recommend a disposable/testnet-instance smoke test before the next paid provision**: confirm the bootstrap log shows the DNS-wait/certbot-retry firing and succeeding once DNS lands, `bootstrap_peers` populated with resolved entries, and `sudo node-status` showing `peerCount >= 1` + syncing without manual SSH.

Closes #807

### DIFF
--- a/infra/baas/README.md
+++ b/infra/baas/README.md
@@ -25,7 +25,8 @@ recipe in `infra/seed/` and `infra/faucet/`:
 
 ## What it does (idempotent, logged to `/var/log/botho-node-bootstrap.log`)
 
-1. **Install deps** — nginx, certbot, curl, jq, ca-certificates.
+1. **Install deps** — nginx, certbot, curl, jq, ca-certificates, dnsutils
+   (`dig`, for DNS-seed TXT resolution).
 2. **Download the prebuilt `botho` linux-aarch64 binary** — from
    `BOTHO_BINARY_URL` if set (the release tarball
    `botho-vX.Y.Z-linux-aarch64.tar.gz`, or a bare-binary mirror), else reuse an
@@ -37,23 +38,49 @@ recipe in `infra/seed/` and `infra/faucet/`:
    `aarch64` and installed mode `0755`.
 3. **Generate identity + config** — a per-node 24-word wallet mnemonic and
    `~/.botho/testnet/config.toml` (testnet, RandomX minting on, faucet off,
-   `quorum.mode = recommended` / `min_peers = 1`, `bootstrap_peers = []`,
-   DNS-seed discovery `seeds.testnet.botho.io` enabled). The wallet mnemonic is
-   preserved across re-runs. The libp2p **node_key** (peer identity) is created
-   and persisted automatically by the binary at `~/.botho/testnet/node_key`, so
-   the peer id is stable across reboots with no extra handling.
+   `quorum.mode = recommended` / `min_peers = 1`, **resolved `bootstrap_peers`**
+   (see below), DNS-seed discovery `seeds.testnet.botho.io` still enabled). The
+   wallet mnemonic is preserved across re-runs. The libp2p **node_key** (peer
+   identity) is created and persisted automatically by the binary at
+   `~/.botho/testnet/node_key`, so the peer id is stable across reboots with no
+   extra handling.
 
-   > **Peering note (learned during verification):** the node's libp2p
-   > transport rejects bare `/dns4/host/tcp/port` entries in explicit
-   > `bootstrap_peers` (`MultiaddrNotSupported` → 0 peers). The working path is
-   > DNS-seed discovery (empty `bootstrap_peers` + `dns_seeds.enabled = true`),
-   > which resolves the seed TXT records to `/ip4/.../p2p/<peer_id>` multiaddrs
-   > the transport accepts. An explicit `BOOTSTRAP_PEERS` override must
-   > therefore supply resolved `/ip4/.../tcp/<port>/p2p/<peer_id>` multiaddrs.
+   > **Peering note (Fix 2, #807):** the node's libp2p transport rejects bare
+   > `/dns4/host/tcp/port` entries in `bootstrap_peers` (`MultiaddrNotSupported`
+   > → 0 peers). The node's **own** DNS-seed parser also emits that rejected
+   > `/dns4` form for the hostname-style `PEER_ID@host:port` records published at
+   > `seeds.testnet.botho.io`, so relying on empty `bootstrap_peers` +
+   > `dns_seeds` discovery idles at `peerCount 0` on the released binary. The
+   > bootstrap therefore **resolves the seed TXT records itself at config-write
+   > time** into the transport-accepted `/ip4/<ip>/tcp/<port>/p2p/<peer_id>`
+   > form (via `dig`) and writes them into `bootstrap_peers`.
+   > `[network.dns_seeds] enabled = true` is kept as an additive fallback for
+   > binaries that later gain IP-form discovery. If TXT resolution fails or
+   > returns nothing, the script falls back to empty `bootstrap_peers` with a
+   > loud `WARN` in the log rather than failing the boot. An explicit
+   > `BOOTSTRAP_PEERS` override still takes precedence and must supply resolved
+   > `/ip4/.../tcp/<port>/p2p/<peer_id>` multiaddrs. This resolution runs only
+   > on **fresh provisions** (the `config.toml`-does-not-exist path); a re-run
+   > on an existing node preserves its config verbatim (same idempotency
+   > semantics as the wallet mnemonic).
 4. **Install + start** the `botho` systemd unit
    (`botho --testnet run --mint --mint-threads N`).
 5. **nginx + TLS + `/rpc` proxy** for `NODE_HOSTNAME` (mirrors
    `seed-nginx.conf`: HTTP→HTTPS redirect, CORS de-duplication, `/rpc/ws`).
+
+   > **certbot retry-until-DNS (Fix 1, #807):** the provisioner now creates the
+   > node's DNS A record minutes *after* launch, so certbot cannot succeed on
+   > the first inline attempt. The bootstrap tries once, and if DNS does not yet
+   > resolve to this instance's public IP it installs a systemd **oneshot +
+   > timer** pair (`botho-tls-retry.service` / `.timer`) that re-invokes
+   > `node-bootstrap.sh --tls-retry-only` every ~5 min. Each tick verifies (via
+   > IMDSv2) that the hostname's A record points at this instance before calling
+   > Let's Encrypt (so it never burns ACME rate limits while DNS is absent);
+   > once a cert is issued it swaps in the TLS site and **self-disables the
+   > timer**. It gives up after a ~2h cap with a loud log line. This replaces
+   > the old "re-run this script after DNS propagates" manual step — no SSH
+   > needed. Re-running the bootstrap by hand rewrites the (deterministic) unit
+   > files without creating duplicate/competing timers.
 6. **Emit node info** to `~/node-info.txt` and install `node-status` for read-back.
 
 ## Inputs (env vars / user-data exports)
@@ -69,7 +96,8 @@ recipe in `infra/seed/` and `infra/faucet/`:
 | `REGION`              | no       | —                  | AWS region the node launched in (informational; the provisioner picks it at run-instances time). Recorded in `node-info.txt`. |
 | `TIER`                | no       | `t4g.medium`       | Instance type/tier (informational; MVP is t4g.medium-only). Recorded in `node-info.txt`. |
 | `NETWORK`             | no       | `testnet`          | Only `testnet` is supported in this slice. |
-| `BOOTSTRAP_PEERS`     | no       | DNS-seed discovery | Comma-separated **resolved** multiaddrs (`/ip4/.../tcp/<port>/p2p/<peer_id>`). Default empty -> DNS-seed discovery (the working path; bare `/dns4` is unsupported). |
+| `BOOTSTRAP_PEERS`     | no       | resolved from `SEED_DOMAIN` | Comma-separated **resolved** multiaddrs (`/ip4/.../tcp/<port>/p2p/<peer_id>`). When set, used verbatim (bare `/dns4` is unsupported). When unset, the bootstrap resolves `SEED_DOMAIN`'s TXT records into `/ip4/...` peers itself (Fix 2, #807); if that yields nothing it falls back to empty + DNS-seed discovery. |
+| `SEED_DOMAIN`         | no       | `seeds.<NODE_DOMAIN>` | DNS-seed domain whose `PEER_ID@host:port` TXT records are resolved into `bootstrap_peers` at config-write time (testnet: `seeds.testnet.botho.io`). |
 | `MINT_THREADS`        | no       | `1`                | RandomX threads (t4g.medium = 2 vCPU). |
 | `CERTBOT_EMAIL`       | no       | `admin@botho.io`   | Let's Encrypt registration email. |
 | `TLS_MODE`            | no       | `webroot`          | `webroot` (needs nginx+DNS) / `standalone` / `skip` (HTTP-only, for local testing). |
@@ -142,13 +170,16 @@ export BOTHO_BINARY_SHA256="019f31e8e29cf482567be1c51f65d499aeffda1b63f57098a991
 
 ### 2. DNS pre-creation (`NODE_HOSTNAME` / `NODE_ID`)
 
-The provisioner must create the `NODE_HOSTNAME` (or the derived
-`node-<NODE_ID>.<NODE_DOMAIN>`) A record pointing at the
-instance's public IP **before/at boot** so Let's Encrypt (`webroot`/
-`standalone`) can validate. If DNS isn't ready when the script runs, certbot is
-skipped gracefully and nginx serves HTTP-only `/rpc`; re-running the script
-after DNS propagates issues the cert and switches to HTTPS (idempotent). For
-DNS-less local testing use `TLS_MODE=skip`.
+The provisioner creates the `NODE_HOSTNAME` (or the derived
+`node-<NODE_ID>.<NODE_DOMAIN>`) A record pointing at the instance's public IP so
+Let's Encrypt (`webroot`/`standalone`) can validate. In the current flow this
+record is backfilled **minutes after launch** (the worker learns the public IP
+only once the instance is running). The bootstrap tolerates this: if DNS isn't
+pointed here yet when the script runs, it serves HTTP-only `/rpc` and installs a
+systemd timer that retries certbot every ~5 min until the A record resolves to
+this instance, then swaps to HTTPS and self-disables (see step 5 above, Fix 1 /
+#807). **No manual re-run is required** — the old "re-run this script after DNS
+propagates" step is obsolete. For DNS-less local testing use `TLS_MODE=skip`.
 
 ## Example user-data
 
--- a/infra/baas/node-bootstrap.sh
+++ b/infra/baas/node-bootstrap.sh
@@ -34,6 +34,14 @@
 #
 # Re-running is safe (idempotent): each step checks current state first.
 #
+# Flags:
+#   --tls-retry-only   internal: run only the certbot-until-DNS retry step.
+#                      Invoked by the botho-tls-retry systemd timer; not for
+#                      manual use.
+#   --dry-run          local, non-EC2 smoke test of the two #807 additions
+#                      (DNS-seed -> bootstrap_peers resolution + systemd
+#                      unit-file rendering). No apt/certbot/systemd side effects.
+#
 # ---------------------------------------------------------------------------
 # PARAMETERS (environment variables / user-data exports)
 # ---------------------------------------------------------------------------
@@ -89,9 +97,15 @@
 #                      node-info.txt. The MVP is t4g.medium-only (#458 §5).
 #   NETWORK            (optional, default "testnet"). Only "testnet" is
 #                      supported by this slice.
-#   BOOTSTRAP_PEERS    (optional) comma-separated libp2p multiaddrs to use as
-#                      bootstrap_peers. Default: the live seed nodes plus DNS
-#                      seed discovery (seeds.testnet.botho.io).
+#   BOOTSTRAP_PEERS    (optional) comma-separated RESOLVED libp2p multiaddrs
+#                      (/ip4/<ip>/tcp/<port>/p2p/<peer_id>) to use as
+#                      bootstrap_peers verbatim. When unset, the script resolves
+#                      SEED_DOMAIN's TXT records into /ip4 peers itself (Fix 2,
+#                      #807); on empty/failed resolution it falls back to an
+#                      empty list + DNS-seed discovery with a loud log.
+#   SEED_DOMAIN        (optional, default "seeds.<NODE_DOMAIN>") DNS-seed domain
+#                      whose PEER_ID@host:port TXT records are resolved into
+#                      bootstrap_peers at config-write time (fresh provisions).
 #   MINT_THREADS       (optional, default 1) RandomX minting threads. t4g.medium
 #                      has 2 vCPU / ~4GB; 1 thread leaves headroom.
 #   CERTBOT_EMAIL      (optional) email for Let's Encrypt registration.
@@ -115,6 +129,10 @@
 #   /home/ubuntu/node-info.txt          machine-readable summary (RPC URL, peer
 #                                      id, mnemonic location, status command)
 #   systemd unit `botho` running `botho --testnet run --mint`
+#   systemd `botho-tls-retry.timer`/.service (only when the first inline certbot
+#     attempt fails because DNS is not yet pointed here) — retries certbot until
+#     DNS resolves, then self-disables. Re-invokes this script with
+#     --tls-retry-only from /usr/local/sbin/botho-node-bootstrap.sh.
 #   Read back any time with:  sudo /usr/local/bin/node-status   (installed here)
 #
 set -euo pipefail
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed with production-caution given this script ships live from `raw.githubusercontent main` at provision time with no CI coverage. Verified every claim myself in the external worktree (`/Volumes/Stripe/botho/issue-807`).

## Relocation claim — VERIFIED byte-for-byte
Extracted the `write_http_only_site` (30-line) and `write_tls_site` (79-line) heredoc bodies from both this branch and `origin/main` and diffed them: **IDENTICAL** in both cases. The large diff is genuinely a nest→top-level relocation of `have_cert`/`write_http_only_site`/`write_tls_site` with zero nginx config drift. No silent proxy/TLS/CORS change.

## Fix 1 (certbot retry-until-DNS)
- `try_issue_tls_cert` writes the TLS site **only after** `have_cert` returns true, so a mid-retry failure never leaves nginx referencing a missing cert — HTTP-only site stays active. Confirmed.
- `try_issue_tls_cert` is always called in `if`-context (Step 5 and `run_tls_retry_only`), which suppresses `set -e` inside it; a failing `dig`/`nginx -t` cannot abort the bootstrap. Verified with a `set -euo pipefail` repro.
- 2h cap arithmetic (`now - first > 7200`, epoch stamp in `/var/lib/botho/`) is correct — no off-by-one, no never-expires (numeric guard fails safe toward continued retry).
- oneshot + timer pair with `daemon-reload` + `enable --now`; deterministic unit content means re-runs rewrite identical files → no duplicate/competing timers. `self-disable` on cert and on cap.
- IMDSv2 token pattern reused correctly for the DNS-match gate (no ACME rate-limit burn while DNS absent).

## Fix 2 (seed TXT resolution)
Exercised `resolve_seed_peers` via `--dry-run` with a mocked `dig`:
- hostname seed → A-record `/ip4/.../tcp/17100/p2p/...`; bare-IP seed passed through. Both correct.
- one unresolvable seed among good ones → loud per-entry `WARN` and the good peer **still renders** (per-entry failure does not kill the list). Confirmed.
- malformed records (`no-@`, `@host` without `:port`) skipped via the `*@*:*` case guard.
- empty TXT → fallback WARN path, empty `bootstrap_peers` array, boot continues.
- generated TOML valid for empty / one / many peers (`dnsutils` present in `NEEDED_PKGS`, `command -v dig` guard preserves fallback).

## Idempotency
Fix 2 runs inside the `[[ ! -f "$CONFIG_FILE" ]]` new-provision guard — existing configs/mnemonics untouched on re-run. Stable script copy at `/usr/local/sbin/botho-node-bootstrap.sh` gives the unit a durable ExecStart path. `SCRIPT_SELF`/`SELF_INSTALL_PATH`/`TLS_RETRY_*` all defined before every use (safe under `set -u`; `--dry-run` and `--tls-retry-only` dispatch verified clean).

## Static checks (run myself)
- `shellcheck 0.11.0` (repo `.shellcheckrc`, `external-sources=true`): **0 findings**.
- `bash -n`: clean.
- `--dry-run` matrix (two seed forms, empty TXT, bad-seed isolation, unknown-arg guard): all pass.

## CI + merge state
- `cargo deny`: pass (the only CI check that touches this PR — the `.sh` file has no automated coverage, as the PR body correctly warns).
- `mergeStateStatus`: CLEAN, mergeable: MERGEABLE.

## README
Accurate: removes the stale "re-run this script after DNS propagates" guidance, documents both fixes, adds `SEED_DOMAIN` + `dnsutils` and the new-provision-only scoping.

## Non-blocking note
If `TLS_RETRY_STAMP` were somehow corrupted to a non-numeric value, the `^[0-9]+$` guard makes the cap check permanently false — the timer would then only stop on cert issuance, never on the 2h cap. Unreachable in practice (the script writes a numeric epoch itself) and it fails safe toward "keep trying," so not blocking. Flagging for awareness only.

**Heeding the PR's own warning:** merge ships this to every future paid node with no release gate. Strongly second the recommendation to run the disposable/testnet-instance smoke (bootstrap log shows DNS-wait → certbot-retry firing and succeeding once DNS lands; `bootstrap_peers` populated; `node-status` shows `peerCount >= 1`) before the next real provision.

Closes #807.

Issue #839, opened 2026-07-13
bridge: enforce OrderStatus::can_transition_to at the DB layer (update_order_status)


rjwalters:
Follow-up from the #838 review (Ethereum wBTH mint path).

#838 added `OrderStatus::can_transition_to` / `BridgeOrder::try_set_status` and enforces the critical edges:
- `MintPending -> Completed` is guarded in `engine.rs::confirm_mint` (explicit `can_transition_to` check) AND by `db.rs::mark_mint_confirmed`'s `WHERE status = 'mint_pending'` clause.
- The reorg unwind is guarded by `db.rs::rollback_mint`'s `WHERE status = 'mint_pending'` clause and refuses confirmed mints.

However `Database::update_order_status` (bridge/service/src/db.rs) is still a raw `UPDATE bridge_orders SET status = ...` with no transition validation. All current call sites use legal edges (DepositConfirmed->MintPending, ->Failed, BurnConfirmed->ReleasePending, AwaitingDeposit->Expired), but nothing stops a future caller (e.g. the #822 release path) from writing an illegal transition or clobbering a terminal state.

Suggested hardening:
- Make `update_order_status` read the current status inside the same transaction and reject the write if `current.can_transition_to(new)` fails (or add a guarded variant and deprecate the raw one).
- Alternatively add `WHERE status IN (...)` clauses derived from the state machine, mirroring `mark_mint_confirmed`.

Low urgency: no live bypass exists today; this is defense-in-depth for the #822/#824 work that will add more writers.

PR #838, opened 2026-07-13
bridge: implement Ethereum wBTH mint via alloy (replace engine.rs TODO)

rjwalters:
Closes #821 (part of epic #816; per ADR 0002 custody and ADR 0005 chain scope).

## What this implements

Replaces `engine.rs::process_mint_order`'s "Would mint … wBTH on Ethereum" stub with a real, exactly-once mint pipeline, following the expanded checklist in the issue comment.

### Fully implemented

**Shared / core (`bridge/core`)**
- `derive_order_id` / `BridgeOrder::order_id_bytes()`: deterministic, domain-tagged 32-byte on-chain order id from the order UUID; the SAME value is bound on both chains (golden-vector test pins it).
- `MintAuthorization` / `AttestationSignature` / `SignatureScheme` (`attestation.rs`): the artifact the #824 threshold-attestation protocol produces and this PR consumes (interface only — the protocol itself is #824).
- `OrderStatus::can_transition_to` + `BridgeOrder::try_set_status`: guarded transitions. `MintPending -> Completed` is the only path to completion, and `MintPending -> DepositConfirmed` is a first-class reorg-unwind edge (clears `dest_tx`/`dest_confirmed_at`) instead of terminal `Failed`.
- `BridgeOrder` gains `mint_authorization` and `dest_confirmed_at` (serde + SQLite round-trip).

**DB (`bridge/service/src/db.rs`)**
- `mints` idempotency table (`order_id` UNIQUE + unique `order_id_hash` index) with `record_mint_submitted` (INSERT OR IGNORE — returns the canonical prior row on duplicates), `get_mint_by_order`, `mark_mint_confirmed` (atomic with order completion), `rollback_mint` (atomic unwind; refuses to touch confirmed mints).
- Column migrations via `PRAGMA table_info` guard for existing databases.

**Engine (`bridge/service/src/engine.rs`)**
- `process_mint_order` split into `submit_mint` (`DepositConfirmed -> MintPending`) and `confirm_mint` (`MintPending -> Completed`), driven as separate retryable stages each tick.
- Exactly-once: tx id persisted to `mints` BEFORE first broadcast; resume-after-crash reuses the recorded tx (never prepares a second one); re-broadcast retries send the SAME signed bytes, capped at `bridge.max_retries`.
- Reorg unwind: `Reorged` rolls back to `DepositConfirmed` and re-submits against the same on-chain order id.
- Unconfigured chain leaves orders retryable in `DepositConfirmed` (never dropped/failed).

[...]

Issue #828, opened 2026-07-13
testing: bridge unit + integration + Ethereum fork + testnet end-to-end (happy paths) (Part of #816)


rjwalters:
**Part of #816 (Phase 3). The core positive-path test bar.**

## Task
Stand up the bridge test infrastructure and cover every happy path. Today there are only ~9 tests across the whole bridge — this issue raises that to a real suite.

- **Unit** tests for `bridge/core` (order state machine transitions, config parsing, chain address types) and `bridge/service` (engine decisions, watcher confirmation logic).
- **Ethereum fork tests** (Hardhat/anvil fork or Foundry) against a deployed `WrappedBTH`: full `bridgeMint` and `bridgeBurn` flows.
- **End-to-end on testnet:** BTH deposit → wBTH mint → hold → wBTH burn → BTH release, asserting amounts (net of demurrage decision #818), stealth-address re-shield (#819), and order lifecycle terminal states.
- **CI wiring:** run the Rust suite + contract suite + a gated (nightly or manual) testnet e2e job.

## Acceptance
- Both round-trip directions pass end-to-end on testnet in CI (or a documented gated job); unit + fork suites run on every PR touching `bridge/` or `contracts/`.

Issue #826, opened 2026-07-13
contracts: harden + review WrappedBTH.sol (threshold minter, reentrancy, circuit-breaker, upgradeability) (Part of #816)


rjwalters:
**Part of #816 (Phase 2). Blocked by #817 (custody) and #821-range attestation design.**

## Task
`contracts/ethereum/contracts/WrappedBTH.sol` is ERC20 + AccessControl + Pausable with per-tx/daily/cooldown rate limits — a good start, but unaudited and single-`MINTER_ROLE`.

- Replace the single `MINTER_ROLE` holder with the **threshold authorization** from #817 (contract-level t-of-n, or `MINTER_ROLE` granted to a Safe/multisig).
- Bind `bridgeMint` to a unique **order id** and reject replays on-chain (idempotent mint).
- Reentrancy review of mint/burn paths; confirm the daily-limit accounting cannot be griefed or bypassed.
- Decide **upgradeability** (immutable vs proxy) and document the trade-off + admin-key custody.
- Emergency **circuit-breaker**: pause path tested; consider an auto-pause on anomalous mint volume.
- Confirm `decimals`/units line up with picocredits end-to-end (no 10^x scaling bug across the boundary).

## Testing (required for DoD)
- Full Hardhat/Foundry unit + invariant test suite (access control, rate limits, pause, replay, reentrancy).
- Fuzz/invariant tests on supply accounting.
- A dedicated contract review (feeds the Phase-3 audit).

## Acceptance
- No single key mints; mints are replay-proof and order-bound; pause works; upgradeability + admin custody are decided and documented.

rjwalters:
*Implementation tasks expanded 2026-07-13 (per ADR 0002 threshold mint authority, ADR 0005 both token programs).*

## Implementation tasks

**Ethereum (`contracts/ethereum/contracts/WrappedBTH.sol`)**
- [ ] Replace single-holder role custody: in the `constructor`, stop granting `DEFAULT_ADMIN_ROLE`/`MINTER_ROLE`/`PAUSER_ROLE` to `msg.sender`; parameterize to grant `MINTER_ROLE` to the validator Gnosis Safe (secp256k1 t-of-n) and `DEFAULT_ADMIN_ROLE` to a separate governance Safe; document Safe address + threshold in NatSpec.
[...]

PR #840, opened 2026-07-13
feat(bridge): BTH reserve release / send with exactly-once claims and threshold attestation

### LINKED ISSUE #822: bridge: implement BTH reserve release / send (replace engine.rs TODO) (Part of #816)
**Part of #816 (Phase 1). Blocked by #817 (custody model) and #818 (demurrage/peg).**

## Task
`engine.rs::process_burn_order` "Implement actual BTH sending" is a stub. Implement the BTH-side release: on a confirmed Ethereum burn, build and threshold-sign a BTH transaction from the reserve paying the user's stealth address (`bridgeBurn`'s `bthAddress`).

- Reserve spend authorized by the #817 t-of-n federation (no single hot key).
- Pay to a **fresh stealth address** to preserve the re-shield property (#819).
- Idempotent + crash-safe: a burn order releases **exactly once** even across service restarts (coordinate with the order-engine durability issue).
- Advance `ReleasePending → Released` only after BTH finality.

## Testing (required for DoD)
- Unit tests for release-tx construction + exactly-once guard.
- Testnet integration: burn on ETH testnet → release on BTH testnet, asserting a fresh stealth output and correct amount (net of the demurrage decision).
- Failure injection: restart mid-release, partial-signer set, replayed burn event → single release.

## Acceptance
- A confirmed burn releases exactly the right BTH once, to a fresh stealth address, authorized by threshold signatures.

rjwalters:
Closes #822

Part of the bridge epic #816. Builds directly on the mint engine merged in #838, reusing its record-before-broadcast idempotency, submit/confirm stage split, and guarded state-machine transitions for the burn/release side.

## What this does

Replaces the `process_burn_order` "Would send" stub in `bridge/service/src/engine.rs` with a real BTH reserve-release path for confirmed wBTH burns (source Ethereum or Solana, `dest_chain == Bth`). The mint path is untouched.

## Fully implemented (live + tested)

- **Exactly-once release claims** (`bridge/service/src/db.rs`): new `release_claims` table keyed on `order_id` (PRIMARY KEY) with a unique index on the deterministic order-id hash. The claim is taken in a `BEGIN IMMEDIATE` transaction BEFORE any signing; the signed tx **hash AND raw bytes** are recorded BEFORE the first broadcast. A concurrent tick or post-restart re-entry either finds no recorded tx (nothing was broadcast — safe to sign fresh) or reuses the recorded tx — the engine **never re-signs with different reserve inputs** (double-spend risk; BTH has no on-chain order-id guard, unlike the mint side's #826).
- **Submit/confirm split** (`bridge/service/src/engine.rs`): `submit_release` (`BurnConfirmed -> ReleasePending`, claim -> attest -> sign -> record -> status -> broadcast with bounded retries) and `confirm_release` (`ReleasePending -> Released` only at the configured finality depth; default 0 = SCP externalization). Resume path re-broadcasts the recorded raw bytes idempotently.
- **Unwind edge** (`bridge/core/src/order.rs`): `ReleasePending -> BurnConfirmed`, guarded by `can_transition_to` and restricted (in docs + engine semantics) to transactions that **provably cannot land** — a merely-unseen tx must stay Pending, because re-signing while the old tx could still land risks a double release. `Failed` releases keep the claim intact so retries reuse the same signed tx.
- **Threshold release authorization** (ADR 0002): new `ReleaseAuthorization` in `bridge/core/src/attestation.rs`, bound to order id + net amount + recipient through a domain-separated digest (`release_payload_digest`, tag `botho-bridge-release-v1`, golden-vector tested, mirroring the operator-signed-action `DOMAIN_SEPARATOR` pattern). `bridge/service/src/release/bth.rs` performs **real Ed25519 verification** (`verify_strict`) of every signature, enforces federation membership against the configured validator key set, a distinct-signer threshold, and a configured threshold floor — all BEFORE any reserve key material would be touched. Replay safety comes from the order-id binding + the claims table + the order state machine (one release per order id); the envelope/nonce collection protocol itself is #824.
- **Config knobs** (`bridge/core/src/config.rs`): `BthConfig.reserve_address`, `release_signers` (hex Ed25519 keys), `release_threshold`, `release_confirmations_required` (default 0 = SCP finality), threaded through `BridgeConfig::default` and TOML-backward-compatible.
- **Fail-safe wiring**: a misconfigured/absent releaser or a failed attestation leaves burn orders in `BurnConfirmed` (retryable) — they are never dropped or failed by configuration problems.

## Stubbed (fail-safe, after the attestation gate)

Following the exact `mint/solana.rs` pattern, the live-node RPC bodies in `bridge/service/src/release/bth.rs` return `NotImplemented` with `TODO(#828)` markers documenting the pinned pipeline:

- `prepare_release` tx construction: reserve UTXO scan (`chain_getOutputs` + ownership scan + `chain_areKeyImagesSpent`), **factor-1/background-only input selection** (ADR 0003), **fresh one-time stealth output** to the burn's `bthAddress` (ADR 0004; never a static payout key), factor-1-preserving change back to the reserve, ring decoy gathering, t-of-n threshold signing (#824) producing the CLSAG-signed tx, self-verify.
- `broadcast`: `tx_submit` wiring ("already known" = success).
- `check_confirmation`: depth polling vs `release_confirmations_required`; `Dropped` only for provably-dead txs.

Because the stub fires only AFTER attestation validation passes and BEFORE anything is recorded, today's deployments keep burn orders safely in `BurnConfirmed` with zero reserve exposure.

Issue-checklist items that inherently require the #828 live-node harness (fresh-stealth distinctness on-chain, factor-1 provenance of actual selected inputs, ETH-testnet -> BTH-testnet integration) are covered at the deterministic layer here (digest binding, validation gates, documented invariants) and complete under #828. Note: this PR keeps the #838 pattern that #839 (DB-layer transition enforcement) will harden — status updates go through guarded `can_transition_to` checks at the engine layer plus status-scoped SQL `WHERE` clauses.

## Test Plan

- [x] `cargo test -p bth-bridge-core -p bth-bridge-service` — 65 tests, all passing:
  - restart mid-release, both crash points: after claim + recorded hash (resumes with the SAME tx, `prepare` never re-called) and after claim but before signing (signs exactly once) — `test_release_resume_after_crash_reuses_recorded_tx`, `test_release_crash_after_claim_before_sign_is_safe`
  - replayed confirmed burn / repeated ticks -> exactly one signed release; a duplicate `record_release_tx` with a different hash never overwrites — `test_release_replayed_burn_single_release`, `test_release_claim_exactly_once`
  - threshold authorization: fewer than `t` signers, duplicate signer, threshold below the configured floor, attestation bound to a different order id / amount / recipient, signatures over a stale digest, non-federation signer, malformed or corrupted signature — all rejected before the construction stub — `release::bth` tests
  - finality gating: stays `ReleasePending` below the requirement, then `Released` with `dest_confirmed_at`; `Failed` keeps the claim; `Dropped` unwinds to `BurnConfirmed` and re-submits fresh — engine + db tests
  - release digest golden vector + field-binding; state-machine edges (`ReleasePending -> BurnConfirmed` unwind clears `dest_tx`, terminal `Released` frozen, burn orders cannot expire)
  - config: legacy TOML without the new knobs still parses; knobs round-trip
- [x] `cargo clippy --all-targets -p bth-bridge-core -p bth-bridge-service` — zero warnings
- [x] `cargo +nightly-2025-12-03 fmt -- --check` (repo's pinned-nightly convention) — clean
- [x] `cargo deny check advisories` — ok (Cargo.lock only adds the existing workspace `ed25519-dalek` to bridge-service)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2035,6 +2035,7 @@ dependencies = [
  "bth-bridge-core",
  "chrono",
  "clap",
+ "ed25519-dalek",
  "hex",
  "rusqlite",
  "serde_json",
--- a/bridge/core/src/attestation.rs
+++ b/bridge/core/src/attestation.rs
@@ -88,6 +88,103 @@ impl MintAuthorization {
     }
 }
 
+/// Domain-separation tag for the BTH reserve-release attestation payload.
+///
+/// Mirrors the operator-signed-action pattern (`botho/src/operator_action.rs`
+/// `DOMAIN_SEPARATOR`, per ADR 0002): every federation signature covers this
+/// tag so a release signature can never be confused with (or replayed as) an
+/// operator action, a mint attestation, or any other Ed25519 payload signed
+/// by a validator node key. Changing this tag invalidates all in-flight
+/// release authorizations.
+pub const RELEASE_ATTESTATION_DOMAIN_TAG: &[u8] = b"botho-bridge-release-v1";
+
+/// Compute the digest the federation signs to authorize one BTH release.
+///
+/// `sha256(domain_tag || order_id || amount_le || recipient_len_le ||
+/// recipient)`
+///
+/// The digest binds the authorization to:
+/// - the deterministic on-chain **order id** (replay safety: the release
+///   engine's `release_claims` table plus the order state machine allow at most
+///   one release per order id, so a replayed authorization cannot trigger a
+///   second reserve spend);
+/// - the exact **net amount** in picocredits; and
+/// - the exact **recipient** BTH address (length-prefixed so the encoding is
+///   unambiguous).
+///
+/// The #824 attestation protocol produces signatures over exactly these
+/// bytes; the release engine verifies them before any reserve key material
+/// is touched.
+pub fn release_payload_digest(order_id: &[u8; 32], amount: u64, recipient: &str) -> [u8; 32] {
+    use sha2::{Digest, Sha256};
+    let mut hasher = Sha256::new();
+    hasher.update(RELEASE_ATTESTATION_DOMAIN_TAG);
+    hasher.update(order_id);
+    hasher.update(amount.to_le_bytes());
+    hasher.update((recipient.len() as u64).to_le_bytes());
+    hasher.update(recipient.as_bytes());
+    hasher.finalize().into()
+}
+
+/// Threshold authorization for a single BTH reserve release, produced by the
+/// #824 attestation protocol and bound to one bridge burn order.
+///
+/// Per ADR 0002, releases are authorized by a t-of-n threshold of the SCP
+/// validators' **Ed25519 node keys** (the scheme is always Ed25519 on the
+/// BTH side, so no scheme field is carried). Each signature covers
+/// [`release_payload_digest`], binding the order id, net amount, and
+/// recipient address — a signature for one order can never authorize a
+/// different amount, recipient, or a second release.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ReleaseAuthorization {
+    /// The deterministic 32-byte order id this authorization is bound to.
+    /// Must equal [`crate::order::BridgeOrder::order_id_bytes`] for the
+    /// burn order being released.
+    #[serde(with = "hex_array_32")]
+    pub order_id: [u8; 32],
+
+    /// Net amount to pay the recipient, in picocredits
+    /// ([`crate::order::BridgeOrder::net_amount`]).
+    pub amount: u64,
+
+    /// The recipient's BTH address (`bridgeBurn`'s `bthAddress`). The
+    /// release transaction pays a **fresh one-time stealth output** derived
+    /// from this address (ADR 0004) — never a static payout key.
+    pub recipient: String,
+
+    /// The threshold `t` required by the federation configuration. Per
+    /// ADR 0002 this is never lower than the SCP safety threshold.
+    pub threshold: u32,
+
+    /// Collected validator Ed25519 signatures over
+    /// [`ReleaseAuthorization::digest`]. Must contain at least `threshold`
+    /// entries from distinct signers to be usable.
+    pub signatures: Vec<AttestationSignature>,
+}
+
+impl ReleaseAuthorization {
+    /// The digest every federation signature must cover.
+    pub fn digest(&self) -> [u8; 32] {
+        release_payload_digest(&self.order_id, self.amount, &self.recipient)
+    }
+
+    /// Whether enough distinct signers have signed to meet the threshold.
+    ///
+    /// This only counts distinct signer identities — cryptographic
+    /// verification of each signature (and federation membership) is the
+    /// release engine's job (`validate_release_attestation`).
+    pub fn meets_threshold(&self) -> bool {
+        let mut signers: Vec<&[u8]> = self
+            .signatures
+            .iter()
+            .map(|s| s.signer.as_slice())
+            .collect();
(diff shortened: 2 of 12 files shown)


rjwalters:
## Judge review: APPROVED

Thorough review of the BTH reserve-release path against issue #822's expanded checklist and ADRs 0002/0003/0004. CI is green on d4dc5a63 (rustfmt, cargo deny, coverage, workspace build 8m10s) and mergeStateStatus is CLEAN against main (a386bd42).

### Verified locally (worktree at d4dc5a63)
- `cargo test -p bth-bridge-core -p bth-bridge-service` — 65 tests (21 core + 44 service), all pass
- `cargo clippy --all-targets` for both crates — zero warnings (after the doc fix below)
- `cargo +nightly-2025-12-03 fmt -- --check` — clean
- `cargo deny check advisories` — ok (Cargo.lock only adds workspace ed25519-dalek to bridge-service)

### Priority areas

**Exactly-once (db.rs)** — `try_claim_release` takes the claim in a `BEGIN IMMEDIATE` transaction before any signing; `record_release_tx` persists hash + raw bytes with `WHERE release_tx_hash IS NULL` so a recorded tx is never overwritten, and the engine explicitly discards its own freshly signed tx (without broadcasting) when it loses the record race. Every crash window fails safe: after claim/before sign (nothing broadcast, fresh sign is safe), after record/before status (resume reuses the recorded tx, `prepare` never re-called), after broadcast (idempotent re-broadcast). Replayed burns / repeated ticks proven to sign exactly once.

**Attestation gate (release/bth.rs)** — `validate_release_attestation` fires first in `prepare_release`, before any reserve logic: `verify_strict` on every signature (one forged signature fails the whole attestation), federation membership pinned when configured, distinct-signer dedup, claimed threshold checked against the config floor. Digest is `sha256(tag || order_id[32] || amount_le[8] || recipient_len_le[8] || recipient)` — fixed-width + length-prefixed, unambiguous, golden-vector pinned.

**State machine (order.rs)** — new `ReleasePending -> BurnConfirmed` unwind via `can_transition_to`, clears `dest_tx`/`dest_confirmed_at`; `rollback_release` only deletes unconfirmed claims and is status-scoped; confirmed releases proven immune to late rollback. `Failed` keeps the claim (documented: retries must reuse the recorded tx, never sign a competing reserve spend) with warn + status visibility.

**Stubs** — all three RPC bodies are `TODO(#828)` `NotImplemented`, fire after the attestation gate and before anything is recorded, so orders stay retryable in `BurnConfirmed` with zero reserve exposure. The deterministic layer (digest, validation, claims, state machine) is fully live — nothing deterministic is stubbed. Deferrals match the pre-agreed #828 scope (UTXO scan, factor-1 selection, fresh stealth, decoys, CLSAG, tx_submit, confirmation polling); the checklist's envelope/nonce machinery is #824's protocol work, with replay safety here via order-id binding + claims table.

### Changes made during review (committed directly)
- d4dc5a63: reworded the release module doc — a wrapped line starting with `+` tripped clippy's `doc_lazy_continuation` (2 warnings on the previous head, in tension with rustfmt's comment wrapping). Doc-only.

### Follow-up filed
- #842 — `BthReleaser::new` accepts `release_threshold = 0` (the serde default) with a configured reserve address; combined with the stub attestation's empty signature set that would pass validation with zero signatures once #828 wires real construction. No exposure today (spend path stubbed after the gate), but the config-time guard should land before or with #828.

Solid, carefully reasoned work — the never-re-sign-after-record invariant and the provably-dead-only unwind restriction are exactly the right posture for a chain with no on-chain order-id guard.

Issue #816, opened 2026-07-13
epic: complete + harden the BTH↔Ethereum wBTH bridge (proof-of-reserves + comprehensive testing)


rjwalters:
## Goal

Take the existing bridge scaffold from a **stub** to a **production-grade, audited, well-tested** BTH↔Ethereum bridge that lets users wrap BTH into an ERC-20 (**wBTH**) for DeFi access and unwrap back to native BTH — with the wrapped/locked reserve recorded and auditable on the Botho chain (proof-of-reserves).

## Current state (audited 2026-07-12)

There is a real skeleton — but it predates today's chain and all money movement is a TODO:

- `bridge/core` — order state machine (Mint/Burn lifecycles), config (BTH/ETH/Solana), chain types. **Structurally there.**
- `bridge/service` — `engine.rs` (mint/release are all `TODO`), `watchers/{bth,ethereum}.rs`, `db.rs`, `main.rs`. **Wiring is stubbed.**
- `contracts/ethereum/contracts/WrappedBTH.sol` — ERC20 + AccessControl + Pausable + per-tx/daily/cooldown rate limits. **Fairly complete, unaudited.**
- `contracts/solana/programs/wbth` — Anchor wBTH program (scope TBD, see decision issue).
- `bth-crypto-secp256k1` — Ethereum key support present.

**Gaps:** custody is a single hot wallet (not a threshold federation); no on-Botho reserve accounting; ~9 tests total; unwired from the node; unmaintained since ~v0.2.0 (predates privacy/demurrage/SCP-federation/picocredits work).

## The Botho-specific hard problems (why this isn't a generic bridge revival)

1. **Custody = the honeypot.** Bridges are the #1 hack target in crypto. A single hot wallet is unacceptable; we need a **threshold-multisig federation** (candidate: reuse the operator-signed-action envelope + nonce + audit-log pattern just shipped in P4.4 / #748–#756).
2. **Privacy collision.** BTH amounts are Pedersen-hidden and senders ring-hidden; wBTH on Ethereum is fully transparent. Locking must **reveal the amount** to mint the right wBTH (a deliberate privacy exit); unwrapping can **re-shield** by releasing to a fresh stealth address, but amount/timing correlation remains. Must be explicit to users.
[...]

PR #844, opened 2026-07-13
bridge: on-Botho reserve accounting + proof-of-reserves (peg invariant)

rjwalters:
Closes #825 (part of epic #816, Phase 2).

The #818 blocker is resolved by ADR 0003 (merged): factor-1-only wrapping means the reserve pays zero demurrage, so the peg invariant is **exact** — `Σ(wBTH outstanding) == locked BTH reserve` — and the default drift tolerance is 0 (modulo in-flight orders). ADR 0005 extends the invariant across both wrapped chains. Everything here is application-level (bridge-controlled outputs + a reconciler); no consensus change.

## Implemented

**Reserve ledger (`bridge/service/src/db.rs`)**
- `reserve_ledger`: locked reserve derived from bridge-controlled outputs (`SUM WHERE locked=1`), never a mutable counter. Confirmed deposits record a locked output for the **net** amount (the mint fee is bridge revenue, not peg backing); confirmed releases spend outputs **FIFO within the chain** for the **gross** burn amount, returning the remainder as a locked change output — mirroring how the reserve wallet returns change (per-chain attribution supports the ADR 0005 per-chain invariant).
- `reserve_snapshots`: one row per reconciliation pass, so drift history is auditable.
- All mutations are exactly-once (idempotent by output id / release order id); an uncoverable spend rolls back atomically.

**Lifecycle wiring (`engine.rs`)**
- `submit_mint` locks the backing exactly once before any supply can be minted; a mint that *fails* after locking is unlocked (funds owed back to the depositor).
- `confirm_release` applies the idempotent reserve spend BEFORE the order leaves `release_pending` (crash-safe replay); a spend shortfall is audited (`reserve_spend_failed`) and surfaces as reconciler drift but never blocks a release that already landed on-chain.
- Reconciler + proof API run as engine tasks honoring the shutdown broadcast.

**Reconciler (`reserve.rs`)** — compares three legs:
1. expected reserve from the DB ledger,
2. on-chain wrapped supply per chain — **Ethereum implemented** (`WrappedBTH.totalSupply()` via the existing alloy provider pattern),
3. actual on-Botho reserve balance — **comparison logic implemented**, custody shortfall flips `pegHealthy` even when supplies match the ledger.

[...]

PR #847, opened 2026-07-13
bridge: federation attestation + authorization protocol (replay-safe, order-bound)

### LINKED ISSUE #824: bridge: cross-chain attestation + authorization protocol (replay-safe, order-bound) (Part of #816)
**Part of #816 (Phase 1). Blocked by #817.**

## Task
Define and implement how a federation authorizes the two privileged actions, since Botho cannot natively verify Ethereum and vice versa:

- **ETH mint authorization:** t-of-n signers attest "BTH deposit X for order O is final" → produces a signed authorization the `WrappedBTH.bridgeMint` path accepts (contract-level threshold, or a Safe holding `MINTER_ROLE`).
- **BTH release authorization:** t-of-n signers attest "wBTH burn for order O is final" → co-sign the reserve release tx.
- **Replay protection + order binding:** every attestation is bound to a unique **order id** and a nonce (reuse the operator-action reserve-then-apply nonce design, #749/#754), so an authorization can be consumed **exactly once** and cannot be replayed across orders or chains. Domain-separated signatures (distinct from wallet + operator-action domains).

## Testing (required for DoD)
- Property tests: no authorization applies to more than one order; below-threshold signer sets never authorize; cross-domain signature reuse fails.
- Adversarial tests: replayed attestation, forged signer, reordered/duplicated approvals.

## Acceptance
- Mint and release each require a valid, order-bound, single-use, t-of-n authorization; there is no replay or cross-order reuse path.

### LINKED ISSUE #842: bridge: reject release_threshold = 0 when the BTH releaser is enabled
Follow-up from the #840 review (BTH reserve release, #822).

`BthReleaser::new` (bridge/service/src/release/bth.rs) validates the reserve address, federation key parsing, and that `release_threshold` does not exceed the signer count — but it accepts `release_threshold = 0` (the serde default) even when `reserve_address` is set and `release_signers` is non-empty.

With threshold 0, `validate_release_attestation` passes an authorization carrying **zero signatures** (0 distinct signers >= threshold 0, and `auth.threshold` 0 >= floor 0) — which is exactly what today's `StubAttestationProvider::authorize_release` produces. There is no reserve exposure right now because `prepare_release` is a `TODO(#828)` `NotImplemented` stub that fires only after the attestation gate, but once #828 wires real transaction construction, an operator who sets `reserve_address` and forgets `release_threshold` would have a releaser that signs reserve spends with no federation authorization at all.

Suggested hardening (cheap, config-time, fail-safe):
- In `BthReleaser::new`, return `ReleaseError::Config` when the releaser is otherwise enabled (reserve address present) but `release_threshold == 0` — burn orders then stay `BurnConfirmed` exactly like every other misconfiguration, per the engine's `build_releaser` fail-safe.
- Optionally keep a documented escape hatch for development (e.g. only permit threshold 0 when `release_signers` is also empty, which is already documented as dev-only no-pinning mode).

Must land before or with #828; #824's protocol work is adjacent but does not currently include this config-time check.

rjwalters:
Closes #824
Closes #842

Part of the bridge epic #816 (ADR 0002 custody). Implements the t-of-n federation attestation protocol: SCP validators sign domain-separated, order-bound, single-use envelopes over confirmed deposits/burns; the bridge verifies, replay-checks, order-binds, and aggregates them to threshold into the `MintAuthorization` / `ReleaseAuthorization` artifacts the mint (#838) and release (#840) paths already consume.

## Implemented (real, tested)

**`bridge/core/src/attestation.rs` — envelope protocol** (mirrors `operator_action.rs`, P4.4)
- `AttestationKind` v1 allowlist: `bridge.mint_wbth` / `bridge.release_bth`, each carrying chain, recipient, amount, order UUID, finalized source tx (+ the Safe nonce for Ethereum mints).
- Per-target-chain envelope domains (`botho-bridge-attest-{eth,sol,bth}-v1`): a signature over one chain's envelope can never verify under another's (cross-domain binding).
- Canonical-JSON envelope with parse-after-verify: verification is over the received bytes; the parser rejects unknown/duplicate keys (at every nesting level), non-integer numerics, non-canonical chain aliases, and unknown versions — one signed byte string can never carry two logical attestations.
- Two detached signatures per envelope: the domain-separated envelope signature (authenticates nonce/freshness/action) and the chain-specific **payload** signature (`release_payload_digest`, `mint_payload_digest`, or the Gnosis Safe EIP-712 SafeTx hash) — the payload signature is what survives into the authorization and is re-verified by the mint/release path before any privileged action.
- Freshness window (300s max lifetime, 30s skew, saturating arithmetic), `check_order_binding` (order id, net amount, recipient, chains, source tx all must match the on-record order), `AttestationSet` (distinct-signer dedupe; Ethereum sets refuse to mix Safe nonces; threshold 0 never authorizes), rejection taxonomy with authenticated/pre-signature split.

**`bridge/core/src/nonce.rs` — durable replay store** (port of `botho/src/operator_nonce.rs`)
- Reserve-then-apply `(signer, nonce)` consumption, atomic owner-only persistence, GC by expiry, corrupt-store = hard error (never a silent wipe).

**`bridge/service/src/attestation.rs` — `FederationAttestationProvider`**
- Fail-closed ingest pipeline (`submit_attestation`): peek routing (public data only) → signature verify per ADR 0002 key domain → parse-after-verify → freshness → durable nonce reserve → order binding → threshold aggregation.
- Key domains per ADR 0002: **Ed25519** (`verify_strict`) for BTH releases + Solana mints; **secp256k1** for Ethereum mints — the payload signature is a Safe owner signature over the EIP-712 SafeTx hash wrapping `bridgeMint(to, amount, orderId)` at the attested Safe nonce, so aggregated signatures are directly consumable by `Safe.execTransaction` (via the existing `safe_tx_hash` helper).
- Local self-attestation goes through the SAME pipeline (no special trust for the node's own envelopes).
- Config: `ethereum.mint_signers`/`mint_threshold`, `solana.mint_signers`/`mint_threshold` (BTH release federation reuses the existing #840 `bth.release_*` fields), local signing key paths, nonce-store path. Misconfiguration installs a fail-closed `DisabledAttestationProvider` — never the permissive dev stub.

**#842 — `BthReleaser::new`** now refuses `release_threshold == 0` when release signers are configured (a zero-threshold federation would authorize reserve spends with zero signatures). The documented dev-only escape hatch (no signers at all) still constructs. Same guard at provider construction (`check_threshold`).

**Engine**: provider selection (federation / disabled / stub) replaces the #824 TODO; every threshold authorization is audit-logged with action, threshold, and signer count.

## Stubbed (fail-safe)

- **Network transport between federation members is TODO(#828)**: `submit_attestation` is the ingest seam the transport endpoint will call; until it lands, a threshold above the locally-signable count simply never authorizes — orders stay in their confirmed, retryable state. No permissive fallback.
- Per-attestation JSONL audit entries ride the tracing log for now; the DB audit row is written at authorization time. Full per-envelope audit persistence can move into the #828 endpoint where the DB handle lives.

## Test Plan

- [x] `cargo test -p bth-bridge-core -p bth-bridge-service` — 119 passed, 0 failed (40 core + 79 service)
- [x] `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets` — 0 warnings
- [x] `cargo +nightly-2025-12-03 fmt --check` — clean
- [x] `cargo deny check advisories` — ok (Cargo.lock adds only the workspace `ed25519-dalek` edge to `bth-bridge-core`)

Coverage of the issue's required adversarial/property tests:
- **Replay**: same envelope twice → `refused:replayed_nonce`; nonce store reopen within the window keeps the slot closed (`nonce::tests::replay_protection_survives_a_restart_within_the_window`).
- **Cross-order reuse**: envelope for order A vs order B → `refused:wrong_order`, even with a fresh nonce (core + pipeline tests).
- **Cross-domain reuse**: signatures fail across BTH/Solana/Ethereum domains; Ethereum-target envelopes are refused on the Ed25519 path (core tests).
- **Below threshold**: t−1 distinct signers never authorize; the t-th flips it (`authorize_release_meets_threshold_and_output_verifies_downstream`, `authorize_mint_eth_collects_safe_owner_signatures_to_threshold`).
- **Byzantine signer**: non-federation key → `refused:unknown_signer` (doesn't count); tampered bytes → `refused:bad_signature`.
- **Double-count resistance**: same signer, fresh nonces → counts once toward t.
- **End-to-end**: a collected `ReleaseAuthorization` passes the release path's own `validate_release_attestation` (#840), and every collected Ethereum payload signature recovers to a Safe owner over the exact SafeTx digest.
- **Structural attacks**: duplicate keys, unknown fields, float amounts, chain aliases, version downgrades all rejected before parse output is used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2018,6 +2018,7 @@ name = "bth-bridge-core"
 version = "0.3.2"
 dependencies = [
  "chrono",
+ "ed25519-dalek",
  "hex",
  "serde",
  "serde_json",
--- a/bridge/core/Cargo.toml
+++ b/bridge/core/Cargo.toml
@@ -19,5 +19,8 @@ toml = { workspace = true }
 hex = { workspace = true }
 sha2 = { workspace = true, features = ["std"] }
 
-[dev-dependencies]
+# Attestation envelope protocol (#824): federation Ed25519 signatures and
+# canonical-JSON envelope parsing, mirroring the operator-signed-action
+# machinery.
+ed25519-dalek = { workspace = true, features = ["std"] }
 serde_json = { workspace = true, features = ["std"] }
--- a/bridge/core/src/attestation.rs
+++ b/bridge/core/src/attestation.rs
@@ -10,12 +10,26 @@
 //!   `MINTER_ROLE` on `WrappedBTH.sol`.
 //! - **Solana**: validators sign natively with their Ed25519 node keys.
 //!
-//! The attestation *protocol* (signature collection, envelopes, nonces) is
-//! issue #824 and lives outside this crate. This module only defines the
-//! artifact that protocol produces — [`MintAuthorization`] — which the mint
-//! submission path (issue #821) consumes.
+//! This module defines both the **artifacts** the protocol produces
+//! ([`MintAuthorization`] / [`ReleaseAuthorization`], consumed by the mint
+//! and release submission paths) and the **attestation envelope protocol**
+//! itself (#824): domain-separated, order-bound, single-use signed
+//! envelopes mirroring the operator-signed-action machinery
+//! (`botho/src/operator_action.rs`, P4.4), aggregated to the federation
+//! threshold by [`AttestationSet`]. Signature *collection over the network*
+//! between federation members is TODO(#828); everything cryptographic —
+//! signing, verification, replay rejection, order binding, thresholding —
+//! is implemented here and in `bridge/service/src/attestation.rs`.
 
+use ed25519_dalek::{Signature as Ed25519Signature, SigningKey, VerifyingKey};
 use serde::{Deserialize, Serialize};
+use serde_json::{Map, Value};
+use uuid::Uuid;
+
+use crate::{
+    chains::Chain,
+    order::{derive_order_id, BridgeOrder, OrderType},
+};
 
 /// The signature scheme used by an attestation, per destination chain.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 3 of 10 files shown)


rjwalters:
## Judge review — APPROVED

Adversarial review of the attestation/authorization protocol (#824 + #842), post-rebase onto the #844 merge (head `0b9f0833`).

**Verification**: `cargo test -p bth-bridge-core -p bth-bridge-service` — 143 passed (41 core + 102 service), 0 failed. `cargo clippy --all-targets` clean, pinned-nightly `fmt --check` clean, `cargo deny check advisories` ok. All 4 CI checks green on the rebased head; `mergeStateStatus: CLEAN`.

### Findings per priority area

**Replay / order binding** — sound. The payload digests (`release_payload_digest` / `mint_payload_digest`) bind domain tag + order id + amount + length-prefixed recipient; golden vectors pinned. `check_order_binding` additionally pins source tx, chains, and order type against the on-record order, so a validly-signed attestation for order A is rejected against order B even with a fresh nonce (`pipeline_rejects_cross_order_reuse_even_with_a_fresh_nonce`). Envelope-level replay dies at the durable `(signer, nonce)` reserve; freshness bounds the window (`expires_at <= now + 330s`, saturating arithmetic).

**Canonical encoding / parse-after-verify** — sound. Verification is over the received bytes under the chain-selected domain; parsing happens only after the envelope signature verifies. Duplicate keys are rejected at every nesting level via the streaming `DeserializeSeed` checker (defeats serde_json last-wins), unknown keys/versions/actions rejected, floats rejected, and `parse_canonical_chain` refuses the aliases `Chain::FromStr` accepts — one canonical byte encoding per logical attestation. The two pre-verify peeks (`signerKeyId`, target chain) drive only key/domain selection; a lying peek changes the signed message and fails verification.

**Aggregation** — sound. `AttestationSet` dedupes by signer id; same signer with fresh nonces counts once (tested at core and pipeline level). Signer ids are matched by exact string equality against `hex::encode` output (lowercase), so case-aliased ids fail as `UnknownSigner` rather than double-counting. `is_threshold_met` hard-refuses threshold 0. Unknown signers and tampered bytes never enter the set.

**Key domains** — sound. Ed25519 uses `verify_strict` (both envelope and payload signatures, same key). secp256k1 recovers the address from the prehash and compares against the config-pinned owner `Address` (canonical, so no checksum-case games). The Ethereum payload signature is re-verified over the SafeTx hash reconstructed from the *parsed* attestation with config-pinned chain id/Safe/wBTH — and the e2e test recovers every collected signature to a Safe owner over that exact digest. Ethereum-target envelopes are refused pre-signature on the Ed25519 path; Safe-nonce mixing within a set is refused.

**NonceStore** — sound. Reserve-then-apply with atomic temp-file+rename persistence (0600), rollback of the in-memory insert on persist failure, corrupt-store = hard error (never a silent wipe), restart-within-window replay test present. GC retention is bounded because freshness caps `expires_at` before reserve. The `\0` join is injective since signer ids are canonical hex.

**Fail-closed** — sound. `DisabledAttestationProvider` errors on every call and is installed on any config error (never the permissive stub). `check_threshold` rejects zero AND unsatisfiable thresholds at provider construction, and **#842's guard in `BthReleaser::new` refuses `release_threshold == 0` when signers are configured**, with the documented no-signers dev escape hatch preserved — both tested. The #828 transport stub can only fail to reach threshold, never authorize.

**Integration** — real, not mocked. `authorize_release_meets_threshold_and_output_verifies_downstream` feeds the collected authorization through the actual #840 `validate_release_attestation` against the pinned federation; the Ethereum test verifies SafeTx owner recovery against `assemble_safe_signatures`' digest. Engine gating audit-logs every threshold authorization.

**Post-rebase sanity (#844)** — the engine diff is additive (provider selection + two audit-log calls); all #844 reserve-ledger tests pass on the merged head.

### Non-blocking follow-up

Filed **#849**: a partially-filled Ethereum `AttestationSet` pinned to a Safe nonce the on-chain Safe advances past can never complete and is never evicted (in-memory, cleared on restart, order stays retryable). Pure liveness, only reachable once #828's multi-node transport exists — suggested fix is set eviction/re-collection when the pinned nonce falls behind.

Issue #824's checklist is fully covered (the per-envelope JSONL audit rides tracing until the #828 endpoint owns the DB handle, as disclosed); #842's ask is implemented at both the releaser and provider layers. Approving.

Issue #853, opened 2026-07-13
bridge: implement live reserve transports — Solana getTokenSupply supply source + BTH reserve-address balance source


rjwalters:
Part of #816. Split out of #846 (item 4): the two reconciler transports are fail-safe `NotImplemented` stubs whose TODOs previously pointed at #828 (the test-infrastructure issue). This issue is their concrete implementation home.

## Task
In `bridge/service/src/reserve.rs`:

1. **`SolSupplySource`** — implement against `SolanaConfig::rpc_url` via JSON-RPC `getTokenSupply` for the configured wBTH SPL mint; return the supply in picocredit-equivalent units consistent with `EthSupplySource`.
2. **`NodeReserveBalanceSource`** — implement against the BTH node transport (view-key scan of the reserve address, mirroring how the wallet derives balances) so the reconciler's custody leg (`reserve_balance_checked`) verifies the actual reserve balance instead of being skipped.

Both must keep the current fail-safe behavior on transport error: `ReserveError::NotImplemented`/RPC failures leave the leg unverified (surfaced via `reserveBalanceChecked` / nullable supplies, #846) rather than reporting a false-healthy peg.

## Acceptance
- Reconciler passes verify both legs against a live (or mocked-RPC) endpoint; `reserve_balance_checked: true` on the proof API when the BTH leg runs.
- The `NotImplemented` stub paths are retired.

Issue #825, opened 2026-07-13
bridge: on-Botho reserve accounting + proof-of-reserves (peg invariant) (Part of #816)


rjwalters:
**Part of #816 (Phase 2). Blocked by #818 (demurrage/peg definition).**

## Task
Record and expose the bridge reserve on the Botho chain so the peg is auditable (the maintainer's requirement: "the Botho chain itself could record funds that have been wrapped").

- Maintain the invariant **`Σ(wBTH outstanding on Ethereum) == locked BTH reserve on Botho`** (over time, per the #818 demurrage decision).
- Track locked reserve as bridge-controlled outputs/account; reconcile continuously against the on-chain wBTH `totalSupply`.
- Surface a **proof-of-reserves** view via the existing metrics-daemon + explorer stack (`/network` dashboard) — locked reserve, wrapped supply, and any drift, updated live.
- Alert on any drift beyond tolerance (a peg break or a custody incident).
- Prefer an **application-level** convention (bridge-controlled outputs + reconciler) over a consensus/protocol construct, to respect the no-hard-forks posture — unless #818 forces protocol support, in which case escalate.

## Testing (required for DoD)
- Property test: the invariant holds across randomized mint/burn/demurrage sequences.
- Integration test: reserve view matches wBTH totalSupply on testnet after a batch of wraps/unwraps.
- Drift-injection test: an unauthorized reserve movement trips the alert.

## Acceptance
- A live, auditable proof-of-reserves showing reserve == wrapped supply (within the demurrage-defined tolerance), with alerting on drift.

rjwalters:
*Implementation tasks expanded 2026-07-13 (per ADR 0003 peg invariant / factor-1 zero-demurrage reserve, ADR 0005 two-chain invariant). Application-level reconciler, no protocol change.*

## Implementation tasks

- [ ] Add a `reserve` module to `bridge/service/src/` (new `reserve.rs`, declared in `main.rs`) owning reserve-tracking, on-chain supply polling, reconciliation, and the proof-of-reserves snapshot type.
- [ ] Extend `Database::migrate` (`bridge/service/src/db.rs`) with a `reserve_ledger` table (`chain`, `bridge_output_id`, `amount_picocredits`, `locked` bool, `order_id`, `created_at`, `spent_at`) and a `reserve_snapshots` table (`taken_at`, `locked_reserve`, `eth_supply`, `sol_supply`, `drift`, `in_tolerance`) with indexes on `chain`/`locked`.
[...]

PR #854, opened 2026-07-13
bridge service: crash-safe idempotent order engine + monitoring / circuit-breaker (#827)

### LINKED ISSUE #827: bridge service: crash-safe idempotent order engine + monitoring / circuit-breaker (Part of #816)
**Part of #816 (Phase 2).**

## Task
Make the `bridge/service` order engine production-operable.

- **Crash-safety / idempotency:** every order transition (`db.rs`) is durable and replay-safe so a restart mid-flow never double-mints or double-releases (coordinate with #822/#823 exactly-once guards).
- **Rate limits + circuit-breaker:** enforce per-order and aggregate caps mirroring the contract limits; a global kill-switch that halts mint/release on anomaly.
- **Monitoring/alerting:** health, order backlog, stuck/expired orders, peg drift (from #825), signer availability.
- **Ops runbook:** key rotation, pause/resume, incident response — mirroring the operator runbooks (#752/#760).

## Testing (required for DoD)
- Chaos/restart tests: kill the service at each state transition; assert exactly-once outcomes.
- Load test the engine + DB.
- Alerting tests: each alert fires on its trigger condition.

## Acceptance
- The engine survives crashes without double-spending either side, enforces limits, and is observable + operable.

### LINKED ISSUE #839: bridge: enforce OrderStatus::can_transition_to at the DB layer (update_order_status)

### LINKED ISSUE #843: bridge: BTH deposit crash window can strand an order at DepositDetected
Found during review of #841 (part of epic #816). Safety is intact — this is a narrow **liveness** gap only.

## The window

`BthWatcher::process_deposit` (bridge/service/src/watchers/bth.rs) runs, in order:

1. `record_deposit_detected` — order `AwaitingDeposit -> DepositDetected` (+ revealed amount + source_tx), SQL-guarded on `status = 'awaiting_deposit'`
2. `mark_deposit_processed` — inserts the `processed_deposits` idempotency row
3. audit entries
4. `update_order_status(DepositConfirmed)`

A crash between step 1 and step 4 leaves the order at `DepositDetected` permanently:

- Crash between 1 and 2: on replay the tx is not marked processed, but the order is no longer `AwaitingDeposit`, so the watcher logs "not awaiting a deposit — skipping" and returns without confirming (and without ever marking the tx processed).
- Crash between 2 and 4: on replay `is_deposit_processed` short-circuits.

The engine's `OrderProcessor` only picks up orders from `deposit_confirmed` onward, so nothing ever advances the stranded order. The depositor's funds are locked with no mint until an operator intervenes.

## Why it is safety-safe

The `status = 'awaiting_deposit'` guard means the detect can never fire twice, so no double-mint is possible; the revealed amount and `source_tx` are already durably recorded.

## Suggested fix

Since detection implies finality in this watcher (scanned blocks always meet the finality requirement), a startup/per-scan sweep that advances any `deposit_detected` order with a recorded `source_tx` to `deposit_confirmed` (with an audit entry) closes the window. Alternatively fold steps 1–4 into one SQLite transaction like `insert_burn_order` does on the burn side.

### LINKED ISSUE #846: bridge: reserve accounting follow-ups — value-based unlock for failed mints, expose verification flags in proof API, snapshot pruning, transport-TODO homes
Follow-ups from the #844 review of the #825 reserve ledger (part of epic #816). None are peg-safety regressions — all fail in the loud/over-collateralized direction — but each deserves a fix:

1. **Failed-mint unlock can miss after FIFO consumed its output.** `unlock_outputs_for_order` unlocks `WHERE order_id = ? AND locked = 1`. If a release's FIFO spend already consumed the failed mint's `dep:` output (its residual value now lives in a `chg:` output attributed to the release order), the unlock finds nothing and the ledger permanently overcounts by the failed mint's net amount → persistent `reserve_drift_alert` (locked − supply > in_flight, since Failed orders leave the allowance) requiring manual reconciliation. Fix: unlock by **value** (FIFO, mirroring `apply_release_spend`, change-output semantics) rather than by specific output id. Narrow window (mint must fail after a deep-FIFO release lands), safe direction, but it is a forever-red false alarm when it hits.
2. **Proof API can't distinguish "custody checked OK" from "custody never checked".** `ReserveProof.reserve_balance_checked` and per-chain `verified` flags exist in the reconciler output (and land in `audit_log` via `reserve_reconcile`), but `reserve_snapshots` and therefore `GET /api/reserve/proof` / `GET /api/metrics/reserve` omit them. A consumer sees `pegHealthy: true` with no signal that the BTH reserve-balance leg (stubbed until the transport lands) was skipped, and can only infer chain verification from nullable supplies. Persist `reserve_balance_checked` (and ideally the per-chain detail) in the snapshot and surface it through both APIs, so the dashboard can render "peg healthy (custody unverified)" honestly.
3. **Snapshot tables are unbounded.** Bridge `reserve_snapshots` grows one row per reconcile pass (default 60s ≈ 525k rows/yr) and the daemon's copy one row per poll. Rows are tiny so this is not urgent, but add retention/pruning (the daemon already has a rollup pattern).
4. **Transport TODOs point at a testing issue.** The Solana `getTokenSupply` supply source and the BTH reserve-address balance source are `TODO(#828)` stubs, but #828 is the test-infrastructure issue. Give the two transports a concrete implementation home (this issue or dedicated ones) so the fail-safe `NotImplemented` paths get retired.

rjwalters:
Closes #827
Closes #839
Closes #843
Closes #846

Part of the bridge epic #816. Makes the `bridge/service` order engine production-operable on top of the exactly-once machinery from #838/#840/#841/#844/#847.

## What's implemented

**Chaos/restart suite (DoD centerpiece)** — `bridge/service/src/chaos_tests.rs`
- Drives mint and burn orders through the full lifecycle over a file-backed SQLite DB with mocked chains, crashing (dropping engine + DB handles) and rebooting via the real startup path (`recover_on_startup`) at **every** durable transition boundary: 6 mint crash points, 2 deposit-detection crash points (#843), 6 release crash points, plus a rolling restart-after-every-tick run.
- Asserts exactly-once everywhere: the mock Ethereum chain enforces the #826 contract order-id guard and counts duplicate-mint *attempts* (must stay 0); BTH deliberately has **no** on-chain guard, so a second landed release tx is the double-release bug the suite hunts; the reserve ledger must be arithmetically exact after every scenario; terminal states are stable across further restarts.
- Load test: 240 concurrent orders with 4 engine tick loops hammering the shared connection — no deadlocks, no lost orders, chain-level exactly-once, exact reserve arithmetic. Runs in ~1s (not `#[ignore]`d).

**DB-layer state machine (#839)** — `update_order_status` reads the current status inside the same immediate transaction and rejects any edge `OrderStatus::can_transition_to` disallows; same-status writes are idempotent refreshes; terminal states cannot be clobbered by any writer through this path.

**Startup recovery (#843)** — `recover_on_startup` (run before any watcher/loop) rolls `DepositDetected` orders with a durable `source_tx` forward to `DepositConfirmed` exactly once (audited `deposit_recovered`), also writing the idempotency row so the watcher's block replay short-circuits — both halves of the crash window closed. Covered by unit + chaos tests.

**Rate limits + circuit breaker**
- Config-driven per-order (`max_order_amount`), per-address daily and global daily volume caps, reserved exactly-once per order in `limit_reservations` (crash/tick replay counts each order once). Daily-window rejections defer (retry next window); rejections are audited `rate_limited` exactly once per order.
- Breaker auto-trips (fail closed) on: reserve drift alert from the reconciler (#825/#844 peg incident), actionable backlog above `max_pending_orders`, and global-daily-cap exhaustion (anomalous volume).
- Kill switch: `bridge.paused = true` config (start paused) + runtime `POST /api/breaker` toggle. Pausing halts the **submit** stages only; **confirm** stages keep running so in-flight orders settle to durable terminal states. Pause state persists in the DB across restarts.

**Monitoring / alerting**
- `GET /health` (503 with reason when paused/DB-down), `GET /api/status` (pause state, per-status order counts, actionable backlog, stuck orders, component health, latest peg verdict), `GET /metrics` (hand-rolled Prometheus text — no new dependency), `POST /api/breaker`.
- Component availability (attestation signer, per-chain minters, releaser) recorded at startup and surfaced.
- Stuck-order alerts: post-deposit orders never auto-expire (funds moved); they alert exactly once per order (`stuck_order_alert`) and stay on the `bridge_stuck_orders` gauge until resolved.
- `bridge/service/alerts.yml`: Prometheus rules (breaker tripped, peg unhealthy, backlog, stuck orders, component down, API down), each pointing at the runbook.
- Alerting tests: every alert fires on its trigger (drift→trip, backlog→trip, global cap→trip, stuck order→audit-once, breaker toggle→health 503, component health surface).

**Reserve follow-ups (#846) — all four items**
1. Value-based unlock for failed mints: `unlock_backing_for_order` unlocks the order's own outputs first, then FIFO-by-value with change-output semantics (mirroring `apply_release_spend`) when a release's FIFO spend already consumed the `dep:` output — no more permanent overcount / forever-red drift alarm. Idempotent by attribution.
2. `reserve_balance_checked` persisted in `reserve_snapshots` and exposed via `GET /api/reserve/proof` (`reserveBalanceChecked`) and `/metrics`, so "peg healthy (custody unverified)" is honestly distinguishable.
3. Snapshot pruning: `reserve.snapshot_retention_days` (default 30, 0 disables), pruned each reconcile pass, latest snapshot always survives.
4. Transport homes: the Solana `getTokenSupply` and BTH reserve-balance stubs are re-homed from the test-infra issue #828 to the new dedicated implementation issue #853.

**Ops runbook** — `docs/operations/runbooks/bridge-order-engine-recovery.md`: breaker/kill-switch operation (API + config + contract-level pause layers), automatic-trip table, resume safety checklist, peg-drift / stuck-order / signer-outage incident response, key rotation (attestation Ed25519 + secp256k1/Safe, reserve spend key, submitter key), SQLite online backup + restore procedure with the one restore hazard called out (post-backup release claims), symptom→action map.

## Deferred (with homes)
- Live reserve transports (Solana supply, BTH reserve balance): #853 (new, split out of #846 item 4).
- Federation envelope transport / live RPC wiring: #828 as before.

## Test Plan
- [x] `cargo test -p bth-bridge-core -p bth-bridge-service` — 169 passed, 0 failed (42 core + 127 service, including the 5 chaos/load tests)
- [x] `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets` — 0 warnings
- [x] `cargo +nightly-2025-12-03 fmt --check` — clean
- [x] `cargo deny` — n/a (no dependency changes)
- [x] Chaos suite: `cargo test -p bth-bridge-service chaos` (mint/deposit/release crash matrices + restart-per-tick) and `load_test_concurrent_orders_exactly_once`

### DIFF
--- a/bridge/core/src/config.rs
+++ b/bridge/core/src/config.rs
@@ -50,6 +50,13 @@ pub struct ReserveSettings {
     /// the server.
     #[serde(default = "default_reserve_api_listen")]
     pub api_listen: String,
+
+    /// Days of reconciliation snapshot history to retain (#846: the table
+    /// grows one row per pass — ~525k rows/yr at the 60s default). Older
+    /// rows are pruned each pass; the most recent snapshot always
+    /// survives. 0 disables pruning.
+    #[serde(default = "default_snapshot_retention_days")]
+    pub snapshot_retention_days: u64,
 }
 
 fn default_reconcile_interval_secs() -> u64 {
@@ -60,12 +67,17 @@ fn default_reserve_api_listen() -> String {
     "127.0.0.1:9741".to_string()
 }
 
+fn default_snapshot_retention_days() -> u64 {
+    30
+}
+
 impl Default for ReserveSettings {
     fn default() -> Self {
         Self {
             tolerance_picocredits: 0,
             reconcile_interval_secs: default_reconcile_interval_secs(),
             api_listen: default_reserve_api_listen(),
+            snapshot_retention_days: default_snapshot_retention_days(),
         }
     }
 }
@@ -239,6 +251,20 @@ pub struct BridgeSettings {
     #[serde(default = "default_max_retries")]
     pub max_retries: u32,
 
+    /// Start with the circuit breaker tripped: the engine pauses the
+    /// submit stages (mints and releases) at startup until an operator
+    /// resumes via the API. Confirmation stages keep running so in-flight
+    /// orders settle. The runtime pause state lives in the database
+    /// (`bridge_state`) and survives restarts independently of this flag.
+    #[serde(default)]
+    pub paused: bool,
+
+    /// Actionable-order backlog above which the circuit breaker trips
+    /// automatically (a flood of orders is an anomaly signal — fail
+    /// closed and let an operator inspect). 0 disables the auto-trip.
+    #[serde(default = "default_max_pending_orders")]
+    pub max_pending_orders: u64,
+
     /// Enable testnet mode
     #[serde(default)]
     pub testnet: bool,
@@ -288,6 +314,10 @@ fn default_order_expiry() -> i64 {
 
 fn default_max_retries() -> u32 {
     3
+}
+
+fn default_max_pending_orders() -> u64 {
+    1_000
 }
 
 /// Gas price strategy for Ethereum transactions.
@@ -376,6 +406,8 @@ impl Default for BridgeConfig {
                 global_daily_limit: default_global_daily_limit(),
                 order_expiry_minutes: default_order_expiry(),
                 max_retries: default_max_retries(),
+                paused: false,
+                max_pending_orders: default_max_pending_orders(),
                 testnet: false,
                 attestation_ed25519_key_file: None,
                 attestation_secp256k1_key_file: None,
@@ -408,6 +440,35 @@ mod tests {
         let config = BridgeConfig::default();
         assert_eq!(config.bridge.fee_bps, 10);
         assert!(!config.bridge.testnet);
+        assert!(!config.bridge.paused);
+        assert_eq!(config.bridge.max_pending_orders, 1_000);
+    }
+
+    #[test]
+    fn test_breaker_knobs_parse() {
+        // A pre-existing config without the breaker knobs still parses
+        // with safe defaults, and the knobs round-trip from TOML.
+        let legacy: BridgeSettings = toml::from_str(
+            r#"
+            mnemonic_file = "m.enc"
+            db_path = "bridge.db"
+            "#,
+        )
+        .unwrap();
+        assert!(!legacy.paused);
+        assert_eq!(legacy.max_pending_orders, 1_000);
+
+        let configured: BridgeSettings = toml::from_str(
+            r#"
+            mnemonic_file = "m.enc"
+            db_path = "bridge.db"
+            paused = true
+            max_pending_orders = 50
+            "#,
+        )
+        .unwrap();
+        assert!(configured.paused);
+        assert_eq!(configured.max_pending_orders, 50);
     }
 
     #[test]
@@ -463,6 +524,7 @@ mod tests {
         let legacy: ReserveSettings = toml::from_str("").unwrap();
         assert_eq!(legacy.tolerance_picocredits, 0);
         assert_eq!(legacy.reconcile_interval_secs, 60);
+        assert_eq!(legacy.snapshot_retention_days, 30);
 
         // The knobs round-trip from TOML; empty api_listen disables.
         let configured: ReserveSettings = toml::from_str(
--- a/bridge/service/alerts.yml
+++ b/bridge/service/alerts.yml
@@ -1,0 +1,81 @@
+# Prometheus alert rules for the BTH bridge service (#827).
+#
+# Scrape target: GET /metrics on `[reserve] api_listen` (default
+# 127.0.0.1:9741). Response procedures for every alert:
+# docs/operations/runbooks/bridge-order-engine-recovery.md
+
+groups:
+  - name: bth-bridge
+    rules:
+      - alert: BridgeBreakerTripped
+        expr: bridge_paused == 1
+        for: 1m
+        labels:
+          severity: critical
+        annotations:
+          summary: "Bridge circuit breaker is tripped (submits halted)"
+          description: >
+            The bridge is paused (kill switch or auto-trip: backlog, global
+            volume cap, or reserve drift). Confirm stages keep settling
+            in-flight orders. Triage per the runbook before resuming via
+            POST /api/breaker.
+
+      - alert: BridgePegUnhealthy
+        expr: bridge_peg_healthy == 0
+        for: 1m
+        labels:
+          severity: critical
+        annotations:
+          summary: "Proof-of-reserves reconciliation reports an unhealthy peg"
+          description: >
+            Wrapped supply vs locked reserve drifted beyond tolerance, or the
+            reserve custody check failed. The reconciler trips the breaker
+            automatically. Treat as a peg incident: freeze, reconcile, then
+            resume.
+
+      - alert: BridgeBacklogHigh
+        expr: bridge_actionable_backlog > 100
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: "Bridge actionable-order backlog is growing"
+          description: >
+            Orders are entering faster than they settle. At
+            bridge.max_pending_orders the breaker trips automatically.
+
+      - alert: BridgeStuckOrders
+        expr: bridge_stuck_orders > 0
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          summary: "Orders stuck past the age threshold"
+          description: >
+            Post-deposit orders have not advanced within
+            bridge.order_expiry_minutes. Funds have moved: they never
+            auto-expire — operator triage required (see the stuck-order
+            section of the runbook).
+
+      - alert: BridgeComponentDown
+        expr: bridge_component_healthy == 0
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          summary: "A bridge component (signer / minter / releaser) is unavailable"
+          description: >
+            The engine fails closed for the affected flow (orders stay
+            retryable, nothing is dropped), but throughput is degraded until
+            the component's configuration is fixed.
+
+      - alert: BridgeApiDown
+        expr: up == 0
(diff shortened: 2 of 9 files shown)


rjwalters:
## Judge review: APPROVED

Thorough review of the full ~3,200-line diff, run against the worktree at the PR head (up to date with main, mergeStateStatus CLEAN).

### Verification
- `cargo test -p bth-bridge-core -p bth-bridge-service`: 169 passed (42 core + 127 service), 0 failed
- `cargo clippy --all-targets` for both crates: 0 warnings; `cargo +nightly-2025-12-03 fmt --check`: clean
- CI: workspace build, deny, fmt all green

### Issue acceptance criteria — all four genuinely met

**#827 (DoD):** The chaos suite is the real thing — `Harness::boot()` reconstructs `Database::open()` over the same SQLite file and runs the production `recover_on_startup` path; crash points cover all 6 mint boundaries, both #843 deposit halves, all 6 release boundaries, plus restart-after-every-tick. Assertions are exactly-once at the chain level (mock ETH enforces the #826 order-id guard and requires `double_mint_attempts == 0`; mock BTH deliberately has no guard so a second landed tx IS the double-release), reserve arithmetic is asserted exact after every scenario, and terminal states are re-checked stable across a further restart. The 240-order concurrent load test asserts no lost orders + chain-level exactly-once + zero backlog. Every alert has a fires-on-trigger test (drift trip, backlog trip, global-cap trip, stuck-once, breaker 503, component surface).

**#839:** `update_order_status` reads current status inside a `BEGIN IMMEDIATE` transaction and rejects edges `can_transition_to` disallows; same-status writes are idempotent refreshes. I checked every raw `UPDATE bridge_orders SET status` in the codebase — all live in db.rs and each is a `WHERE status = '<from>'`-guarded edge that matches an allowed transition (`record_deposit_detected`, `mark_mint_confirmed`, `rollback_mint`, `mark_release_confirmed`, `rollback_release`). No writer can bypass the state machine or clobber a terminal state.

**#843:** Both crash halves closed — `recover_on_startup` runs before any watcher/loop, writes the missing `processed_deposits` idempotency row AND rolls `DepositDetected`+`source_tx` forward, audited `deposit_recovered` once. Chaos tests cover both windows.

**#846:** All four items. (1) `unlock_backing_for_order` is value-based with change semantics mirroring `apply_release_spend`; `test_unlock_backing_by_value_after_fifo_consumed_dep_output` reproduces exactly the #844-review scenario. (2) `reserve_balance_checked` persisted (with a proper `ensure_column` migration) and surfaced via `/api/reserve/proof` + `/metrics`. (3) Pruning keeps `MAX(id)`, default 30d, 0 disables. (4) Transports retargeted to #853 (verified open).

### Priority-area findings
- **Rate limits:** `check_and_reserve_limits` is a single IMMEDIATE transaction keyed on order_id — crash/tick replay returns `AlreadyReserved` (never double-counts), `Rejected` reserves nothing (read-only rollback, no leaked reservation). A reserved order that later fails keeps consuming that day's window — conservative fail-closed, fine.
- **Breaker:** submit-halt / confirm-continue verified at the code-path level (`process_pending_orders` gates only stages 1+3 on `is_paused`) and by `test_kill_switch_halts_submits_but_confirms_settle`. State is DB-durable across restarts; reconciler re-trips on every unhealthy pass so a premature resume can't stick.
- **Kill-switch API:** `POST /api/breaker` is unauthenticated but the listener defaults to `127.0.0.1:9741` (empty disables), and both the code docs and runbook state the posture explicitly. Safe as shipped; filed **#855** to warn on non-loopback binds (remote-Prometheus foot-gun) — non-blocking.
- **Monitoring queries:** all aggregates/bounded scans over indexed columns; no N+1.
- **Alert-once:** `stuck_order_alert` audits once per order via `has_audit_for_order`; Prometheus-level alerting still re-arms via the `bridge_stuck_orders` gauge, so no alert-loss on re-stick.
- **Runbook:** accurate against the code, including the one restore hazard (post-backup release claims) and the "never delete release_claims rows" invariant.

### Follow-ups filed
- #855 — warn/refuse non-loopback `api_listen` bind; plus a defense-in-depth nit (terminal same-status refresh can overwrite `dest_tx`).

No blocking defects. Approving.

Issue #865, opened 2026-07-13
epic: Phase 4 — end-to-end testnet bridge + wBTH on testnet DeFi (both chains)


rjwalters:
**Part of #816.** The Phase 0–3 sweep (PRs #832–#864) landed the bridge's cryptographic + safety machinery, but **every chain-touching edge was a fail-safe stub**. This epic drives the bridge to live, demonstrable end-to-end testnet flows, culminating in **wBTH trading on testnet DeFi across three venues: Uniswap (Ethereum), a Solana DEX, and Hyperliquid spot**.

## Goal
Demonstrate, entirely in the testnet realm:
`BTH (testnet) → deposit → wBTH minted on Sepolia/devnet → traded on DeFi (Uniswap / Solana DEX / Hyperliquid spot) → burned → BTH released`.

## Critical path

**Stage A — wire the stubs (implementation, in flight):**
- [x] #858 — federation attestation transport (keystone) — merged #873
- [x] #850 — Solana wBTH Anchor program hardening — merged #871
- [x] #856 (merged #878) — live BTH node transports: deposit scan + reserve release
- [x] #857 (merged #875) — live Solana transports: mint tx assembly/submission + burn watcher
- [x] #853 (merged #880) — live reserve verification legs (Solana getTokenSupply, BTH reserve balance) [after #856+#857]

**Stage B — deploy to testnet (gated on Stage A):**
- [x] #866 — deploy WrappedBTH to Ethereum Sepolia — merged #1027-era (PR #1027; 1 Safe holds all roles, verified)
- [x] #867 — deploy the wbth Anchor program to Solana devnet — done (program CZDnzeywrqEM…)
- [ ] #868 — run a testnet federation end-to-end round trip on both chains

**Stage C — DeFi demonstrations (headline; gated on Stage B):**
- [x] **Uniswap (Ethereum):** #869 — Uniswap v3 wBTH/WETH pool on Sepolia + swap — done (pool 0x16C4fDbe…)
- [x] **Solana:** #870 — Orca wBTH/WSOL pool on devnet + swap — done (pool 9Yog17D3…)
- [ ] **Hyperliquid spot (via Ethereum→HyperEVM):**
  - [x] #876 — move Ethereum wBTH → HyperEVM on testnet (Wormhole NTT) — done (round trip proven, PR #1027)
  - [ ] #877 — Hyperliquid HIP-1 spot listing + HIP-2 liquidity + swap demo [needs #876]

## Architecture note (Hyperliquid)
[...]

Issue #856, opened 2026-07-13
bridge: implement live BTH node transports — deposit scan (NodeBthClient) + release tx construction/submission/confirmation


rjwalters:
Part of #816. Split out of #828 (the test-infrastructure issue), which was the placeholder TODO target for these transports. #828 delivers the test harness and CI; this issue is the concrete implementation home for the BTH-node side.

## Task

1. **`NodeBthClient`** (`bridge/service/src/watchers/bth.rs`): implement against `BthConfig::ws_url` / `rpc_url` — stream (or poll) finalized blocks, view-key-scan outputs to the bridge deposit address, extract amount + memo + cluster factor into `BthBlock`/`BthDeposit`. The watcher state machine (cursor, dedupe, factor-1 gating per ADR 0003) is already implemented and tested against a mock client; only the transport is missing.
2. **`BthReleaser` live stages** (`bridge/service/src/release/bth.rs`): the three `NotImplemented` stubs —
   - construct + CLSAG-sign the release transaction spending reserve-owned factor-1 outputs to a FRESH one-time stealth address (ADR 0004), change back to the reserve address (ADR 0003 provenance);
   - submit via the node's `tx_submit` JSON-RPC;
   - poll block inclusion/depth for `ReleasePending -> Released` (SCP finality by default, `release_confirmations_required` override).

Everything above the transport seam (attestation gating #824, engine retries/unwind, chaos suite) is in place and tested; the stubs are fail-safe (`NotImplemented` keeps orders in their confirmed, retryable states).

## Acceptance
- A testnet BTH deposit drives `AwaitingDeposit -> DepositConfirmed` with the real node (no mock).
- A burn order completes `BurnConfirmed -> ReleasePending -> Released` on testnet, paying a fresh stealth output with change to the reserve.
- The testnet e2e runbook's release-leg steps (docs/bridge/testnet-e2e-runbook.md, from #828) execute end-to-end.

Issue #848, opened 2026-07-13
bridge: Ethereum attestation sets pinned to a stale Safe nonce can wedge an order (liveness, fail-safe)


rjwalters:
Follow-up from the #847 review (federation attestation protocol, #824). No safety exposure — every failure mode below fails closed — but two related liveness gaps should be handled when the #828 transport lands:

1. **Partial set pinned to a stale Safe nonce never completes.** `AttestationSet` for an Ethereum mint is bound to the Safe nonce of its first accepted attestation (`bridge/service/src/attestation.rs`, `AttestationSet::matches` refuses to mix Safe nonces — correct, signatures over different nonces cannot share one `execTransaction`). But the set is only removed from the tracker via `take_if_threshold_met`. If the Safe's on-chain nonce advances while a set is below threshold (e.g. another Safe tx executes), every subsequent attestation at the fresh nonce is refused (`refused:invalid_payload`, nonce mixing) and the set can never reach threshold — the order stays `DepositConfirmed` until a process restart clears the in-memory sets. Consider: evict/rotate a set whose Safe nonce is behind the current on-chain nonce (the collected signatures are unusable anyway).

2. **`MintAuthorization` does not carry the attested Safe nonce.** `EthMinter::prepare_mint` reads the *current* on-chain nonce and builds `execTransaction` with the collected signatures; if the Safe nonce moved between collection and submission, the Safe rejects the signatures (fail-safe) and the engine retries — but the retry needs a full re-collection at the fresh nonce, and after `take_if_threshold_met` the set is already gone, silently discarding threshold signatures. Carrying `safe_nonce` on the authorization would let the minter detect the mismatch pre-broadcast and trigger re-collection deliberately.

Minor config nit that can ride along: duplicate entries in `mint_signers`/`release_signers` inflate `n` for the `threshold <= n` construction check (`check_threshold`). Dedup by signer identity means the effective federation is smaller, so a threshold equal to the padded `n` becomes unsatisfiable — fail-safe, but a construction-time duplicate-signer rejection would surface the misconfiguration immediately.

Refs: #824, #828, PR #847.

---

## Curator Enhancement

*Curating per user request. All code claims below verified against `origin/main`.*

### Problem Statement

[...]

PR #873, opened 2026-07-13
feat(bridge): federation attestation envelope transport between validator bridge nodes (#858)

### LINKED ISSUE #858: bridge: federation attestation envelope transport between validator bridge nodes (#824 follow-up)
Part of #816. Split out of #828 (the test-infrastructure issue), which was the placeholder TODO target. The #824 attestation pipeline is complete and tested — signing, fail-closed verification (signer selection, domain-separated signatures, parse-after-verify, freshness, durable nonce replay protection, order binding, threshold aggregation with distinct-signer dedupe) — but envelopes currently only reach a node from its OWN local signer; peer envelopes are injected directly in tests. A threshold above 1 therefore never authorizes in production (fail-safe by design).

## Task
Implement envelope exchange between federation members:

1. An authenticated endpoint on the bridge service that accepts an `AttestationEnvelope`, routes the peeked order id to the on-record order, and calls `FederationAttestationProvider::submit_attestation` (the designated seam — see its doc comment in `bridge/service/src/attestation.rs`).
2. Outbound: when a node self-attests (mint or release), it pushes/gossips its envelope to the other federation members; on `authorize_*` threshold misses it may request missing envelopes.
3. For Ethereum mints, peers must bind the SAME Safe nonce: the requesting node should communicate the set's nonce (see the stale-nonce liveness issues #848/#849).

Transport security posture: envelopes are self-authenticating (signature over domain-separated bytes + replay nonce), so the transport needs integrity/anti-DoS but not confidentiality; mTLS or the existing node p2p layer are both acceptable.

## Acceptance
- A 2-node bridge federation (threshold 2) authorizes a mint and a release with NO direct test injection — envelopes travel over the wire.
- The `TODO(federation transport)` markers in `bridge/service/src/attestation.rs` are retired.
- Adversarial cases (replayed envelope, wrong-order envelope, unknown signer) are rejected at the endpoint with the existing refuse reasons.

rjwalters:
Closes #858

The KEYSTONE of the Phase-4 bridge milestone (#865): the #824 attestation pipeline was complete and tested, but envelopes only ever reached a node from its OWN local signer (peers were injected directly in tests), so any threshold > 1 never authorized in production. This wires envelope exchange between the t-of-n validator bridge nodes (ADR 0002) over a real, authenticated transport.

Envelopes are self-authenticating — a signature over domain-separated bytes plus a single-use replay nonce, verified fail-closed regardless of how they arrive — so the transport adds integrity/anti-DoS but NOT confidentiality. The endpoint is pure transport in front of the existing verify pipeline; it never weakens verification.

## Implemented

- **Inbound endpoint** (`bridge/service/src/federation.rs`): authenticated `POST /api/attest` that
  1. gates on a shared bearer secret (anti-DoS, before any signature work),
  2. peeks the (unverified) order id purely to route to the on-record order (new `peek_order_id` in `bridge/core/src/attestation.rs`),
  3. hands the RECEIVED envelope + order to `FederationAttestationProvider::submit_attestation` — the full fail-closed pipeline (signer selection → signature → parse-after-verify → freshness → durable nonce reserve → order binding → aggregation).
  The verdict maps to HTTP status (200 accepted / 202 benign / 400 malformed·bad-sig·unknown-signer / 401 bad bearer / 404 unknown order / 409 replay·wrong-order·stale / 503 internal).
- **Outbound push** (`PeerBroadcaster`): when a node self-attests (mint or release) it pushes the accepted envelope to every configured peer's inbound endpoint. Fire-and-forget — a slow/unreachable peer never wedges or fails the local authorize path (the peer self-attests and pushes back; re-authorization re-pushes). Minimal raw-socket HTTP/1.1 client (no new dependency).
- **Ethereum Safe nonce**: the provider stamps its Ethereum self-attestation with the set's `safe_nonce()` and pushes THAT envelope, so peers bind the SAME Safe nonce; a peer whose set already pinned a different nonce refuses (`refused:invalid_payload`) rather than mixing incompatible signatures. Coordinates with the #848/#849 stale-nonce seam without regressing it.
- **Config**: new `[federation]` section (`peers`, `attest_listen`, `inbound_auth_token`, `peer_auth_token`, `peer_push_timeout_secs`). Legacy configs parse with safe defaults (transport disabled → fail-safe).
- **Engine wiring** (`bridge/service/src/engine.rs`): the concrete provider is shared between the `OrderProcessor` and the inbound endpoint; `PeerBroadcaster` is attached when peers are configured; the endpoint is spawned when a listen address is set. Empty listen / no peers = fail-safe (threshold > 1 never authorizes).
- Retired the `TODO(#858)` / `TODO(federation transport)` markers.

## Deferred

- BTH / Solana chain transports for the payload SUBMISSION side remain stubbed elsewhere (#856/#857/#853) — out of scope; this issue is envelope transport between federation nodes, which is chain-agnostic and covers Ethereum + BTH + Solana attestations uniformly.
- mTLS is left to a reverse-proxy / transport layer in front of the plain endpoint for v1 (the doc comments call this out); the bearer fence + self-authenticating envelopes are the v1 posture the issue explicitly permits.
- Pull-based "request missing envelopes on a threshold miss" is not implemented; push + periodic re-authorization (which re-pushes) already converges to threshold. Noted as an optional follow-up.

## Test Plan

- `cargo test -p bth-bridge-core -p bth-bridge-service` — 66 core + 140 service tests pass (0 failed).
- **2-node no-injection DoD** (`federation_transport_tests`): two providers, each with its own DB + real inbound endpoint + real `PeerBroadcaster` to the other; envelopes travel over the wire (bound listeners break the endpoint/provider cycle). `two_node_release_authorizes_over_the_wire_no_injection` and `two_node_eth_mint_authorizes_over_the_wire_no_injection` both reach threshold 2 and produce authorizations that satisfy the downstream release validator / are Safe-owner signatures over the shared-nonce SafeTx digest. No test calls `AttestationSet::insert` across nodes.
- **Adversarial endpoint tests**: replayed envelope (409 `replayed_nonce`, never double-counts), wrong-order (409 `wrong_order`) + unknown-order (404), unknown signer (400 `unknown_signer`, not counted), bad/missing bearer (401 before any signature work).
- `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets` — clean.
- `cargo +nightly-2025-12-03 fmt --check` — clean.
- No new dependencies → `Cargo.lock` / `deny.toml` untouched (`cargo deny` not required).

Scope: `bridge/` only. Does not modify the main worktree.

### DIFF
--- a/bridge/core/src/attestation.rs
+++ b/bridge/core/src/attestation.rs
@@ -17,9 +17,12 @@
 //! envelopes mirroring the operator-signed-action machinery
 //! (`botho/src/operator_action.rs`, P4.4), aggregated to the federation
 //! threshold by [`AttestationSet`]. Signature *collection over the network*
-//! between federation members is TODO(#858); everything cryptographic —
-//! signing, verification, replay rejection, order binding, thresholding —
-//! is implemented here and in `bridge/service/src/attestation.rs`.
+//! between federation members is the #858 transport in
+//! `bridge/service/src/federation.rs` (an authenticated `POST /api/attest`
+//! endpoint in front of this pipeline, plus outbound peer push); everything
+//! cryptographic — signing, verification, replay rejection, order binding,
+//! thresholding — is implemented here and in
+//! `bridge/service/src/attestation.rs`.
 
 use ed25519_dalek::{Signature as Ed25519Signature, SigningKey, VerifyingKey};
 use serde::{Deserialize, Serialize};
@@ -841,6 +844,33 @@ pub fn peek_target_chain(envelope: &str) -> Result<Chain, AttestationRejectReaso
             "action `{other}` is not in the v1 allowlist"
         ))),
     }
+}
+
+/// Peek the bound order UUID out of the (still-unverified) envelope bytes,
+/// ONLY to route the received envelope to the on-record order the ingest
+/// pipeline then binds it against (#858 transport). Like [`peek_signer_key_id`]
+/// this is a hint that drives no security decision on its own: the routed
+/// order is re-checked field-by-field by `check_order_binding` AFTER the
+/// signature verifies, so a lying `orderId` selects an order the signature —
+/// or the order binding — then rejects.
+pub fn peek_order_id(envelope: &str) -> Result<Uuid, AttestationRejectReason> {
+    let value: Value = serde_json::from_str(envelope).map_err(|_| {
+        AttestationRejectReason::Malformed("envelope is not valid JSON".to_string())
+    })?;
+    let order_id = value
+        .as_object()
+        .and_then(|o| o.get("params"))
+        .and_then(|p| p.as_object())
+        .and_then(|p| p.get("orderId"))
+        .and_then(|v| v.as_str())
+        .ok_or_else(|| {
+            AttestationRejectReason::Malformed(
+                "envelope missing string `params.orderId`".to_string(),
+            )
+        })?;
+    Uuid::parse_str(order_id).map_err(|_| {
+        AttestationRejectReason::Malformed("`params.orderId` is not a valid UUID".to_string())
+    })
 }
 
 /// Parse a chain from its canonical wire string ONLY (`ethereum` /
--- a/bridge/core/src/config.rs
+++ b/bridge/core/src/config.rs
@@ -22,6 +22,83 @@ pub struct BridgeConfig {
     /// Reserve reconciliation / proof-of-reserves settings (#825).
     #[serde(default)]
     pub reserve: ReserveSettings,
+
+    /// Federation attestation transport settings (#858): how this node
+    /// exchanges signed attestation envelopes with the other validator
+    /// bridge nodes.
+    #[serde(default)]
+    pub federation: FederationSettings,
+}
+
+/// Federation attestation envelope transport settings (#858).
+///
+/// The #824 attestation pipeline verifies, replay-checks, order-binds, and
+/// thresholds signed envelopes, but each node only ever sees its OWN local
+/// signer's envelope until they are exchanged over the network. This section
+/// configures that exchange between the t-of-n validator bridge nodes (per
+/// ADR 0002 the SCP validator set doubles as the federation).
+///
+/// Envelopes are self-authenticating — every envelope carries a signature
+/// over domain-separated bytes plus a single-use replay nonce, verified
+/// fail-closed by the ingest pipeline regardless of how it arrived. The
+/// transport therefore needs integrity / anti-DoS but NOT confidentiality: a
+/// shared-secret bearer token gates the inbound endpoint (rejecting
+/// unauthenticated floods before any signature work) and authenticates
+/// outbound pushes to peers. The endpoint is pure transport in front of the
+/// existing verify pipeline; it never weakens verification.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FederationSettings {
+    /// Base URLs of the OTHER federation members' bridge services (e.g.
+    /// `https://bridge-2.example:9742`). When a node self-attests it pushes
+    /// its envelope to each peer's `/api/attest` endpoint. Empty disables
+    /// outbound push (single-node development, or threshold 1).
+    #[serde(default)]
+    pub peers: Vec<String>,
+
+    /// Listen address for the inbound attestation endpoint
+    /// (`POST /api/attest`). Empty string disables the inbound endpoint —
+    /// this node then only ever counts its own local signer, so any
+    /// threshold above 1 never authorizes (fail-safe). Distinct from the
+    /// proof-of-reserves API so the two surfaces can bind separately.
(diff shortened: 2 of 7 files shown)


rjwalters:
## Judge review — APPROVED ✅

Adversarial, security-focused review of the #858 federation attestation transport (the keystone that turns tested-in-isolation crypto into a working multi-validator federation). No rebase needed — the branch was already up to date with `main` (0 behind, 1 ahead); #871's Solana changes touched `contracts/solana`/`solana.rs`/`watchers/solana.rs` and do not intersect this PR's `federation.rs`/`engine.rs`/`config.rs`/`attestation.rs`.

### Verify pipeline is NOT bypassed or weakened (the crux)
The endpoint is pure transport. `attest` (federation.rs) calls the SAME `FederationAttestationProvider::submit_attestation` seam that local self-attestation uses; `verify_and_aggregate` in `bridge/service/src/attestation.rs` is unchanged in its fail-closed ordering (signer selection over public data → signature verify over received domain-separated bytes → parse-after-verify → freshness → durable nonce reserve → order binding → threshold dedupe). `verify_and_parse_ed25519` / `verify_and_parse_secp256k1` / `check_attestation_freshness` / `check_order_binding` / the nonce store are all untouched. A replayed envelope → 409 `replayed_nonce` (never double-counts), wrong-order → 409, unknown-order → 404, unknown-signer → 400, tampered bytes → bad_signature — each exercised through the REAL endpoint→provider path in `federation_transport_tests`.

### Per-area findings
- **Auth**: `POST /api/attest` gates on the shared bearer token BEFORE any peek/DB/signature work; missing/wrong → 401 (verified by `endpoint_rejects_bad_bearer`, and `distinct_signers == 0` after). `token_matches` is length-independent (leaks length, not content — acceptable since the token is a DoS fence, not the security boundary; the envelope signature is). An unauthenticated caller can't reach the pipeline when a token is configured; even with a stolen token, acceptance still requires a valid federation signature.
- **peek_order_id**: new in `bridge/core/src/attestation.rs`, mirrors `peek_signer_key_id`/`peek_target_chain` exactly — reads UNVERIFIED bytes purely to route to the on-record order, drives no security decision. A lying id routes to an order that `check_order_binding` then rejects (proven by `endpoint_rejects_wrong_order`: a same-id/different-amount DB record verifies the signature then fails binding → 409).
- **Safe nonce binding**: the provider stamps its self-attestation with the set's `safe_nonce()` and pushes THAT envelope; a peer whose set pinned a different nonce refuses `invalid_payload` rather than mixing signatures (`two_node_eth_mint_*` binds nonce 7 on both nodes; `pipeline_rejects_eth_attestation_at_a_different_safe_nonce` covers the mismatch). #848/#849 seam preserved.
- **Fire-and-forget outbound**: `PeerBroadcaster::broadcast` `tokio::spawn`s each push; `push_one`'s whole attempt (connect+write+read) is wrapped in `tokio::time::timeout`. A slow/unreachable peer can't wedge or fail local authorize — the two-node tests assert `authorize_release` returns `Err` (fail-safe) at 1/2 and only succeeds once the peer's envelope arrives over the wire.
- **Raw-socket HTTP assessment**: the INBOUND endpoint (the network-reachable attack surface) is **axum**, not hand-rolled — same stack as `api.rs`, with axum 0.6's default 2MB `Json` body limit. Only the OUTBOUND client is a minimal raw-socket HTTP/1.1 POST (no new dep), talking to a configured trusted-peer allowlist. This sidesteps the classic hand-rolled-server footguns (smuggling/chunked/header-injection on inbound). Good call.

### No-injection DoD tests are genuine
`two_node_release_*` and `two_node_eth_mint_*`: each node has its OWN provider + DB + bound axum endpoint and a real `PeerBroadcaster` to the other's URL. Envelopes travel real TCP; tests only OBSERVE via read-only `distinct_signers_for_test` + poll. No `AttestationSet::insert` across nodes. Both reach threshold 2 and produce authorizations that pass `validate_release_attestation` / are Safe-owner sigs over the shared-nonce SafeTx digest — proving real federation signatures crossed the wire.

### Verification (worktree, after confirming no rebase)
- `cargo test -p bth-bridge-core -p bth-bridge-service` → 66 core + 140 service pass, 0 failed (incl. both no-injection + all adversarial endpoint tests)
- `cargo clippy ... --all-targets` → clean
- `cargo +nightly-2025-12-03 fmt --check` → clean
- CI: build (9m19s), bridge tests (2m35s), deny, rustfmt all **pass**; mergeStateStatus **CLEAN**

### Follow-up (non-blocking, v1 testnet scope)
- #874: cap the outbound peer-push response read (`read_to_end` is timeout-bounded but unbounded within the window; body is discarded, so low sev). mTLS/pull-based recovery are already noted as deferred in the PR.

Config is fail-safe (empty `attest_listen`/no peers → threshold >1 never authorizes; legacy configs parse with safe defaults). Engine shares the concrete provider between order processor and endpoint correctly. Approving.

PR #875, opened 2026-07-13
bridge: live Solana transports — mint tx assembly/submit/confirm + burn watcher RPC scan (Closes #857, #872)

rjwalters:
Implements the live Solana wBTH transports for the Phase-4 bridge milestone (#865), targeting the just-merged hardened `wbth` Anchor program (#850).

Closes #857
Closes #872

## What landed (implemented vs live-gated)

**Fully implemented + unit-tested (mock RPC):**
- **`solana_rpc` (new module):** a lightweight raw JSON-RPC client (`reqwest`) + a self-contained **legacy Solana transaction/message assembler** — compact-u16 shortvec encoding, the account-ordering + privilege-merge rules, `find_program_address` (off-curve PDA derivation via `ed25519-dalek`), base58 (`bs58`), and a local base64 encode/decode. A `SolanaRpc` trait abstracts the transport for testability. **No `solana-sdk`/`solana-client`** — see below.
- **`SolMinter` (`mint/solana.rs`):** assembles `bridge_mint` with the exact account metas the hardened program declares (bridge PDA, per-order marker PDA `[b"order", order_id]`, mint, recipient ATA, recipient, mint-authority signer, token + system programs); resolves the wBTH mint from the on-chain `Bridge` account; fetches a recent blockhash; signs with the local mint-authority key; `sendTransaction(base64(raw))` with **idempotent** "already processed" re-broadcast. Confirmation polls `getSignatureStatuses` honoring `SolanaConfig::commitment` (default `Finalized`). The pinned instruction-data encoding was **reused unchanged**.
- **Burn watcher (`watchers/solana.rs`):** live `getSignaturesForAddress` + `getTransaction` scan at **Finalized only** (the reorg guard), feeding the existing Anchor event decoder + exactly-once order path; finalized burns confirm immediately (no orphan/canonical-recheck analogue needed).
- **`MintError::NotImplemented` / Solana `WatchError::NotImplemented` paths retired.**

**Live-gated (`#[ignore]`), honest status:**
- `solana_devnet_tests.rs` — a documented `solana-test-validator`/devnet drill (blockhash smoke + live-PDA `prepare_mint`, optional broadcast). This env has **no** `solana`/`anchor`/validator, so the live path is validated only by construction + the mocked suites. **No devnet pass is claimed.**

## Custody model (ADR 0002)
[...]

Issue #868, opened 2026-07-13
bridge(ops): run a testnet federation end-to-end round trip (BTH↔Sepolia↔devnet)


rjwalters:
**Part of #865 (Phase 4). Depends on #858, #856, #857 (transports) + the Sepolia/devnet deploys.**

Stand up a running bridge federation on testnet and prove the full round trip on **both** chains.

## Tasks
- [ ] Deployment topology doc + config: N bridge-service instances (a real threshold federation, not single-signer) wired to BTH testnet, Sepolia, and Solana devnet, each with its validator key.
- [ ] Wire the reserve wallet on BTH testnet (funded from the faucet) and the deployed WrappedBTH (Sepolia) + wbth program (devnet).
- [ ] Execute and script the round trip on **each** chain:
  - BTH deposit → federation attestation (threshold, #858) → wBTH mint (Sepolia via Safe / devnet via multisig) → confirm.
  - wBTH burn → federation release attestation → BTH reserve release (fresh stealth output, ADR 0004) → confirm.
- [ ] Assert exactly-once + factor-1 amounts + proof-of-reserves invariant (`Σ wBTH == locked reserve`) holds live across the run.
- [ ] Extend the existing testnet-e2e runbook (`docs/bridge/testnet-e2e-runbook.md`, from #828) into a reproducible driver.

## Acceptance
- A reproducible testnet run (with tx links) doing deposit→mint→burn→release on BOTH Ethereum and Solana, against a real threshold federation, with the peg invariant verified live.

rjwalters:
Unblocking via /loom:sweep all: declared dependencies #856, #857, #858 (transports) and #866, #867 (Sepolia/devnet deploys) are all CLOSED. Routing to builder for the federation e2e round-trip driver.

rjwalters:
**Builder drill log (2026-07-17).** PR #1053 delivers the scriptable maximum: the federation driver (`scripts/bridge-testnet-federation.sh`), the runbook Phase C, and a LIVE 3-instance / threshold-2 federation run against the live endpoints.

**Proven live:**
[...]

Issue #830, opened 2026-07-13
security: dedicated bridge security audit before any mainnet value (Part of #816)


rjwalters:
**Part of #816 (Phase 3 — the gate). Blocks any real value on the bridge.**

## Task
A dedicated security audit of the **entire** bridge surface — the highest-risk component in the system — before it custodies any real funds.

Scope:
- **Custody / key management** (#817): threshold model, key storage/HSM, rotation, quorum-loss handling.
- **Attestation/authorization protocol** (#821): replay, order-binding, cross-domain reuse, threshold soundness.
- **`WrappedBTH.sol`** (#826): access control, reentrancy, upgradeability, rate-limit + supply accounting — ideally an **external** contract audit.
- **Watchers + engine** (#823/#824/#827): reorg safety, exactly-once, crash-safety.
- **Reserve accounting / peg** (#825/#818): can the invariant be violated or the proof-of-reserves be spoofed?
- **Privacy boundary** (#819): does the bridge leak more than the intended amount?

Follow the internal `AUDIT.md` process, extended with a bridge section; then commission an **external** audit (this is where the external-audit budget, #616, is best spent first). Add the bridge to the internal audit cadence.

## Acceptance
- Internal + external audits with no unresolved Critical/High findings.
- A published threat model + audit report under `audits/` / `docs/security/`.
- Explicit sign-off gate: **no mainnet bridge value until this closes.**

rjwalters:
Reopening — the sign-off gate is not yet satisfied. The internal audit (merged in #864, `audits/2026-07-13-bridge.md`) is only the **first** of the two audits this issue requires.

## Status of the acceptance criteria

- [x] **Internal audit, no unresolved Critical/High** — done (#864). 0 Critical / 0 High across all seven bridge sections; attestation protocol and `WrappedBTH.sol` read in full and re-verified against code.
- [ ] **External audit, no unresolved Critical/High** — **not commissioned.** Operator action, gated on the external-audit budget #616.
- [x] **Published threat model + audit report** — `docs/security/bridge-threat-model.md` (#829) + `audits/2026-07-13-bridge.md` (#864).
- [ ] **Explicit sign-off gate: no mainnet bridge value until this closes** — **gate holds; issue stays OPEN.**

## Remaining before mainnet value

1. **External contract + protocol audit** (first spend of #616) — the primary blocker.
[...]

PR #878, opened 2026-07-13
bridge: live BTH node transports — deposit scan + release tx (#856)

### LINKED ISSUE #856: bridge: implement live BTH node transports — deposit scan (NodeBthClient) + release tx construction/submission/confirmation

rjwalters:
Closes #856

Wires the two remaining BTH-node transports over the seam the tested state machines already drive — the largest transport gap in the Phase-4 bridge milestone (#865). A real BTH deposit can now trigger a mint, and a burn can drive a real reserve release.

## What's live-wired

**Deposit scan** (`NodeBthClient`, `watchers/bth.rs`)
- Polls finalized blocks via `getChainInfo` + `chain_getOutputs`.
- View-key-scans each output for reserve ownership (`TxOutput::belongs_to`), decrypts the destination memo (order UUID), and reads factor-1 eligibility (ADR 0003).
- Only reserve-owned outputs become `BthDeposit`s, so the cursor/dedupe/factor-1-gating state machine (already tested) drives it unchanged.
- Watch-only (no reserve keys configured) stays idle and fail-safe, exactly as the prior stub did.

**Release** (`BthReleaser`, `release/bth.rs`) — three live stages after the #824 attestation gate:
- `prepare_release`: scan the recent reserve window for spendable, factor-1, reserve-owned outputs (spent/pending dropped via `chain_areKeyImagesSpent`), select inputs, gather ring decoys, and build + CLSAG-sign a tx paying a **FRESH one-time stealth output** to the recipient (ADR 0004) with **change back to the reserve** (ADR 0003). Self-verified against the node's verifier before returning.
- `broadcast`: submit the bincode wire bytes via `tx_submit`; idempotent on an "already known" / duplicate-key-image rejection of our own recorded tx.
- `check_confirmation`: poll `getTransaction` for `release_confirmations_required` depth (0 = SCP externalization finality). An unseen tx stays `Pending` for re-broadcast — never a false `Dropped` (BTH has no on-chain order-id guard, so re-signing while the old tx could land risks a double release).

**Fail-safe posture preserved:** every transport/RPC failure maps to a retryable `Rpc` error, never a false success — orders stay in their confirmed retryable states, and a released tx reuses the recorded bytes on resume (exactly-once intact).

## Reused crypto/wallet crates (no reimplementation)

- `bth-transaction-clsag` — `Transaction`/`TxOutput`/`ClsagRingInput`/`RingMember`/`MemoPayload`, `belongs_to`, `recover_spend_key`, `decrypt_memo`, and the exact bincode wire format `tx_submit` deserializes.
- `bth-account-keys` — `AccountKey`/`PublicAddress` (reserve + recipient).
- `bth-crypto-keys` — Ristretto keys.
- `bth-crypto-ring-signature` — `KeyImage` (node-identical spent-status derivation).

New thin modules: `bth_rpc` (JSON-RPC client), `bth_scan` (scan + tx construction), `bth_keys` (reserve key loading).

## Node change (additive, privacy-safe)

Exposes the encrypted memo ciphertext (`eMemo`) in the node's `chain_getOutputs` so thin clients can read the destination memo. Only the recipient's view key can decrypt it, so publishing the ciphertext is identical to what already travels on-chain. Existing node RPC tests (122) + `rpc_integration` (46) still pass.

## Test Plan

**Runs in every `cargo test` (mock / native, honest coverage):**
- `bth_rpc`: little-endian amount decode, `chain_getOutputs` shape decode (coinbase / cluster tags / eMemo).
- `bth_scan`: deposit ownership detection + memo decrypt + factor-1 flag; **release tx construction** — recipient gets a fresh stealth output the reserve cannot detect, change returns to the reserve, two releases to the same address use distinct one-time keys, and the node verifier accepts the signed tx; insufficient-inputs / too-few-decoys rejections.
- `bth_keys`: reserve key load / absent-disabled / malformed rejection.
- `release::bth`: attestation gate fires before construction; a configured-but-unreachable RPC yields a **retryable `Rpc` error, never a false success**.
- Existing watcher/engine/chaos/adversarial suites unchanged and green.

Result: `cargo test -p bth-bridge-core -p bth-bridge-service` → 66 + 155 pass, 3 ignored. `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets` clean. `cargo +nightly-2025-12-03 fmt --check` clean. `cargo deny check advisories` ok.

**Live-node (NOT claimed to pass here — no node available):**
- `bridge/service/src/bth_fork_tests.rs` holds `#[ignore]`d integration tests that drive the REAL transport against a running node. They **self-skip** without `BRIDGE_BTH_RPC_URL` + reserve key env vars (never a false pass). Run instructions mirror the Ethereum `fork_tests.rs` pattern and are documented in `docs/bridge/testnet-e2e-runbook.md` (Layer 1.5). These have not been executed against a live node in this change.


rjwalters:
## Judge review: APPROVED ✅ (testnet-track)

Thorough adversarial review of the two live BTH-node transports plus the consensus-adjacent node RPC change. Rebased onto `origin/main` (over #875 Solana transports, merged concurrently), resolved additively, and drove full CI to green before approving.

### Node RPC `eMemo` change — additive + privacy-safe (highest scrutiny)
`botho/src/rpc/mod.rs` `handle_get_outputs` gains a single new `eMemo` JSON field:
```rust
let e_memo = output.e_memo.as_ref().map(|m| hex::encode(m.as_bytes()));
```
- **Additive.** No existing RPC field changes; no consensus/validation/serialization path touched. It is purely an RPC-response enrichment. Node `rpc::` (122) + `rpc_integration` (46) stay green.
- **Privacy-safe.** `output.e_memo: Option<EncryptedMemo>` is a *pre-existing on-chain field* (`transaction/core/src/tx.rs:280`). `EncryptedMemo(GenericArray<u8, U66>)` is the AES-256-CTR **ciphertext**, keyed by the TxOut shared secret (HKDF-SHA512 over the ECDH secret from the recipient's view key — `transaction/core/src/memo.rs`). There is no plaintext accessor without the shared secret; `as_bytes()` returns the raw 66-byte ciphertext buffer. Publishing it is byte-identical to what already travels on-chain — a non-recipient learns nothing new. Confirmed it is the same committed ciphertext surfaced, not a re-encryption or plaintext.

### CLSAG release assembly — sound (`bth_scan::build_release_tx` + `release/bth.rs`)
No crypto reimplemented — the audited `bth-transaction-clsag` primitives are reused; risk is in the assembly, which checks out:
- **Fresh stealth output per release** (`TxOutput::new(amount, recipient)`, ADR 0004) — two releases to the same address use distinct one-time keys (asserted), and the recipient output is undetectable by the reserve.
- **Change back to the reserve** (`account.default_subaddress()`, ADR 0003); sub-dust change folded into fee (never an unspendable output); checked-arithmetic balance equation (no value burned/lost).
- **No provenance laundering** — recipient output carries no cluster tags; reserve-owned non-factor-1 outputs are excluded from *both* inputs and decoys.
- **Self-verified before broadcast** — `is_valid_structure()` + `verify_ring_signatures()` on the same code path the node runs.
- **No double-spend across releases** — `chain_areKeyImagesSpent` drops spent **and** pending inputs before selection.
- **Attestation-gated first** — no reserve key material is touched before a threshold federation attestation bound to the exact order id/amount/recipient verifies (threshold-0-with-signers refused, #842).

### Exactly-once / fail-safe
- `broadcast` idempotent (reuses recorded bytes; "already/exists/duplicate/key image" node rejection of our own tx = success). Hash mismatch → retryable, never trusts a divergent hash.
- `check_confirmation` holds `Pending` on an unseen tx rather than fabricating `Dropped` (BTH has no on-chain order-id guard — correctly avoids double release).
- Every RPC/transport error → retryable `Rpc`, never a false success (tested against an unreachable node).

### Deposit scan
View-key scan identifies reserve-owned outputs, decrypts the order-UUID memo, applies the factor-1 gate with audit on rejection (ADR 0003); cursor/dedupe state machine unchanged and green (restart-resume + cursor-rewind replay tests pass).

### GPL-3.0 boundary
`bth-transaction-clsag` (GPL-3.0, first-party path dep) is allow-listed in `deny.toml`; `cargo deny check` (advisories/bans/licenses/sources) all ok. Boundary matches existing workspace precedent.

### Rebase + CI fixes committed by the judge
1. `style(rpc)`: rustfmt collapsed the `eMemo` binding under pinned nightly-2025-12-03 (the branch's original fmt failed CI).
2. `style(bth-crypto)`: adding `bth-crypto-ring-signature` + `bth-transaction-types` to the bridge's dep graph pulls them into the bridge CI clippy job (`-D warnings`), surfacing two **pre-existing, previously-latent** lints — `needless_range_loop` in CLSAG sign/verify and `should_implement_trait` on `Network::from_str`. Suppressed with targeted `#[allow]` (audited crypto stays byte-identical; crate tests 47 + 69 pass unchanged). Also resolved a duplicate `reqwest`/`bs58` key in the auto-merged `Cargo.toml`.

Local verification (post-rebase): `bth-bridge-core` 66 + `bth-bridge-service` 181 pass (5 live tests `#[ignore]`-gated); node `rpc::` 122 + `rpc_integration` 46 pass; clippy `-D warnings` clean on bridge crates; pinned-nightly fmt clean; `cargo deny` ok. **CI all green, `mergeStateStatus: CLEAN`.**

### Judgement
Testnet-track — the live BTH paths are honestly `#[ignore]`-gated (self-skip without `BRIDGE_BTH_RPC_URL` + reserve keys) and otherwise mock/native-tested; correct-by-reading. The mainnet gate (**#830**, external bridge security audit) holds. Devnet/testnet-node execution gap is tracked by **#868** (testnet federation end-to-end round trip), which is the validation gate for exercising these transports against a live node. No follow-ups required.

PR #885, opened 2026-07-13
fix(bridge): bound the outbound federation peer-push response read

### LINKED ISSUE #874: Bridge federation: bound the outbound peer-push response read (#858 follow-up)
## Original Issue

Follow-up from PR #873 (#858 federation attestation transport) review.

`PeerBroadcaster::push_one` in `bridge/service/src/federation.rs` reads the peer's HTTP response with `read_to_end` into a `Vec`. The whole attempt (connect + write + read) is bounded by `peer_push_timeout_secs` (default 5s), and only the status line is parsed (the body is discarded), so this is not an attack surface — peers are a configured trusted allowlist and this is an outbound client. But a malicious/misconfigured peer could stream data for the duration of the timeout to grow transient memory.

Low priority hardening: cap the response read (e.g. `take(N)`) since we only need the status line. Non-blocking; the inbound endpoint (the real attack surface) is axum with the default 2MB body limit.

Context: the #858 transport is v1 testnet scope — integrity/anti-DoS, not confidentiality; bearer fence + self-authenticating envelopes are the posture. mTLS is deferred to a reverse proxy per the PR doc comments.

---

### Problem Statement

`PeerBroadcaster::push_one` (`bridge/service/src/federation.rs`, lines ~347-362) writes an HTTP/1.1 request to a federation peer and then reads the entire response into an unbounded `Vec` via `AsyncReadExt::read_to_end`:

```rust
let mut response = Vec::new();
stream
    .read_to_end(&mut response)   // <-- unbounded
    .await
    .map_err(|e| format!("read {}: {e}", peer.authority))?;
let text = String::from_utf8_lossy(&response);
let status: u16 = text
    .split_whitespace()
    .nth(1)                       // only the status line is used
    .and_then(|s| s.parse().ok())
    .unwrap_or(0);
```

Only the status line (`HTTP/1.1 <status> ...`) is parsed; the rest of the response is discarded. The whole attempt is wrapped in `tokio::time::timeout(timeout, attempt)` (`peer_push_timeout_secs`, default 5s), so a slow or malicious peer cannot hang the task indefinitely — but it *can* stream response bytes for the full timeout window, growing transient per-task heap unboundedly. With one spawned task per peer, a misconfigured/hostile allowlist entry could inflate memory during the timeout.

This is **low-priority hardening**, not an active vulnerability: peers are a configured trusted allowlist, this is an outbound client (not the inbound attack surface), and the inbound endpoint is axum with the default 2MB body limit. The fix bounds the transient memory to a fixed cap.

### Implementation Guidance

Cap the read using `AsyncReadExt::take(N)` before `read_to_end`. `take` is already available — `use tokio::io::{AsyncReadExt, AsyncWriteExt};` is imported at line 53. Example:

```rust
const MAX_PEER_RESPONSE_BYTES: u64 = 8 * 1024; // 8 KiB — status line + headers only
// ...
let mut response = Vec::new();
(&mut stream)
    .take(MAX_PEER_RESPONSE_BYTES)
    .read_to_end(&mut response)
    .await
    .map_err(|e| format!("read {}: {e}", peer.authority))?;
```

Notes:
- Use `(&mut stream).take(...)` (or `Read::take` on a borrowed reader) so the `TcpStream` is not consumed if any later use is added; the `Connection: close` request header means we only read once and drop the stream anyway.
- **Suggested cap: 8 KiB.** We only need the first line (`HTTP/1.1 200 OK`). 8 KiB comfortably covers the status line plus any response headers a well-behaved peer might send before the body, while capping worst-case transient memory at 8 KiB/peer. A smaller value (e.g. 1 KiB) also works since only the status line is parsed; 8 KiB is a safe margin. Do not go so small that a legitimate multi-header response truncates before the status line is readable (the status line is always first, so even 512 B is functionally sufficient — the cap is purely a memory bound, not a correctness constraint).
- Truncation is harmless: parsing uses `split_whitespace().nth(1)`, which only needs the status line at the very start of the stream. A capped read still contains it.
- Keep the `String::from_utf8_lossy` + status-parse logic unchanged.

### Acceptance Criteria

- [ ] `push_one` reads at most a fixed, bounded number of bytes from the peer response (no unbounded `read_to_end` directly on the raw stream).
- [ ] The cap is a named constant (e.g. `MAX_PEER_RESPONSE_BYTES`) with a comment explaining only the status line is needed.
- [ ] Status-line parsing behavior is unchanged: a normal peer returning `HTTP/1.1 200 OK` still yields `status == 200`.
- [ ] A peer that streams more than the cap does not cause `push_one` to allocate beyond the cap, and `push_one` still returns/logs a status (or the timeout/error path) without panicking.
- [ ] Existing federation tests continue to pass; the test-only `read_to_end` at line ~808 (test harness, not production path) is out of scope.

### Affected Files

- `bridge/service/src/federation.rs` - In `push_one` (~lines 347-362), wrap the read in `AsyncReadExt::take(MAX_PEER_RESPONSE_BYTES)`; add the `MAX_PEER_RESPONSE_BYTES` constant. Add/extend a unit test asserting the read is bounded and status parsing is unaffected.

### Test Plan

- [ ] Manual verification: run the bridge service against a local federation peer; confirm a normal push still logs `pushed attestation to <peer> -> HTTP 200` (behavior unchanged).
- [ ] Automated tests: add a unit test in `bridge/service/src/federation.rs` (`#[cfg(test)]`) that spins up a local TCP listener which, after receiving the request, writes a valid status line followed by more than `MAX_PEER_RESPONSE_BYTES` of body bytes; assert `push_one` completes and the parsed status is correct while the buffered response length is `<= MAX_PEER_RESPONSE_BYTES`.
- [ ] Edge cases: peer that closes immediately after the status line (short read < cap) still parses correctly; peer that sends exactly `MAX_PEER_RESPONSE_BYTES`; peer that sends nothing then the timeout fires (existing timeout path still logs).
- [ ] `cargo test -p botho-bridge-service` (or the service's crate name) passes.

rjwalters:
## Summary

`PeerBroadcaster::push_one` in `bridge/service/src/federation.rs` read the peer's HTTP response with an unbounded `read_to_end` into a `Vec`. The whole attempt is bounded by `peer_push_timeout_secs` (default 5s) and only the status line is parsed (the body is discarded), so a slow or malicious peer could stream response bytes for the full timeout window to grow transient per-task heap unboundedly.

This caps the read at a fixed `MAX_PEER_RESPONSE_BYTES` (8 KiB) using `AsyncReadExt::take` before `read_to_end`. Low-priority hardening — non-behavioral for well-behaved peers.

## Changes

- Added `MAX_PEER_RESPONSE_BYTES: u64 = 8 * 1024` constant with a comment explaining only the status line is needed.
- Extracted the bounded read + status-line parse into a `read_status_line` helper that wraps the reader in `.take(MAX_PEER_RESPONSE_BYTES)` before `read_to_end`, keeping the existing `String::from_utf8_lossy` + `split_whitespace().nth(1)` parse unchanged.
- Added four `#[cfg(test)]` unit tests: normal short response parses 200; a flood 4x the cap keeps the buffer capped and still parses 202; a status-line-only short read parses 500; an empty response defaults to 0.

| Bounded read (no unbounded `read_to_end` on the raw stream) | ✅ | `read_status_line` wraps the reader in `.take(MAX_PEER_RESPONSE_BYTES)` |
| Cap is a named constant with an explanatory comment | ✅ | `MAX_PEER_RESPONSE_BYTES` documents that only the status line is needed |
| Status-line parsing unchanged (`HTTP/1.1 200 OK` -> 200) | ✅ | `read_status_line_parses_normal_short_response` asserts 200 |
| Over-cap stream does not allocate beyond the cap and still yields a status | ✅ | `read_status_line_is_bounded_on_flood` asserts buffer len == cap and status == 202 |
| Existing federation tests still pass; test-harness `read_to_end` out of scope | ✅ | `cargo test -p bth-bridge-service` — 191 pre-existing + 4 new pass; test-only `read_to_end` untouched |

## Test Plan

- `cargo check -p bth-bridge-service` — clean.
- `cargo clippy -p bth-bridge-service --all-targets` — no warnings.
- `cargo test -p bth-bridge-service` — 195 passed, 0 failed (191 prior + 4 new).

Closes #874

### DIFF
--- a/bridge/service/src/federation.rs
+++ b/bridge/service/src/federation.rs
@@ -278,6 +278,43 @@ pub struct PeerBroadcaster {
     timeout: std::time::Duration,
 }
 
+/// Maximum number of bytes read from a peer's push response.
+///
+/// `push_one` only parses the HTTP status line (`HTTP/1.1 <status> ...`) and
+/// discards the body, so the read is capped here rather than using an unbounded
+/// `read_to_end`. 8 KiB comfortably covers the status line plus any response
+/// headers a well-behaved peer sends, while bounding the worst-case transient
+/// heap a slow/malicious peer could grow during the push timeout window.
+const MAX_PEER_RESPONSE_BYTES: u64 = 8 * 1024;
+
+/// Read a peer's push response and parse the HTTP status code from the status
+/// line (`HTTP/1.1 <status> ...`).
+///
+/// Only the status line is needed, so the read is capped at
+/// [`MAX_PEER_RESPONSE_BYTES`] via [`AsyncReadExt::take`] rather than an
+/// unbounded `read_to_end`. This bounds the transient per-task heap a
+/// slow/malicious peer could grow by streaming body bytes for the whole push
+/// timeout window. Truncation is harmless: the status line is always first, so
+/// the capped buffer still contains everything the parse needs. A response with
+/// no parseable status yields `0`.
+async fn read_status_line<R>(reader: R) -> std::io::Result<u16>
+where
+    R: tokio::io::AsyncRead + Unpin,
+{
+    let mut response = Vec::new();
+    reader
+        .take(MAX_PEER_RESPONSE_BYTES)
+        .read_to_end(&mut response)
+        .await?;
+    let text = String::from_utf8_lossy(&response);
+    let status: u16 = text
+        .split_whitespace()
+        .nth(1)
+        .and_then(|s| s.parse().ok())
+        .unwrap_or(0);
+    Ok(status)
+}
+
 /// A parsed peer base URL split into the pieces the raw-socket client needs.
 #[derive(Clone)]
 struct PeerEndpoint {
@@ -348,17 +385,9 @@ impl PeerBroadcaster {
                 .write_all(request.as_bytes())
                 .await
                 .map_err(|e| format!("write {}: {e}", peer.authority))?;
-            let mut response = Vec::new();
-            stream
-                .read_to_end(&mut response)
+            let status = read_status_line(&mut stream)
                 .await
                 .map_err(|e| format!("read {}: {e}", peer.authority))?;
-            let text = String::from_utf8_lossy(&response);
-            let status: u16 = text
-                .split_whitespace()
-                .nth(1)
-                .and_then(|s| s.parse().ok())
-                .unwrap_or(0);
             Ok::<u16, String>(status)
         };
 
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Sound, low-risk outbound hardening. The refactor into `read_status_line` is faithful: the original unbounded `read_to_end` + `from_utf8_lossy` + `split_whitespace().nth(1)` + `unwrap_or(0)` parse is moved verbatim, with the only functional change being `.take(MAX_PEER_RESPONSE_BYTES)` (8 KiB) wrapping the reader before `read_to_end`.

Verification:
- **Timeout preserved** — `read_status_line(&mut stream)` is called inside the `attempt` closure, which `tokio::time::timeout` still wraps in full (`federation.rs:365`). Body still discarded.
- **No false truncation** — the status line is always first and a status line + headers is far under 8 KiB, so the capped buffer always contains the bytes `nth(1)` needs. Well-behaved peers unaffected.
- **Flood test is meaningful** — `read_status_line_is_bounded_on_flood` sends 4x the cap, asserts `buf.len() == MAX_PEER_RESPONSE_BYTES` exactly, and asserts the helper still parses status 202 from the capped buffer. That is precisely the property the hardening claims.
- `AsyncReadExt` (providing `.take()`/`.read_to_end()`) already imported at line 53.

## Test Execution

**Test plan from PR description:**
1. `cargo check -p bth-bridge-service` — ✅ Verified via CI (`cargo build --workspace` passes)
2. `cargo clippy -p bth-bridge-service --all-targets` — ✅ CI clean (rustfmt + build pass)
3. `cargo test -p bth-bridge-service` (195 passed) — ✅ Verified: `cargo test (bridge crates)` CI check passes

**CI**: All 4 checks pass (build, cargo deny, bridge tests, rustfmt). Merge state CLEAN. `Closes #874` present.

Non-blocking note: the flood test asserts the cap boundary via a separate inline `take().read_to_end()` rather than through the helper (the helper only returns the parsed status, not the buffer), so the boundary assertion re-implements the read path. Acceptable — the same test's second half exercises the real helper for the parse.

PR #888, opened 2026-07-13
fix(bridge): recover Ethereum mints wedged behind a stale Safe nonce

### LINKED ISSUE #848: bridge: Ethereum attestation sets pinned to a stale Safe nonce can wedge an order (liveness, fail-safe)

### LINKED ISSUE #849: bridge: evict stale Safe-nonce attestation sets (liveness, follow-up to #824)
Follow-up from the #847 review (federation attestation protocol, #824).

`FederationAttestationProvider` keys Ethereum mint `AttestationSet`s to the Safe nonce the first accepted attestation was bound to (`bridge/service/src/attestation.rs`, `AttestationSet::matches`). Refusing to mix Safe nonces inside one set is correct — signatures over different SafeTx nonces cannot be combined into one `execTransaction`.

However, sets are only ever removed by `take_if_threshold_met`. If the Safe's on-chain nonce advances past a partially-filled set's pinned nonce (e.g. an unrelated Safe transaction executes while the federation is below threshold), then:

- attestations at the *current* on-chain nonce are refused (`refused:invalid_payload` mixing refusal), and
- the pinned set can never reach threshold at a nonce the Safe will accept,

so the `(order, bridge.mint_wbth)` set is wedged until a process restart (sets are in-memory). The order stays `DepositConfirmed` and retryable — fail-safe, no custody exposure — but it is a permanent local liveness stall.

Suggested fix, in or alongside the #828 transport work (where multi-signer sets first become reachable):

- when `authorize_mint` finds an existing set whose `safe_nonce()` is behind the current on-chain `Safe.nonce()`, drop the set and re-collect at the fresh nonce (the durable nonce store already forces peers to re-sign with fresh envelope nonces, so this is safe);
- optionally add set expiry tied to `MAX_ATTESTATION_LIFETIME_SECS` so abandoned sets are GC'd.

Not a security defect: today the only signer is local self-attestation, which either meets threshold immediately or the deployment is already waiting on #828.

rjwalters:
## Summary

The #824 federation aggregator binds each Ethereum mint `AttestationSet` to the Safe nonce of its first attestation (correct — signatures over different SafeTx nonces cannot share one `execTransaction`). But sets were only removed via `take_if_threshold_met`, so once the Safe's on-chain nonce advanced past a below-threshold set's pinned nonce (an unrelated Safe tx executed), every fresh-nonce attestation was refused and the order wedged at `DepositConfirmed` until a process restart. A second window existed after threshold: `take_if_threshold_met` removed the set and `prepare_mint` re-read the live nonce with no cross-check, so a nonce move between collection and broadcast silently discarded threshold signatures.

This is a **liveness** fix. Every failure mode here was already fail-safe (fails closed); no nonce-mixing or threshold check is weakened — this adds recovery, a pre-broadcast cross-check, and a construction-time guard.

## Changes

- **Stale-nonce eviction (`bridge/service/src/attestation.rs`, `authorize_mint` Ethereum arm):** reads the live `Safe.nonce()` outside the tracker lock, and drops an `AttestationSet` pinned **strictly behind** it (re-checking under the lock before removal), letting it re-collect at the fresh nonce. An equal nonce (healthy) and a behind-read (transient stale RPC) never evict.
- **`MintAuthorization.safe_nonce: Option<u64>` (`bridge/core/src/attestation.rs`):** carries the Safe nonce the collected signatures are bound to (`None` for Ed25519 / Solana). `#[serde(default)]` keeps authorizations persisted before this field deserializable.
- **Pre-broadcast cross-check (`bridge/service/src/mint/ethereum.rs`):** `prepare_mint` compares `auth.safe_nonce` to the live `Safe.nonce()` via a pure `check_attested_nonce` helper and returns a new retryable `MintError::StaleNonce` **before** persisting or broadcasting any tx, so the engine re-authorizes and re-collects instead of broadcasting doomed signatures.
- **Duplicate-signer rejection at construction (`from_config`):** dedups eth owners by parsed `Address` (checksum-insensitive) and ed25519 signers by verifying-key bytes, so a `threshold == padded_n` misconfig fails at startup instead of wedging at runtime.

| Stale set evicted + re-collected at fresh nonce, recovers without restart (closes #849) | ✅ | `authorize_mint_eth_evicts_a_set_pinned_behind_the_onchain_nonce` |
| `MintAuthorization` carries `safe_nonce` (`None` for Ed25519) | ✅ | assertions in `authorize_mint_eth_collects...` (`Some(7)`) and `authorize_mint_solana_carries_no_safe_nonce` (`None`) |
| `prepare_mint` detects mismatch pre-broadcast, returns a distinct error | ✅ | new `MintError::StaleNonce`; `test_check_attested_nonce_detects_mismatch_pre_broadcast` |
| Duplicate signer identities rejected at construction with a clear message | ✅ | `from_config_rejects_duplicate_ed25519_signers`, `from_config_rejects_duplicate_eth_owners_checksum_insensitive` |
| Existing attestation tests still pass | ✅ | `pipeline_rejects_eth_attestation_at_a_different_safe_nonce` + `authorize_mint_eth_collects_safe_owner_signatures_to_threshold` green |
| Edge cases: behind-read / equal nonce no-op; Solana unaffected | ✅ | `authorize_mint_eth_does_not_evict_on_equal_or_stale_read`, `authorize_mint_solana_carries_no_safe_nonce` |
| PR body includes both closes | ✅ | below |

## Test Plan

- `cargo test -p bth-bridge-service -p bth-bridge-core` — green (66 core + 197 service tests, 0 failures).
- `cargo clippy -p bth-bridge-service -p bth-bridge-core --all-targets` — clean.
- `cargo fmt` applied.

Closes #848
Closes #849

### DIFF
--- a/bridge/core/src/attestation.rs
+++ b/bridge/core/src/attestation.rs
@@ -89,6 +89,21 @@ pub struct MintAuthorization {
     /// Collected validator signatures. Must contain at least `threshold`
     /// entries from distinct signers to be usable.
     pub signatures: Vec<AttestationSignature>,
+
+    /// The Gnosis Safe nonce every collected Ethereum payload signature is
+    /// bound to (`None` for Ed25519 / Solana mints, which are blockhash-bound
+    /// rather than nonce-bound).
+    ///
+    /// The Safe owner signatures were produced over a SafeTx digest that
+    /// includes this nonce; `Safe.execTransaction` only accepts them while it
+    /// is the Safe's *current* on-chain nonce. Carrying it here lets the
+    /// minter cross-check the live nonce **before broadcast** and trigger a
+    /// deliberate re-collection at the fresh nonce instead of broadcasting
+    /// signatures the Safe will reject (see #848). `#[serde(default)]` keeps
+    /// authorizations persisted before this field was added deserializable
+    /// (they decode as `None`).
+    #[serde(default)]
+    pub safe_nonce: Option<u64>,
 }
 
 impl MintAuthorization {
@@ -1388,6 +1403,7 @@ mod tests {
             scheme: SignatureScheme::Secp256k1,
             threshold: 2,
             signatures: vec![sig(1)],
+            safe_nonce: Some(7),
         };
         assert!(!auth.meets_threshold());
 
@@ -1406,10 +1422,30 @@ mod tests {
             scheme: SignatureScheme::Ed25519,
             threshold: 3,
             signatures: vec![sig(1), sig(2)],
+            safe_nonce: None,
         };
         let json = serde_json::to_string(&auth).unwrap();
         let back: MintAuthorization = serde_json::from_str(&json).unwrap();
         assert_eq!(auth, back);
+
+        // Ethereum mint authorizations carry the bound Safe nonce.
+        let eth_auth = MintAuthorization {
+            order_id: [3u8; 32],
+            scheme: SignatureScheme::Secp256k1,
+            threshold: 2,
+            signatures: vec![sig(1), sig(2)],
+            safe_nonce: Some(42),
+        };
+        let json = serde_json::to_string(&eth_auth).unwrap();
+        let back: MintAuthorization = serde_json::from_str(&json).unwrap();
+        assert_eq!(eth_auth, back);
+        assert_eq!(back.safe_nonce, Some(42));
+
+        // Authorizations persisted before `safe_nonce` existed still
+        // deserialize (the field defaults to `None`).
+        let legacy = r#"{"order_id":"0909090909090909090909090909090909090909090909090909090909090909","scheme":"ed25519","threshold":3,"signatures":[]}"#;
+        let back: MintAuthorization = serde_json::from_str(legacy).unwrap();
+        assert_eq!(back.safe_nonce, None);
     }
 
     #[test]
--- a/bridge/service/src/attestation.rs
+++ b/bridge/service/src/attestation.rs
@@ -133,6 +133,7 @@ impl AttestationProvider for StubAttestationProvider {
             scheme,
             threshold: 0,
             signatures: vec![],
+            safe_nonce: None,
         })
     }
 
@@ -314,6 +315,32 @@ fn check_threshold(threshold: u32, n: usize, what: &str) -> Result<(), String> {
     Ok(())
 }
 
+/// Reject a federation configured with duplicate signer identities.
+///
+/// Duplicate owner addresses / signer pubkeys inflate the raw list length
+/// `n` used by [`check_threshold`], so a `threshold == padded_n` config would
+/// pass construction yet be unsatisfiable at runtime: the aggregator dedups
+/// by signer identity ([`AttestationSet::insert`]), so the *effective*
+/// federation is smaller than `n` and the threshold can never be reached —
+/// the order wedges forever, fail-safe but silent. Rejecting the misconfig at
+/// construction surfaces it immediately (#848).
+///
+/// `identities` must already be normalized (parsed/trimmed) so that surface
+/// variants like `"0xAbc"` and `"0xabc"` collide on the same canonical bytes.
+fn reject_duplicate_signers<T: Ord + Clone>(identities: &[T], what: &str) -> Result<(), String> {
+    let mut seen = identities.to_vec();
+    seen.sort();
+    if seen.windows(2).any(|w| w[0] == w[1]) {
+        return Err(format!(
+            "{}: duplicate signer identities configured — every federation \
+             member must be distinct (a repeated signer inflates the \
+             configured count and makes a threshold equal to it unsatisfiable)",
+            what
+        ));
+    }
+    Ok(())
+}
+
 /// The signer identity string for an Ethereum owner address: lowercase hex,
 /// no 0x prefix (40 chars).
 fn eth_signer_key_id(address: &Address) -> String {
@@ -333,11 +360,6 @@ impl FederationAttestationProvider {
         let eth = if eth_cfg.mint_signers.is_empty() {
             None
         } else {
-            check_threshold(
-                eth_cfg.mint_threshold,
-                eth_cfg.mint_signers.len(),
-                "ethereum.mint_threshold",
-            )?;
             let mut owners = Vec::with_capacity(eth_cfg.mint_signers.len());
             for s in &eth_cfg.mint_signers {
                 owners.push(
@@ -346,6 +368,14 @@ impl FederationAttestationProvider {
                         .map_err(|e| format!("bad ethereum mint signer {}: {}", s, e))?,
                 );
             }
+            // Dedup on the parsed 20-byte address (not the raw string), so
+            // checksummed / lowercase spellings of one owner collide.
+            reject_duplicate_signers(&owners, "ethereum.mint_signers")?;
+            check_threshold(
+                eth_cfg.mint_threshold,
+                owners.len(),
+                "ethereum.mint_threshold",
+            )?;
             let safe: Address = eth_cfg
                 .safe_address
                 .as_deref()
@@ -373,13 +403,18 @@ impl FederationAttestationProvider {
         let sol = if config.solana.mint_signers.is_empty() {
             None
         } else {
+            let signers = parse_ed25519_federation(&config.solana.mint_signers, "solana mint")?;
+            reject_duplicate_signers(
+                &signers.iter().map(|k| k.to_bytes()).collect::<Vec<_>>(),
+                "solana.mint_signers",
+            )?;
             check_threshold(
                 config.solana.mint_threshold,
-                config.solana.mint_signers.len(),
+                signers.len(),
                 "solana.mint_threshold",
             )?;
             Some(Ed25519Federation {
-                signers: parse_ed25519_federation(&config.solana.mint_signers, "solana mint")?,
+                signers,
                 threshold: config.solana.mint_threshold,
             })
         };
@@ -387,13 +422,18 @@ impl FederationAttestationProvider {
         let bth = if config.bth.release_signers.is_empty() {
             None
         } else {
+            let signers = parse_ed25519_federation(&config.bth.release_signers, "bth release")?;
+            reject_duplicate_signers(
+                &signers.iter().map(|k| k.to_bytes()).collect::<Vec<_>>(),
+                "bth.release_signers",
+            )?;
             check_threshold(
                 config.bth.release_threshold,
-                config.bth.release_signers.len(),
+                signers.len(),
                 "bth.release_threshold",
             )?;
             Some(Ed25519Federation {
-                signers: parse_ed25519_federation(&config.bth.release_signers, "bth release")?,
+                signers,
                 threshold: config.bth.release_threshold,
             })
         };
@@ -992,6 +1032,49 @@ impl AttestationProvider for FederationAttestationProvider {
                         .to_string()
                 })?;
 
+                // Stale-nonce eviction (#848 / #849). A partial set is pinned
+                // to the Safe nonce of its first attestation. If an unrelated
+                // Safe transaction executed while the set was below threshold,
+                // the Safe's on-chain nonce advances past the pinned nonce and
+                // every fresh-nonce attestation is (correctly) refused as a
+                // nonce mismatch — the set can never reach a usable threshold
+                // and the order wedges at DepositConfirmed until a restart.
+                //
+                // Read the live nonce OUTSIDE the tracker lock (RPC), then
+                // re-lock to evict only when the pinned nonce is STRICTLY
+                // behind. An equal nonce is the healthy case; a set pinned
+                // AHEAD of the read (a transient stale on-chain read) is left
+                // untouched — we never evict on a nonce that only appears to
+                // have moved. Dropping the set is safe: the collected
+                // signatures bind the stale nonce and are unusable anyway, and
+                // the durable nonce store forces each peer to re-sign with a
+                // fresh envelope nonce on re-collection (no replay risk).
+                let pinned_nonce = {
+                    let tracker = self.tracker.lock().expect("tracker lock poisoned");
+                    tracker
+                        .sets
+                        .get(&(order.id, "bridge.mint_wbth"))
+                        .and_then(|s| s.safe_nonce())
+                };
+                if let Some(pinned) = pinned_nonce {
+                    let live = fed.nonce_source.safe_nonce().await?;
+                    if pinned < live {
+                        let mut tracker = self.tracker.lock().expect("tracker lock poisoned");
+                        // Re-check under the lock: only evict if the set is
+                        // still pinned to the same stale nonce (it may have
+                        // been re-collected or taken between the RPC read and
+                        // re-acquiring the lock).
+                        if tracker
+                            .sets
+                            .get(&(order.id, "bridge.mint_wbth"))
+                            .and_then(|s| s.safe_nonce())
+                            == Some(pinned)
+                        {
+                            tracker.sets.remove(&(order.id, "bridge.mint_wbth"));
+                        }
+                    }
+                }
+
                 // Self-attest with the local Safe owner key, binding to the
                 // set's Safe nonce (or the current on-chain nonce for a
                 // fresh set) so all collected signatures share one SafeTx.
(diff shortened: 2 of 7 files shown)


rjwalters:
✅ **Approved!**

Reviewed with extra rigor as security- and consensus-sensitive federation code. This is a liveness fix that adds recovery + a pre-broadcast cross-check + a construction-time guard without weakening any safety invariant. Every critical focus area verified:

**1. Nonce-mixing refusal UNCHANGED (the load-bearing invariant).** `AttestationSet::matches` (`bridge/core/src/attestation.rs:1287-1304`) is not in the diff — signatures bound to different Safe nonces still cannot be combined into one `execTransaction`. `insert` still calls `matches` first. Regression test `pipeline_rejects_eth_attestation_at_a_different_safe_nonce` passes.

**2. Eviction is strictly-behind + TOCTOU-safe.** `bridge/service/src/attestation.rs:1055-1073`: live `Safe.nonce()` read outside the tracker lock, `if pinned < live` (strict — equal never evicts), then re-lock and re-check `safe_nonce() == Some(pinned)` before `remove`. A behind-read (transient stale RPC) and an equal nonce are both no-ops (`authorize_mint_eth_does_not_evict_on_equal_or_stale_read`). The under-lock re-check closes the read→re-lock window; worst case in any race is a spurious re-collection (liveness), never a safety break, because `matches` still guards every insert.

**3. `safe_nonce: Option<u64>` backward-compat + None safety.** `#[serde(default)]`; legacy JSON without the field deserializes to `None` (asserted). `check_attested_nonce` (`bridge/service/src/mint/ethereum.rs:527`) only cross-checks when `Some` — `None` (Solana blockhash-bound / legacy) returns `Ok(())` and bypasses nothing harmful. Fresh eth mints now always populate `safe_nonce` via eviction + re-collection, so the silent-discard window is closed rather than reopened.

**4. `MintError::StaleNonce` retryable + fires pre-persist/pre-broadcast.** `check_attested_nonce` runs inside `prepare_mint` (ethereum.rs:392) before any tx assembly; `prepare_mint` (engine.rs:754) runs strictly before `record_mint_submitted` (the exactly-once persist, line 760) and before broadcast (line 790). The `Err` propagates through `submit_mint` and is only logged (engine.rs:496-498) — the order stays `deposit_confirmed` and re-authorizes next tick. No partial state, no double-spend.

**5. Duplicate-signer dedup correct + non-over-rejecting.** Eth dedups by parsed `Address` (checksum-insensitive), ed25519 by `to_bytes()`, in all three `from_config` arms, before `check_threshold`. Distinct signers still construct (asserted). `from_config_rejects_duplicate_eth_owners_checksum_insensitive` + `from_config_rejects_duplicate_ed25519_signers` pass.

**6. `threshold <= n` preserved.** `check_threshold` still errors on `threshold > n` and `threshold == 0`, now fed the deduped/parsed count.

## Test Execution
- `cargo test -p bth-bridge-service -p bth-bridge-core` — **263 green** (66 core + 197 service, 0 failures), all named regression tests present and passing.
- `cargo clippy -p bth-bridge-service -p bth-bridge-core --all-targets` — clean.
- CI: all 5 checks SUCCESS (cargo test, cargo build --workspace Linux, cargo deny, rustfmt, hardhat). Merge state CLEAN.

PR body uses `Closes #848` and `Closes #849`. LGTM.

PR #891, opened 2026-07-13
feat(bridge): detect + audit federation attestation equivocation

### LINKED ISSUE #859: security(bridge): active equivocation detection + alerting for federation attestations
Follow-up from #829 (bridge adversarial suite, Phase 3 of #816).

## Context
The adversarial tests confirm equivocation is **neutralized** but not **detected**:

- `check_order_binding` pins every attested field (order id, amount, recipient, chains, source tx) to the on-record order, so a signer cannot produce two binding-valid conflicting attestations for one order.
- `AttestationSet::insert` deduplicates by signer identity, so a single equivocating signer can never inflate the `t`-of-`n` threshold or move conflicting funds.

See `bridge/core/src/adversarial_tests.rs`:
- `equivocating_signer_cannot_inflate_the_distinct_signer_count`
- `conflicting_payload_for_the_same_order_cannot_cross_order_binding`

and the threat model: `docs/security/bridge-threat-model.md` (threat #7).

## Gap
The service does **not** raise an auditable alarm when a federation member submits *conflicting* bytes for the same order id — it silently keeps the first valid submission and rejects later ones via the normal replay / binding / dedup paths. A Byzantine (or compromised-key) validator equivocating is a governance-relevant signal that today produces no distinct audit event.

## Proposed work
- In the `bridge/service` attestation ingest pipeline (`verify_and_aggregate`), when a *verified* envelope from a known signer conflicts with what that signer already attested for the same `(order, action)` (different payload digest / different SafeTx nonce for the same logical order), emit a dedicated `attestation_equivocation` audit event (with signer key id + order id), distinct from `replayed_nonce` / `wrong_order` / `invalid_payload`.
- Consider tripping the circuit breaker (or at least a high-severity alert) on equivocation, since it implies a misbehaving or compromised federation key.
- Add a service-level test asserting the equivocation audit fires exactly once per equivocating signer.

## Acceptance
- A verified signer presenting conflicting attestations for one order produces an `attestation_equivocation` audit event; the funds-safety behavior is unchanged (still neutralized).

---

## Curator Enhancement (2026-07-13)

Verified against `origin/main` (post-#888, which reworked stale-nonce eviction and added `safe_nonce` on `MintAuthorization`). Line numbers below are from `origin/main`.

### Root cause: where the equivocation goes silent today

The pipeline is `FederationAttestationProvider::verify_and_aggregate` in `bridge/service/src/attestation.rs` (fn at **line 641**). Its Step 7 aggregation (**lines ~738-777**) calls `AttestationSet::insert`. The set lives in `bridge/core/src/attestation.rs` (**struct at line 1258, `insert` at line 1310**):

```rust
pub fn insert(&mut self, parsed: &ParsedAttestation, signature: AttestationSignature) -> Result<bool, String> {
    self.matches(parsed)?;                                    // order/action/chain/safe_nonce guard
    if self.entries.iter().any(|(id, _)| *id == parsed.signer_key_id) {
        return Ok(false);                                     // <-- SILENT: already-counted signer
    }
    self.entries.push((parsed.signer_key_id.clone(), signature));
    Ok(true)
}
```

When a signer that already counted submits again, `insert` returns `Ok(false)` **without comparing the stored signature/digest to the new one**. In `verify_and_aggregate`, `Ok(false)` (a "new signer" false) flows into `AttestationOutcome::accept(...)` (**line 760-763**) — so an equivocating resubmission currently reports as an accepted, benign duplicate. There is no signal distinguishing "same signer re-sent identical bytes" (benign retry) from "same signer sent CONFLICTING bytes" (equivocation).

Two distinct conflict shapes to handle:
1. **Same set, conflicting payload signature** — the signer already counted, and the newly-presented `AttestationSignature` (or the payload digest it covers) differs from the stored one for that signer. This is the genuinely-silent case: it reaches `insert`, returns `Ok(false)`, and is reported as a benign duplicate today.
2. **Different SafeTx / Safe nonce (Ethereum mints)** — `AttestationSet::matches` (**line 1288**) already rejects a different `safe_nonce` with an `Err(...)`, which `verify_and_aggregate` maps to `AttestationRejectReason::InvalidPayload` (**line 769**). This conflict is *caught* but is currently indistinguishable from a generic invalid payload; the equivocation classifier should recognize it when the signer has already counted for that `(order, action)`.

### The audit-event mechanism to extend

Two seams cooperate. The **reject-reason taxonomy** is `AttestationRejectReason` in `bridge/core/src/attestation.rs` (**enum at line 544**), with `.tag()` (**line 590**) producing the stable machine tags (`replayed_nonce`, `wrong_order`, `invalid_payload`, ...) and `.is_authenticated()` (**line 578**) gating audit logging to post-signature outcomes. `AttestationOutcome` (**line 631**) carries `tag`, `signer_key_id`, `order_id`, `action` — the exact fields an equivocation audit needs.

The **audit sink** is `Database::log_audit(order_id, action, details)` in `bridge/service/src/db.rs` (**line 2056**), which INSERTs into `audit_log`; the matching count helper for tests is `Database::count_audit_action(action)` (**line 858**). Existing callers (`engine.rs` `breaker_tripped`/`breaker_resumed`/`reserve_*` etc.) show the pattern.

The **db-bearing callsite** for inbound peer envelopes is the `attest` handler in `bridge/service/src/federation.rs` (**fn at line 131**). It holds `AttestState { provider, db, inbound_auth_token }` and already calls `state.provider.submit_attestation(&envelope, &order)` (**line ~207**) then maps `outcome.tag` to HTTP status. This is where `log_audit` must fire, because `verify_and_aggregate`/the provider deliberately have no `Database` handle. `submit_attestation_at` is also called from the provider's own self-attest and mint/release authorize paths (**lines 836, 1107**); equivocation there implies a local-key or logic bug and should log too — but detection must be a property of the outcome so all callsites can react uniformly.

### Recommended approach (detection-only; funds-safety invariant unchanged)

1. **Detect in `core`.** Give `AttestationSet` the ability to distinguish a benign re-count from a conflicting one — e.g. add a method that, for an already-counted signer, compares the newly-presented signature/payload digest against the stored entry and returns an "equivocation detected" signal (or extend `insert`'s return to a small enum: `NewSigner` / `DuplicateBenign` / `Equivocation`). Do **not** change what counts toward threshold: an equivocating resubmission must still count **zero** (still `Ok(false)`-equivalent). The two existing adversarial tests must continue to pass unchanged.
2. **Classify in the pipeline.** In `verify_and_aggregate` Step 7, when the set reports equivocation, surface it on the `AttestationOutcome` — either a new `AttestationRejectReason::Equivocation { .. }` variant with `.tag() == "equivocation"` (note: it must be treated as *authenticated* so it is audit-eligible; and it must NOT change funds behavior — the first submission already counted), or a dedicated boolean/field on `AttestationOutcome`. Prefer the outcome-field approach if you want the first submission to remain "accepted" while still flagging the conflict; prefer a distinct refuse-tag if the conflicting *second* submission should surface as a refusal. Either way the machine tag must be **distinct** from `replayed_nonce` / `wrong_order` / `invalid_payload`.
3. **Emit the audit event.** In the `attest` handler (and any other db-bearing callsite), when the outcome carries the equivocation signal, call `db.log_audit(Some(&order.id), "attestation_equivocation", &details)` where `details` includes the signer key id and order id (no secret material — signer key id is a public identity, consistent with `.tag()`'s "no secret / no attacker free-text" rule).
4. **(Optional, gated by acceptance)** Consider tripping the breaker or raising a high-severity alert. The breaker seam is `engine.rs` (`log_audit(None, "breaker_tripped", reason)` at **line 569**) and `alerts.yml`. Recommend **audit-only in the first PR** and filing breaker-trip as a follow-up, because auto-halting the whole bridge on a single equivocation report is a policy decision (a compromised key equivocating is exactly when you may want the *other* honest signers to still make progress). Document the choice in the PR.

### Funds-safety invariant (MUST hold)

Detection is **observation only**. After this change:
- An equivocating signer still counts **once** toward `t` (never inflates the threshold).
- No conflicting attestation can move funds (order binding + Safe-nonce binding unchanged).
- The two `adversarial_tests.rs` regression tests still pass verbatim.
The only behavioral delta is the new `attestation_equivocation` row in `audit_log` (and optionally an alert/breaker trip if that scope is approved).

## Affected Files

- `bridge/core/src/attestation.rs` - `AttestationSet::insert` / `matches` (lines 1288/1310): add conflict-vs-benign discrimination for an already-counted signer (compare stored signature/digest); optionally a new `AttestationRejectReason::Equivocation` variant + `.tag()`/`.is_authenticated()`/`.message()` arms (lines 544/578/590/606).
- `bridge/service/src/attestation.rs` - `verify_and_aggregate` Step 7 (lines 738-777): classify the equivocation signal onto `AttestationOutcome`; do not change threshold counting.
- `bridge/service/src/federation.rs` - `attest` handler (line 131): call `db.log_audit(Some(&order.id), "attestation_equivocation", ...)` when the outcome flags equivocation; leave HTTP status mapping funds-safe.
- `bridge/service/src/adversarial_tests.rs` - add a service-level test asserting the audit fires exactly once per equivocating signer.
- `bridge/core/src/adversarial_tests.rs` - the two existing equivocation tests must still pass; extend/add a core-level assertion that `insert` now reports conflict vs benign duplicate.
- `docs/security/bridge-threat-model.md` - update threat #7 "Residual / follow-up" (lines 126-134) and the Follow-ups entry (lines 187-189) once detection lands.
- `bridge/service/src/db.rs` - reference only: `log_audit` (line 2056), `count_audit_action` (line 858) for the test.

## Test Plan

- [ ] Automated (core): a `bridge/core/src/adversarial_tests.rs` test where the SAME signer submits a CONFLICTING payload signature/digest for one `(order, action)` — assert the set reports equivocation, `distinct_signers()` stays at 1, and `is_threshold_met(2)` is still false.
- [ ] Automated (service): in `bridge/service/src/adversarial_tests.rs`, drive two conflicting verified envelopes from one signer through `submit_attestation_at` / the `attest` handler and assert `db.count_audit_action("attestation_equivocation") == 1` (fires **exactly once** per equivocating signer — a third conflicting submission from the same signer should not double-log, or the test must pin the exact expected count).
- [ ] Automated (regression): `equivocating_signer_cannot_inflate_the_distinct_signer_count` and `conflicting_payload_for_the_same_order_cannot_cross_order_binding` still pass unchanged.
- [ ] Edge cases: (a) benign identical re-submission (same signer, same bytes) must NOT log `attestation_equivocation`; (b) Ethereum-mint different-`safe_nonce` from an already-counted signer is classified as equivocation, not generic `invalid_payload`; (c) two DIFFERENT honest signers for the same order do not trigger the alarm.
- [ ] Manual verification: run `cargo test -p bth-bridge-core -p bth-bridge-service` (adjust crate names to the actual `Cargo.toml`) and confirm the new audit row appears; grep the `audit_log` for `attestation_equivocation`.
- [ ] Invariant check: confirm no funds-path behavior changed (threshold still requires `t` distinct signers; conflicting attestation never authorizes).

rjwalters:
## Summary

Equivocation by a federation signer (submitting CONFLICTING bytes for the same order) was *neutralized* but *silent*: `AttestationSet::insert` returned `Ok(false)` for an already-counted signer without comparing the stored signature/digest, so a Byzantine or compromised-key validator equivocating produced no distinct audit signal. This PR adds **detection-only** classification that emits a dedicated `attestation_equivocation` audit event — funds-safety behavior is unchanged.

## Changes

- **`bridge/core/src/attestation.rs`**
  - New `InsertOutcome` enum (`NewSigner` / `DuplicateBenign` / `Equivocation`).
  - `AttestationSet::insert_classified` compares the newly-presented payload signature against the stored entry for an already-counted signer: identical bytes = benign re-send, differing bytes = equivocation. `insert` is now a thin wrapper preserving its `Result<bool, String>` contract (both already-counted cases still map to `Ok(false)`).
  - `flag_equivocation(signer)` records flagged signers so the alarm fires at most once per signer; `AttestationOutcome.equivocation` flag + `with_equivocation()` builder.
- **`bridge/service/src/attestation.rs`** — `verify_and_aggregate` Step 7 classifies both conflict shapes onto the outcome without changing threshold counting: (1) conflicting payload signature bytes via `insert_classified`; (2) an already-counted signer presenting a different Safe nonce (which `matches` rejects as `Err`) is flagged as equivocation rather than generic `invalid_payload`.
- **`bridge/service/src/federation.rs`** — the db-bearing `attest` handler emits `db.log_audit(Some(order.id), "attestation_equivocation", details)` (signer key id + order id) when the outcome flags equivocation; a failed audit insert never changes the funds-safe HTTP verdict.
- **`bridge/core/src/lib.rs`** — export `InsertOutcome`.
- **`docs/security/bridge-threat-model.md`** — threat #7 updated to "neutralized AND detected"; follow-up entry marked landed (breaker-trip deferred as policy).

### Scope note (breaker / alerting)

Per the curator's guidance this PR is **audit-only**. Auto-tripping the circuit breaker on equivocation was deliberately deferred: a single compromised key equivocating is exactly when the *other* honest signers may need to keep making progress. Left as a policy follow-up (documented in the threat model).

| Verified signer with conflicting attestations for one order produces an `attestation_equivocation` audit event | ✅ | `endpoint_audits_equivocation_exactly_once_per_signer` asserts `count_audit_action("attestation_equivocation") == 1` after a conflicting Safe nonce from a counted owner |
| Audit event is DISTINCT from `replayed_nonce` / `wrong_order` / `invalid_payload` | ✅ | New audit action string `attestation_equivocation`; the refusal still carries its own `invalid_payload` tag |
| Fires EXACTLY ONCE per equivocating signer | ✅ | Third conflicting submission (Safe nonce 9) keeps the count at 1; `flag_equivocation` dedups per signer |
| Benign identical re-send does NOT log | ✅ | Step 4 of the service test: re-sending nonce-7 bytes stays 200 with no new audit row |
| Two DIFFERENT honest signers do NOT trigger the alarm | ✅ | Step 5: owner B reaches 2/2, audit count stays 1 |
| Funds-safety unchanged (still neutralized, counts once) | ✅ | `distinct_signers` stays correct throughout; core test asserts `is_threshold_met(2)` false and `insert` still `Ok(false)` |
| Existing equivocation regression tests pass unchanged | ✅ | `equivocating_signer_cannot_inflate_the_distinct_signer_count` and `conflicting_payload_for_the_same_order_cannot_cross_order_binding` pass verbatim |

## Test Plan

- `cargo test -p bth-bridge-core -p bth-bridge-service` → 67 + 201 pass, 0 fail.
- `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets` → clean.
- New tests: `equivocating_signer_is_classified_distinctly_from_a_benign_resend` (core), `endpoint_audits_equivocation_exactly_once_per_signer` (service, drives the real `POST /api/attest` handler + counts `audit_log` rows).

Closes #859

### DIFF
--- a/bridge/core/src/adversarial_tests.rs
+++ b/bridge/core/src/adversarial_tests.rs
@@ -28,7 +28,7 @@ use crate::{
     attestation::{
         attestation_signed_message, canonical_attestation_envelope, release_payload_digest,
         sign_attestation_ed25519, AttestationEnvelope, AttestationKind, AttestationRejectReason,
-        AttestationSet, AttestationSignature, ATTEST_DOMAIN_BTH, ATTEST_DOMAIN_ETH,
+        AttestationSet, AttestationSignature, InsertOutcome, ATTEST_DOMAIN_BTH, ATTEST_DOMAIN_ETH,
         ATTEST_DOMAIN_SOL, MINT_ATTESTATION_DOMAIN_TAG_ETH, MINT_ATTESTATION_DOMAIN_TAG_SOL,
         RELEASE_ATTESTATION_DOMAIN_TAG,
     },
@@ -238,13 +238,15 @@ fn equivocating_signer_cannot_inflate_the_distinct_signer_count() {
     // count as two toward the threshold. The aggregation primitive
     // deduplicates by signer identity, so the second submission counts zero.
     //
-    // NOTE ON DETECTION (follow-up filed): `AttestationSet` is *resistant* to
+    // NOTE ON DETECTION (#859): `AttestationSet` is *resistant* to
     // equivocation (a single malicious signer can never move the threshold),
-    // but it does not *flag* the equivocation as an auditable event — it
-    // silently keeps the first submission. Active detection (raising an
-    // equivocation alarm when a signer presents conflicting bytes for the
-    // same order id) is tracked as a hardening follow-up. See the threat
-    // model doc: docs/security/bridge-threat-model.md.
+    // and now also *classifies* it — `insert_classified` reports
+    // `Equivocation` when a counted signer re-sends CONFLICTING bytes so the
+    // service can raise an `attestation_equivocation` audit alarm. This test
+    // pins the funds-safe behavior (`insert` still returns `Ok(false)`, the
+    // count stays at 1);
+    // `equivocating_signer_is_classified_distinctly_from_a_benign_resend` below
+    // pins the new classification. See docs/security/bridge-threat-model.md.
     let sk = signing_key(5);
     let order = burn_order();
     let kind = release_kind(&order);
@@ -277,6 +279,59 @@ fn equivocating_signer_cannot_inflate_the_distinct_signer_count() {
         !set.is_threshold_met(2),
         "a single equivocating signer can never satisfy a 2-of-n threshold"
     );
+}
+
+#[test]
+fn equivocating_signer_is_classified_distinctly_from_a_benign_resend() {
+    // #859: the set must DETECT (not just neutralize) equivocation. A counted
+    // signer that re-sends IDENTICAL bytes is a benign duplicate; a counted
+    // signer that presents CONFLICTING payload bytes for the same
+    // (order, action) is an equivocation — a governance-relevant alarm. In
+    // BOTH cases the threshold is unchanged (still counts once): detection is
+    // observation only, so funds-safety is preserved.
+    let sk = signing_key(9);
+    let order = burn_order();
+    let kind = release_kind(&order);
+
+    let env = sign_attestation_ed25519(&kind, &sk, "eq-c", now(), now() + 120).unwrap();
+    let parsed = env.verify_and_parse_ed25519(&sk.verifying_key()).unwrap();
+
+    let mut set = AttestationSet::for_attestation(&parsed);
+    let real_sig = AttestationSignature {
+        signer: sk.verifying_key().as_bytes().to_vec(),
+        signature: hex::decode(&env.payload_signature_hex).unwrap(),
+    };
+
+    // First submission: a new distinct signer counts.
+    assert_eq!(
+        set.insert_classified(&parsed, real_sig.clone()),
+        Ok(InsertOutcome::NewSigner),
+    );
+    assert_eq!(set.distinct_signers(), 1);
+
+    // Benign re-send: same signer, IDENTICAL bytes. Not an alarm, counts once.
+    assert_eq!(
+        set.insert_classified(&parsed, real_sig.clone()),
+        Ok(InsertOutcome::DuplicateBenign),
+    );
+    assert_eq!(set.distinct_signers(), 1);
+
+    // Equivocation: same signer, CONFLICTING payload signature bytes.
+    let mut conflicting = real_sig.clone();
+    conflicting.signature[0] ^= 0xff;
+    assert_eq!(
+        set.insert_classified(&parsed, conflicting.clone()),
+        Ok(InsertOutcome::Equivocation),
+        "a counted signer presenting different payload bytes must be flagged as equivocation"
+    );
+
+    // Funds-safety invariant: the equivocation NEVER inflated the count and
+    // NEVER satisfied a 2-of-n threshold, and the funds-path `insert` still
+    // reports the neutralizing `Ok(false)` for the conflicting resubmission.
+    assert_eq!(set.distinct_signers(), 1);
+    assert!(!set.is_threshold_met(2));
+    assert_eq!(set.insert(&parsed, conflicting), Ok(false));
+    assert_eq!(set.distinct_signers(), 1);
 }
 
 #[test]
--- a/bridge/core/src/attestation.rs
+++ b/bridge/core/src/attestation.rs
@@ -648,6 +648,14 @@ pub struct AttestationOutcome {
     pub signers: u32,
     /// The configured federation threshold `t`.
     pub threshold: u32,
+    /// Detection-only flag: this VERIFIED envelope came from a signer that had
+    /// already attested this `(order, action)` with CONFLICTING bytes (a
+    /// different payload signature, or — for Ethereum mints — a different Safe
+    /// nonce). It is a governance-relevant equivocation signal the db-bearing
+    /// callsite audits (`attestation_equivocation`). It NEVER changes funds
+    /// behavior: the first submission already counted and an equivocating
+    /// resubmission still counts zero toward `t`.
+    pub equivocation: bool,
 }
 
 impl AttestationOutcome {
@@ -663,7 +671,15 @@ impl AttestationOutcome {
             order_id: Some(parsed.action.order_uuid()),
             signers,
             threshold,
+            equivocation: false,
         }
+    }
+
+    /// Flag this outcome as carrying an equivocation signal (detection only —
+    /// the threshold decision is unchanged). Chainable on `accept`/`refuse`.
+    pub fn with_equivocation(mut self) -> Self {
+        self.equivocation = true;
+        self
     }
 
     /// Build a refusal outcome. `parsed` is `Some` once the envelope parsed
@@ -684,6 +700,7 @@ impl AttestationOutcome {
             order_id: parsed.map(|p| p.action.order_uuid()),
             signers,
             threshold,
+            equivocation: false,
         }
     }
 }
@@ -1250,6 +1267,25 @@ pub fn check_order_binding(
     Ok(())
 }
 
+/// The classified result of inserting a verified attestation into an
+/// [`AttestationSet`]. The threshold decision is carried by the variant, and
+/// an already-counted signer is further split into a benign re-send vs. an
+/// equivocation so the ingest pipeline can raise a detection alarm WITHOUT
+/// changing what counts toward `t`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InsertOutcome {
+    /// A new distinct signer was recorded and now counts once toward `t`.
+    NewSigner,
+    /// An already-counted signer re-sent IDENTICAL payload bytes (a benign
+    /// retry / duplicate delivery). Counts zero; not an alarm.
+    DuplicateBenign,
+    /// An already-counted signer presented CONFLICTING payload bytes for the
+    /// same `(order, action)` — an equivocation. Still counts zero (the
+    /// threshold is never inflated), but is a governance-relevant signal the
+    /// caller should audit.
+    Equivocation,
+}
+
 /// Collects verified per-signer attestations for ONE `(order, action)` and
 /// answers the threshold question. Deduplicates by signer identity: the
 /// same federation member attesting twice (even with distinct nonces)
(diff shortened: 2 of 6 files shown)


rjwalters:
✅ **Approved!** Detection-only equivocation classification is correct and funds-safety is demonstrably unchanged.

## Security review (funds-safety path)

**`insert` contract preserved verbatim.** `insert` is now a thin wrapper over `insert_classified`: `NewSigner → Ok(true)`, `DuplicateBenign | Equivocation → Ok(false)`, and `Err` still propagates via `?`. The set only pushes to `entries` on a new distinct signer, so threshold counting, dedup-by-signer, and the `matches`-first order-binding refusal are behaviorally identical. The new core test pins this directly: after a conflicting resubmission `insert(...) == Ok(false)`, `distinct_signers() == 1`, `!is_threshold_met(2)`.

**Regression tests unmodified.** Diffed `adversarial_tests.rs` against main — zero assertion lines were removed or changed. The only edit inside `equivocating_signer_cannot_inflate_the_distinct_signer_count` is a doc comment; `conflicting_payload_for_the_same_order_cannot_cross_order_binding` does not appear in the diff at all. Both pass verbatim.

**Audit fires exactly once per signer.** `flag_equivocation` dedups via `flagged_equivocators` and returns `true` only the first time. Both conflict shapes route through it: (1) conflicting payload signature bytes in `insert_classified`, (2) conflicting Safe nonce (rejected by `matches` as `Err`) in the service layer. Service test steps 2→3 confirm the count stays at 1 across Safe nonce 8 then 9 — no spam.

**Benign resend does not alarm.** Identical bytes → `DuplicateBenign`, no flag (core test step 2 + service test step 4). A second honest signer reaching threshold does not alarm (service test step 5).

**Failed audit insert cannot change the verdict.** `db.log_audit` error is `warn!`-logged and swallowed in step 3a of the `attest` handler; the HTTP status is computed afterward from `outcome.accepted`. Observability only.

**Both shapes distinguished from replay/wrong-order/invalid-payload.** Distinct `attestation_equivocation` action string; the Safe-nonce shape still carries its own `invalid_payload` refusal (test asserts 409 + `invalid_payload` body AND exactly one audit row). The `already_counted` snapshot is taken before `insert_classified`, which is correct since the `Err` path (matches fails first) never mutates `entries`.

**Breaker-trip deferral is the right call** and correctly documented in the threat model as a policy follow-up — halting the whole bridge on one equivocation report can strand the honest signers.

## Test Execution
- `cargo test -p bth-bridge-core -p bth-bridge-service` → **67 + 201 pass, 0 fail** (matches PR claim). New tests `equivocating_signer_is_classified_distinctly_from_a_benign_resend` and `endpoint_audits_equivocation_exactly_once_per_signer` both green; both regression tests green.
- `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets` → clean.

## CI Status
- cargo test (bridge crates) → pass
- cargo build --workspace (Linux) → pass
- cargo deny → pass
- rustfmt --check → pass
- mergeStateStatus → CLEAN

Closes #859 via auto-close syntax. Clean, well-documented, funds-safety demonstrably preserved.

Issue #872, opened 2026-07-13
Solana wBTH: add Solana CI (anchor build/test) + reconcile watcher burn-event field to timestamp


rjwalters:
Follow-up from the review of #871 (Solana wBTH Anchor program hardening, closes #850). Non-blocking for testnet-track merge; must be closed before the mainnet gate (#830).

## 1. No Solana CI covers `contracts/solana/**`

`contracts-ci.yml` (#861) path-filters on `contracts/ethereum/**` plus two Ethereum Rust files only — it does NOT trigger on `contracts/solana/**`. As a result:

- `anchor build` / `anchor test` never run in CI.
- The `tests/wbth.ts` ts-mocha suite added in #871 is **unvalidated by execution** (the review env had no `solana`/`anchor`/`cargo-build-sbf`; only `cargo build --lib -p wbth` ran, clean).

Add a Solana CI job (solana toolchain + Anchor 0.29 + `solana-test-validator`) that runs `anchor build` and `anchor test` on `contracts/solana/**` changes. Alternatively/additionally, confirm #867's devnet deploy exercises the suite and record the run.

## 2. Watcher burn-event type drift (pre-existing)

`bridge/service/src/watchers/solana.rs::parse_bridge_burn_event` decodes the trailing `BridgeBurnEvent` field as `slot: u64`. The program (#871) emits `timestamp: i64` in that position. Borsh byte layout is identical (8 LE bytes) so decoding does not break, but the field is semantically mislabeled. Reconcile the watcher's `SolanaBurn.slot` -> `timestamp` (i64) when #857 wires the transport.

Refs: #850, #871, #857, #867, #830.

Issue #850, opened 2026-07-13
contracts(solana): harden wbth Anchor program (multisig authority, order-id replay guard, limits, upgrade authority) (Part of #816)


rjwalters:
Split out of #826, which hardened the Ethereum side (`contracts/ethereum/contracts/WrappedBTH.sol`). Per ADR 0005 the Solana `wbth` Anchor program (`contracts/solana/programs/wbth/src/lib.rs`) is also v1 scope and needs the equivalent hardening, per ADR 0002 (validators' Ed25519 keys sign Solana mint authorizations natively).

## Tasks (from the #826 implementation checklist)

- [ ] Replace the single-signer `authority` on `Bridge` with a validator multisig: change the `has_one = authority` gate in `BridgeMint`/`AdminOnly` to an SPL/Squads multisig (validators' Ed25519) or an on-chain t-of-n check; keep the mint-authority PDA (`seeds=[b"bridge"]`) as the `MintTo` signer.
- [ ] Add an order-id replay guard to `bridge_mint`: an `order_id: [u8;32]` arg + a per-order PDA (`init`, `seeds=[b"order", order_id]`) so a duplicate fails at `init` (mirror the Ethereum `processedOrders` guard; coordinate with the #824 attestation order id).
- [ ] Fix arithmetic + daily-reset safety: replace `.checked_add(amount).unwrap()` with `ok_or(BridgeError::…)?`; reconsider the slot-based reset window in favor of `Clock::unix_timestamp` for the 24h reset.
- [ ] Enforce per-tx max + review burn/pause: confirm `bridge_burn` honors `paused`, validate `bth_address` length bounds, and constrain the `user: UncheckedAccount` in `BridgeMint` so mints can't be redirected.
- [ ] Decide Anchor upgradeability: keep the BPF upgrade authority (and document who holds it) or make it immutable via `set-upgrade-authority --final`; record alongside the Ethereum decision (Ethereum chose IMMUTABLE, see contracts/ethereum/README.md).
- [ ] Confirm units/config parity: `mint::decimals = 12`, daily-limit/max literals in picocredits identical to the EVM contract (1M BTH per-tx / 10M BTH daily), `Bridge::LEN` accounts for new replay-guard fields.
- [ ] Consider the auto-pause breaker added on the Ethereum side (daily volume >= threshold flips `paused`) for parity.

## Testing (DoD)

[...]

PR #894, opened 2026-07-13
ci(solana): add anchor build/test workflow for contracts/solana

### LINKED ISSUE #883: ci(solana): add anchor build/test workflow for contracts/solana/** (orphaned by #872 partial close)
## Gap

#872 ("add Solana CI + reconcile watcher burn-event field") was closed **COMPLETED** by PR #875 (`Closes #872`), but #875 only did **part 2** (the burn-event `slot:u64`→`timestamp:i64` field reconcile). **Part 1 — the Solana CI — was never implemented.** This re-tracks it.

## Problem
- `.github/workflows/contracts-ci.yml` (#861) path-filters on `contracts/ethereum/**` + two Ethereum Rust files only. It does **not** trigger on `contracts/solana/**`.
- No workflow anywhere runs `anchor build` / `anchor test` (verified: `grep -l anchor` across all 14 `.github/workflows/*.yml` → none; `bridge-ci.yml` only path-filters `contracts/ethereum/contracts/**`).
- Consequences:
  - The hardened wBTH Anchor program (#850, `contracts/solana/programs/wbth/src/lib.rs`) — a security-critical mint-authority contract — has **no CI proving it even compiles as an SBF program** (only a local host `cargo build --lib -p wbth` was ever run).
  - The `tests/wbth.ts` ts-mocha suite (#850) and the `solana_devnet_tests.rs` drills (#857) are **unvalidated by execution** — no environment in CI or review has run them.

## Task
- [ ] Add a Solana CI job (Solana toolchain + Anchor 0.29 + `solana-test-validator`) that runs `anchor build` and `anchor test` on `contracts/solana/**` changes (mirror the `contracts-ci.yml` structure/path-filter pattern).
- [ ] Ensure the #850 ts-mocha suite runs green in that job (first real execution — expect to fix any test issues surfaced).
- [ ] Confirm the instruction-data / account-meta assumptions the Rust minter (`bridge/service/src/mint/solana.rs`) makes actually hold against a built program.

## Priority
Non-blocking for testnet-track, but a hard prerequisite before the mainnet gate (#830) and ideally landed **before** the #867 devnet deploy so the program is validated first. Also related: #879 (assert multisig-authority init — now CLOSED), #867 (devnet deploy exercises the suite for real).

Refs: #872 (partial close), #850, #857, #861, #867, #879, #830.

---

### Verified facts against `origin/main`

- **No Solana workflow exists.** All 14 repo workflows enumerated; none run `anchor`. `bridge-ci.yml` and `contracts-ci.yml` both path-filter Ethereum only. Gap confirmed, not a duplicate.
- **`contracts/solana/` layout** (all committed to `origin/main`):
  - `Anchor.toml` — `[scripts] test = "yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts"`; three program IDs all `wBTH111111111111111111111111111111111111111`; `provider.cluster = "Localnet"`, `provider.wallet = "~/.config/solana/id.json"`.
  - `package.json` — `@coral-xyz/anchor ^0.29.0`, `@solana/spl-token ^0.4.0`, `ts-mocha`, `mocha`, `chai`, `typescript ^5.3.0`. **`"test": "anchor test"`**. Note the Anchor test *script* inside `Anchor.toml` invokes **`yarn run ts-mocha`**, but `package.json`'s own `test` calls `anchor test`.
  - `Cargo.toml` (workspace) — `members = ["programs/wbth"]`, `resolver = "2"`, release profile has `overflow-checks = true`, `lto = "fat"`.
  - `programs/wbth/Cargo.toml` — `anchor-lang = "0.29"`, `anchor-spl = "0.29"`, `hex = "0.4"`; `[lib] name = "wbth"`, `crate-type = ["cdylib", "lib"]`.
  - `programs/wbth/src/lib.rs` — `declare_id!("wBTH111111111111111111111111111111111111111")`; program module is `pub mod wbth_bridge`.
  - `tests/wbth.ts` — imports `Wbth` from `../target/types/wbth` and uses `anchor.workspace.Wbth` + `anchor.AnchorProvider.env()`. **Requires `anchor build` to have generated `target/types/wbth.ts` first**, and requires the env vars `ANCHOR_PROVIDER_URL` + `ANCHOR_WALLET` (which `anchor test` sets automatically when it spawns the validator).
  - `tsconfig.json` — `module: commonjs`, `types: [mocha, chai, node]`.
- **Anchor version = 0.29** (README "Anchor CLI 0.29"; both crates pinned `0.29`; `@coral-xyz/anchor ^0.29.0`). CI must pin the Anchor CLI to **exactly 0.29.0** — a newer CLI generates an incompatible IDL and diverges from the pinned crates.
- **No lockfile is committed** (`package-lock.json` / `yarn.lock` both absent under `contracts/solana/`). `contracts-ci.yml`'s Ethereum job uses `npm ci` with `cache-dependency-path` — that pattern **cannot be copied verbatim** here because `npm ci` requires a lockfile. See "Lockfile decision" below.
- **Rust minter assumptions to validate** (`bridge/service/src/mint/solana.rs`), so a build-time drift fails CI instead of a stuck mint:
  - `encode_bridge_mint_instruction_data(amount, order_id)` = `anchor_discriminator("bridge_mint")` (8 bytes) ++ `amount.to_le_bytes()` (u64 LE) ++ `order_id` (32 raw bytes). This matches `bridge_mint(ctx, amount: u64, order_id: [u8; 32])` and Anchor's borsh arg encoding.
  - The `build_bridge_mint_instruction` account-meta order — `bridge_pda(w)`, `order_marker_pda(w)`, `mint(w)`, `recipient_ata(w)`, `recipient(r)`, `mint_authority(w,signer)`, `TOKEN_PROGRAM_ID(r)`, `SYSTEM_PROGRAM_ID(r)` — matches the `BridgeMint<'info>` accounts struct field order exactly (bridge, order_marker, mint, user_token_account, user, mint_authority, token_program, system_program). The associated-token program is NOT in the meta list (the program constrains the ATA but does not CPI-create it), which is correct for a mint into a pre-existing ATA.
  - Seeds: `BRIDGE_PDA_SEED = b"bridge"`, `ORDER_MARKER_SEED = b"order"` match the program's `seeds = [b"bridge"]` and `seeds = [b"order", order_id.as_ref()]`.
  - `Bridge` layout offset assumption (`extract mint`: discriminator(8) + mint_authority(32) + admin_authority(32) + pauser_authority(32) → mint at offset 104) matches the `#[account] Bridge` field order and `Bridge::LEN`.

### Implementation Guidance

Add a new workflow `.github/workflows/solana-contracts-ci.yml` (keep it separate from the Ethereum `contracts-ci.yml` so a failing/slow Solana job never gates Ethereum PRs). Mirror the `contracts-ci.yml` shape: `on: pull_request` with a `paths:` filter, `workflow_dispatch`, a single job with `defaults.run.working-directory: contracts/solana`, and an explicit `timeout-minutes`.

**Path filter** (mirror the Ethereum pattern — contract dir + the Rust files whose assumptions the suite pins + the workflow itself):

```yaml
on:
  pull_request:
    paths:
      - 'contracts/solana/**'
      - 'bridge/service/src/mint/solana.rs'
      - '.github/workflows/solana-contracts-ci.yml'
  workflow_dispatch:
```

**Job skeleton to follow** (versions are illustrative — pin to the compatible set; see risks):

```yaml
jobs:
  anchor:
    name: anchor build + test
    runs-on: ubuntu-latest
    timeout-minutes: 45   # SBF build + validator spin-up + toolchain install is slow
    defaults:
      run:
        working-directory: contracts/solana
    steps:
      - uses: actions/checkout@v4

      # Rust toolchain the Anchor/SBF build needs.
      - uses: dtolnay/rust-toolchain@stable

      - name: Install Solana CLI (pinned)
        run: |
          sh -c "$(curl -sSfL https://release.anza.xyz/v1.18.26/install)"
          echo "$HOME/.local/share/solana/install/active_release/bin" >> "$GITHUB_PATH"

      - name: Install Anchor CLI 0.29.0 (pinned via avm)
        run: |
          cargo install --git https://github.com/coral-xyz/anchor avm --locked --force
          avm install 0.29.0
          avm use 0.29.0

      - name: Setup Node
        uses: actions/setup-node@v4
        with:
          node-version: 22

      - name: Install JS deps
        run: npm install    # NOTE: no lockfile committed — see "Lockfile decision"

      - name: Generate a local keypair for the provider wallet
        run: solana-keygen new --no-bip39-passphrase -o "$HOME/.config/solana/id.json"

      - name: anchor build
        run: anchor build

      - name: anchor test
        run: anchor test     # spawns solana-test-validator, runs ts-mocha
```

**Cache** the `~/.cargo`, `~/.local/share/solana`, and Anchor `avm` installs (via `actions/cache` or `Swatinem/rust-cache`) to keep re-runs under the timeout; the first cold run installing the Solana + Anchor toolchains is the slow path.

### Lockfile decision (must resolve before green)

`contracts-ci.yml`'s `npm ci` + `cache-dependency-path` pattern needs a committed lockfile. `contracts/solana/` has none. Pick one and do it in this PR:
- **(A, recommended)** `npm install` (no `ci`, no npm cache) — simplest, tolerates the missing lockfile. Downside: non-reproducible dep resolution, no `setup-node` cache.
- **(B)** Commit a `package-lock.json` in `contracts/solana/`, then use `npm ci` + `cache: npm` + `cache-dependency-path: contracts/solana/package-lock.json` exactly like the Ethereum job. Reproducible; preferred if you want parity with #861. If chosen, add `contracts/solana/package-lock.json` to the Affected Files.

The `Anchor.toml [scripts] test` line invokes **`yarn`**; either (a) add yarn to the job, or (b) rewrite that script line to `npm exec ts-mocha ...` / `npx ts-mocha ...` so the toolchain stays npm-only. Recommend (b) for consistency with the Ethereum job.

### Expected first-execution fixups (flag, don't pre-solve)
- `target/types/wbth.ts` is generated from program module `wbth_bridge` but the package/lib is `wbth`; the test imports type `Wbth` from `../target/types/wbth`. If `anchor build` emits the IDL under a different basename, the import path or the `[lib] name`/module name may need reconciling. Verify the generated file is `target/types/wbth.ts` exporting `Wbth`.
- The zero-order-id and duplicate-order-id assertions in `tests/wbth.ts` are the first real exercise of the `init`-based replay guard; surfaced failures are in-scope to fix here.

## Affected Files

- `.github/workflows/solana-contracts-ci.yml` - **NEW** workflow: path-filter `contracts/solana/**` + `bridge/service/src/mint/solana.rs`; job installs pinned Solana + Anchor 0.29 toolchains, runs `anchor build` then `anchor test`.
- `contracts/solana/Anchor.toml` - the `[scripts] test` line may need `yarn`→`npx`/`npm exec` so CI stays npm-only (or add yarn to the job).
- `contracts/solana/package.json` - reference; may add a lockfile-adjacent script; if option (B) chosen, a sibling `package-lock.json` is added.
- `contracts/solana/tests/wbth.ts` - reference; may need fixes surfaced by the first real `anchor test` run.
- `contracts/solana/programs/wbth/src/lib.rs` - reference only; validate the built program matches the Rust minter's assumptions. Do NOT change the `bridge_mint` arg order or seeds without updating `bridge/service/src/mint/solana.rs`.
- `bridge/service/src/mint/solana.rs` - reference only; the assumptions (discriminator, account-meta order, seeds, `Bridge` mint offset) this workflow validates. No change expected unless the build surfaces drift.

## Test Plan

- [ ] Manual verification: Trigger the new workflow via `workflow_dispatch` (and via a PR touching `contracts/solana/**`); confirm `anchor build` produces `contracts/solana/target/types/wbth.ts` and `anchor test` runs the full `tests/wbth.ts` suite green.
- [ ] Automated tests: The `tests/wbth.ts` ts-mocha suite is the automated test — it must pass in CI (first real execution). No new Rust unit tests required; if the build surfaces a drift against `bridge/service/src/mint/solana.rs`, add/adjust a pin there.
- [ ] Edge cases:
  - [ ] Path filter fires on `contracts/solana/**` and on `bridge/service/src/mint/solana.rs`, and does NOT fire on unrelated changes (verify with a no-op PR).
  - [ ] Anchor CLI pinned to exactly `0.29.0` (matches the `0.29` crate + `^0.29.0` JS dep); a version bump must be a deliberate, reviewed change.
  - [ ] Cold-cache run completes within `timeout-minutes`; warm-cache run is materially faster.
  - [ ] Instruction discriminator/account-meta parity: the suite's successful `bridge_mint` proves the account order the Rust minter builds is accepted by the built program.

## Risks

- **`anchor test` in CI is slow and flaky.** It installs a full Solana + Anchor toolchain (multi-minute cold), compiles the SBF program, and spins up `solana-test-validator`. Mitigations: **pin exact toolchain versions** (Solana CLI + Anchor 0.29.0 via `avm`), aggressively cache cargo/solana/avm, set a generous `timeout-minutes` (≈45), and keep the job in its own workflow so its flakiness never blocks Ethereum/bridge PRs. Consider `workflow_dispatch` + path-filter only (not on every PR) if flakiness proves unmanageable, but path-filtered is the target.
- **Toolchain drift.** A floating Solana/Anchor install will silently regenerate an incompatible IDL; the pin is load-bearing, not cosmetic.

rjwalters:
## Summary

Adds a new CI workflow that runs `anchor build` then `anchor test` against the wBTH Anchor program on `contracts/solana/**` changes. Before this, the security-critical mint-authority program (#850) had **no CI proving it compiles as an SBF program**, and the `tests/wbth.ts` ts-mocha suite had never been executed anywhere. This re-tracks part 1 of #872 (orphaned by that issue's partial close — PR #875 only landed part 2).

The workflow is kept **separate** from the Ethereum `contracts-ci.yml` so the slow Solana toolchain install + SBF compile + `solana-test-validator` spin-up never gates Ethereum/bridge PRs.

## Changes

- **NEW `.github/workflows/solana-contracts-ci.yml`**
  - Path filter: `contracts/solana/**`, `bridge/service/src/mint/solana.rs` (the Rust minter whose discriminator / account-meta / seed / `Bridge`-layout assumptions the suite pins), and the workflow file itself. Plus `workflow_dispatch`.
  - Pins **Solana CLI 1.18.26** and **Anchor 0.29.0** (via `avm`) — matching the `0.29` crates and `^0.29.0` JS dep. The pins are load-bearing: a floating install regenerates an IDL incompatible with the pinned crates.
  - Caches cargo registry/target and the Solana/avm toolchains so warm re-runs stay under the 60-minute timeout.
  - Generates a provider keypair at `~/.config/solana/id.json` (the `Anchor.toml [provider].wallet`), then runs `anchor build` (which also generates `target/types/wbth.ts` the suite imports) and `anchor test`.
- **`contracts/solana/Anchor.toml`** — rewrote the `[scripts] test` line from `yarn run ts-mocha ...` to `npx ts-mocha ...` so the toolchain stays npm-only, consistent with the rest of the repo (blocker B).

### Blocker resolutions (curator-flagged, required for a green run)

- **(A) No committed JS lockfile** under `contracts/solana/` → the job uses `npm install` (not `npm ci`, which requires a lockfile) and skips the npm lockfile cache. Chose the recommended option A over committing a lockfile to keep the PR minimal; cargo + Solana/Anchor toolchains are cached instead.
- **(B) `Anchor.toml [scripts] test` invoked `yarn`** → rewritten to `npx ts-mocha` (npm-consistent), rather than installing yarn in the job.

| Solana CI job runs `anchor build` + `anchor test` on `contracts/solana/**` | ✅ | Workflow has both steps under a path-filtered `pull_request` + `workflow_dispatch` job |
| Mirror `contracts-ci.yml` structure / path-filter pattern | ✅ | Same shape: `pull_request.paths` + `workflow_dispatch`, single job, `defaults.run.working-directory`, `timeout-minutes` |
| Pin Solana toolchain + Anchor 0.29 + test-validator | ✅ | `SOLANA_VERSION=1.18.26`, `ANCHOR_VERSION=0.29.0` via `avm`; `anchor test` spawns `solana-test-validator` |
| Blocker A — lockfile | ✅ | `npm install`, no `npm ci`, no npm lockfile cache |
| Blocker B — yarn in Anchor.toml | ✅ | `[scripts] test` now `npx ts-mocha` |
| Rust minter parity (discriminator / arg order) | ✅ (static) | `encode_bridge_mint_instruction_data` = `discriminator("bridge_mint")` ++ `amount` LE ++ `order_id`; program sig is `bridge_mint(ctx, amount: u64, order_id: [u8;32])`. Curator verified account-meta order. Runtime parity is proven by the first CI `anchor test`. |

## Test Plan

**The first CI run on this PR is the real validation of `anchor build` / `anchor test`** — this is the authoritative gate, since the full Solana/Anchor toolchain + validator cannot be run in the builder environment.

What I verified statically here:
- YAML parses cleanly (`yaml.safe_load`); job/paths/steps structure matches the Ethereum workflow's shape.
- `Anchor.toml` `[scripts] test` and `package.json` are now npm-consistent (no `yarn`).
- `cargo build --lib -p wbth` still succeeds locally (host build, cfg warnings only) after the `Anchor.toml` edit — the `[scripts]` change does not affect the Rust build.
- Program `[lib] name = "wbth"` matches the test's `../target/types/wbth` import and the exported `Wbth` type, so no IDL-basename reconciliation is expected.
- Minter discriminator/arg encoding matches the program's `bridge_mint` signature.

**Watch the first CI run for these curator-flagged first-execution fixups** (in-scope to address in follow-up if they surface):
- Whether `anchor build` emits `target/types/wbth.ts` exporting `Wbth` at the path the suite imports.
- The zero-order-id / duplicate-order-id replay-guard assertions (first real exercise of the `init`-based marker).
- Cold-cache run completing within `timeout-minutes` (60).

Closes #883


rjwalters:
❌ **Changes Requested — CI Failing (anchor build)**

This PR's own first CI run is the authoritative validation gate, and it **failed** at the `anchor build` step (run 29294391533, `anchor build + test` job → FAILURE). The YAML structure itself is sound (see review below), but the build does not yet go green, so this is not mergeable.

## The failure

```
error: failed to parse lock file at: /home/runner/work/botho/botho/contracts/solana/Cargo.lock

Caused by:
  lock file version `4` was found, but this version of Cargo does not understand
  this lock file, perhaps Cargo needs to be updated?
```

## Root cause

- No `Cargo.lock` is committed under `contracts/solana/`, so `anchor build` generates a fresh one.
- The workflow installs the **latest stable Rust** globally (`dtolnay/rust-toolchain@stable`). Stable Rust now emits **Cargo lockfile version 4**.
- `anchor build` shells out to the SBF toolchain bundled with the **pinned Solana 1.18.26** (`cargo-build-sbf`), whose bundled Cargo is older and only understands lockfile version ≤ 3. It cannot parse the v4 lock the host stable toolchain wrote, and aborts.

This is a host-Rust vs. pinned-SBF-Cargo version skew — a different first-execution fixup than the three the curator flagged (it fails before we ever reach IDL basename / replay-guard / validator-timeout).

## Suggested fixes (pick one)

1. **Pin the host Rust toolchain** to a version compatible with Solana 1.18.26's SBF Cargo (e.g. `dtolnay/rust-toolchain@1.79.0` or similar), so the generated lock stays at version 3. This is the most surgical fix and matches the "pins are load-bearing" intent already stated in the workflow header.
2. **Force lockfile v3**: run `cargo generate-lockfile` / build with `-Znext-lockfile-bump`-free settings, or commit a `Cargo.lock` pinned to `version = 3` under `contracts/solana/` and build with `--locked`. Committing a v3 lock also gives reproducibility parity with the Ethereum `npm ci` path.
3. Bump to a Solana release whose bundled Cargo understands lockfile v4 (larger blast radius — would break the 0.29-crate IDL pinning the PR relies on, so **not** recommended).

Option 1 (pin host Rust) is likely the smallest, lowest-risk change.

Once `anchor build` compiles, watch for the still-unvalidated downstream fixups the curator predicted: IDL type basename (`target/types/wbth.ts` exporting `Wbth`), zero/duplicate order-id replay-guard assertions, and cold-cache validator timeout under the 60-min cap.

## What is already correct (no changes needed here)

- YAML mirrors `contracts-ci.yml` shape: path-filtered `pull_request` + `workflow_dispatch`, single job, `defaults.run.working-directory: contracts/solana`, `timeout-minutes`.
- Path filters (`contracts/solana/**`, `bridge/service/src/mint/solana.rs`, the workflow file) are appropriate.
- Blocker A: `npm install` (not `npm ci`) correctly handles the absent JS lockfile.
- Blocker B: `Anchor.toml [scripts] test` rewritten `yarn` → `npx ts-mocha` is correct and npm-consistent; `cargo build --lib -p wbth` is unaffected by this edit.
- `[lib] name = "wbth"` matches the test's `../target/types/wbth` import and `Wbth` type — no IDL basename reconciliation needed.
- `Closes #883` present.

Please address the lockfile/toolchain skew and re-request review. I'll re-evaluate once `anchor build` + `anchor test` go green.

rjwalters:
Feedback addressed: pinned the host Rust toolchain (fix #1, Judge's recommended smallest change).

`dtolnay/rust-toolchain@stable` -> `dtolnay/rust-toolchain@1.79.0`

Root cause confirmed: current stable Rust (>= 1.83) writes Cargo lockfile version 4 by default; with no `Cargo.lock` committed under `contracts/solana/`, `anchor build` regenerated a v4 lock, which Solana 1.18.26's bundled `cargo-build-sbf` (older Cargo, understands lockfile <= v3) could not parse -> build abort. Pinning host Rust to 1.79.0 keeps the generated lock at v3 while still building avm/anchor 0.29 cleanly.

Scope: only the toolchain-pin line changed (plus an explanatory comment). Everything you verified as correct (YAML shape, path filters, `npm install`, `npx ts-mocha` in Anchor.toml, `[lib] name = wbth`) is untouched.

This needs the Solana CI to re-run and go green — please re-evaluate once `anchor build` + `anchor test` pass. Watch for the downstream fixups the curator flagged (IDL basename, replay-guard assertions, cold-cache validator timeout) that were previously unreachable behind the build failure.

rjwalters:
**Sweep disposition: blocked after one Doctor→Judge cycle (single-cycle cap reached).**

The Doctor's toolchain pin (`dtolnay/rust-toolchain@stable` → `@1.79.0`) did **not** clear the failure. CI run `29294966098` against the fixed head `a7755208` still fails at `anchor build` with the identical error:

```
error: failed to parse lock file at: contracts/solana/Cargo.lock
Caused by: lock file version `4` was found, but this version of Cargo does not understand this lock file
```

**Why the pin didn't work:** pinning the host default toolchain to 1.79 is insufficient — some step in the `avm`/anchor install path (or a rustup override) still uses a newer cargo that regenerates `contracts/solana/Cargo.lock` at v4 before Solana 1.18.26's bundled `cargo-build-sbf` (Cargo ≤ v3) reads it.

**Recommended next fix (focused follow-up, needs live CI iteration):**
- Commit a `version = 3` `contracts/solana/Cargo.lock` and build with `--locked` / `anchor build -- --locked`, so no step regenerates a v4 lock; **and/or**
- Force the SBF build's own toolchain, e.g. set `RUSTUP_TOOLCHAIN`/`cargo-build-sbf --tools-version`, or export `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS` and ensure `rustup default 1.79` (not just the action) so every cargo invocation in the job emits v3.
- Then let the downstream curator-predicted fixups (IDL basename, replay-guard assertions, cold-cache validator timeout) surface on the next green build.

The **workflow YAML itself is correct** (Judge verified structure/path-filters/blocker-A+B); only the lockfile-version interaction remains. Leaving the PR open for a focused follow-up; the sweep does not run a second Doctor pass per its latency cap.

rjwalters:
## Judge review: APPROVED

Reviewing the doctor's 2nd-cycle fix (head `62aa81a7`). CI is fully green including the 11-minute `anchor build + test` job (the point of #883), merge state CLEAN.

### Verified (not on faith)

**Rust minter ↔ renamed program consistency (the risk surface):**
- The `#[program] mod wbth_bridge → wbth` rename is purely for Anchor 0.29 IDL/TypeScript basename parity. I confirmed it has **zero wire-encoding effect**:
  - Instruction discriminator = `sha256("global:bridge_mint")[..8]` — hashes the *instruction name*, not the module. (`bridge/service/src/mint/solana.rs:84`)
  - Event discriminator = `sha256("event:BridgeBurnEvent")[..8]` — hashes the *event struct name*, not the module. (`bridge/service/src/watchers/solana.rs:378-382`)
  - Both the instruction name (`bridge_mint`) and event struct (`BridgeBurnEvent`) are unchanged. Minter and watcher agree with the on-chain program.
- Grepped the whole tree: the only remaining `wbth_bridge` string is the explanatory NOTE comment in `lib.rs:66` documenting the rename. The minter/watcher doc comments now correctly say `wbth`.
- `cargo test -p bth-bridge-service`: **209 passed, 0 failed** on the branch. Bridge stays green.

**Cargo.lock v3:** `version = 3` header, anchor-lang 0.29.0 + family all pinned to 0.29.0 matching the manifest, real 2624-line resolved graph — not a hand-hacked version field. The workflow's `cargo metadata --locked` (pre-build) + `git diff --exit-code -- Cargo.lock` (post-build) guards both passed in CI, empirically proving `--locked` holds. `rust-version = 1.72.0` MSRV pin is the mechanism that lets the resolver pick SBF-compilable versions.

**ts suite — no assertion weakening:**
- Event name `bridgeBurnEvent → BridgeBurnEvent`: a **correctness** fix (Anchor 0.29 emits PascalCase IDL names; the old camelCase listener would never have fired). Assertions on captured amount/address remain.
- `20_000_000 → 15_000_000` BTH daily limit: verified math — 20M×10^12 = 2.0e19 > u64::MAX (1.8447e19) and fails borsh client-side; 15M×10^12 = 1.5e19 fits. Assertion `b.dailyMintLimit.eq(newLimit)` intact and still exercises a larger-than-default limit.
- 38 assertions present, all negative-path `assert.fail` guards intact (attacker/zero/over-max/duplicate/non-admin/pause enforcement).

**Workflow YAML (156 lines):** path filters correct (`contracts/solana/**` + `mint/solana.rs` + self); toolchain pins coherent (Solana 1.18.26 / Anchor 0.29.0 / host Rust 1.79.0 / avm `--tag v0.29.0`); plain `pull_request` (no `pull_request_target`), no secrets, no over-broad triggers. Well-formed steps (my initial diff view was mangled by a local grep filter; the on-disk file is clean).

**.gitignore:** `!/contracts/solana/Cargo.lock` exception is necessary to commit the lock past the blanket `Cargo.lock` ignore. Coherent.

### Minor nit (non-blocking)
`watchers/solana.rs` is not in the path filter. Reasonable — the anchor suite exercises the *mint* path, and the watcher's burn-event decoding is independently pinned by its own Rust unit tests (`test_parse_bridge_burn_event_*`). Not worth a cycle.

Approving.

PR #923, opened 2026-07-14
fix(web): stop RemoteNodeAdapter fabricating tx fields + hashRate

rjwalters:
## Summary

`RemoteNodeAdapter` (`web/packages/adapters/src/remote.ts`) hardcoded several fields that render to users as if they were live readings — the #541–#544 fabrication pattern reproduced in the web adapter. This makes the return types honest (nullable/optional where the node genuinely exposes nothing) and updates consumers to render the absence gracefully rather than papering over it with fake defaults.

Refs: #908, PR #910, `docs/design/post-ct-analytics.md` §5 + D1, `docs/design/block-explorer-network-stats.md` (derive-or-n/a rule).

## Changes

- **hashRate (`getNetworkStats`)**: `NetworkStats.hashRate` is now `string | null`. The adapter returns `null` (the seed read RPC has no hash rate) and the desktop dashboard "Hash Rate" card (`dashboard.tsx`) renders `n/a` via `formatHashRate` instead of a fabricated `0 H/s`.
- **amount / type (`getTransaction`)**: `Transaction.amount` and `Transaction.type` are now optional. The adapter omits both — the node exposes no per-tx direction, and per-tx amounts are never public (deprecation **D1**). The explorer `transaction-detail.tsx` **deletes the "Amount" row** entirely and falls back to a neutral icon/label when direction is unknown.
- **timestamp (`getTransaction`)**: `Transaction.timestamp` is now optional. Instead of `Date.now()`, the adapter resolves the real containing-block timestamp when the tx is in a block, else leaves it undefined; `transaction-detail.tsx` renders `—`.
- **WS path**: `parseTransactionEvent` had the same amount/type/timestamp fabrications — fixed consistently.
- `transaction-row.tsx` (wallet history, which supplies real values) guards the now-optional fields; the wallet UI is unchanged in practice.
- Tests: added adapter tests (`getNetworkStats` null hashRate; `getTransaction` no fabricated amount/type, timestamp resolved-from-block / undefined-otherwise) and a `transaction-detail` test (no Amount row, `—` for absent timestamp, neutral type).

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| hashRate no longer a fabricated constant; dashboard shows n/a (not "0 H/s") | ✅ | adapter returns `null`; `formatHashRate(null) => 'n/a'`; adapter test asserts `stats.hashRate` is null |
[...]

Issue #913, opened 2026-07-14
web: RemoteNodeAdapter fabricates tx fields and hashRate rendered as live data


rjwalters:
Follow-up from the #541-class wiring sweep in `docs/design/post-ct-analytics.md` (#908, PR #910). The `RemoteNodeAdapter` (`web/packages/adapters/src/remote.ts`) fabricates several fields that are rendered to users as if they were live data:

## Fabrications

1. **`remote.ts:269`** — `getNetworkStats()` hardcodes `hashRate: '0'`. This is **rendered live** by the desktop dashboard:
   - `web/packages/desktop/src/hooks/useNetworkStats.ts:34` calls `adapter.getNetworkStats()`
   - `web/packages/desktop/src/contexts/connection.tsx:59` (also :119/:170/:211) — desktop connects to seed/remote nodes via `RemoteNodeAdapter`
   - `web/packages/desktop/src/pages/dashboard.tsx:75` renders a "Hash Rate" stat card from `stats.hashRate` via `formatHashRate` — the dashboard shows "0 H/s" as if it were a real reading.

   Fix: derive a real value or render "n/a"/"—" per the derive-or-n/a rule in `docs/design/block-explorer-network-stats.md`; do not display a fabricated constant.

2. **`remote.ts:489`** — `getTransaction()` returns `amount: BigInt(0)`, rendered as a real "Amount: 0 BTH" row (`web/packages/features/src/explorer/components/transaction-detail.tsx:64`). Fix = deprecation **D1** in the fate map: delete the Amount row (the node has never exposed per-tx amounts, and under CT it never will).

3. **`remote.ts:494`** — `timestamp: Date.now()` fabricates the tx timestamp as "now". Render "—" or resolve the block timestamp.

4. **`remote.ts:488`** — `type: 'receive'` hardcoded for every transaction.

## Why this matters

These are exactly the #541–#544 failure mode (fields hardcoded to look live) reproduced in the web adapter. The judge review on PR #910 confirmed the hashRate stub is not dormant — the desktop dashboard renders it today.

Refs: #908, PR #910, `docs/design/post-ct-analytics.md` §5 and D1.

rjwalters:
[...]

PR #928, opened 2026-07-14
fix(web): stop parseBlockEvent fabricating reward/size on WS block events (closes #924)

### LINKED ISSUE #924: web: parseBlockEvent fabricates reward:0 / size:0 on live WS block events (#541-class)
Follow-up found during #913 (PR #923). Distinct surface from the four RemoteNodeAdapter fabrications fixed there.

`parseBlockEvent` in `web/packages/adapters/src/remote.ts` (WebSocket block-event path) hardcodes `reward: BigInt(0)` and `size: 0` because the WS event payload omits them. These feed the explorer block list via `onNewBlock`, so a live WebSocket block would flash a fabricated **"+0" reward** and size 0 to users — the same #541–#544 failure mode (fields hardcoded to look live).

## Fix
Same approach as #913: make `Block.reward`/`Block.size` optional/nullable when the WS event genuinely omits them, and have the explorer block list render absence gracefully (—/omit) rather than a fabricated 0. Alternatively, if the node CAN include reward/size in the WS block event, wire them through instead of defaulting.

## Affected files
- `web/packages/adapters/src/remote.ts` — `parseBlockEvent`
- `web/packages/core/src/types/index.ts` — Block type (if made optional)
- explorer block-list component(s) consuming `onNewBlock`
- adapter + component tests

Refs: #913, PR #923, docs/design/post-ct-analytics.md (fabrication inventory / derive-or-n/a rule).

rjwalters:
Closes #924

## Problem
`parseBlockEvent` in `web/packages/adapters/src/remote.ts` (the WebSocket block-event path) hardcoded `reward: BigInt(0)` and `size: 0`. The node's `NewBlock` WS payload carries **only** `height/hash/timestamp/tx_count/difficulty` (`botho/src/rpc/websocket.rs::WsEvent::NewBlock`) — no reward, no size. These blocks flow into the explorer block list via `onNewBlock`, so a live WebSocket block would flash a fabricated **"+0" reward** and **"0 B" size** to users — the #541-class "fields hardcoded to look live" failure mode.

## WS event shape (checked first)
The WS block event **cannot** carry reward/size — the Rust `WsEvent::NewBlock` variant does not include them. So the honest fix is to omit them, not wire them through (mirrors #913 / PR #923).

## Fix (mirrors #913, PR #923)
- `Block.reward` and `Block.size` are now genuinely **optional** in `@botho/core` (not sentinel defaults). Absence means "not carried by this source", never a real zero.
- `parseBlockEvent` **omits** reward/size instead of fabricating `0`/`+0`.
- `mapRpcBlock` (RPC `getBlock` path) likewise omits `size` — the read RPC does not expose it; it keeps the honest `reward` from `mintingReward`.
- Explorer **block list** renders `—` (not `+0`) when reward is absent.
- Explorer **block detail** renders `—` for absent size/reward.

## Tests
- `remote.test.ts`: new `parseBlockEvent` group asserting reward/size are omitted when the WS payload lacks them, and that the fields the event *does* carry are still mapped; plus a `getBlock` assertion that size is omitted.
- `block-list.test.tsx` (new): a block with a reward renders `+<amount>`; a block that omits its reward (live WS block) renders `—` and never `+0`.
- `pnpm test:run` (adapter + block-list): 28 passed.
- `pnpm -r typecheck`: green across all 8 packages.

## Deploy
botho.io redeploy needed to reach the live web wallet / explorer.

### DIFF
--- a/web/packages/adapters/src/__tests__/remote.test.ts
+++ b/web/packages/adapters/src/__tests__/remote.test.ts
@@ -290,6 +290,49 @@ describe('RemoteNodeAdapter.getBlock enriched fields (#700)', () => {
     expect(block!.transactions).toEqual([])
     expect(lastCall('getBlockByHash')?.params.hash).toBe('f'.repeat(64))
   })
+
+  it('omits size — the read RPC does not expose it (not a fabricated 0) (#924)', async () => {
+    const adapter = await connectedAdapter({ getBlockByHeight: rpc(legacyBlock) })
+    const block = await adapter.getBlock(42)
+    expect(block!.size).toBeUndefined()
+  })
+})
+
+describe('RemoteNodeAdapter.parseBlockEvent (WS block event)', () => {
+  // The node's NewBlock WS payload carries only height/hash/timestamp/tx_count/
+  // difficulty (botho/src/rpc/websocket.rs::WsEvent::NewBlock). It has no reward
+  // or size, so the adapter must omit them rather than fabricate 0/+0 (#924).
+  const wsBlock = {
+    height: 100,
+    hash: 'a'.repeat(64),
+    timestamp: 1751900000,
+    tx_count: 3,
+    difficulty: 54321,
+  }
+
+  /** Exercise the private WS parser via bracket access. */
+  function parse(adapter: RemoteNodeAdapter, data: Record<string, unknown>) {
+    return (adapter as unknown as {
+      parseBlockEvent(d: Record<string, unknown>): import('@botho/core').Block
+    }).parseBlockEvent(data)
+  }
+
+  it('omits reward and size — the WS event carries neither (not fabricated 0) (#924)', async () => {
+    const adapter = await connectedAdapter({})
+    const block = parse(adapter, wsBlock)
+    expect(block.reward).toBeUndefined()
+    expect(block.size).toBeUndefined()
+  })
+
+  it('still maps the fields the WS event does carry', async () => {
+    const adapter = await connectedAdapter({})
+    const block = parse(adapter, wsBlock)
+    expect(block.height).toBe(100)
+    expect(block.hash).toBe('a'.repeat(64))
+    expect(block.timestamp).toBe(1751900000)
+    expect(block.transactionCount).toBe(3)
+    expect(block.difficulty).toBe(54321n)
+  })
 })
 
 describe('RemoteNodeAdapter.getNetworkStats', () => {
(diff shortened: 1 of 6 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed against #924 and the #913/PR#923 honest-nullable convention. This is a clean mirror of that pattern on a distinct surface.

### Verified
- **Honest omission, not sentinels.** `Block.reward`/`Block.size` are now genuinely optional (`reward?: Amount`, `size?: number`) in `@botho/core` — not `| 0` sentinels. `parseBlockEvent` omits both keys rather than fabricating `0`/`+0`.
- **WS event genuinely lacks reward/size.** Confirmed in `botho/src/rpc/websocket.rs` — `WsEvent::NewBlock` carries only `height/hash/timestamp/tx_count/difficulty`. There is no reward or size to wire through, so omission is the correct fix, not laziness.
- **mapRpcBlock in scope and correct.** The RPC `getBlock` path was fabricating `size: 0`; it now omits it and keeps the honest `reward` from `mintingReward`. Same bug class as #924, reasonable extension. (The RPC block response does not expose serialized size.)
- **Graceful rendering.** block-list renders `—` (dim) instead of `+0` when reward is absent; block-detail renders `—` for absent size/reward and drops the success color. No new fabrication.
- **No consumers broken.** The only other Block-field consumers are block-list.tsx and block-detail.tsx (both handled). LocalNodeAdapter.parseBlock is a separate local-source path that always sets the fields and is unaffected by the type widening — correctly left out of scope.
- **Tests genuinely assert, not gutted.** New `parseBlockEvent` group asserts `reward`/`size` are `undefined` while still mapping height/hash/timestamp/tx_count/difficulty; block-list test asserts `—` renders AND that `+0`/`^\+0` never appear; getBlock test asserts size omitted. Existing tests updated, not weakened.
- **Consistency + rebase.** Reuses the exact convention #923 established on these files; no conflict. Branch is not fast-forward-rebased on origin/main tip but `git merge-tree` shows a clean merge and `mergeStateStatus: CLEAN / MERGEABLE`.
- **CI + local.** `gh pr checks`: both jobs pass (cargo deny; typecheck + vitest + worker dry-run). Reran locally: `pnpm -r typecheck` green across all 8 packages; `vitest run remote.test block-list.test` = **28 passed**, matching the claim.

### Faith (not independently verified)
- That the RPC `getBlock` response has no size field is asserted from the mapper/comment and the passing test, not by reading the Rust RPC serializer directly — but consistent with the WS finding and the existing `// Not provided` note.

Solid, honest fix. Approving.

Issue #925, opened 2026-07-14
fix: price the demurrage class transition on deflating spends (background-reset leak, #834)


rjwalters:
**MAINNET BLOCKER.** Surfaced by the #834 monetary-integrity investigation. Verdict doc: `docs/research/demurrage-background-reset-leak.md`. Executable evidence: `cluster-tax/src/demurrage.rs::tests::demurrage_background_reset_leak_is_real`.

## The leak (confirmed)

A wealthy holder can reset a coin's demurrage class to background via an **ordinary spend**, paying **zero** demurrage and escaping **all future** stock-level demurrage. The sequence:

1. Re-spend a wealthy coin wealthy→wealthy to reset `created_at` (standard UTXO behavior; new output's `created_at` = current height).
2. Immediately spend the now-YOUNG wealthy coin → a fully background-tagged output. Legal deflation (`check_cluster_tag_inheritance` only rejects inflation). The ring floor `max(claimed, ring_centroid_implied_factor)` still raises the FACTOR to the wealthy value — but demurrage is `∝ elapsed`, and `elapsed ≈ 0`, so the charge is **0**.
3. Future spends of the reset coin pick background decoys ⇒ ring floor is now 1× ⇒ zero demurrage forever.

Observed numbers (1000-BTH coin in a 10M-BTH cluster, 2%/yr, default curve, picocredits):

| Step | factor | elapsed | charge |
|------|--------|---------|--------|
| Honest hold 1yr then spend | 5745 | 1yr | 18,980,000,000,000 (~18.98 BTH) |
| Spend young wealthy → background (the leak) | 5745 (floored, but useless) | ~0 | **0** |
| Future spend of reset background coin, 1yr | 1000 | 1yr | **0** |
| **Total to escape the wealthy class** | | | **0** |

## Why the existing floors miss it

`ring_centroid_implied_factor` (B2) floors the FACTOR; `ring_elapsed_quantile@max` (B1) defends the ELAPSED term against fresh-decoy dilution of an OLD real input. Neither helps when the real input is genuinely YOUNG — the charge is proportional to elapsed, so a fresh coin pays ~0 at any factor. Churn-invariance holds only WITHIN a wealth class; the class transition itself is unpriced.

## Why this is mainnet-blocking

[...]

Issue #831, opened 2026-07-13
consensus: demurrage-settlement operation — pay-to-reclassify a coin to factor-1 (fees → lottery), the wrap on-ramp (Part of #816, from #818)


rjwalters:
**Part of #816. Spawned by the #818 decision. This is the one CONSENSUS-level piece of the bridge.**

## What
A new **demurrage-settlement operation**: a holder pays a charge to reclassify a wealthy-cluster coin down to **factor-1 / background**, in exchange for wrap eligibility (#818). The settlement fee is routed to the **redistribution lottery pool**, the same sink demurrage charges already feed (`cluster-tax/src/demurrage.rs`).

This is the sanctioned — and only — way to shed cluster provenance. Cluster-tag inheritance (#713) otherwise forbids reducing a coin's cluster mass; the settlement op carves a controlled exception whose price funds redistribution.

## Why it's consensus-level (not bridge plumbing)
It introduces new transaction semantics + a new validation rule and interacts directly with the audited cluster-tag inheritance bound (#713). It must be deterministic (pure integer/BTreeMap, like the rest of the tag path) and cannot open an inflation/laundering vector. Pre-mainnet, so it is not a live-chain fork (consistent with the no-hard-forks-before-mainnet posture), but it needs consensus-grade rigor.

## Design questions to resolve
- **Charge formula.** What does it cost to reclassify to factor-1 — the full prospective demurrage obligation, a multiple, or a settlement of accrued-to-date? Must be deterministic and not gameable by churn (cf. the existing "churning doesn't escape" invariant).
- **Tag rewrite rule.** How the output's `ClusterTagVector` is set to background, and how this composes with the #713 per-ring-max inheritance bound (the settlement is the ONE place cluster mass may legitimately drop — bound it tightly so it can't be a general laundering primitive).
- **Ring-signature interaction.** The real input is hidden; specify how the charge is computed over the ring (reuse the value-weighted centroid / max-quantile age machinery from #596/#582).
- **Fee routing.** Confirm the settlement fee flows through the standard split into the lottery pool.

## Testing (required for DoD)
- Determinism: proposer and validators reach identical verdicts (pure integer math).
- Property tests: settlement never reduces cluster mass by more than paid-for; never a net inflation of any cluster; churn-invariance preserved.
- Adversarial: attempt to launder cluster provenance cheaply, double-settle, or under-pay; all rejected.
- Integration with #825 reserve accounting (a settled-then-wrapped coin is factor-1 and peg-neutral).

[...]

PR #931, opened 2026-07-14
test(cluster-tax): demonstrate + document demurrage background-reset leak (closes #834)

### LINKED ISSUE #834: investigate: does a normal spend-to-background already leak future demurrage? (ring floor charges only accrued)
Surfaced by the #831 architect grounding (see ADR 0003). **Independent of the bridge** — flagged for a monetary-integrity review.

## Observation

Botho's demurrage anti-gaming rests on the **ring floor** (`cluster-tax/src/demurrage.rs` `ring_centroid_implied_factor` / `Ledger::ring_centroid_floored_factor`), applied **at spend time**: a spender can tag outputs "background," but the demurrage charge on that spend is floored by the wealth implied by the coin's **ring members' public tags**, which the spender cannot rewrite.

However:
- `check_cluster_tag_inheritance` (`botho/src/ledger/store.rs:116`) only rejects **inflation** (`actual > expected`; with `CONSENSUS_TAG_DECAY_RATE = 0`, `expected = input_mass`). Dropping a coin's cluster mass to background is **deflation — already legal**.
- The demurrage charged at a spend is for **`elapsed`** (age since creation), floored by the ring factor. For a **young** wealthy coin, `elapsed ≈ 0` ⇒ the ring-floored charge ≈ 0.

So the sequence "spend a young wealthy coin → produce a fully-background output" appears to pay **~zero** demurrage on that spend and yields a genuinely background-tagged coin. If future spends of that coin pick background decoys, its ring floor is now background (1×) ⇒ it escapes **future** demurrage.

## The question

Does this constitute a **future-demurrage leak** — i.e., can a wealthy holder reset a coin's demurrage class to background cheaply via an ordinary spend, escaping the stock-level charge the cluster-tilted mechanism is designed to extract?

This would appear to be in tension with the **churn-invariance** property (`demurrage.rs` header: "Total demurrage paid over any holding period is invariant to churn frequency"). Either (a) churn-invariance only holds within a wealth class and the class-transition is the leak, or (b) something already prevents/prices the transition that this analysis missed. **Needs a definitive answer.**

## Scope / method
- [ ] Confirm or refute empirically: extend the sim harness (`cluster-tax/src/simulation/`, reuse `calculate_gini`) with a whale agent that periodically spends-to-background, and measure whether total lifetime demurrage paid falls materially vs. a non-resetting whale.
- [ ] If confirmed a leak: propose a fix (e.g. a minimum ring-floor persistence / cluster-mass stickiness, or charging capitalized rather than accrued demurrage whenever cluster mass drops) and weigh it against ring-signature privacy and liveness (must not false-reject valid spends).
- [ ] If refuted: document why (the mechanism that already prices the transition) so the invariant is clearly recorded.

## Relationship to #831
**#831 does not depend on this.** The settlement flag + capitalized charge close the escape specifically for the **wrapping** on-ramp regardless of the answer here. This issue is about whether the same escape exists purely **on-chain**, which is a broader monetary-integrity question.

rjwalters:
## VERDICT: the leak is REAL (issue #834)

A wealthy holder **can** reset a coin's demurrage class to background via an **ordinary spend**, paying **~0 demurrage** and escaping **all future** stock-level demurrage. This is option (a) from the issue: churn-invariance holds only *within* a wealth class, and the **class transition is the unpriced leak**.

### The mechanism

The consensus demurrage charge (`Ledger::consensus_fee_floor`, `botho/src/ledger/store.rs`) is:
```
demurrage_charge(output_sum, max(claimed_factor, ring_centroid_implied_factor), ring_elapsed_quantile@max, rate, bpy)
```
and `charge = value × rate × (factor−1)/(max−1) × elapsed / blocks_per_year`.

The charge is **proportional to `elapsed`**. Two floors defend it — `ring_centroid_implied_factor` (B2, floors the FACTOR) and `ring_elapsed_quantile@max` (B1, defends ELAPSED against fresh-decoy dilution of an OLD real input). The leak slips **between** them: when the real input is genuinely **young**, `elapsed ≈ 0`, so the factor floor multiplies a near-zero quantity and does nothing. Spending a young wealthy coin to a background output is **legal deflation** (`check_cluster_tag_inheritance` only rejects inflation), pays 0, and yields a genuinely-background coin whose future ring floor is 1×.

### Observed charges (from the regression test)

1000-BTH coin in a 10M-BTH cluster, 2%/yr, default curve, picocredits:

| Step | factor | elapsed | charge |
|------|--------|---------|--------|
| Honest: hold 1yr, spend as wealthy | 5745 (floor) | 1yr | **18,980,000,000,000** (~18.98 BTH) |
| Spend young wealthy → background (**the leak**) | 5745 (floored, useless) | ~0 | **0** |
| Future spend of reset background coin, 1yr | 1000 (1×) | 1yr | **0** |
| **Total paid to escape the wealthy class** | | | **0** |

### Symmetry with #831 / ADR 0003

This is the exact escape #831's settlement op **prices** for the wrapping on-ramp (capitalized future demurrage routed to the lottery pool). ADR 0003 established that the only sanctioned way to shed cluster provenance is to *pay* for it. Leaving the same downgrade **free** on the pure on-chain path would make #831 decorative — which is why this question is load-bearing.

## Deliverables

- **Executable evidence**: `cluster-tax/src/demurrage.rs::tests::demurrage_background_reset_leak_is_real` — permanent regression test that walks the exact sequence and asserts every number above. It will begin failing the moment the transition is priced (the intended close signal).
- **Verdict doc**: `docs/research/demurrage-background-reset-leak.md`.
- **Header updated**: the churn-invariance scope limit is now recorded next to the claim in `demurrage.rs`, referencing the fix.

## Fix issue (MAINNET BLOCKER)

**#925** — *price the demurrage class transition on deflating spends* (`loom:triage`, `priority:high`, `loom:urgent`). Proposed mechanism: charge capitalized future demurrage when output cluster mass drops below the ring-implied input floor, mirroring #831's settlement charge.

## Testing

`cargo test -p bth-cluster-tax` → 432 + 5 passed, 0 failed (incl. the new test).

Closes #834

### DIFF
--- a/cluster-tax/src/demurrage.rs
+++ b/cluster-tax/src/demurrage.rs
@@ -31,6 +31,28 @@
 //! invariant to churn frequency; churning just adds transaction fees on
 //! top (which also feed the lottery pool).
 //!
+//! ### Scope of the churn-invariance claim: it holds WITHIN a wealth class
+//!
+//! Churn-invariance is invariance of *accrued* demurrage under re-spending at
+//! the *same* cluster factor. It does NOT price a **class transition**. The
+//! charge is proportional to `elapsed` (age-since-creation), so a coin spent
+//! while young (`elapsed ≈ 0`) pays ≈0 *regardless of factor* — the
+//! ring-centroid factor floor ([`ring_centroid_implied_factor`]) raises the
+//! factor a background-tagged output claims, but multiplies a near-zero
+//! elapsed. A holder can therefore spend a freshly-created wealthy coin to a
+//! fully background-tagged output (legal deflation under
+//! `check_cluster_tag_inheritance`, which only rejects inflation), pay ≈0, and
+//! land a genuinely-background coin whose future ring floor is 1× — escaping
+//! all FUTURE stock-level demurrage. This is a real, unpriced future-demurrage
+//! leak, demonstrated by [`tests::demurrage_background_reset_leak_is_real`] and
+//! written up in `docs/research/demurrage-background-reset-leak.md`. It mirrors
+//! exactly the escape #831's settlement op closes for the wrapping on-ramp: the
+//! settlement op prices the same class transition by charging *capitalized
+//! future* demurrage (ADR 0003), whereas an ordinary spend charges only
+//! *accrued-to-date* demurrage. The proposed on-chain fix (price the class
+//! transition = charge capitalized future demurrage on deflation) is tracked as
+//! a mainnet blocker in issue #925.
+//!
 //! ## Determinism
 //!
 //! CONSENSUS-CRITICAL: pure integer arithmetic throughout.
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review — APPROVED (verdict independently CONFIRMED)

This is a monetary-integrity investigation, so I did not rubber-stamp the verdict — I independently re-derived the mechanism from the cited code paths and re-ran the arithmetic. **The leak is real.** The demonstration is faithful to the real consensus path, not a test artifact.

### Independently verified against the code

1. **The zero-charge crux.** `demurrage_charge` (`cluster-tax/src/demurrage.rs:89`) short-circuits to `0` whenever `elapsed_blocks == 0`, *before* the factor is read. So the ring-centroid factor floor (5745×) genuinely multiplies a near-zero quantity for a young coin — confirmed, exactly as the doc claims. The floor defends the factor term; nothing defends the fact that a fresh coin has no age to charge.

2. **Deflation is legal.** `check_cluster_tag_inheritance` (`botho/src/ledger/store.rs:172`) rejects only `actual > expected + tolerance`. With `CONSENSUS_TAG_DECAY_RATE = 0` (`store.rs:66`), `expected = input_mass`, so dropping output cluster mass to background (`actual = 0`) is never rejected. The spend-to-background is permitted by design — confirmed.

3. **Age reset is standard UTXO mechanics + no maturity gate.** Every new output sets `created_at: new_height` (`store.rs:1054` coinbase, `1088` regular outputs, `1142` lottery payout). There is no lower-bound age gate on spending a fresh coin — `min_utxo_age` (`store.rs:3047`) gates *lottery eligibility only*. So Step 1 (re-spend to reset age) → Step 2 (immediately spend young→background) is genuinely reachable; the "young wealthy coin" is not blocked by any maturity rule. This closes probe #4 — the leak is **not** narrower than claimed.

4. **The test drives the REAL consensus primitives, not a reimplementation.** `demurrage_background_reset_leak_is_real` calls `ring_centroid_implied_factor` and `demurrage_charge` directly — the exact functions `Ledger::consensus_fee_floor` (`store.rs:1784`) composes via `ring_centroid_floored_factor` (`store.rs:2434`, which returns `claimed_factor.max(implied)` and delegates to `ring_centroid_implied_factor`). The one modeling shortcut — passing `elapsed = 0` for Step 2 rather than routing through `ring_elapsed_quantile@max` — is sound: the max-quantile age floor surfaces the oldest ring member, but cannot manufacture age a genuinely young real input does not have (all members young ⇒ max age ≈ 0). The B1 floor defends against fresh-decoy dilution of an *old* real input, which is orthogonal to this leak. The test's own docstring states this scope honestly.

5. **Arithmetic checks out.** I independently reproduced the honest baseline: `1000 BTH × 2% × (5745−1000)/5000 × 1yr = 18.98 BTH` = `18,980,000,000,000` pico — matches the asserted value exactly. Not a hardcoded fiction.

### Ran it myself
- `cargo test -p bth-cluster-tax --lib demurrage_background_reset_leak_is_real` → **ok** (1 passed).
- CI: `cargo build --workspace (Linux)` pass, `cargo deny` pass, `rustfmt --check` pass. Merge state CLEAN / MERGEABLE.

### Verdict-doc & fix-issue accuracy
- `docs/research/demurrage-background-reset-leak.md` correctly picks option (a), cites the right code paths, and its symmetry claim is grounded: ADR 0003 Alternative 4 literally calls the unpriced downgrade "a straight redistribution leak, not a paid transition" — verified in `docs/decisions/0003-...md`. The reasoning that #831 becomes decorative without this fix is correct.
- The churn-invariance header caveat is placed adjacent to the existing claim (`demurrage.rs:34+`), scoped precisely ("holds WITHIN a wealth class"), and points at the fix issue.
- **#925** is filed correctly as the mainnet-blocking fix (`priority:high` + `loom:urgent`), with a proposed mechanism (capitalized future demurrage on downgrade, mirroring #831) and an acceptance criterion to update this test to assert the priced charge — the intended close signal.

### Verified vs. faith
- **Verified myself:** the elapsed=0 short-circuit, the inflation-only inheritance check, `CONSENSUS_TAG_DECAY_RATE=0`, the created_at reset on all output paths, absence of a spend maturity gate, the fee-floor→floored-factor→implied-factor call chain, the honest-baseline arithmetic, the target test passing, CI green, ADR 0003 language.
- **Taken on faith (low risk):** the full 432+5 workspace test count (I confirmed the target test + CI build; local full-suite runs were reliably backgrounded by the harness but the target test and CI build both pass).

Approving. The test correctly documents a real, currently-unpriced monetary-integrity defect, and the fix is properly tracked as a mainnet blocker.

PR #936, opened 2026-07-14
docs(research): calibrate SETTLEMENT_HORIZON_BLOCKS via simulation

### LINKED ISSUE #833: monetary-policy: calibrate SETTLEMENT_HORIZON_BLOCKS via simulation (blocks #831 promotion)
**Blocks promotion of #831** (the demurrage-settlement op). Part of epic #816. Decision context: ADR 0003.

## The parameter

The settlement charge in #831 is a **capitalized** demurrage over a fixed horizon:

```
settlement_charge = demurrage_charge(settled_value, ring_floored_factor,
                                     SETTLEMENT_HORIZON_BLOCKS, rate_bps, blocks_per_year)
```

`SETTLEMENT_HORIZON_BLOCKS` sets **how much a wealthy coin pays to permanently exit demurrage by wrapping**. It is a monetary-policy magnitude, NOT a builder default:

- **Too low** → wrapping is a cheap demurrage escape; wealthy holders wrap to dodge redistribution, eroding the cluster-tilted Gini gain the whole mechanism exists to produce.
- **Too high** → wrapping wealthy value is punitively expensive; the bridge on-ramp is unusable for exactly the holders who most want DeFi liquidity.

The right value balances "wrapping isn't a redistribution loophole" against "the bridge is usable."

## This must be calibrated by simulation

Botho already has the harness (`cluster-tax/src/simulation/`):
- `constrained_analysis::parameter_sweep` — the sweep driver (starting point; the runner must be verified/extended).
- `emission_sweep.rs` / `decoy_quantile_sweep.rs` — precedent sweeps that gate on the **`Δgini > 0.05`** criterion using the shared `metrics::calculate_gini` (no second Gini impl).
- Agent-based population: `agents/{whale,retail,merchant,mixer,minter,market_maker}.rs` via `runner.rs` + `state.rs`.

### Proposed experiment
Add a `settlement_horizon_sweep` (sibling to `decoy_quantile_sweep`) that, over a range of `SETTLEMENT_HORIZON_BLOCKS` values and with a **whale agent that wraps to escape demurrage** when it is cheaper than holding:
1. **Δgini vs the no-settlement baseline** — wrapping must not erode the redistribution gain below the design floor (`Δgini > 0.05`, same criterion the demurrage kernel was validated against).
2. **Whale escape rate** — fraction of wealthy value that exits via wrapping at each horizon (should fall as the horizon rises).
3. **Churn-invariance / no-cheaper-than-holding** — settling+wrapping must not cost *less* than holding-and-paying ordinary demurrage over the same horizon (else settlement is a strictly-dominant escape). This is the key soundness gate.
4. **Lottery-pool revenue** — settlement charges routed to the pool at each horizon (the redistribution offset to escaped demurrage).

Output a table (like `format_parameter_sweep`) mapping horizon → (Δgini, escape rate, churn-invariance margin, pool revenue) so the maintainer can pick the point.

## Deliverables
- [ ] A `settlement_horizon_sweep` simulation with a demurrage-escaping whale agent, reusing `calculate_gini` and the `Δgini > 0.05` gate.
- [ ] A results table + short write-up (under `cluster-tax/src/simulation/` or `docs/design/`) recommending a `SETTLEMENT_HORIZON_BLOCKS` value with the trade-off curve.
- [ ] **Maintainer decision** on the final value, recorded in ADR 0003 (or a new ADR) — this is the gate that unblocks #831 for `loom:issue`.

## Acceptance
- A simulated horizon value where wrapping does not push Δgini below the design floor AND settlement is never cheaper than holding-and-paying (no strictly-dominant escape), with the maintainer's ratified constant.

rjwalters:
Closes #833.

Calibrates `SETTLEMENT_HORIZON_BLOCKS` — the price a wealthy coin pays to permanently exit demurrage by wrapping (ADR 0003 / #831) — via a reproducible simulation, and records the recommendation. **This is the calibration sim + recommendation; the final constant is a maintainer ratification at review** (that ratification, not this PR, unblocks #831 and #925).

## What's here

- **`cluster-tax/src/simulation/settlement_horizon_sweep.rs`** — a focused, deterministic Monte-Carlo (sibling to the shipped `decoy_quantile_sweep`), reusing the consensus `demurrage_charge` kernel, the real `ClusterFactorCurve`, and the shared `calculate_gini` (no second impl). Seeded `0xB07A_0833`.
- **`cluster-tax/src/bin/sim.rs`** — a `settlement-horizon-sweep` subcommand that regenerates every doc table. No hand-computed numbers.
- **`docs/research/settlement-horizon-calibration.md`** — the curves, the recommendation, sensitivity notes, and the shared-vs-separate-horizon verdict re #925.

## Method

Sweeps horizons of **1mo / 6mo / 1yr / 2yr / 5yr** of blocks (5s reference) × wealth-factor classes derived from the real curve, measuring:
1. **wrap cost** as % of settled value per factor class,
2. **churn-invariance margin** (the #831 soundness gate: settling must never be cheaper than holding over the horizon),
3. **dGini erosion** if the top decile wraps out vs. holds — must keep the residual Δgini above the **0.05 design floor**,
4. **lottery-pool revenue** captured, plus an **adoption sensitivity** table (25% / 50% / 100% of the decile exiting).

## Key findings

- The charge is **linear in elapsed**, so the horizon (in years) *is* the break-even holding period — the churn-invariance margin is exactly `1.0000` at every horizon × class. The soundness gate holds structurally, at any horizon.
- Under a **100% exit of the top decile**, only the **5-year** horizon keeps the residual redistribution gain above the 0.05 floor (Δgini_escape = 0.0796). Shorter horizons collapse it.
- Under **partial adoption (25/50%)**, every horizon clears the floor — the horizon is a worst-case guardrail.

## Recommendation

**`SETTLEMENT_HORIZON_BLOCKS = 31_536_000` (5 years at 5s = `5 × BLOCKS_PER_YEAR`).** Only horizon robust to a full top-decile exit; whale wrap cost 1.9%–9.5% of the coin (a usable one-time on-ramp fee, far below the cost of not wrapping). **2yr** is a defensible accessibility-leaning alternative if a simultaneous 100% exit is judged unrealistic.

## Shared vs. separate horizon (#925)

**Recommend ONE shared horizon** for both escape doors: #831 (wrap-to-factor-1) and #925 (spend-to-background) price the identical "permanently exit demurrage" transition with the identical capitalized-future-demurrage formula. Two different horizons would let the cheaper door dominate — reintroducing exactly the asymmetry #925 exists to close. Have both read the single constant. Rationale in the doc §6.

## Testing

- `cargo test -p bth-cluster-tax --features cli` green (new module: 6 tests — soundness gate, monotonic wrap cost, erosion ordering, reproducibility).
- `cargo clippy` clean on the new module.
- Reproduce the doc numbers: `cargo run -p bth-cluster-tax --features cli --bin cluster-tax-sim -- settlement-horizon-sweep`

No consensus code touched — the horizon constant lands via #831 / #925 after ratification.


rjwalters:
## Judge review: APPROVED

Reviewed as a research/sim deliverable — judging whether the simulation is sound and whether the recommendation *follows from* it (the operator ratifies the value). It does.

### Reproducibility — verified
- Ran `cargo run -p bth-cluster-tax --features cli --bin cluster-tax-sim -- settlement-horizon-sweep` on the branch (worktree at HEAD), twice: **byte-for-byte identical** (seed `0xB07A_0833`, deterministic).
- Every table in `docs/research/settlement-horizon-calibration.md` **regenerates verbatim** from the sim output — factor classes, wrap-cost table, erosion table, and adoption sensitivity all match to the last digit. No hand-computed numbers.
- The `demurrage_charge` kernel is genuinely linear in `elapsed_blocks` (confirmed against `cluster-tax/src/demurrage.rs`), and the initial/baseline Gini (0.8174) reproduces under an independent Gini implementation.

### Recommendation follows from the data — yes
The erosion sweep is computed by the sim, not asserted: at 100% top-decile exit only the **5yr** horizon keeps Δgini_escape above the 0.05 floor (0.0796), with 2yr at 0.0318 and 1yr at 0.0159. The floor crossing and the monotone ordering are real and regenerated, not cherry-picked. The 0.05 floor matches the project's ratified `Δgini > 0.05` criterion (M2 / decoy-quantile / emission sweeps) — the right precedent. Whale wrap cost 1.9%–9.5% is the sim's own output. Recommendation is well-founded; the 2yr accessibility alternative is honestly caveated.

### Soundness gate (churn-invariance) — sound, with a note
The "margin = 1.0000" column is, as coded, tautological (settlement and hold costs use identical kernel args), so numerically it verifies determinism rather than the economic property. But the doc does **not** overclaim: it correctly frames the no-strictly-dominant-escape property as the *analytic* fact that settlement charge = exactly H-years of demurrage up front, so settling is cheaper than holding iff the holder would have held < H — which is precisely why H is the break-even. The load-bearing claim is the analysis, and it is correct; the numerical column is a (weak but honest) restatement.

### Shared-vs-separate horizon (#925) — correct
The argument that one horizon must serve both #831 (wrap-to-factor-1) and #925 (spend-to-background) is sound: identical formula, identical factor input, identical lottery sink, identical economic meaning; two different prices let the cheaper door dominate — reintroducing the exact asymmetry #925 exists to close ("the fix should MIRROR #831"). Load-bearing design conclusion holds.

### Model validity
Population, escape scenario, and cost model are reasonable and conservative: wrapped-out value still counts toward final Gini as 1:1 wBTH (does not fake-inflate the erosion), per-capita redistribution over all residents, baseline == initial because no-demurrage has no pool. Coherent.

### Scope / hygiene
- Sim + doc only; no consensus code touched. New module: 6 tests pass; clippy clean on the new module.
- **Trivial fix applied by the judge**: the PR was failing `rustfmt --check (pinned nightly)` — pure mechanical formatting drift (comment re-wrap, import granularity, mod ordering) in the three PR files. Applied `cargo +nightly-2025-12-03 fmt`, confirmed **sim output byte-for-byte unchanged** and all tests still green, committed (`7069077a`) and pushed. All three CI checks now pass; merge state CLEAN.
- Linking comments confirmed posted on both #831 and #925.

Verified (not taken on faith): reproducibility + doc-number regeneration, kernel linearity, Gini baseline, the floor-crossing that drives the recommendation, CI now green, merge CLEAN. Taken as reasonable modeling judgment (not independently re-derived): the specific whale/background population sizing and the 10-year run length — standard choices consistent with the sibling sweeps.

Approving. The final `SETTLEMENT_HORIZON_BLOCKS` value remains a maintainer ratification (recorded in ADR 0003), which is what unblocks #831 and #925 for implementation.

Issue #902, opened 2026-07-14
spec epic: reconcile confidential amounts with value-dependent economics (demurrage, lottery, cluster tags)


rjwalters:
## Context (ADR 0006, decision 1)

The target spec keeps **confidential amounts** (Pedersen + Bulletproofs; fees public), but Botho's novel economics are value-dependent and the whitepaper currently does not specify how validators verify them over hidden amounts:

1. **Demurrage**: charge ∝ value × rate × factor × time-held. With a public fee and hidden value, a naive `fee = k·v` reveals `v = fee/k` exactly. Needs e.g. a committed demurrage term with an inequality proof (`C_fee − k·C_v ≥ 0` via a range proof on the difference), accepting the documented bounded leakage.
2. **Cluster-tag blending**: output tag vectors are value-weighted blends of input tags — validators can't compute the weights without the values. Needs a ZK blend-correctness proof (or a redesign of tag propagation that is value-free).
3. **Lottery**: per-block winners drawn with odds ∝ value × inverse-factor tilt. Weighted sampling over hidden values is the hard one — a Merkle-sum tree over amount commitments (proof-of-liabilities-style) would let the seed select a cumulative-range ticket and the winner prove coverage without revealing amounts; needs full specification (insertion/update costs, proof sizes, grinding interaction with the finalized-hash seed).

Also to specify: what the public fee reveals (the existing §9 metadata-leakage table needs updating once the demurrage term is committed), and an **implementation-status appendix** so auditors can see live-testnet vs target-spec at a glance (today: public amounts, classical stealth, no ML-KEM ct on outputs).

## Blockers / tooling

This is long-form normative-spec authoring with a hard requirement that claims stay consistent with the implementation — exactly the `anvil:spec` artifact class requested upstream (rjwalters/anvil — normative spec with a `code_ref` consistency audit, the mirror image of primer's `spec_ref`). Park the authoring on that request the way #881 was parked on anvil#686. The design thinking (gadget selection) can proceed before the tooling lands.

Blocks: the CT implementation epic. Not a loom:issue — needs the Anvil authoring loop + human design ratification.

rjwalters:
Upstream artifact-class request filed: rjwalters/anvil#697 (anvil:spec — normative spec with code_ref consistency audit). This epic's authoring phase is parked on it, per the #881↔anvil#686 pattern; design work on the ZK gadgets can proceed meanwhile.

rjwalters:
[...]

PR #975, opened 2026-07-15
research: CT-economics gadgets — demurrage + tag-blend over hidden amounts (#902)

rjwalters:
Design-research memo for the **two remaining** confidential-amounts economics gadgets in #902 (the lottery is already solved & ratified — Path C, value-free, no ZK sort). No consensus code changes. Informs #902; does not close the epic.

Memo: `docs/research/ct-economics-gadgets.md` (mirrors the `ct-compatible-lottery-selection.md` style).

## Gadget 1 — demurrage over hidden value: SOLVED, cheaply
The charge `demurrage = k·V` has a **public** coefficient `k` (factor, elapsed, rate, horizon are all public; only `V` is hidden), and #925's capitalized reset + #831's settlement fold into the same `k*·V`. Construction: commit `C_d`, prove `q·d − p·V ≥ 0` (charge ≥ owed) and `fee − base − d ≥ 0` (public fee covers it) — **one aggregated Bulletproof over 2 scalars, ~0.7 KB, a few ms to verify, batchable per block**. Reuses the ADR-0006 range-proof machinery; no new crypto. Value-free reframe does *not* dissolve it, but the gadget is trivial.

**Residual leak (updates §9 `tab:metadata-leak`):** factor-conditional. Factor-1 (background/commerce) coins → `k=0` → **zero** value leakage. Wealthy coins at min-fee → value revealed nearly exactly (intrinsic public-fee leak); the one-sided inequality lets a privacy-seeking prover **overpay to a coarse bucket**, capping the leak at the fee-quantization granularity.

## Gadget 2 — cluster-tag blending: LARGELY DISSOLVES
Consensus **already** does not verify the exact value-weighted blend — `validate.rs` enforces a mass **upper bound** ("output tag masses must not exceed the decayed input tag mass"); deflation is legal and *priced* (ring floor + #925). So the requested "blend-correctness ZK proof" is a problem consensus doesn't have. The inflation bound goes **value-free** (`weight_out(c) ≤ max_i weight_i(c)`, public tags only, no proof — over-permissive only in the fee-*raising* direction). The load-bearing anti-deflation floor goes value-free via an **order statistic** over ring members' implied factors (the exact age-quantile fix). This is the same value-free reframe that dissolved the lottery.

## The one irreducible core (shared by both)
The factor curve's input is **cluster wealth**, a value aggregate `Σ v·weight`. Options: **(A)** octave-bucket proof over a committed aggregate — the curve is *already* log2-structured, so ~2 range proofs place `W` in a public octave (~1–2 KB/spend); **(B)** a **value-free provenance factor** keyed on public mint/import/hop signals (generalizing ADR 0007), which dissolves even the octave proof. Recommend a calibration sim on (B) — if it holds the Δgini/Sybil bars, the last hard gadget evaporates.

## Crisp decisions surfaced
- **D1** — mandate demurrage-fee quantization (hard privacy floor) vs optional overpayment. Recommend mandate.
- **D2** — domestic factor under CT: octave-bucket proof (A) vs value-free provenance factor (B). Highest-leverage decision; recommend sim-testing B first.
- **D3** — inflation bound: value-free max-weight (default) vs exact homomorphic committed inequality.

## Bottom line
**#902 is close to buildable.** With D1–D3 ratified, none of the three sub-problems needs a Merkle-sum tree, a weighted-sampling ZK sort, or any proof system beyond the Bulletproofs ADR 0006 already commits to. Normative authoring (commitment layout, `p/q` rounding, proof aggregation grouping, octave-proof statement, `code_ref` audit) is parked on the `anvil:spec` loop, as #902 records.


rjwalters:
✅ **Approved!**

Research/docs memo — reviewed for soundness of reasoning and fidelity to the cited code, not runtime behavior. Both central claims independently verified against source; the memo is careful and appropriately hedged. Adds only `docs/research/ct-economics-gadgets.md` (+178) and a README index line; no consensus code touched; informs (does not close) #902. CI: `cargo deny` passes; merge state CLEAN.

## Gadget 1 (demurrage) — public-`k` claim HOLDS
`cluster-tax/src/demurrage.rs::demurrage_charge` computes `charge = transfer_value × rate_bps × progressivity × time_fraction / …` (lines 137–142) — genuinely linear in the hidden `V` with coefficient `k = rate·progressivity·elapsed/bpy`. `rate_bps` and `blocks_per_year` are policy constants (public). The **only** value-dependent input to `k` is `factor` (via `progressivity = factor − FACTOR_SCALE`), which traces to hidden cluster wealth — and the memo does **not** hand-wave this: §4 explicitly carves the factor out as "the one irreducible core." So "`k` public, `V` hidden" is stated with the correct caveat, not a missed dependency.
- `#925` capitalized reset (`capitalized_reset_charge` = `demurrage_charge(floor) − demurrage_charge(output)`) and `#831` settlement (`demurrage_settlement_charge`) are both reused verbatim and are linear in value, so the "fold into one `k*·V`" claim is accurate.

## Gadget 2 (tag-blend) — "dissolves" claim CONFIRMED in code
This was the load-bearing claim, so I verified it directly. `transaction/core/src/validation/validate.rs::validate_cluster_tag_inheritance` (lines 494–510) enforces **only** an upper bound:
```rust
if actual > expected + tolerance {
    return Err(TransactionValidationError::ClusterTagInflation(...));
}
```
where `expected = decayed input mass`. There is **no branch that rejects `actual < expected`** — deflation is legal. Consensus enforces `output tag mass ≤ decayed input tag mass`, **not** the exact value-weighted blend. The memo's conclusion — that the requested "blend-correctness ZK proof" is a problem consensus does not have — is correct. (`demurrage.rs` module docs at line 46 independently corroborate: "only rejects inflation".)

## Factor core & Option A — accurately characterized
`fee_curve.rs` is genuinely log2/octave-structured (`log2_fp`, `LOG2_WMID_FP`, 1× below `W_MID>>12` / 6× at `W_MID<<12`), so the "octave-bucket proof = 2 range proofs" framing (Option A) is grounded. `mempool.rs::effective_cluster_wealth_from_outputs` confirms the factor's input is a value aggregate — the honestly-stated irreducible core. D1–D3 and the recommended Option-B calibration sim are presented as open decisions, not settled facts; the residual-leakage table and §6 open-questions are candid. No overclaiming — the "close to buildable" headline is correctly gated on "with D1–D3 ratified."

## Two minor citation nits (non-blocking — for the deferred `anvil:spec` code_ref audit)
1. §1 calls `elapsed` "already value-free … CT-clean today" via `ring_elapsed_quantile`, but the value-free order statistic's own doc says "once wired" (#577) and the currently-wired demurrage `elapsed` input is the value-**weighted** `ring_elapsed_centroid`. The design point (elapsed *can* be value-free with no gadget) stands; "today" slightly overstates the current wiring.
2. §5.1 cites the error variant as `MassInflation`; the actual variant is `ClusterTagInflation` (`validation/error.rs:170`).

Neither affects any conclusion. Both are exactly the kind of thing the memo already parks on the `anvil:spec` normative loop. Solid research record.

Issue #965, opened 2026-07-15
web: v2 address QR + TS browser derivation — follow-ups from #963 (PQ address rollout)


rjwalters:
Two web-side follow-ups flagged during #963 (address-string v2, PR #964). Out of scope for the Rust rollout sub-issues; grouped here.

## 1. QR path can't fit a v2 address in one symbol
A `botho://2/` address is ~4.4k base58 chars (view+spend+ML-KEM-1184+ML-DSA-1952 ≈ 3.2 KB), well past a single QR symbol's ~2953-byte cap. Files: `web/packages/*` `ReceiveModal.tsx`/`RequestModal.tsx` (`qrcode.react`). Options: structured-append (multi-symbol QR), a compressed/hash-pointer QR + fetch, or drop QR for full addresses and QR only a short address hash / payment id. **Needs a UX decision** — note for maintainer.

## 2. TS browser derivation still emits v1
`web/packages/core/src/wallet/address.ts` still derives + encodes v1 (view+spend only). It must call the new wasm `encodeAddress`/`decodeAddress` exports (from #963's wasm-signer) AND derive the ML-KEM/ML-DSA keys in JS (or via wasm) so browser-generated addresses are v2. Until then, web-wallet addresses are incompatible with the 6.0.0 chain.

## Dependency
After the Rust rollout (#958) lands. #2 blocks web-wallet usability on 6.0.0; #1 is a UX polish that can trail.

PR #982, opened 2026-07-15
fix(web-wallet): derive SafeQR fit cap from the rendered EC level

### LINKED ISSUE #979: SafeQR: cap uses EC-level-L ceiling (2953) but default level is M (~2331)
Found during review of PR #977.

`web/packages/web-wallet/src/components/SafeQR.tsx` gates rendering on
`value.length <= SINGLE_QR_BYTE_CAP` where `SINGLE_QR_BYTE_CAP = 2953` (the
version-40 **EC-level-L** byte-mode ceiling), but the component's default
`level` is `'M'` (version-40 byte capacity ~2331). A payload in the
~2331..2953 band would pass the `fits` guard yet make `qrcode.react` throw
during render at level M/Q/H.

Not a fund-safety issue for the v2 rollout: v2 addresses are ~4.4 KB (always
fall back to the copy panel) and any short alias is well under 2331, so no
current caller lands in the danger band. Filing so the cap is made
level-aware (or clamped to the passed `level`) before a mid-size payload is
ever handed to SafeQR.

## Suggested fix
Derive the cap from the actual `level` prop (L/M/Q/H version-40 byte
capacities: 2953 / 2331 / 1663 / 1273), or pin `level='L'` when rendering
full-size payloads.

rjwalters:
## Summary
`SafeQR` (added in #965) decided whether a payload fits one QR symbol using a
fixed `SINGLE_QR_BYTE_CAP = 2953` — the version-40 **EC-level-L** byte-mode
ceiling — but the component renders at its default `level='M'`, whose version-40
byte capacity is only **2331**. A payload in the 2332..2953 band passed the
`fits` guard and was then handed to `qrcode.react`, which throws during render at
level M/Q/H. No current caller lands in that band (v2 addresses are ~4.4 KB and
always fall back; short aliases are well under 2331), so this is a latent
correctness fix, not a fund-safety issue — filed to close the gap before any
mid-size payload is handed to `SafeQR`.

## Fix (option 3 — drift-proof)
Derive the cap from the actual `level` prop so the two can never diverge:

- `QR_V40_BYTE_CAP` — version-40 byte capacities per EC level (L/M/Q/H =
  2953/2331/1663/1273).
- `singleQrByteCap(level)` — single source of truth for the fit decision; the
  component computes `cap = singleQrByteCap(level)` and gates on it.
- `SINGLE_QR_BYTE_CAP` retained (marked `@deprecated`) as the L ceiling for
  back-compat; a test pins it to L and asserts it is **not** the default-level
  (M) cap, guarding against a re-introduction of the drift.

**Invariant restored:** the fit-check cap always corresponds to the EC level the
QR is actually rendered at, so no scannable-but-wrong QR can be produced.

## Changes
- `web/packages/web-wallet/src/components/SafeQR.tsx`: level-keyed cap lookup;
  guard uses `singleQrByteCap(level)`; deprecate flat constant; updated docs.
- `web/packages/web-wallet/src/components/SafeQR.test.tsx`: new test file (10 tests).

| Fit-check cap corresponds to the rendered EC level | ✅ | `singleQrByteCap(level)` drives both the guard and the render; test asserts caps per level |
| Payload just above corrected cap → copy panel | ✅ | 2332 B (M cap + 1) and mid-band 2600 B render the `role="note"` copy panel |
| Payload just below → QR | ✅ | 2331 B (exactly M cap) renders an `<svg>` QR |
| Cap matches rendered level | ✅ | `level="L"` renders 2600 B as a QR; default M rejects it |
| Existing SafeQR behavior preserved | ✅ | empty/short-address cases still behave; full web suite green |

## Test Plan
- `pnpm test:run SafeQR` — 10/10 pass.
- `pnpm -r typecheck` — clean across all 8 web packages.
- `pnpm test:run` (full web suite, matches web CI) — 926 passed, 10 skipped.

Closes #979

### DIFF
--- a/web/packages/web-wallet/src/components/SafeQR.test.tsx
+++ b/web/packages/web-wallet/src/components/SafeQR.test.tsx
@@ -1,0 +1,99 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Tests for SafeQR's capacity-aware fit guard (#965, #979).
+ *
+ * The core invariant (#979): the fit-check cap MUST correspond to the EC level
+ * the symbol is actually rendered at. The component defaults to level 'M'
+ * (version-40 byte capacity 2331), so a payload in the 2332..2953 band — which
+ * the old level-L cap (2953) wrongly accepted — must fall back to the copy
+ * panel instead of being handed to `qrcode.react` (where it throws at level M).
+ */
+import { describe, it, expect, afterEach } from 'vitest'
+import { render, screen, cleanup } from '@testing-library/react'
+import {
+  SafeQR,
+  singleQrByteCap,
+  QR_V40_BYTE_CAP,
+  SINGLE_QR_BYTE_CAP,
+} from './SafeQR'
+
+afterEach(() => cleanup())
+
+/** The rendered QR lives inside an <svg>; the fallback is a role="note" panel. */
+function hasQr(container: HTMLElement): boolean {
+  return container.querySelector('svg') !== null
+}
+function hasCopyPanel(): boolean {
+  return screen.queryByRole('note') !== null
+}
+
+describe('singleQrByteCap — cap matches the rendered EC level', () => {
+  it('returns the version-40 byte capacity for each EC level', () => {
+    expect(singleQrByteCap('L')).toBe(2953)
+    expect(singleQrByteCap('M')).toBe(2331)
+    expect(singleQrByteCap('Q')).toBe(1663)
+    expect(singleQrByteCap('H')).toBe(1273)
+  })
+
+  it('is the single source of truth for the level table', () => {
+    for (const level of ['L', 'M', 'Q', 'H'] as const) {
+      expect(singleQrByteCap(level)).toBe(QR_V40_BYTE_CAP[level])
+    }
+  })
+
+  it('keeps the deprecated flat constant pinned to the level-L ceiling only', () => {
+    // Back-compat: SINGLE_QR_BYTE_CAP is the L ceiling and must NOT be used as
+    // the default-level (M) cap — that mismatch is the #979 drift.
+    expect(SINGLE_QR_BYTE_CAP).toBe(QR_V40_BYTE_CAP.L)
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Verified the SafeQR level/cap drift fix (closes #979).

**Correctness — drift genuinely closed:** The fit guard now computes `cap = singleQrByteCap(level)` and renders `<QRCodeSVG level={level} />` at that *same* `level`, so the cap and the rendered EC level are structurally impossible to diverge. `QR_V40_BYTE_CAP` keys the version-40 byte-mode ceilings L/M/Q/H = 2953/2331/1663/1273, and `singleQrByteCap` is the single source of truth for both the guard and the level table. No mid-size payload in the old 2332..2953 danger band can pass the guard and then throw at level M/Q/H.

**Deprecated constant handled correctly:** `SINGLE_QR_BYTE_CAP` is retained as `QR_V40_BYTE_CAP.L` (marked `@deprecated`), no longer wired into the guard — it's the L ceiling only, and a test pins it there.

**Tests are genuine boundary checks, not tautologies:**
- 2331 B (exactly M cap) → renders an `<svg>` QR
- 2332 B (M cap + 1, in the drift band) → copy panel
- 2600 B mid-band → copy panel at default M; renders a QR when `level="L"` (cap widens consistently)
- `SINGLE_QR_BYTE_CAP === QR_V40_BYTE_CAP.L` and `!== QR_V40_BYTE_CAP.M` (guards against re-introducing the drift)
- empty + short-address edge cases preserved

**No regression / scope clean:** Diff is exactly the two SafeQR files (`.tsx` + new `.test.tsx`); no stray pnpm-workspace.yaml or other artifacts. Protocol untouched.

**Verification:**
- Local: `SafeQR.test.tsx` 10/10 pass (vitest, jsdom)
- CI: `typecheck + vitest + worker dry-run` pass, `cargo deny` pass
- Merge state: CLEAN

Issue #970, opened 2026-07-15
feat(crypto)!: wallet scanner decapsulates on live path — hybrid scan/spend + input recovery — #958 sub-issue 6


rjwalters:
Sub-issue 6 of the #958 universal-PQ rollout. Builds on 1–5 (all output producers now emit ML-KEM ciphertexts + published-key addresses). This is the **receiving side**: the real scan/detect/spend path must compute the hybrid one-time key (classical DH + ML-KEM decapsulation) so a wallet actually finds and spends its funds. Without this, funds sent as hybrid outputs are undetectable.

## Scope
- `botho/src/wallet.rs::belongs_to` — compute the hybrid one-time key via `belongs_to_hybrid`/`recover_spend_key_hybrid` (from #957), decapsulating the output's `kem_ciphertext` with the wallet's derived ML-KEM secret.
- `botho-wallet/src/transaction.rs` — `scan_outputs`/`check_ownership` (~567–640); fold in the `scan_pq_outputs` logic (~717) so there's ONE hybrid scan path, retiring the separate `is_pq_output` branch (every output is hybrid now).
- **Input-spend recovery** (deferred here from sub-issue 4): making `recover_spend_key` on INPUTS hybrid-aware so a wallet can spend a hybrid UTXO it received. This is the piece sub-issue 4 explicitly left for the scanner.
- `botho-exchange-scanner` and `bridge/service/src/bth_scan.rs` scanners must not silently miss hybrid funds.
- RPC output DTO: unify `pq_ciphertext`/`is_pq_output` → the single `kem_ciphertext` field.

## Out of scope
Consensus enforcement that rejects ciphertext-less outputs (sub-issue 7 — the final cutover).

## Acceptance
- [ ] End-to-end: send (sub-issue 4) → **scanner detects** the output → **spend it** (input hybrid-recovery) → funds move. This is the full round trip that proves the rollout works.
- [ ] Minting reward + lottery payout (sub-issue 5) are detected+spendable by their owner.
- [ ] Exchange + bridge scanners detect hybrid outputs (no silently-missed funds).
- [ ] Single hybrid scan path; `is_pq_output` branch retired; RPC field unified.
- [ ] Protocol stays 6.0.0. Verify cargo test --workspace --no-run + fuzz + no-default-features.

Issue #955, opened 2026-07-15
feat(consensus): migrate lottery to value-free Path C selection (uniform + circulation window + carry-forward reward-cap) — resolves #902 lottery


rjwalters:
Implements the ratified **Path C** lottery selection (see #902 ratification comment; research + sims in PR #952 / `docs/research/ct-compatible-lottery-selection.md`). Replaces the current `ClusterWeighted` (value-reading) draw with a value-free rule that stays Sybil-safe and CT-clean.

## Design (ratified)
- **Uniform** draw from a **last-N-blocks circulation window** (N ≈ 10k blocks); replaces `SelectionMode::ClusterWeighted` default.
- **No fee floor.**
- **Reward cap** `R = min(fee_pool, ρ·base_fee)`, ρ = count of in-window outputs.
- **Carry-forward** the over-cap excess across blocks (persistent accumulator; per-block payout stays = cap).
- Winner selection remains seed-driven (finalized-hash) over the eligible window set.

## Scope / files
- `cluster-tax/src/lottery.rs` (`SelectionMode`, `draw_winners`), `botho/src/consensus/lottery.rs` (draw + carry-forward accumulator + reward cap), eligibility (window = recency, replaces/augments `min_utxo_age`).
- Reference sims: `lottery_selection_sweep`, `lottery_sybil_brake_sweep`, `lottery_reward_cap_sweep`.

## Sequencing
Rides the **CT-amounts batch** (#904 CT half), NOT the 6.0.0 reset — the value-weighted lottery is fine while amounts are public; Path C is only required once amounts are hidden. Consensus-breaking → protocol bump with that batch.

## Acceptance
- [ ] Uniform windowed draw with carry-forward reward-cap; ClusterWeighted retired as default.
- [ ] Determinism (pure integer); splitting a position yields net-zero expected lottery gain (regression test from the sim).
- [ ] Carry-forward accumulator conserves supply; per-block payout ≤ cap.
- [ ] Δgini improvement preserved vs the sim baseline.

rjwalters:
**Sequencing update:** pulled forward into the **6.0.0 reset batch** (was: CT-amounts batch). Path C is value-free — independent of confidential amounts — so it can land on 6.0.0 now and get live-validated with the rest of the batch, rather than waiting for a separate later reset. Protocol stays 6.0.0 (do NOT bump).

PR #983, opened 2026-07-15
test: scan hybrid ML-KEM coinbase outputs in tx_lifecycle_integration

### LINKED ISSUE #981: test: tx_lifecycle_integration stale-e2e failure on main (miner-balance check, pre-lottery)
Multiple recent PRs (#973, #955/#980) noted `tx_lifecycle_integration` failing **identically on clean origin/main** at a miner-balance check *before* any lottery/consensus logic — i.e. not introduced by those PRs. It's a stale end-to-end test that no longer matches current emission/fee behavior. Low priority (not a required merge-gating check), but it's recurring noise in reviews. Fix: update the e2e miner-balance expectation to current 6.0.0 emission/fee/lottery behavior, or mark it `#[ignore]` with a note if the e2e harness is stale. Surfaced repeatedly during the 6.0.0 batch.

rjwalters:
## Summary

`tx_lifecycle_integration` failed on clean `origin/main` at the pre-lottery
miner-balance assertion — recurring review noise flagged by #973 and #955/#980.
This is a **stale test-harness fix, not a real code regression**: the test's
wallet-scan and spend helpers were never updated for protocol 6.0.0's hybrid
post-quantum coinbase outputs.

## The exact assertion that failed

```
thread 'test_basic_tx_lifecycle_utxo_creation_and_consumption' panicked at
botho/tests/tx_lifecycle_integration.rs:427:
Miner should have at least 1 block reward, got 0 picocredits
```

All 14 tests in the file failed, cascading from this same root cause (miner
scanned as balance 0 -> no UTXOs -> index-out-of-bounds / "Should have N UTXOs").

## Diagnosis: stale test expectation (harness), NOT a code bug

The balance was exactly `0`, not a shifted-but-nonzero number — the wallet scan
found *nothing*, which is deeper than a stale emission figure. Root cause:

- Protocol 6.0.0 (#957/#958) made coinbase (minting) rewards **hybrid ML-KEM
  stealth outputs**. The one-time `target_key` folds in an ML-KEM-768 shared
  secret bound to `MINTING_OUTPUT_INDEX` (`MintingTx::coinbase_stealth_fields`).
- The test's `scan_wallet_utxos` still used the **classical**
  `belongs_to`/`recover_spend_key`, which cannot detect a hybrid output -> miner
  balance scanned as 0.
- After fixing detection, `create_signed_transaction` still recovered the
  spend key classically, yielding the wrong scalar -> `Invalid CLSAG signature`
  on block apply.

The **production** scan/spend paths already use the correct hybrid detectors
(`block.rs` coinbase test at `belongs_to_hybrid`/`recover_spend_key_hybrid`;
`botho-wallet` scan at `belongs_to_hybrid`). So the code's behavior — including
the miner balance it produces — is correct under current 6.0.0
emission/fee/lottery economics (at epoch 0 `lottery_emission_bps == 0`, so
`miner_payout == full reward == TEST_BLOCK_REWARD`). Only the test harness was
stale.

## Changes

- `scan_wallet_utxos`: resolve ownership + one-time key via the unified
  `belongs_to_account` / `recover_spend_key_for` dispatchers (#970), which use
  the hybrid path for ciphertext-bearing outputs (coinbases) and the classical
  path for the classical change/received outputs these tests mint. The ML-KEM
  keypair (`pq_account_key().pq_kem_keypair()`) is derived once per scan;
  the correct output index is passed (`MINTING_OUTPUT_INDEX` for coinbases).
- `create_signed_transaction`: recover the real input's one-time key via
  `recover_spend_key_for` at the UTXO's own `id.output_index`, so a spent
  coinbase produces a valid CLSAG signature.
- Both are `#[cfg(feature = "pq")]`-gated with a classical fallback so
  `--no-default-features` builds still compile.
- Import `MINTING_OUTPUT_INDEX`.

No production code touched — only `botho/tests/tx_lifecycle_integration.rs`.

## Confirmation this does not mask a regression

The fix updates only *how the test scans/spends* wallet outputs to match the
6.0.0 hybrid scheme the production node already uses. The economic quantities
the tests assert are unchanged (block reward, fee burn split, change amounts,
recipient amounts all verify against real ledger state). The miner balance the
code produces is the correct current value; there is no papered-over bug.

| Target test passes | ✅ | `cargo test --test tx_lifecycle_integration` -> 14 passed; 0 failed |
| Diagnosed stale-test vs real-bug | ✅ | Stale harness (classical scan vs hybrid 6.0.0 outputs); production paths already correct |
| Not masking a regression | ✅ | Only scan/spend helpers changed; economic assertions unchanged and green |
| Workspace compiles | ✅ | `cargo test --workspace --no-run` = OK |
| Formatted | ✅ | `cargo +nightly-2025-12-03 fmt --all` clean |

## Test Plan

```
cargo test --test tx_lifecycle_integration   # 14 passed; 0 failed
cargo test --workspace --no-run              # OK
cargo +nightly-2025-12-03 fmt --all
```

Closes #981

### DIFF
--- a/botho/tests/tx_lifecycle_integration.rs
+++ b/botho/tests/tx_lifecycle_integration.rs
@@ -25,7 +25,7 @@ use serial_test::serial;
 use tempfile::TempDir;
 
 use botho::{
-    block::{Block, BlockHeader, BlockLotterySummary, MintingTx},
+    block::{Block, BlockHeader, BlockLotterySummary, MintingTx, MINTING_OUTPUT_INDEX},
     consensus::{BlockBuilder, LotteryFeeConfig},
     ledger::Ledger,
     mempool::{Mempool, MempoolError},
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Confirmed independently: this is a genuine **stale-test-harness fix**, not a papered-over regression.

## Diagnosis verification (the load-bearing check)

**Root cause is real and correctly identified.** The test's `scan_wallet_utxos` used classical `belongs_to`/`recover_spend_key`, which return `None` for a hybrid ML-KEM coinbase output (`kem_ciphertext.is_some()`), so the miner scanned as balance 0 → no UTXOs → the `:427` assertion and cascading index panics. Not an emission drift (would have been nonzero) — a total scan miss.

**Production wallets are NOT broken — verified by reading the real paths, not the PR text:**
- `botho/src/wallet.rs:373,397` — production scan already uses `belongs_to_account` / `recover_spend_key_for`, the exact unified dispatchers this test now adopts.
- `botho/src/rpc/mod.rs:3857` — RPC wallet scans the coinbase at `MINTING_OUTPUT_INDEX` via the hybrid path.
- `botho/src/block.rs:2294` — the minter self-scan uses `belongs_to_hybrid`/`recover_spend_key_hybrid` at `MINTING_OUTPUT_INDEX`.
- `transaction/clsag/src/lib.rs:800-831` — `belongs_to_account`/`recover_spend_key_for` dispatch to the hybrid variant when a ciphertext is present, classical otherwise. The test now exercises the real production dispatch, not a bypass.

**Economic expectation is the correct 6.0.0 value, not a fabricated number.** `miner_payout = reward − lottery_emission_share`, and `MonetaryPolicy::lottery_emission_bps(0) == 0` (asserted at `cluster-tax/src/monetary.rs:762`; ramp starts at height 100). So at epoch 0 `miner_payout == full reward`. The "at least 1 block reward" assertion is right.

## Checks
- CI green — build/deny/rustfmt/fuzz-build-guard all pass; merge state CLEAN (skipping = non-applicable conditional fuzz jobs, no failures).
- Local: `cargo test --test tx_lifecycle_integration` → **14 passed; 0 failed**, incl. the formerly-failing `test_basic_tx_lifecycle_utxo_creation_and_consumption`.
- `git diff --name-only` = **only** `botho/tests/tx_lifecycle_integration.rs`. No production/consensus file touched.
- Fix is `#[cfg(feature="pq")]`-gated with a classical `--no-default-features` fallback; coinbase scanned at `MINTING_OUTPUT_INDEX`, spend recovered at the UTXO's own `id.output_index`.

Clean, well-documented, and it makes the test track the production scheme rather than masking anything. Closes #981.

PR #985, opened 2026-07-16
research: CT provenance-factor calibration — D2 for #902

### LINKED ISSUE #902: spec epic: reconcile confidential amounts with value-dependent economics (demurrage, lottery, cluster tags)

rjwalters:
## Headline

**D2 → Option B (value-free provenance factor), but ONLY with the `EpochOrigin` rule. The naive age/hop provenance factors FAIL the Sybil bar.**

Design-research + calibration sim for **decision D2** of the confidential-amounts epic (#902) — the single highest-leverage decision: under CT the factor curve's input, cluster wealth `Σ v·weight`, is the one quantity that genuinely goes hidden. Option A = octave-bucket range proof over committed wealth (~1-2 KB/spend). Option B = a value-free provenance factor from public signals (dissolves the last gadget if it holds). No consensus code change.

## Reproduce
```
cargo run -p bth-cluster-tax --features cli --bin cluster-tax-sim -- provenance-factor-sweep
```
Sim: `cluster-tax/src/simulation/provenance_factor_sweep.rs` (deterministic, seed `0xB07A_0902_D2`, byte-for-byte reproducible). Memo: `docs/research/ct-provenance-factor-calibration.md`. Reuses the real `ClusterFactorCurve`, `demurrage_charge`, `TagVector::mix` blend, and shared `calculate_gini` — no second implementation.

## The bars, per candidate

| rule | Δgini recovery (honest) | Sybil: whale fake a low factor for free? | Δgini when gamed | verdict |
|---|---|---|---|---|
| **ValueWeighted** (baseline, hidden) | 100.0% | no (split-invariant value aggregate) | +0.1017 | resistant but **reads hidden value → this is what Option A proves** |
| **Age** | 118.0% | **YES** — churn resets age → 1× (free) | **−0.0249** | **BROKEN** |
| **HopCount** | 119.5% | **YES** — self-hop wash → 1× (free) | **−0.0248** | **BROKEN** |
| **EpochOrigin** (ADR-0007 generalized) | **100.2%** | **no** — split/churn/self-hop leave it at 5.745×; only *real* circulation lowers it | **+0.1019** | **HOLDS both bars** |

- **The Sybil axis is the crux and it discriminates cleanly.** Age/HopCount look *best* on honest redistribution (the trap) but a free churn/wash collapses them to 1×, and their gamed Δgini goes **negative** (a regressive lottery). They are the factor-side restatement of Path C's proven-impossible value-free split-invariant weight.
- **`EpochOrigin` escapes the trap** because its factor keys on the **public** mint-epoch pool wealth (coinbase amounts are public), decays **only by circulation** (the real value-weighted `TagVector::mix` — self-spends with 0 incoming background value are a no-op), and is floored at `F=1.5×`. Split/churn/self-hop cannot move it (5.745× → 5.745×); only genuinely acquiring background value does (→ 1.474×, at real cost). It is also the most *faithful* value-free rule (0.238× mean factor error; exact on every wealthy tier).

## Bottom line

- **Option B dissolves the last hard CT gadget — with `EpochOrigin`.** CT-clean, leakage-neutral, 100.2% Δgini recovery, gaming-invariant redistribution, unmovable by any free attack. With the factor value-free, the CT economics reconcile with **no ZK beyond the range proofs ADR 0006 already mandates**.
- **Option A (octave-proof) does not need to stand — but remains the fallback** for a maintainer who rejects the one honest residual: **drip-mining across epochs** (accumulate a large low-factor position by minting slowly across many epochs). That residual is bounded by `K` and already throttled by emission economics — the *identical* residual ADR 0007 already ratified for the bridge boundary — but it is the conscious tradeoff Option B requires.

Recommended spec params (if Option B): domestic mint-epoch origin factor `max(F, ClusterFactorCurve(Σ public coinbase in [mK,(m+1)K)))` blended toward background by circulation; `F=1.5×`, `K=17,280` (1 day), inherited from ADR 0007.

No `Closes #902` — this surfaces D2 for ratification; the normative authoring stays on the `anvil:spec` loop. `cargo test -p bth-cluster-tax` green (7 new tests), `cargo build -p bth-cluster-tax --features cli` clean, fmt clean.


rjwalters:
✅ **Approved!** Research record informing #902 (D2). Applying `loom:pr` (no `Closes` — correctly surfaces the decision for ratification rather than closing the epic).

## Verification

**CI + determinism**
- `gh pr checks 985`: cargo build (Linux), cargo deny, rustfmt — all **pass**; `mergeStateStatus: CLEAN`.
- `cargo test -p bth-cluster-tax`: 7 new provenance tests green.
- Ran `cluster-tax-sim provenance-factor-sweep` **twice, diff-clean (byte-for-byte identical)** under seed `0xB07A_0902_D2`. Every number in the memo regenerates exactly from a fresh run (5.745×, 1.474×, 100.2% / 118.0% / 119.5%, 0.238× MAE, −0.0249 / −0.0248 gamed Δgini).

**Real-kernel reuse (no second economics)** — confirmed the sweep calls the shipped `ClusterFactorCurve::factor`, `demurrage_charge`, `TagVector::mix`, and `simulation::metrics::calculate_gini`. No reimplemented curve/blend/charge/Gini.

**The load-bearing claim — the gameability test is real, not a strawman.** I traced `apply_attack` and the factor rules independently:
- **Age collapses under the modeled attack**: `ChurnYoung` sets `age_blocks → 0`, and `age_factor(0) = 1×`. Real.
- **HopCount collapses**: `SelfHopWash` sets `hops → HOP_REF(20)`, and `hop_factor(20) = 1×`. Real.
- **EpochOrigin genuinely resists the *same* attacks** because both of its inputs are un-manufacturable for free:
  1. `mint_epoch_wealth_bth` (Σ public coinbase in the origin epoch) is an **immutable public historical fact** — no split/churn/hop can alter which epoch a coin was minted in, so the curve input stays at 10M → 5.745×.
  2. `origin_weight_ppm` decays **only** through the real `TagVector::mix`. I verified against `tag.rs:116` that `mix` with `incoming_value = 0` yields `new_weight = self_value·self_weight / self_value = self_weight` — a **genuine no-op in the shipped kernel**, not a sim assumption. A self-spend carries no background value → no decay. This is why churn/wash/split leave EpochOrigin at 5.745× → 5.745×.
- **The one attack that moves it (`RealCirculate`) drives the REAL `mix`** and only reaches the floor 1.474×, at the cost of receiving genuine background value over 9 spends. Track 3 then holds EpochOrigin at its honest factor legitimately — a consequence of the Track-2 result, not an assumption.

**Honest Δgini numbers are fair** — produced by the shared Gini kernel over the tiered population; the 118–120% age/hop over-recovery is correctly framed as *the trap* (coarseness maxing wealthy tiers to 6×), and it inverts to negative once the free attack is applied.

**No overclaiming** — the drip-mine-across-epochs residual is disclosed prominently in §5 (bounded by K/emission, the identical residual ADR 0007 already ratified), Option A is retained as the explicit fallback, and the memo notes EpochOrigin bottoms at F=1.5×, not 1×.

Minor, non-blocking: §4.3 prose ("1× value over 9 spends") slightly understates the `RealCirculate` cost — the loop actually receives ≈9× the position's value to dilute origin weight to 0.1. This strengthens the "not free" conclusion rather than weakening it, so no change requested.

Independent read on the crux: **the sim genuinely proves EpochOrigin resists the manufacture-a-low-factor attack that kills age/hop.** The whale has no free lever on either EpochOrigin input (immutable epoch-pool wealth + value-gated circulation decay backed by the real no-op-under-zero-incoming `mix`), while age/hop key on per-coin freshness the holder resets at will. The D2 → Option B / EpochOrigin recommendation rests on a sound, faithfully-modeled attack.

Issue #1000, opened 2026-07-16
consensus(hardening): a consensus-incompatible peer's height advertisement must not hold the propose-gate closed


rjwalters:
Surfaced by the #998 6.0.0-reset incident. Two stale 4.0.0 nodes advertising a higher (old-chain) height held the 6.0.0 minter's propose-gate closed — the minter withheld its proposal (`should_propose_this_round` → `!initial_sync_complete`) because a **consensus-incompatible** peer claimed a higher height, wedging block production until the zombies were firewalled off.

The node already DISCONNECTS consensus-incompatible peers (protocol major mismatch), but their height gossip / connect churn influences sync-completion + quorum reconfiguration BEFORE the disconnect. Fix: exclude consensus-incompatible peers' chain-height advertisements from the sync-complete / propose-gate determination (and ideally reject their connect before it churns quorum), so old-protocol zombies on the network can't stall a freshly-reset chain. Add a test: a minter with only incompatible peers advertising higher height still proposes + externalizes. Pre-mainnet robustness — relevant to every future reset.

Issue #968, opened 2026-07-15
feat(consensus)!: minting + lottery outputs attach ML-KEM ciphertexts (fold into MintingTx hash) — #958 sub-issue 5


rjwalters:
Sub-issue 5 of the #958 universal-PQ rollout. Builds on 1–4. The **coinbase output producers** — minting reward + lottery payout stealth outputs in `block.rs` — must also carry ML-KEM ciphertexts (the send path did transfers in sub-issue 4; this does the two consensus-produced output paths).

## Scope
- `botho/src/block.rs`: `MintingTx::new` (~231) / `to_tx_output` (~325); `LotteryPayout::to_tx_output` (~395). Both go to a `PublicAddress` that now publishes a KEM key → encapsulate + attach the 1088-byte ciphertext via the #957 hybrid path.
- `botho/src/node/minter.rs` (~703): the minter uses its OWN published address (coinbase has no external sender) — encapsulate to self so the minter can scan its reward.
- **Consensus-critical:** `MintingTx::hash()` (~283) currently hashes `target_key` only — **fold `kem_ciphertext` into the minting-tx hash** so the ciphertext is consensus-bound (matches how #831 folded settlement into the signing hash). The lottery-payout hash likewise. Must NOT desync the header↔minting-tx link (the per-height seed).

## Out of scope
Receiving scanner (6), consensus enforcement of "every output has a ciphertext" (7). Just make minting + lottery EMIT them and bind them in the hash.

## Acceptance
- [ ] Minting reward + lottery payout outputs carry a valid 1088-byte ML-KEM ciphertext derived from the destination's published KEM key.
- [ ] Minter encapsulates to its own address; can `belongs_to_hybrid`-recover its reward.
- [ ] `kem_ciphertext` is folded into `MintingTx::hash()` (and lottery payout hash); block-hash determinism preserved; header↔minting-tx not desynced.
- [ ] Protocol stays 6.0.0. Verify with `cargo test --workspace --no-run` + fuzz workspace + `--no-default-features`.

PR #1001, opened 2026-07-16
test(consensus): fresh-genesis liveness + lottery-winner + federated-minter coverage (#998)

### LINKED ISSUE #998: CRITICAL: protocol 6.0.0 chain wedges at height 5 on fresh genesis (minter re-submits forever, slot never externalizes)
**Found during the v0.6.0 / protocol-6.0.0 testnet reset (2026-07-16).** A fresh 6.0.0 genesis externalizes blocks 0→4, then **wedges permanently at height 5**. Runtime-only — no unit test caught it; the testnet reset did (working as intended, pre-mainnet).

## Symptom
- Fresh 6.0.0 genesis (all 3 fleet nodes on v0.6.0, `network=botho-testnet`) climbs to `chainHeight=4`, then stops.
- The minter (faucet) submits a NEW minting tx for `height=5` to consensus every ~0.5s — each a distinct `tx_hash`/`priority` (fresh RandomX nonce on minimal fresh-chain difficulty) — and computes a valid `Block reward: 50.000000 BTH`, but **slot 5 never externalizes**. `scpSlotPhase=NominatePrepare`, `slotStalled` eventually true, `lastExternalizedSecondsAgo` climbs unbounded.
- **Persists even SOLO**: set the minter to `[network.quorum] mode=explicit threshold=1 members=[]` (self-quorum, `Quorum satisfied: 1-of-1`) → STILL does not externalize slot 5. A 1-of-1 node is its own quorum, so this rules out multi-node/quorum-agreement as the cause.
- **Not the zombie peers**: two stale 4.0.0 nodes from the prior 5-node testnet reconnect every ~5s (correctly disconnected as consensus-incompatible, but each connect churns a quorum reconfiguration). Firewalling gossip (17100) to only the 3 fleet IPs removed the churn — slot 5 STILL doesn't externalize. So the zombies are noise, not the root cause.
- No explicit reject/invalid/panic at INFO level. The failure is silent between "Submitting minting tx" and a (never-occurring) externalize.

## Prime suspects (6.0.0 batch changes touching minting/consensus/lottery)
1. **#955 Path C lottery** — `draw_winners`/`sybil_reward_cap`/carry-forward accumulator; the lottery draws from a last-N circulation window. An empty-window / height-5 edge case in the draw or the `min(sybil_reward_cap, block_reward)` cap could make block-building or `validate_block_lottery` fail. **Top suspect** (lottery runs per block; #955 changed the draw + added the accumulator + a `draw_winners` sampling fix).
2. **#968 minting-output ML-KEM ciphertext + `MintingTx::hash()` fold** — if the coinbase's ML-KEM ciphertext isn't attached/valid, #973 enforcement (`validate_minting_tx`, `KEM_CIPHERTEXT_ENFORCED` at protocol≥6) rejects the minter's OWN block → mints forever. (But heights 1-4 externalized with the same path, so height-dependent.)
3. **#925/#831 demurrage** or emission calc at a height-dependent boundary.

Note "height 5" may be an off-by-a-few artifact (externalizes a burst then wedges); the invariant is **wedge on fresh genesis after a few blocks**.

## Repro (offline, no live infra)
Boot a local `botho-testnet` (`botho/src/bin/botho_testnet.rs`) with `--mint` on a fresh ledger and let it mint; it should wedge after a few blocks the same way. Bisect the 6.0.0 batch (#955 / #968 / #973 / #925) — likely `git revert`-test #955 first (highest suspicion). Add a regression test that mints N>5 blocks on a fresh genesis and asserts height advances.

## Impact
- **Testnet is DOWN** (wedged at height 4) pending the fix. Fleet is on v0.6.0/6.0.0.
- **Mainnet-blocking** for the 6.0.0 protocol line — must be fixed + a new reset before the testnet is usable on 6.0.0.

## Status of the live fleet
faucet (172.31.45.251, minter, now solo-quorum + gossip-firewalled), seed (172.31.42.3), seed2 (172.31.42.218) — all v0.6.0, wedged at height 4. Left running for diagnosis.

rjwalters:
Coverage tests from the #998 investigation. The height-5 "wedge" during the 6.0.0 reset was **operational** (stale 4.0.0 zombie nodes holding the minter's propose-gate closed + a degenerate `members=[]` quorum misconfig), NOT a 6.0.0 code bug — see #998. These tests **rule out the code suspects** and fill real gaps; no production code changes.

## Tests added
- `botho/tests/issue_998_fresh_genesis_liveness.rs` — mints **12 successive hybrid-ML-KEM-coinbase blocks** on a fresh 6.0.0 genesis through the real `add_block` path; asserts height advances every block. (The exact scenario the reset exercised — proves the batch doesn't wedge a fresh chain.)
- `botho/src/consensus/block_builder.rs` — `winner_drawing_block_passes_validator_gates`: the lottery **winner** path, **never previously tested** (all prior tests use `min_utxo_age=720` so nothing is ever eligible), now passes `validate_block_lottery` + `validate_lottery_output_bindings`. Real gap closed.
- `botho/src/consensus/service.rs` — `single_minter_federated_advances_past_height_5` (2-of-2, one minter, advances past height 8 once SCP timeouts fire) + `federated_recovers_from_delayed_minting_tx_registration` (SCP recovers from the gossip/validity-cache race).

## Related
- Operational resolution + root cause: #998 (testnet recovered, live on 6.0.0 at height 9+).
- Hardening follow-up: #1000 (incompatible-peer height advertisement must not hold the propose-gate).

Closes #998

### DIFF
--- a/botho/src/consensus/block_builder.rs
+++ b/botho/src/consensus/block_builder.rs
@@ -483,6 +483,154 @@ mod tests {
         }
     }
 
+    /// REPRODUCTION (issue #998): once the lottery first draws WINNERS (fleet
+    /// runs with a low `min_utxo_age`, so eligible coinbase outputs appear a
+    /// few blocks in), the block the proposer builds via `apply_lottery`
+    /// must pass the validator gates `validate_block_lottery` AND
+    /// `validate_lottery_output_bindings` — otherwise the proposer's own
+    /// `add_block` rejects its block and the chain wedges (the minter re-mints
+    /// the same height forever). No existing test exercised the winner path
+    /// (all use min_utxo_age=720, so nothing is ever eligible).
+    #[test]
+    fn winner_drawing_block_passes_validator_gates() {
+        use crate::{
+            block::BlockLotterySummary,
+            consensus::lottery::{
+                validate_block_lottery, validate_lottery_output_bindings, LotteryFeeConfig,
+            },
+            transaction::{TxOutput, UtxoId},
+        };
+        use bth_cluster_tax::{LotteryCandidate, LotteryDrawConfig, TagVector};
+        use bth_transaction_types::ClusterTagVector;
+
+        // Lottery config with a LOW age gate (the fleet's testnet setting) so the
+        // eligible-output / winner path is actually exercised.
+        let config = LotteryFeeConfig {
+            pool_fraction_permille: 800,
+            draw_config: LotteryDrawConfig {
+                min_utxo_age: 0,
+                min_utxo_value: 1,
+                winners_per_draw: 4,
+                ..Default::default()
+            },
+        };
+
+        let height = 5u64;
+        let prev_hash = [7u8; 32];
+
+        // A single eligible coinbase UTXO carrying a 6.0.0 hybrid ML-KEM
+        // ciphertext, created early so it is in-window and mature at height 5.
+        let winner_utxo_id = UtxoId::new([0xAAu8; 32], 0);
+        let winner_out = TxOutput {
+            amount: 50_000_000_000_000,
+            target_key: [0x11u8; 32],
+            public_key: [0x22u8; 32],
+            e_memo: None,
+            cluster_tags: ClusterTagVector::empty(),
+            kem_ciphertext: Some(vec![0x33u8; 1088]),
+        };
+        let winner_utxo = Utxo {
+            id: winner_utxo_id,
+            output: winner_out.clone(),
+            created_at: 0,
+        };
+        let candidate = LotteryCandidate::new(
+            winner_utxo_id.to_bytes(),
+            winner_out.amount,
+            1000,
+            &TagVector::new(),
+            0,
+        );
+        let candidates = vec![candidate];
+
+        // A carryover pool large enough that the per-block payout is nonzero, so
+        // the draw actually produces winners.
+        let stored_pool: u128 = 10_000_000_000_000;
+
+        // Minting-only block (no transfers/fees) at height 5.
+        let minting_tx = MintingTx {
+            block_height: height,
+            reward: 50_000_000_000_000,
+            minter_view_key: [1u8; 32],
+            minter_spend_key: [2u8; 32],
+            target_key: [3u8; 32],
+            public_key: [4u8; 32],
+            kem_ciphertext: Some(vec![9u8; 1088]),
+            prev_block_hash: prev_hash,
+            difficulty: u64::MAX,
+            nonce: 1,
+            timestamp: 1_000_000,
+        };
+        let block = Block {
+            header: BlockHeader {
+                version: 1,
+                prev_block_hash: prev_hash,
+                tx_root: [0u8; 32],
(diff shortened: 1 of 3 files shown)


rjwalters:
✅ **Approved!**

Pure consensus coverage for the exact scenario the 6.0.0 reset exercised — no production code changed (both `block_builder.rs` and `service.rs` additions live entirely inside their existing `#[cfg(test)] mod tests` blocks; the third file is a new integration test). All four tests are genuine and non-tautological.

## Verification

- **CI green.** All required checks pass on the latest commit: `rustfmt --check`, `cargo build --workspace (Linux)` (compiles the new tests), `cargo deny`, benchmarks, Fuzz Build Guard. Merge state `CLEAN`/`MERGEABLE`.
- **Fresh-genesis liveness** (`issue_998_fresh_genesis_liveness.rs::fresh_6_0_0_genesis_advances_past_height_6`): genuinely mines and applies **12 successive blocks** through the real `ledger.add_block` path (not a mock), asserts a real trivial-difficulty PoW (`verify_pow`) on each, and asserts `chain_state.height == height` every block plus final `> 6`. The coinbase is a real hybrid ML-KEM output — the test asserts `kem_ciphertext.is_some()`, so it exercises the #968/#969 coinbase + hash fold + lottery accounting across many blocks. Passed locally (3.18s). This would catch a future add_block-level liveness regression on a fresh 6.0.0 chain.
- **Lottery winner path** (`block_builder.rs::winner_drawing_block_passes_validator_gates`): sets `min_utxo_age=0` (the real gap — all prior tests use 720 so nothing is ever eligible), supplies a genuinely-eligible in-window coinbase candidate + a nonzero carryover pool, and **guards against a vacuous pass** by asserting `!lottery_outputs.is_empty()` before running the block through `validate_block_lottery` AND `validate_lottery_output_bindings` (#973 ML-KEM binding gate). Real winner path, previously untested. Passed.
- **Federated SCP** (`service.rs`): `single_minter_federated_advances_past_height_5` drives a real 2-of-2 quorum with a single minter across 8 heights via real `tick()`/timeout-driven pumping, asserting both nodes externalize the same value, exactly one coinbase per height, and `current_slot() > 5`. `federated_recovers_from_delayed_minting_tx_registration` reproduces the gossip/validity-cache race (registers the tx on B only after it dropped the first nominate) and asserts recovery. Both passed (18.6s / 4.9s) — timeout-bounded pumps with generous 120-iteration budgets, not trivially-passing shells.
- **Determinism:** deterministic fixtures (fixed keys, seeded mnemonic, trivial PoW). The federated tests use real wall-clock SCP timeouts — the standard harness pattern for this suite (`#593` regression sits right below) — with large retry budgets, so no flakiness observed.

## Fixed during evaluation
- `rustfmt --check` was failing on the incoming commit (import ordering: `botho_wallet` before `bth_account_keys`; one doc-comment wrap in the liveness test). Applied `rustfmt` (formatting only, no behavior change), committed, and pushed — CI is now green.

These close a real regression gap for the fresh-genesis wedge scenario. LGTM.

Issue #1017, opened 2026-07-16
bridge: parameterize DEX slippage bounds (zero-min is testnet-only) before mainnet


rjwalters:
Follow-up from PR #1014 judge review.

The testnet DeFi bootstrap scripts use zero slippage bounds — Uniswap `amountOutMinimum`/`amount*Min = 0` in `contracts/ethereum/scripts/live-defi-roundtrip.ts`, and `slippageTolerance`/no min-out in the Solana Orca scripts. Fine for fresh testnet pools (no MEV, deterministic), **unsafe for mainnet** (sandwich/frontrun).

**Acceptance:** make slippage an env-overridable knob with a safe default and an explicit mainnet TODO/guard in:
- `contracts/ethereum/scripts/live-defi-roundtrip.ts` (swap + mint LP mins)
- `contracts/solana/scripts/devnet-orca-pool.ts` + `devnet-orca-swap.ts`
Preserve current testnet behavior by default; document that mainnet must set non-zero bounds.

Part of the mainnet custody/hardening pass (alongside Squads multisigs + revoke Solana upgrade authority).

rjwalters:
Done via PR #1018 (merged). Slippage is now env-overridable — Ethereum `SLIPPAGE_BPS`/`SWAP_MIN_OUT`, Orca `SLIPPAGE_BPS` (default 1%). Testnet defaults preserve prior behavior; mainnet must set non-zero bounds (tracked for the hardening pass, #1019).

Issue #1016, opened 2026-07-16
bridge/solana CI: switch npm install -> npm ci now that a lockfile is committed


rjwalters:
Follow-up from PR #1014 judge review.

`.github/workflows/solana-contracts-ci.yml` (~lines 115-120) has a now-stale comment: 'No JS lockfile is committed under contracts/solana/, so npm ci cannot be used.' PR #1014 committed `contracts/solana/package-lock.json`.

**Acceptance:** update the stale comment and switch `npm install` -> `npm ci` for deterministic, lockfile-pinned installs. Confirm the lockfile-sync guard (~line 145) still makes sense (npm ci does not rewrite the lock). CI green.

Judge verified the committed lockfile resolves cleanly under plain install (no ERESOLVE).

rjwalters:
Done via PR #1018 (merged). Solana CI now runs `npm ci`; `package-lock.json` re-synced so it installs cleanly (verified). Stale 'no lockfile' comment removed.

Issue #1015, opened 2026-07-16
bridge/solana: Anchor.toml [programs.mainnet] holds the devnet program id


rjwalters:
Follow-up from PR #1014 judge review.

`contracts/solana/Anchor.toml` `[programs.mainnet]` currently points at the **devnet** program id `CZDnzeywrqEM5ereWJmtYKUQ9uJXxX2PydqqKTQStxxE` (a global sed during #867 set all three clusters). That is misleading — a future `anchor deploy --provider.cluster mainnet` would target the devnet keypair.

**Acceptance:** remove the `[programs.mainnet]` section (add it back with the real id at mainnet-deploy time). Keep `[programs.localnet]` (needed by `anchor test`) and `[programs.devnet]`. `anchor build`/`anchor test` still pass.

Testnet-only cleanup; no on-chain impact.

PR #1014, opened 2026-07-16
feat(bridge): wBTH testnet DeFi bootstrap on Ethereum + Solana

rjwalters:
## Summary

Drives the mainnet liquidity-launch sequence **end to end on both testnets**, against the real deployed tokens and custody — the pre-mainnet rehearsal for bootstrapping wBTH liquidity on a DEX.

### Ethereum (Sepolia) — Uniswap v3 round trip
- **Etherscan verify (#1013):** wire `ETHERSCAN_API_KEY` (V2 single key) into `hardhat.config.ts`. wBTH is [verified](https://sepolia.etherscan.io/address/0x49b985ec427ee771a601f11b18f7d4402fa2dd7b#code).
- **`scripts/live-defi-roundtrip.ts`** — mint 100k wBTH via the real **2-of-3 Gnosis Safe** `execTransaction(bridgeMint)` relayed by the role-less deployer (ADR-0002), create the wBTH/WETH Uniswap v3 pool, add liquidity, swap, and `bridgeBurn` the proceeds. A faithful ethers port of the fork-tested Rust harness, driving the already-deployed token.
- **`scripts/validate-mint-sim.ts`** — dry-runs the Safe mint against live state (sig-recovery + `eth_call`, zero spend) before broadcasting.

On-chain (Sepolia): wBTH `0x49b985ec…dd7b`, Safe `0x61274F55…3947`, Uniswap pool [`0x16C4fDbe…8D53`](https://sepolia.etherscan.io/address/0x16C4fDbe2b7497EA67f1DC8205dd2F5B31458D53).

### Solana (devnet) — Orca Whirlpool round trip
- Set the real program id (`declare_id!` + `Anchor.toml`) for the deployed program `CZDnzeywrqEM5ereWJmtYKUQ9uJXxX2PydqqKTQStxxE` (#867).
- **`scripts/devnet-bridge-setup.ts`** — `initialize` the bridge + 12-decimal wBTH SPL mint (bridge PDA authority), then `bridge_mint` 100k wBTH through the multisig-gated path (raw web3.js instructions + Anchor discriminators — no anchor CLI, which avoids the toolchain-switching).
- **`scripts/devnet-orca-pool.ts` + `devnet-orca-swap.ts`** — create the wBTH/WSOL Orca Whirlpool (#870), seed liquidity, and swap to prove it trades.
- Add `contracts/solana/.gitignore` (exclude `node_modules/`, `target/`).

[...]

PR #1018, opened 2026-07-16
chore(bridge): address PR #1014 judge follow-ups (#1015/#1016/#1017)

rjwalters:
Sweeps the three non-blocking follow-ups the judge left on #1014.

- **#1015** — drop `[programs.mainnet]` from `contracts/solana/Anchor.toml`; it held the **devnet** program id (a global sed set all three clusters), which would misdirect a mainnet `anchor deploy`. Re-add with the real id at mainnet time; `[programs.localnet]` keeps the devnet id so `anchor test` resolves.
- **#1016** — switch Solana CI JS install to `npm ci` (deterministic, lockfile-pinned) and refresh the stale "no lockfile" comment. Re-synced `package-lock.json` — the original `--legacy-peer-deps` install left it missing a transitive dep (`fastestsmallesttextencoderdecoder`), which `npm ci` rejects; verified `npm ci` now passes locally.
- **#1017** — parameterize DEX slippage. Ethereum bootstrap applies `SLIPPAGE_BPS` to the Uniswap liquidity add (`amount*Min`) and `SWAP_MIN_OUT` to the swap; Orca scripts read `SLIPPAGE_BPS` (default 1%). Testnet behavior unchanged by default; mainnet must set non-zero bounds.

All operational-tooling changes; no core protocol or on-chain impact.

### DIFF
--- a/.github/workflows/solana-contracts-ci.yml
+++ b/.github/workflows/solana-contracts-ci.yml
@@ -112,12 +112,11 @@ jobs:
         with:
           node-version: 22
 
-      # No JS lockfile is committed under contracts/solana/, so `npm ci` (which
-      # requires one) cannot be used here; `npm install` tolerates its absence.
-      # If a package-lock.json is committed later, switch to `npm ci` + npm
-      # cache for parity with contracts-ci.yml (#861).
+      # `contracts/solana/package-lock.json` is committed (#1014), so use
+      # `npm ci` for a deterministic, lockfile-pinned install that never
+      # rewrites the lock (parity with contracts-ci.yml, #861/#1016).
       - name: Install JS dependencies
-        run: npm install
+        run: npm ci
 
       # `anchor test` uses ~/.config/solana/id.json as the provider wallet
       # (Anchor.toml [provider].wallet) and funds it from the spawned validator.
--- a/contracts/ethereum/scripts/live-defi-roundtrip.ts
+++ b/contracts/ethereum/scripts/live-defi-roundtrip.ts
@@ -40,6 +40,15 @@ const WBTH_LIQ = 100_000_000_000_000_000n; // 10^17 pico = 100,000 wBTH
 const WETH_LIQ = 100_000_000_000_000_000n; // 10^17 wei  = 0.1 WETH
 const SWAP_IN = 10_000_000_000_000_000n; //   10^16 wei  = 0.01 WETH
 const FEE = 3000; // 0.30% tier
+
+// Slippage bounds. Zero is safe HERE only because this script is Sepolia-only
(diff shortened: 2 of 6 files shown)


rjwalters:
**APPROVE** — Loom Judge review

Clean sweep of the three non-blocking #1014 follow-ups. Operational-tooling only; no core protocol impact. Verified each item independently:

**#1015 — Anchor.toml mainnet section removal**
`[programs.localnet]` and `[programs.devnet]` are both retained, so `anchor test` (Localnet provider) still resolves the `wbth` program id. Dropping `[programs.mainnet]` (which held the *devnet* id) correctly removes the mainnet-deploy misdirection footgun. The `anchor build + test` CI job passes green, confirming test resolution is intact.

**#1016 — `npm ci` + lockfile resync**
Verified the committed `contracts/solana/package-lock.json` is genuinely in sync with `package.json`: `npm ci --dry-run` resolves cleanly (exit 0, 188 packages, no `ELOCKVERIFY`). The added `fastestsmallesttextencoderdecoder` peer dep and the `typescript` `dev:true` removal are the correct resync. CI's anchor job (which now runs `npm ci`) passes, so the hard-fail path is not triggered.

**#1017 — slippage parameterization**
Default testnet behavior confirmed unchanged:
- Uniswap: `SWAP_MIN_OUT` defaults to `0n` = prior `amountOutMinimum: 0n`. For the liquidity add, `minBound(desired)` at `SLIPPAGE_BPS=0` evaluates to `desired`. Reproduced the exact v3 TickMath/LiquidityAmounts math for the full-range 1:1 pool with equal 10^17 desired amounts: the mint consumes *exactly* `amount0 = amount1 = 10^17`, so `amount0Min = amount1Min = desired` is satisfied — no revert. Behavior preserved.
- Orca: `SLIPPAGE_BPS` defaults to `100` → `Percentage.fromFraction(100, 10_000)` = 1%, identical to the prior hardcoded `fromFraction(1, 100)`.

**Security:** no secrets committed. Scripts reference the gitignored `.secrets/bridge-testnet` path only; diff scan for key material is clean.

**CI:** anchor build + test ✅ · cargo deny ✅ · hardhat compile + test ✅ · mergeStateStatus CLEAN.

Nit (non-blocking, no change requested): the `// 0 = no bound` comment on `SLIPPAGE_BPS` is slightly imprecise — `0` bps is actually the *tightest* bound (`min == desired`), not "no bound." It's harmless here because the fresh-pool mint consumes exactly the desired amounts, but a future reader porting this to a non-1:1 pool should be aware that `0` will bind at full-desired. Worth a one-line clarification if the file is touched again.

Approving. 🤖 Loom Judge

PR #1022, opened 2026-07-16
fix(build): make BUILD_TIME honor SOURCE_DATE_EPOCH for reproducible builds

### LINKED ISSUE #996: ci(release): Reproducibility Check job fails on v0.6.0 (artifacts publish fine)
The release.yml **Reproducibility Check** job failed on the v0.6.0 tag (and prior tags showed similar). All platform artifacts — incl. `botho-v0.6.0-linux-aarch64.tar.gz` + checksums — publish correctly; only the byte-for-byte reproducibility meta-verification fails. Non-blocking for testnet deploys but should be diagnosed: either the build isn't bit-reproducible (SOURCE_DATE_EPOCH / CARGO_HOME / path embedding drift) or the check itself is too strict. Investigate what differs between the two reproducibility builds and fix the build determinism or relax the check.

rjwalters:
## Summary

The `release.yml` **Reproducibility Check** job failed on the v0.6.0 tag (and prior tags) even though every platform artifact published correctly. Root cause: `botho/build.rs` embedded `chrono::Utc::now()` as the `BUILD_TIME` compile-time env. The two builds the check compares (the matrix `build` job vs. the `reproducibility-check` rebuild) run at different wall-clock instants, so each baked a *different* timestamp string into the `botho` binary — different bytes, different SHA256, `MISMATCH`.

This was a genuine build-determinism defect, not an over-strict check. The other two shipped binaries (`botho-wallet`, `botho-exchange-scanner`) have no `build.rs` timestamp and were already reproducible; only `botho` mismatched.

## Changes

- **`botho/build.rs`**: `BUILD_TIME` now honors `SOURCE_DATE_EPOCH` when set — `scripts/build-release.sh` already exports it from the git commit time (`git log -1 --format=%ct`). This makes `BUILD_TIME` a pure function of the source commit. Wall-clock `Utc::now()` is retained as the fallback for ordinary developer builds. Added `cargo:rerun-if-env-changed=SOURCE_DATE_EPOCH` so the pin is respected on change.
- **`.gitignore`**: ignore the reproducible-build scratch dirs `/target-release/` and `/.cargo-home/`. `build-release.sh` creates these at the repo root; `**/target/` does not match `target-release`, so they showed as untracked, marking the tree dirty and mislabeling tagged release binaries as `<hash>-dirty` via the `git status --porcelain` check in `build.rs`.

## Root Cause

`chrono::Utc::now()` in `build.rs` is evaluated at compile time. `SOURCE_DATE_EPOCH` is the standard reproducible-builds mechanism for pinning build timestamps; the release script set it but nothing consumed it. Now `build.rs` consumes it.

| Identify what differs between the two reproducibility builds | ✅ | `BUILD_TIME` embedded via `chrono::Utc::now()` differs per build instant |
| Fix build determinism OR relax the check | ✅ | Fixed determinism (correct fix): `build.rs` now honors `SOURCE_DATE_EPOCH` |
| Build still compiles | ✅ | `cargo check -p botho` → `Finished dev profile` (34.5s) |
| BUILD_TIME deterministic under the release script | ✅ | See Test Plan below |

## Test Plan

Forced `build.rs` to rerun and inspected the emitted `BUILD_TIME` (via `cargo check -vv`):

- Run 1 — `SOURCE_DATE_EPOCH=<commit epoch>` → `BUILD_TIME=2026-07-16T16:54:41Z`
- Run 2 — same `SOURCE_DATE_EPOCH`, wall clock ~2s later → `BUILD_TIME=2026-07-16T16:54:41Z` (identical)
- Run 3 — no `SOURCE_DATE_EPOCH` (dev fallback) → `BUILD_TIME=2026-07-16T17:54:32Z` (wall clock, the pre-fix behavior)

Runs 1 and 2 prove `BUILD_TIME` is now independent of wall-clock time and equal to the git commit time under the release script, which is exactly what makes the two CI builds byte-identical.

## What remains CI-gated

I verified the determinism of the `BUILD_TIME` env at the `build.rs` level, not a full byte-for-byte SHA256 match of two complete release binaries (that requires the RandomX C++/cmake release build twice, impractical to fully reproduce in this environment). Final confirmation that the **Reproducibility Check** job goes green requires a tagged run (or a `workflow_dispatch` run of `release.yml`) on a commit containing this fix. Based on the root-cause analysis, `BUILD_TIME` was the sole non-deterministic input to the `botho` binary, so this is expected to resolve the mismatch — but it should be confirmed on the next release tag.

Closes #996

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,13 @@ junit*
 **/cargo/
 **/cmake-build-debug/
 
+# Reproducible release build scratch dirs (scripts/build-release.sh).
+# These live at the repo root and would otherwise show as untracked, marking
+# the working tree "dirty" and mislabeling tagged release binaries as
+# "<hash>-dirty" (see botho/build.rs and #996).
+/target-release/
+/.cargo-home/
+
 # Coverage reports
 coverage-report/
 
--- a/botho/build.rs
+++ b/botho/build.rs
@@ -37,8 +37,24 @@ fn main() {
         git_hash_short.to_string()
     };
 
-    // Get build timestamp
-    let build_time = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string();
+    // Get build timestamp.
+    //
+    // For reproducible release builds we must NOT embed the wall-clock time:
+    // two builds of the same commit happen at different instants, so a
+    // `chrono::Utc::now()` timestamp bakes a different string into the binary
+    // and breaks byte-for-byte reproducibility (the Reproducibility Check job
+    // in release.yml compares the original build against a rebuild — see #996).
+    //
+    // Honor SOURCE_DATE_EPOCH (a Unix timestamp) when set, which the reproducible
+    // build script exports from the git commit time. This makes BUILD_TIME a pure
+    // function of the source commit. Fall back to the current time for ordinary
+    // developer builds where SOURCE_DATE_EPOCH is unset.
+    let build_time = std::env::var("SOURCE_DATE_EPOCH")
+        .ok()
+        .and_then(|epoch| epoch.trim().parse::<i64>().ok())
+        .and_then(|secs| chrono::DateTime::<chrono::Utc>::from_timestamp(secs, 0))
+        .map(|dt| dt.format("%Y-%m-%dT%H:%M:%SZ").to_string())
+        .unwrap_or_else(|| chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string());
 
     println!("cargo:rustc-env=GIT_HASH={}", git_hash);
     println!("cargo:rustc-env=GIT_HASH_SHORT={}", git_hash_display);
@@ -47,4 +63,6 @@ fn main() {
     // Rerun if git HEAD changes
     println!("cargo:rerun-if-changed=../.git/HEAD");
     println!("cargo:rerun-if-changed=../.git/index");
+    // Rebuild when the reproducibility pin changes so BUILD_TIME stays in sync.
+    println!("cargo:rerun-if-env-changed=SOURCE_DATE_EPOCH");
 }


rjwalters:
## Judge Review: APPROVE

Reviewed the reproducibility fix for #996. This is a correct, minimal, well-targeted change. All CI green (`cargo build --workspace`, `cargo deny`, `rustfmt`, fuzz guards), mergeStateStatus CLEAN.

### 1. build.rs correctness — verified
- `SOURCE_DATE_EPOCH` is parsed as `i64` unix seconds and fed to `chrono::DateTime::<chrono::Utc>::from_timestamp(secs, 0)`, then formatted with the **same** `%Y-%m-%dT%H:%M:%SZ` layout the old `Utc::now()` path used — so the embedded string format is unchanged.
- Confirmed `DateTime::<Utc>::from_timestamp(secs: i64, nsecs: u32) -> Option<Self>` exists in the pinned **chrono 0.4.45** (registry source line 803; stabilized in 0.4.31). The fully-qualified turbofish call form is correct, and `chrono` is a `[build-dependencies]` of `botho` with the default `clock` feature, so both `from_timestamp` and the `Utc::now()` fallback compile.
- `cargo:rerun-if-env-changed=SOURCE_DATE_EPOCH` is present — necessary, since this build script already has `rerun-if-changed` directives (which otherwise pin the rerun set and would ignore env changes).
- Fallback chain is robust: unset env → `.ok()` None; malformed value → `parse` None; out-of-range epoch → `from_timestamp` None. All three land on `unwrap_or_else(|| Utc::now()...)`, preserving exact pre-fix dev-build behavior. No regression to normal `cargo build`.

### 2. Does this achieve reproducibility — well-founded
- Both the matrix `build` job and the `reproducibility-check` job invoke `scripts/build-release.sh`, which exports `SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)` from the **same** checked-out commit → identical epoch → identical BUILD_TIME → identical bytes.
- Other known non-determinism vectors are already handled by the script: `--remap-path-prefix` (Rust paths), `-ffile-prefix-map` on CFLAGS/CXXFLAGS (RandomX C++/DWARF paths, #640), `CARGO_INCREMENTAL=0`, `LC_ALL=C.UTF-8`, `TZ=UTC`. BUILD_TIME was the remaining live input for the `botho` binary.
- Verified the claim that `botho-wallet` and `botho-exchange-scanner` have **no build.rs** (neither crate contains one; `BUILD_TIME`/`env!` is only referenced by `botho`), so they were already reproducible — consistent with the observed single-binary MISMATCH.

### 3. .gitignore change — correct and safe
- `/target-release/` and `/.cargo-home/` exactly match `CARGO_TARGET_DIR`/`CARGO_HOME` set by build-release.sh at the repo root. Leading-slash anchoring scopes them to root only; `**/target/` genuinely does not match `target-release`. These are script-created ephemeral scratch dirs with no tracked contents, so nothing meaningful is masked. This also fixes the secondary `<hash>-dirty` mislabeling of tagged release binaries (the `git status --porcelain` check in build.rs).

### 4. Secrets / regressions — none
No secrets; dev-build path unchanged; change is 2 files, +27/-2.

### Caveat (accepted)
Final byte-for-byte green of the **Reproducibility Check** job requires a tagged / `workflow_dispatch` release run (the RandomX C++ build can't practically run twice in the review env). The builder was upfront about this. Given the root-cause analysis and the build.rs-level determinism proof (identical BUILD_TIME across two runs at different wall-clock instants under a fixed SOURCE_DATE_EPOCH), the fix is well-founded to merge with confirmation on the next release tag.

LGTM.

PR #1023, opened 2026-07-16
fix(consensus): ignore consensus-incompatible peers in sync/propose-gate

### LINKED ISSUE #1000: consensus(hardening): a consensus-incompatible peer's height advertisement must not hold the propose-gate closed

rjwalters:
## Summary

A consensus-incompatible peer (protocol major mismatch) is disconnected, but its chain-height advertisement can land — via a `GetStatus` round-trip or a gossiped tip — **before** the async disconnect completes. A stale old-protocol node advertising a higher (old-chain) height was then picked as the sync catch-up target, so `ChainSyncManager` never reached `Synced`, `initial_sync_complete` stayed `false`, and `should_propose_this_round` held the minter's propose-gate closed. In the #998 6.0.0-reset incident two stale 4.0.0 zombies wedged block production this way until they were firewalled off.

## Root cause

`ChainSyncManager` recorded every peer's advertised height in `peer_statuses` and `best_peer()` picked the highest. An incompatible peer's (old-chain) height therefore drove the state machine into `Downloading` toward a target it could never reach, so `is_synced()` never became `true` → `initial_sync_complete` never flipped → the propose-gate (`should_propose_this_round` → `!initial_sync_complete`) stayed closed. The existing major-mismatch disconnect fired too late: the height had already churned sync (and quorum) state.

## Changes

- `botho/src/network/sync.rs`
  - New `incompatible_peers: HashSet<PeerId>` on `ChainSyncManager`.
  - `mark_incompatible(peer)` records the peer AND scrubs any height it already contributed (routes through the disconnect teardown, resetting a `Downloading` target sourced from it) — closes the status-before-detect race. Marking persists across reconnect (protocol-major compat is a stable property of a peer identity), so a reconnecting zombie cannot re-churn sync/quorum on every connect.
  - `on_status` / `note_gossiped_tip` ignore incompatible peers' heights.
  - `best_peer()` excludes incompatible peers.
  - `tick()` Discovery/Synced branches treat incompatible peers as absent: they are never round-tripped for status, and a minter whose ONLY peers are incompatible zombies concludes initial sync via the no-compatible-peer → `Synced` path instead of waiting forever.
- `botho/src/commands/run.rs`
  - The main-loop `PeerVersionIncompatible` handler calls `sync_manager.mark_incompatible(peer)` **before** disconnecting.

| Criterion (from #1000) | Status | Verification |
|---|---|---|
| Locate sync-completion / propose-gate + incompatible-peer disconnect | ✅ | Root cause traced through `ChainSyncManager` → `is_synced()` → `set_initial_sync_complete` → `should_propose_this_round` |
| Exclude incompatible peers' height from sync-complete / propose-gate | ✅ | `on_status` / `note_gossiped_tip` / `best_peer` / `tick` all skip them |
| Reject/ignore their height before it churns quorum | ✅ | `mark_incompatible` runs on `PeerVersionIncompatible` before disconnect and scrubs already-recorded height; marking persists across reconnect |
| Regression test: minter with only incompatible higher-height peers still completes sync, proposes, externalizes | ✅ | 7 `ChainSyncManager` unit tests + 2 service integration tests wiring a real `ChainSyncManager` to the consensus gate |
| Build + relevant tests pass | ✅ | See Test Plan |

## Test Plan

- `cargo build -p botho` — clean (pre-existing warnings only).
- `cargo test -p botho --lib network::sync` → **81 passed; 0 failed**.
- `cargo test -p botho --lib consensus::service` → **37 passed; 0 failed** (incl. the two new #1000 tests).
- `cargo test -p botho --lib network::` → **721 passed** / `consensus::` → **98 passed**.
- `cargo test -p botho --test issue_998_fresh_genesis_liveness --test chain_sync_catchup_integration` → **1 + 10 passed**.
- `cargo clippy -p botho --lib` — no new warnings in touched files.

New tests: `test_incompatible_peer_status_is_ignored`, `test_mark_incompatible_scrubs_already_recorded_height`, `test_minter_with_only_incompatible_peers_reaches_synced`, `test_incompatible_peer_not_requested_in_discovery`, `test_incompatible_peer_excluded_from_best_peer`, `test_incompatible_gossiped_tip_is_ignored`, `test_incompatible_marking_persists_across_reconnect`, `incompatible_peer_height_does_not_hold_propose_gate_closed`, `minter_externalizes_after_sync_completes_despite_incompatible_peers`.

Closes #1000

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -1638,6 +1638,15 @@ async fn run_async(mut config: Config, config_path: &Path, mint: bool) -> Result
                                 local_version = %local_version,
                                 "Disconnecting consensus-incompatible peer (protocol major mismatch)"
                             );
+                            // Record the incompatibility BEFORE disconnecting so
+                            // the peer's chain-height advertisement is scrubbed
+                            // and any late in-flight status / gossip is ignored.
+                            // The disconnect below is async; without this, a
+                            // stale old-protocol node advertising a higher
+                            // (old-chain) height could hold the propose-gate
+                            // closed on a freshly-reset chain until it is
+                            // firewalled off (#998 / #1000).
+                            sync_manager.mark_incompatible(peer);
                             let _ = swarm.disconnect_peer_id(peer);
                         }
 
--- a/botho/src/consensus/service.rs
+++ b/botho/src/consensus/service.rs
@@ -2029,6 +2029,93 @@ mod tests {
         );
     }
 
+    /// Issue #1000 regression: a consensus-incompatible peer's (old-chain)
+    /// height advertisement must NOT hold the startup propose-gate closed.
+    ///
+    /// The #998 6.0.0-reset incident: two stale 4.0.0 nodes advertising a
+    /// higher old-chain height held the fresh minter's propose-gate closed
+    /// (`should_propose_this_round` -> `!initial_sync_complete`), because the
+    /// sync state machine picked the zombie's height as the catch-up target and
+    /// never reached `Synced`. This wires a REAL [`ChainSyncManager`] to the
+    /// consensus gate exactly as the run loop does and proves the gate opens.
+    #[test]
+    fn incompatible_peer_height_does_not_hold_propose_gate_closed() {
+        use crate::network::{ChainSyncManager, SyncAction};
+        use libp2p::PeerId;
+
+        // A freshly-reset minter that expects peers (min_peers == 1) booted
+        // behind the fleet, so the startup sync gate holds the propose-gate
+        // closed until catch-up completes.
+        let mut svc = solo_service();
+        svc.set_min_peers(1);
+        svc.set_connected_peers(1);
+        svc.set_initial_sync_complete(false);
+        assert!(
+            !svc.should_propose_this_round(),
+            "propose-gate must stay closed while initial sync is incomplete"
+        );
+
+        // Its ONLY peer is a consensus-incompatible zombie advertising a much
+        // higher (old-chain) height. Model the run loop: the zombie completes a
+        // status round-trip (recording its height), then is flagged
+        // incompatible and disconnected.
+        let mut sync = ChainSyncManager::new(0);
+        let zombie = PeerId::random();
+        sync.on_status(zombie, 10_000, [9u8; 32]);
+        sync.mark_incompatible(zombie);
+
+        // The sync state machine must still reach Synced — initial sync
+        // completes — even though the only peer advertises a higher height.
+        let action = sync.tick(&[zombie]);
+        assert!(
+            matches!(action, Some(SyncAction::Synced)),
+            "sync must complete despite an incompatible higher-height peer, got {action:?}"
+        );
+        assert!(sync.is_synced());
+
+        // The run loop flips the consensus gate once the sync manager is Synced.
+        svc.set_initial_sync_complete(sync.is_synced());
+        assert!(
+            svc.should_propose_this_round(),
+            "propose-gate must open once initial sync completes; an incompatible \
+             peer's height must not hold it closed (#998 / #1000)"
+        );
+    }
+
+    /// Issue #1000 regression (liveness): once initial sync completes despite
+    /// an incompatible higher-height peer, a solo-capable minter proposes
+    /// and externalizes a block.
+    #[test]
+    fn minter_externalizes_after_sync_completes_despite_incompatible_peers() {
+        use crate::network::{ChainSyncManager, SyncAction};
+        use libp2p::PeerId;
+
+        // Drive a real sync manager whose only peer is an incompatible zombie
+        // advertising a higher old-chain height; it must still complete initial
+        // sync rather than chasing the zombie's height forever.
+        let mut sync = ChainSyncManager::new(0);
+        let zombie = PeerId::random();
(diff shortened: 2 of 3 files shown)


rjwalters:
## Loom Judge review — APPROVE

Rigorous pass on this consensus-critical liveness fix (#1000 / #998 6.0.0-reset wedge). Verdict: **APPROVE**. Findings below are non-blocking.

### Correctness of the state-machine changes (sync.rs) — verified
- The exclusion is precisely scoped: peers only enter `incompatible_peers` via `mark_incompatible`, which is only called on `PeerVersionIncompatible` (protocol-major mismatch). A **compatible** peer is never added, so no real higher-height compatible peer is ever ignored.
- Discovery correctly returns `RequestStatus` for any compatible peer still lacking status **before** it reaches the decision point, so a node with a real compatible peer that is genuinely behind/pending cannot prematurely conclude `Synced`. Only when every compatible connected peer has answered (vacuously true when all peers are zombies) does it decide, and `best_peer()` returning `None` routes to the genesis/`Synced` branch. Transitions Discovery→Downloading→Synced hold up.
- The `has_connected_peer = !connected_peers.is_empty()` guard preserves the min-peers wait for a genuinely isolated node (empty set → stays Discovery, gate stays closed).

### The race — both halves handled
- `mark_incompatible` → `on_peer_disconnected` scrubs the already-recorded `peer_statuses` entry, drops overlap counts, and — critically — if we were `Downloading` **from** that peer, resets to `Discovery` and clears `pending_request`. In-flight-target reset confirmed (`test_mark_incompatible_scrubs_already_recorded_height`).
- Late in-flight `on_status` / `note_gossiped_tip` from a marked peer are ignored (early return / loop-skip).
- Marking persists (never removed from the set), so a reconnecting zombie can't re-poison the target (`test_incompatible_marking_persists_across_reconnect`).

### Does it weaken liveness/safety? — no
`should_propose_this_round` gates on three **independent** conditions; this PR only relaxes the `initial_sync_complete` sub-gate. The participation gate (`connected_peers >= min_peers`, #428) is unchanged. Confirmed the incompatible peer is removed from `discovery.peers` at detection (discovery.rs:1566), so `peer_count()` drops it and `set_connected_peers` re-closes the gate for an only-zombie node once the async `ConnectionClosed` lands. Externalization is further protected by the transitional-solo guard (#433) and by SCP itself (a 4.0.0 zombie cannot supply valid 6.0.0 SCP votes). So an only-zombie node does **not** sustainably produce blocks — the builder's claim holds. In the realistic mixed case (compatible + zombie peers) the compatible peers correctly drive both sync and quorum.

### Findings (all non-blocking)

**Low — test naming oversells externalization coverage.** `minter_externalizes_after_sync_completes_despite_incompatible_peers` uses `solo_service()` (`min_peers == 0`), so it externalizes via the genuine-solo carve-out, not the zombie-exclusion path. It does not exercise a `min_peers >= 1` node's participation-gate interaction with zombies. Consider a case that externalizes with `min_peers >= 1` and a real compatible peer alongside the zombie. Not a correctness gap — the only-zombie externalize path is *correctly* blocked by other guards.

**Low — run.rs does not refresh the participation count on detection.** The `PeerVersionIncompatible` handler calls `mark_incompatible` + `disconnect_peer_id` but not `consensus.set_connected_peers(discovery.peer_count())`. Since `discovery.peers` already dropped the zombie, the consensus participation count still includes it until the async `ConnectionClosed`. No unsafe externalization results (transitional-solo + SCP quorum cover it), but adding that refresh would close the window cleanly and matches this PR's own "act before the disconnect completes" philosophy.

**Info — permanent marking vs. in-place upgrade.** A peer that upgrades in place (same `PeerId`, now-compatible protocol) stays ignored until local restart. Deliberate tradeoff (prevents reconnect re-poisoning), acceptable given the in-memory set clears on restart and the reset context — worth the one-line doc note it already has.

**Info (pre-existing, out of scope).** `note_gossiped_tip` still attributes a gossiped height to all *compatible* connected peers; this PR strictly improves it by skipping zombies.

### Housekeeping
- No secrets; changes are idiomatic with the surrounding code (HashSet, `debug!`, matches existing teardown routing).
- CI: all checks pass/skipping, none failing; merge state CLEAN / MERGEABLE.
- New unit + integration tests exercise the wedge (higher-height zombie → still `Synced` → gate opens) and the record-before-mark race.

Approving. Nice, well-documented fix on a subtle liveness bug.

PR #1024, opened 2026-07-16
feat(bridge): provision full-loop reserve/user keys at runtime (no committed secret)

rjwalters:
## Summary

The full-loop `wrap → mint-wBTH → burn → release-BTH` e2e's BTH leg self-skips
(green) until a funded **factor-1 reserve wallet's** key files are provisioned.
This wires that provisioning **at harness runtime** — no private key is
committed to the repo — and closes the protocol-6.0.0 gap that would otherwise
stop the reserve from detecting its own outputs.

### Key design decision: no committed secret, reserve == the pre-funded miner

A committed reserve spend key is a smell, and a *freshly random* reserve owns no
on-chain outputs (so the CLSAG release has nothing to spend). Instead the driver
exports the **node's own deterministic mining wallet** as the bridge reserve:

- The reserve keys derive from the harness's in-code disposable node mnemonic
  (`generate_deterministic_mnemonic`), so they detect that node's coinbase +
  lottery outputs. Nothing new is committed.
- The reserve already owns spendable **factor-1** (zero-cluster-weight) outputs:
  lottery emission pays `LotteryOutput::to_tx_output(ClusterTagVector::empty())`
  (ADR 0003), which is exactly what `BthReleaser::prepare_release` filters for
  (`bth.rs:352`, `scanned.owned.factor_one`). Coinbase is cluster-tagged (not
  factor-1); the lottery-emission payouts are the factor-1 source.
- The **user** wallet is freshly random at runtime (it only ever receives the
  released stealth output, ADR 0004).

## Changes

- **`botho/src/bin/botho_testnet.rs`** — new `gen-bridge-keys --node N --out DIR`
  subcommand. Writes 32-byte-hex classical view/spend key files + a 64-byte-hex
  ML-KEM/ML-DSA BIP39 seed file for the reserve (node-N wallet) and a fresh user
  wallet, each 0600, and prints the `export BRIDGE_BTH_*` lines (quoted) the
  driver `eval`s. The published v2 address is derived the same way
  `bth-bridge-service`'s `ReserveKeys::public_address` does, so it matches what
  the reserve advertises. + 4 unit tests.
- **`bridge/service/src/e2e_full_loop_tests.rs`** — thread the reserve/user PQ
  seeds (`BRIDGE_BTH_{RESERVE,USER}_PQ_SEED`) into `BthConfig.pq_seed_file`.
  On protocol 6.0.0 every value output is a hybrid ML-KEM output; without the
[...]

Issue #999, opened 2026-07-16
bridge(e2e): provision a funded factor-1 reserve wallet keyfile so the full-loop BTH leg runs unattended in CI


rjwalters:
From #993 (PR #997): the full-loop wrap→mint→burn→release e2e's BTH leg **self-skips** (green) until a funded factor-1 reserve wallet's view/spend key files are provisioned to the CI environment. The wiring is complete; the loop runs the moment the keys exist. This is the 'main integration risk' the architect flagged. Provision a dedicated reserve keypair (view+spend), fund it with factor-1 reserve BTH on the local testnet the CI harness boots, and wire the key files into the `full-loop` job in bridge-e2e.yml so the BTH release leg runs for real (not self-skipped). Until then the Ethereum leg runs fully and the BTH leg is capability-proven by unit test. NOTE: gated on the 6.0.0 testnet being healthy (#998 wedge fix) since the harness boots a real botho node.

PR #1028, opened 2026-07-16
fix(bridge): unblock the full-loop BTH release leg so it runs green

### LINKED ISSUE #1025: bridge(e2e): unblock the full-loop BTH leg — driver --nodes 1 guard + verify factor-1 accrual
Follow-up from PR #1024 (#999). The reserve-key provisioning now lands, but two gaps still stand between us and the full-loop BTH leg running **green** (not self-skipped) in CI. Both flagged by the judge.

## L1 — driver requests an invalid node count
`scripts/bridge-e2e-full-loop.sh:111` calls `botho-testnet start --nodes 1`, but `cmd_start` guards `(2..=10)` (`botho/src/bin/botho_testnet.rs:223`) and exits 1. This is **pre-existing** (already on main before #1024), so #1024 didn't regress it — but the job can't reach the provisioning/test steps until it's resolved.
- **Fix options:** relax the guard to allow `--nodes 1` for the single-node harness, OR change the driver to `--nodes 2`. A single-node 6.0.0 chain must be able to boot + mine + externalize for the harness (the #1000 wedge fix — now merged — is a prerequisite; confirm a lone minter externalizes with the participation/solo guards).

## M2 — factor-1 accrual is unverified at runtime
PR #1024's "reserve == node-0 miner" rests on lottery emission being the factor-1 source. The rationale cited `botho/src/block.rs` **test** code; in production, lottery payouts inherit the winning UTXO's cluster tags (`botho/src/ledger/store.rs:1279`). So whether node-0's miner wallet actually accrues **spendable factor-1** balance sufficient for the wrap amount + fee is only determinable at runtime.
- **Acceptance:** boot the harness, mine enough to accrue factor-1 emission, and confirm the reserve can cover the wrap. If the miner wallet does NOT accrue factor-1, provision the reserve from an actual factor-1 emission output instead.

## Definition of done
The `full-loop` job in `.github/workflows/bridge-e2e.yml` runs the BTH release leg **for real** (deposit→mint→burn→release, peg invariant asserted) and passes — no self-skip. This completes what #999 set up.

Depends on: #1000 (merged, wedge fix). Related: #993/#997 (full-loop harness), #1024 (key provisioning).

rjwalters:
## Summary

Makes the `full-loop` bridge e2e (`scripts/bridge-e2e-full-loop.sh` /
`.github/workflows/bridge-e2e.yml`) run the **BTH release leg for real** —
deposit → mint-wBTH → burn → release-BTH, peg invariant asserted — instead of
self-skipping. Completes what #999/#1024 set up (#1025).

Verified end-to-end locally against the actual driver (real Botho 2-node testnet
+ local Hardhat node), not just a green CI file:

```
==> Warming the chain up to height >= 50 (decoy anonymity set)
    tip height: 50
==> Provisioning bridge reserve + user keys (runtime keygen; no secret committed)
==> Funding the reserve with a factor-1 settlement output (#1025)
Reserve settlement <tx> confirmed at block 52
==> Running the full-loop e2e against ETH=http://127.0.0.1:8545 BTH=http://127.0.0.1:27200
full loop OK: wrapped 1000000000000 pc → wBTH minted (2/2 EIP-712) → burned →
  released to a fresh stealth output (2/2 Ed25519); peg returned to 0
test result: ok. 1 passed; 0 failed; ...
==> Bridge full-loop e2e finished
```

## L1 decision — `--nodes 2` (reserve = node 0), NOT relaxing the guard

The driver called `botho-testnet start --nodes 1`, which `cmd_start` rejects
(`2..=10`). But relaxing that guard would **not** help: the testnet config
hardcodes `min_peers = 1`, so a lone node is blocked from minting/externalizing
by the #428 participation gate (`should_propose_this_round`: a peer-expecting
node must have `connected_peers >= min_peers`) — independent of the #1000
SYNC-gate fix. A single node with this config never produces a block.

So the driver now uses `--nodes 2` (reserve = node 0), which satisfies both the
participation gate (1 peer each) and the 2-of-2 SCP quorum. **Runtime-verified**:
a 2-node cluster externalizes (height 0→1→2→…); a lone node stays at 0. No
consensus safety guard was weakened. (Also dropped `--wait-consensus` from the
driver: a fresh chain's first block exceeds the harness's 30 s consensus-wait
timeout while RandomX difficulty calibrates, so the height-poll warmup is the
robust wait.)

## M2 finding — the reserve does NOT naturally accrue factor-1 (fixed)

#1024's "reserve == node-0 miner accrues factor-1" premise is **false**, and
this is only knowable at runtime. `chain_getOutputs` on a fresh chain shows every
coinbase carries `clusterTags: [[<id>, 1000000]]` — 100 % cluster weight, so
`explicit_cluster_weight() == 1_000_000` and `factor_one == false`. The miner
accrues **only** cluster-tagged coinbases; lottery *emission* is zero in the
bootstrap epoch (`lottery_emission_bps(h<halving)=0`). The releaser's `factor_one`
filter (`release/bth.rs`) would therefore find nothing to spend.

**How the reserve is now sourced:** an explicit demurrage-settlement (#831) — the
consensus-sanctioned reclassification to factor-1/background, the same on-ramp a
real wrap deposit uses. In the bootstrap epoch the capitalized settlement charge
is zero (`demurrage_rate_bps=0`), so it costs only the base fee.

- `wallet::create_settlement_transaction` — reuses the CLSAG ring builder but
  keeps outputs background and stamps `SettlementInfo` (ordinary transfers are
  unchanged).
- testnet-only RPC `dev_settleToBackground` — settles the node wallet's own
  coinbase to a factor-1 output back to itself.
- `botho-testnet fund-reserve` — calls that RPC, retrying through chain warmup,
  and waits for it to mine. The driver runs it after `gen-bridge-keys`.

**Runtime-verified**: after `fund-reserve`, `chain_getOutputs` shows 2 empty-tag
(factor-1) reserve-owned outputs, and the releaser spends them
("Prepared BTH release … spending 1 reserve input(s)").

## Additional gaps found by running the release for real (all fixed)

Once M2/L1 let the release reach `prepare_release`, four more pre-existing gaps
in the release *output* path surfaced (unreachable before, because the leg
always self-skipped):

1. **v2 recipient decode** — `decode_recipient_address` only accepted a bare
   64-byte classical address; the user's address is a v2 (ML-KEM/ML-DSA) address
   (~3.2 KB). Now uses `bth_address_codec::decode_address` (v1+v2), 64-byte path
   kept as legacy fallback.
2. **Hybrid recipient output** — pay an ADR-0004 hybrid stealth output when the
   recipient publishes an ML-KEM key (the user scans it back with its ML-KEM
   secret), classical fallback for v1.
3. **Hybrid change output** — on 6.0.0 every value output must carry an ML-KEM
   ciphertext (`KEM_CIPHERTEXT_ENFORCED`, #973) or the node rejects the tx at
   intrinsic validation. The change was classical (`TxOutput::new`) and silently
   stranded the release tx in the mempool; now a hybrid self-send encapsulated to
   the reserve's own ML-KEM key.
4. **Decoy pool filter** — `chain_getOutputs` replays every historical output,
   incl. the genesis placeholder (h0, amount 0, all-zero target key) not in the
   UTXO set; using it as a ring member made the node reject the release block
   ("ring member target_key not in UTXO set", C3). Zero-amount/all-zero-key
   outputs are now skipped as decoys.

Plus two harness-timing fixes: the release confirms only once its tx is **mined**
(a full SCP round, unlike the Hardhat-auto-mined mint), so the release
`drive_until` budget went 60→600 ticks (returns early on success — only bounds a
failure); and the node's **anonymous RPC rate limit** (100/min) is lifted on
non-production networks — the reconciler + release drive + the user's 200-block
scan-back otherwise 429 mid-loop (mainnet unchanged).

| Criterion (issue #1025) | Status | Verification |
|---|---|---|
| L1: driver requests a valid node count that actually externalizes | ✅ | `--nodes 2`; runtime-verified blocks externalize, lone node cannot (min_peers=1 gate) |
| M2: factor-1 accrual verified at runtime; reserve funded from a real factor-1 output | ✅ | Proven coinbases are 100 % cluster-tagged; reserve funded via settlement → 2 factor-1 outputs, spent by the releaser |
| Full-loop BTH release leg runs for real (no self-skip), peg asserted | ✅ | `full loop OK … peg returned to 0`; `test result: ok. 1 passed` via `./scripts/bridge-e2e-full-loop.sh` |
| No committed secrets | ✅ | All keys generated at runtime (`gen-bridge-keys` / `fund-reserve`); nothing written to the repo |

## Test Plan

- Ran the real driver `./scripts/bridge-e2e-full-loop.sh` on a clean chain: 2-node
  testnet → warmup to h50 → `gen-bridge-keys` → `fund-reserve` (settlement mined
  at block 52) → full-loop e2e test **passed** (all four assertions: peg intake,
  release to a fresh ADR-0004 stealth output the user scans back, proof-of-reserves
  drift == 0, t-of-n federation authorization).
- `cargo build --release` (node + harness) and `cargo test -p bth-bridge-service --no-run` clean.
- `cargo test -p botho --lib wallet::` — 18 passed (ordinary-transfer path unchanged).

Closes #1025

### DIFF
--- a/botho/src/bin/botho_testnet.rs
+++ b/botho/src/bin/botho_testnet.rs
@@ -1086,6 +1086,12 @@ fn start_node_process(node: &NodeState, config_path: &Path, verbose: bool) -> Re
             "run",
             "--mint",
         ])
+        // These are throwaway local harness nodes: opt in to the dev/test RPC
+        // surface (`dev_settleToBackground` for reserve funding + the lifted
+        // anonymous rate limit for the heavy e2e drive loops, #1025). A live
+        // public testnet node does NOT set this, so it keeps its 100/min
+        // anonymous limit and never exposes the mutating dev RPC (M1/L1).
+        .env("BOTHO_ENABLE_DEV_RPC", "1")
         .stdout(Stdio::from(log_file.try_clone()?))
         .stderr(Stdio::from(log_file))
         .spawn()
--- a/botho/src/bin/botho_testnet.rs
+++ b/botho/src/bin/botho_testnet.rs
@@ -145,11 +145,14 @@ enum Commands {
     /// deterministic node mnemonic (already in-code and disposable) and the
     /// user keys are freshly random at runtime.
     ///
-    /// The RESERVE is the node's own pre-funded mining wallet, so it already
-    /// owns spendable **factor-1** (zero-cluster-weight) outputs from lottery
-    /// emission (`LotteryOutput::to_tx_output(ClusterTagVector::empty())`),
-    /// which is exactly what the CLSAG release path spends (ADR 0003). The
-    /// USER wallet only receives the released stealth output (ADR 0004).
+    /// The RESERVE is the node's own pre-funded mining wallet. NOTE (#1025): a
+    /// freshly-mined node accrues ONLY 100%-cluster-tagged coinbases
+    /// (`MintingTx::to_tx_output` tags each with a new cluster) and lottery
+    /// EMISSION is zero in the bootstrap epoch — so the reserve does NOT
+    /// naturally own the **factor-1** (zero-cluster-weight) outputs the CLSAG
+    /// release path spends (ADR 0003). Run `fund-reserve` after this to settle
+    /// a coinbase into a spendable factor-1 output. The USER wallet only
+    /// receives the released stealth output (ADR 0004).
     GenBridgeKeys {
         /// Node index whose deterministic wallet becomes the bridge reserve.
         #[arg(long, default_value_t = 0)]
@@ -158,6 +161,35 @@ enum Commands {
         /// Directory to write the key files into (created if missing).
         #[arg(long)]
         out: PathBuf,
+    },
+
+    /// Fund the bridge reserve with a spendable **factor-1** (background)
+    /// output by settling the node wallet's own coins (#1025).
+    ///
+    /// A freshly-mined node accrues ONLY 100%-cluster-tagged coinbases, never
+    /// factor-1, so the bridge releaser (which spends only factor-1 reserve
+    /// outputs, ADR 0003) would find nothing. This calls the node's
+    /// `dev_settleToBackground` RPC (testnet-only) to emit an explicit
+    /// demurrage-settlement that reclassifies a coinbase to factor-1, then
+    /// waits for it to be mined — retrying while the chain is still warming up
+    /// (insufficient ring decoys / no mature coinbase yet).
+    FundReserve {
+        /// Node index whose wallet is the reserve (matches `gen-bridge-keys`).
+        #[arg(long, default_value_t = 0)]
+        node: usize,
+
+        /// Picocredits to settle to factor-1 (must cover the wrap amount +
+        /// release fee). Default 1 BTH.
+        #[arg(long, default_value_t = 1_000_000_000_000)]
+        amount: u64,
+
+        /// Base gossip port (RPC = base + 100 + node); locates the node RPC.
+        #[arg(long, default_value_t = DEFAULT_BASE_PORT)]
+        base_port: u16,
+
+        /// Max seconds to wait for the settlement to succeed and be mined.
+        #[arg(long, default_value_t = 900)]
+        timeout_secs: u64,
     },
 }
 
@@ -235,6 +267,12 @@ fn main() -> Result<()> {
         Commands::RestartNode { node } => cmd_restart_node(node, args.verbose),
         Commands::Clean => cmd_clean(args.verbose),
         Commands::GenBridgeKeys { node, out } => cmd_gen_bridge_keys(node, &out),
+        Commands::FundReserve {
+            node,
+            amount,
+            base_port,
+            timeout_secs,
+        } => cmd_fund_reserve(node, amount, base_port, timeout_secs),
     }
 }
 
@@ -832,8 +870,9 @@ fn cmd_gen_bridge_keys(node: usize, out: &Path) -> Result<()> {
         let _ = fs::set_permissions(out, fs::Permissions::from_mode(0o700));
     }
 
-    // RESERVE = the node's own deterministic (pre-funded, factor-1-accruing)
-    // wallet. Deriving from the SAME mnemonic the harness boots node `node`
+    // RESERVE = the node's own deterministic (pre-funded) wallet. It accrues
+    // cluster-tagged coinbases; `fund-reserve` settles one into factor-1.
+    // Deriving from the SAME mnemonic the harness boots node `node`
     // with means the reserve keys detect that node's coinbase + lottery
     // outputs — no secret is introduced.
     let reserve_mnemonic = generate_deterministic_mnemonic(node)?;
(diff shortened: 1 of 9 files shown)


rjwalters:
## Loom Judge review — REQUEST CHANGES

Thorough pass on this consensus-adjacent PR (new RPC + wallet value handling + release path). The security-critical design is sound, but **CI is red on a real code defect**, so this cannot be approved yet.

### Blocker (must fix)

**B1 — clippy `len_zero` fails `cargo test (bridge crates)` (`-D warnings`).**
`bridge/service/src/release/bth.rs:402`
```rust
out.amount > 0 && out.target_key.trim_matches('0').len() != 0
```
Clippy rejects `.len() != 0`. Apply its suggestion:
```rust
out.amount > 0 && !out.target_key.trim_matches('0').is_empty()
```
This is why the bridge job is red; it also means the local `cargo test -p bth-bridge-service --no-run` in the Test Plan did not run clippy. Please run `cargo clippy -p bth-bridge-service --all-targets -- -D warnings` before re-requesting.

**B2 — `cargo build --workspace (Linux)` failed with `No space left on device`.**
This is a CI-runner disk-full (infra), not a code error — but the check must be green before merge. Re-run once B1 is pushed and confirm it passes.

### Security verification (all PASS — recorded for the audit trail)

1. **Mainnet firewall — PASS.** `Network` is a binary enum `{Mainnet, Testnet}`. The RPC guard `state.network_type != Network::Testnet` and the rate-limit lift `!config.network_type.is_production()` (`is_production == matches!(Mainnet)`) both reduce to *Testnet-only*. `dev_settleToBackground` returns `-32000` on mainnet; mainnet's 100/min anonymous limit is unchanged. No path exposes the mutating RPC or unlimited RPC on mainnet.
2. **`create_settlement_transaction` — PASS.** It only reclassifies: forces empty `ClusterTagVector`, sets `settled_value = Σ output.amount`, stamps `new_settlement`. Consensus `Ledger::verify_settlement` enforces exactly that (all-background outputs + certified value == output sum), and the fee floor (#925/C7) prices the *capitalized* future demurrage on any genuine downgrade — so this is **not** a cheap demurrage-escape on mainnet; it is zero-cost only in the bootstrap epoch (`demurrage_rate_bps == 0`). Value conservation stays with the Pedersen/CLSAG layer — no minting. The only caller is the testnet-gated RPC.
3. **No committed secrets — PASS.** All key material is runtime-generated (`gen-bridge-keys` / `fund-reserve`); nothing written to the repo.
4. **Hybrid outputs / decoy filter — PASS (modulo B1).** Change value is passed through correctly (`change_output(change)`); the change is a hybrid self-send encapsulated to the reserve's own ML-KEM key (satisfies `KEM_CIPHERTEXT_ENFORCED` 6.0.0). The decoy filter correctly excludes the genesis placeholder: node emits `targetKey` via `hex::encode` (bare hex, no `0x`), so an all-zero key trims to `""`. No privacy weakening (only a non-UTXO placeholder is dropped).
5. **Budget / driver changes — PASS.** Confirmation budget 60→600 is test-only and returns early on success; `--nodes 2` is driver-only.
6. **Test meaningfulness — PASS.** `drive_until(..Released..)` now exercises the real release leg and asserts the peg returns to 0.

### Non-blocking findings (recommend addressing)

**M1 (Medium, testnet hardening) — the rate-limit lift is scoped to the whole Testnet *network type*, not to localhost.** The live public testnet (multiple internet-facing nodes) would drop from 100/min anonymous to effectively unlimited (`KeyTier::Custom(1_000_000)`). Combined with the unauthenticated, CPU-heavy `dev_settleToBackground` (full UTXO scan + CLSAG ring build), that is a DoS-amplification vector on live testnet nodes. The comment's rationale ("pointless for a localhost dev/testnet node") argues for a **bind-address / localhost** or dedicated dev-flag gate rather than a network-type gate. Mainnet is unaffected, so this is not a merge blocker, but please consider tightening.

**L1 (Low) — `dev_settleToBackground` is unauthenticated and mutating on every Testnet node** (self-send only, disposable funds), so it is reachable on the live testnet, not just the harness. Acceptable given testnet semantics; noting for awareness alongside M1.

Fix B1, get B2 green, and I'll re-review promptly. Nice work running the leg for real and disproving the #1024 factor-1 premise at runtime.

rjwalters:
## Doctor — review feedback addressed

Thanks for the thorough pass. B1 fixed, and I also tightened M1/L1 since these RPCs run on internet-facing live testnet nodes. Two commits pushed (`6e8fdb61`, plus a rustfmt reflow `39bda7ed`).

### B1 — clippy `len_zero` (blocker) — FIXED
`bridge/service/src/release/bth.rs:402` now reads:
```rust
out.amount > 0 && !out.target_key.trim_matches('0').is_empty()
```
Semantics identical (keep outputs whose `target_key` is not all-zero-hex, dropping the genesis placeholder decoy). The exact CI command is green locally:
```
cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets -- -D warnings   # clean
```

### M1 / L1 — dev RPC + rate-limit lift no longer scoped to the whole Testnet network type
You were right that gating on the *network type* leaves every public testnet node exposed. Both are now additionally gated behind an explicit dev-RPC opt-in, following the repo's existing `BOTHO_ENABLE_MAINNET` env-flag idiom (new `config::is_dev_rpc_enabled()`, reads `BOTHO_ENABLE_DEV_RPC`, off by default):

- **`dev_settleToBackground`** (`botho/src/rpc/mod.rs`): now `state.network_type != Network::Testnet || !crate::config::is_dev_rpc_enabled()` → error. So it is unreachable on mainnet *and* on a default public testnet node; only a node with the flag set exposes it.
- **Anonymous rate-limit lift** (`botho/src/commands/run.rs`): now `!is_production() && is_dev_rpc_enabled()`. A default public testnet node keeps its 100/min anonymous limit; only a flagged node lifts it.
- **Harness wiring** (`botho/src/bin/botho_testnet.rs`): the `botho-testnet` harness sets `BOTHO_ENABLE_DEV_RPC=1` on the throwaway nodes it spawns, so the full-loop e2e still funds the reserve and drives the heavy loops without 429s.

Net: mainnet behavior unchanged (still firewalled by network type); a live public testnet node no longer drops its rate limit or exposes the mutating self-send RPC by default; the local harness is unaffected.

### Verification
- `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets -- -D warnings` — clean.
- `cargo check -p botho --bins` — clean (both `botho` + `botho-testnet` build).
- `cargo +nightly-2025-12-03 fmt --all --check` — clean (after the reflow commit).
- **Full-loop harness re-run WITH the new gating** — `./scripts/bridge-e2e-full-loop.sh` on a fresh 2-node chain:
  - `Reserve settlement … confirmed at block 52` → `dev_settleToBackground` correctly reachable *because the harness set the flag* (proves the gate lets the harness through).
  - `full loop OK: wrapped 1000000000000 pc → wBTH minted (2/2 EIP-712) → burned → released to a fresh stealth output (2/2 Ed25519); peg returned to 0`
  - `test result: ok. 1 passed; 0 failed`

### B2 (disk-full runner)
Infra, not code — this push re-triggers the workspace build; it should pass now that the runner has space. Flagging for a re-run rather than a code change.

rjwalters:
## Re-review verdict: APPROVE (delta only)

Focused re-review of the two loom-doctor commits (`6e8fdb61` + `39bda7ed`) addressing the prior blockers. I did not re-review the whole PR — only the delta.

### B1 — clippy `len_zero` fix (bridge/service/src/release/bth.rs)
`.trim_matches('0').len() != 0` → `!...trim_matches('0').is_empty()`. `len() != 0` and `!is_empty()` are exactly equivalent for a `&str`. Semantics unchanged: `is_spendable_decoy` still keeps outputs whose `target_key` is not all-zero-hex (dropping the genesis placeholder decoy). ✓

### M1/L1 — dev-RPC / rate-limit gate (the important part)
- **Off by default:** `is_dev_rpc_enabled()` mirrors the existing `is_mainnet_enabled()` idiom exactly — unset env → `Err` → `unwrap_or(false)` → false. Only `"1"` or (case-insensitive) `"true"` parse truthy; `"0"`/`"false"`/`""` all → false. No truthy-parse footgun. ✓
- **Dev RPC** (`rpc/mod.rs` `handle_dev_settle_to_background`): errors unless `network_type == Testnet` **AND** `is_dev_rpc_enabled()`. Mainnet → `!= Testnet` → hard error regardless of the flag. ✓
- **Rate-limit lift** (`commands/run.rs`): `!network_type.is_production() && is_dev_rpc_enabled()`. `Network` has only `Mainnet`/`Testnet`, so `!is_production()` == Testnet; mainnet is `is_production()` → no lift regardless of flag. A public testnet node with the flag unset keeps its 100/min anonymous limit. ✓
- **No residual exposure path:** a live public testnet node (flag unset) neither exposes `dev_settleToBackground` nor drops its rate limit. Mainnet is hard-off on both gates independent of the flag. ✓

### Harness wiring (bin/botho_testnet.rs)
`.env("BOTHO_ENABLE_DEV_RPC", "1")` is set on the per-child `Command` in `start_node_process` — scoped to the throwaway spawned node processes only. It does not mutate the parent env and does not persist to any operator-run node. The e2e therefore still exercises the dev RPC + lifted limit inside the harness. ✓

### CI
All required checks green. The earlier `cargo build --workspace (Linux)` red was the disk-full infra flake (B2); the re-run passes. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

### Non-blocking nit (future-proofing only)
The dev RPC uses strict `!= Network::Testnet` while the rate-limit uses `!is_production()`. Identical today (two variants); if a future `Devnet`/`Local` variant is ever added they'd diverge. Not a security regression (both still require the opt-in flag) — noting only.

Delta is sound. Approving.

PR #1041, opened 2026-07-16
feat(bridge): add isolated public order API for wallet export/unwrap

rjwalters:
## Summary

Implements the #1036 keystone that makes the already-merged, feature-flagged-inert wallet export (#1031/#1035) and unwrap (#1032/#1040) panels functional: a **separate, browser-reachable order API** for user-initiated mint/release orders — bound on its **own listener**, isolated from the operational kill-switch surface.

Part of epic #1029.

## Security design (the whole point)

**Two routers, two binds.** The existing ops router (`bridge/service/src/api.rs`) co-hosts `POST /api/breaker` — an unauthenticated pause/RESUME kill switch — plus `/api/status`, `/metrics`, `/api/reserve/proof`. It stays loopback-only and **unchanged**. The new public surface (`bridge/service/src/public_api.rs`) is a wholly separate `public_router()` on a separate `TcpListener` (`public_api.listen`), serving **only** user-facing order endpoints. There is no code path from the public router to the breaker/ops control, and a test pins it (`test_public_router_does_not_serve_ops_endpoints` asserts 404 for `/api/breaker`, `/api/status`, `/metrics`, `/api/reserve/proof`).

**CORS.** Exact-match origin allow-list (`public_api.cors_allowed_origins`) — no wildcards. Unlisted origins get no `Access-Control-Allow-Origin`; `OPTIONS` preflight is answered directly; `Vary: Origin` is always set. Hand-rolled middleware — no new dependency/advisory surface.

**Rate limiting + anti-spam.** Per-IP fixed-window counter (via `ConnectInfo`) with a tight cap on order-create specifically (`create_rate_limit_per_min`, default 10/min) plus a general cap (default 120/min), a request body-size cap (`max_body_bytes`, default 8 KiB), and strict input validation (chain enum, per-chain address format via `ChainAddress::validate`, u64 amount parse, per-order cap, amount-above-fee). Order-create only ever produces an `AwaitingDeposit` order; no value moves until a real BTH deposit is confirmed and the engine's daily caps + circuit breaker still gate settlement.

Threat model is documented in the module header and `PublicApiSettings` doc-comments (incl. the trusted-reverse-proxy caveat for per-IP limits).

## `failed`-status normalization

[...]

Issue #1036, opened 2026-07-16
bridge(service): expose a public, CORS'd, rate-limited order API for the wallet export flow (Tier 1 backend)


rjwalters:
Fast-follow from #1031 (PR #1035). The Tier 1 export UI is built + feature-flagged but INERT until a browser-reachable bridge order API exists.

## The problem the builder found
`bridge/service/src/api.rs` today exposes only ops endpoints (`/health`, `/metrics`, `/api/status`, `/api/breaker`, `/api/reserve/proof`), is **loopback-only by design**, and critically **co-hosts `POST /api/breaker` — an unauthenticated kill switch**. Publishing that router as-is would expose the kill switch. So Tier 1 does NOT consume it directly.

## What's needed
A SEPARATE, security-reviewed public surface for user-initiated mint orders — do NOT just CORS-enable the existing ops router:
- `POST /api/bridge/orders` (create mint order: destination chain + user EVM/SVM address + amount → returns deposit address + memo/order UUID)
- `GET /api/bridge/orders/{id}` (status → `AwaitingDeposit → DepositDetected → DepositConfirmed → MintPending → Completed`, strings matching Rust `OrderStatus::Display`).
- On a distinct bind/router from `/api/breaker` (the kill switch must NOT be reachable from the public surface), with CORS for the web origin + rate limiting + abuse protections.
- Matches the typed client contract already written in `web/packages/features/src/bridge/bridge-client.ts`.

## Acceptance
The web export flow works end-to-end against a running bridge service: create order → deposit → track to Completed, with `VITE_BRIDGE_API_BASE` pointed at the public endpoint. The breaker kill switch is NOT exposed on that surface. Security-reviewed before it goes live.

Blocks: the Tier 1 export flow (#1031) becoming functional. Part of epic #1029.

rjwalters:
Contract note (from the #1035 judge, folding in dup #1038): Rust `OrderStatus::Display` emits `failed: {reason}` and the enum serializes as a **snake_case struct variant**, while the TS client type (`bridge-client.ts`) expects a bare `status: "failed"` + separate `failureReason`. Reconcile the serde representation with the TS contract when building this endpoint (either flatten on the Rust side or adapt the TS decoder).

Issue #1029, opened 2026-07-16
epic: wBTH bridge + trading in the web wallet (on-ramp, not a multichain wallet)


rjwalters:
Make the Botho web wallet the **on-ramp** between native BTH and the wBTH DeFi venues we've bootstrapped — WITHOUT turning it into a multichain wallet.

## Guiding principle (custody model)
The web wallet is a Botho-only privacy wallet: client-side WASM ring-sig signing, encrypted vault, talks only to Botho nodes (`@botho/adapters`). It has NO EVM/Solana custody and will not gain any. So its honest role is:
- **Export (BTH→wBTH):** the wallet *constructs + signs + submits* the BTH deposit (its wheelhouse) and tracks the mint order. wBTH lands in the user's OWN EVM/SVM wallet (MetaMask/Phantom/Hyperliquid).
- **Unwrap (wBTH→BTH):** the wallet provides the Botho *release destination* + guides the user to burn wBTH in their own counterparty wallet, then tracks the release and receives the BTH. NO EVM signing in the Botho wallet.
- **Discover:** route users to OUR seeded pairs (Uniswap/Orca/Hyperliquid) via deep-links; we do not embed a DEX.

We explicitly do NOT build: EVM/Solana wallet-connect, in-wallet DEX swaps, or counterparty-chain tx signing.

## Deployed venues to surface (testnet)
- Ethereum Sepolia: wBTH `0x49b985ec…`, Uniswap v3 wBTH/WETH pool `0x16C4fDbe2b7497EA67f1DC8205dd2F5B31458D53`.
- Solana devnet: wBTH mint `F7LsiATxVQxnDEBWemfuq1BgFDYbuzqMMJ5eZjaB7LFX`, Orca wBTH/WSOL `9Yog17D3nt1v9cREJBh1Ddeo6fRGuS48hbQcaH9WH1JS`.
- Hyperliquid HyperEVM: PeerToken `0x230f154…` (HIP-1 spot pending #877).

## Architecture landing spots (from the web map)
- New route `/trade` (or `/bridge`) in `web/packages/web-wallet/src/App.tsx` (dual-host aware nav).
- New feature module `web/packages/features/src/bridge/` (components + hooks + types), mirroring `network/`.
- New locale namespace `web/packages/web-wallet/src/locales/{en,es,zh}/bridge.json`.
- Reuse `@botho/ui` + `theme.css`; model on `reserve-proof-card.tsx` + `send-modal.tsx`.
[...]

PR #1043, opened 2026-07-17
feat(wasm-signer): embed destination memo on deposit for bridge matching

### LINKED ISSUE #1037: wallet(wasm-signer): embed the destination memo on-chain in buildSendTransaction (bridge deposit matching)
Fast-follow from #1031 (PR #1035). The bridge watcher matches deposits by the decrypted destination memo (`bridge/service/src/bth_scan.rs`), but `@botho/wasm-signer`'s `buildSendTransaction` does not currently attach a destination memo, and `wallet.send()` previously ignored its `memo` arg.

PR #1035 threaded the order memo through `wallet.send()` (so no call-site change is needed later), but the actual **on-chain memo embedding in the WASM signer is not done** — so a bridge deposit built by the wallet is not yet matchable by the watcher.

## What's needed
- `buildSendTransaction` (Rust→WASM, `web/packages/wasm-signer/`) attaches the provided destination memo to the output/tx in the exact format the bridge watcher's `bth_scan.rs` decrypts + matches to an order UUID.
- Verify round-trip: a wallet-built deposit with an order memo is detected + bound to the order by the watcher.
- Keep normal (non-bridge) sends unaffected when no memo is provided.

## Acceptance
A wallet-constructed BTH deposit carries the order memo on-chain such that the bridge watcher view-key-matches it to the mint order. Verified against the testnet bridge (dovetails with the backend API, other Tier-1 fast-follow).

Part of epic #1029; needed for the Tier 1 export (#1031) to work end-to-end.

rjwalters:
## Summary

Second Tier-1 fast-follow of epic #1029 (with the merged #1036 backend and the #1041 public order API): make the wallet export flow functional end to end by embedding the bridge **order memo** on-chain in the WASM signer, so a wallet-built BTH deposit is matchable by the bridge deposit-scan watcher.

Before this change, `@botho/wasm-signer`'s `buildSendTransaction` attached no destination memo and `wallet.send(to, amount, memo?)` dropped its `memo` arg, so a deposit built by the wallet was invisible to `bridge/service/src/bth_scan.rs::scan_deposit_output`.

## Exact memo format (wallet embed ⇄ watcher read)

- The public order API (`bridge/service/src/public_api.rs`) returns `memo` as `hex::encode(order.memo)` — a **64-byte** value (128 hex chars). `BridgeOrder::generate_memo` puts the mint-order **UUID in the first 16 bytes**, the remaining 48 zero.
- The watcher reads it via `TxOutput::decrypt_memo(account)` → filters `!is_unused()` → `*memo_data()` (the 64 data bytes), then `BridgeOrder::order_id_from_memo` recovers the UUID from the first 16 bytes.
- So the wallet must produce an output whose `memo_data()` equals those 64 raw bytes, with a non-zero (Destination) memo type. Because the bytes are a raw UUID (not UTF-8), the existing `MemoPayload::destination(&str)` cannot carry them — this PR adds `MemoPayload::destination_bytes(&[u8;64])` (type `0x0200`, 64 raw bytes) for exactly this.

## Where it is embedded

`buildAndSign` (`web/packages/wasm-signer/src/core.rs`) puts the memo on the **recipient output (index 0)** via `TxOutput::new_hybrid_to_address(amount, recipient, 0, Some(memo), ..)`. The memo is AES-CTR-encrypted to the recipient's classical DH shared secret (view key) — the same `e_memo` any memo uses — so **only the bridge reserve (deposit account) can read it**. The change output carries no memo.

## Watcher-match proof

- **Rust (bridge)** `bth_scan::scan_matches_wallet_built_deposit_memo_to_order`: builds a real `BridgeOrder`, takes its `generate_memo()`, embeds it via `MemoPayload::destination_bytes` on a hybrid deposit output to the reserve (the identical construction path the WASM uses), scans with the reserve account+KEM secret, and asserts `BridgeOrder::order_id_from_memo(scanned.memo)` == `order.id`.
- **Rust (wasm-signer)** `core::bridge_order_memo_is_embedded_on_recipient_output`: full `build_and_sign` with a memo; the deposit account decrypts the recipient output's memo and recovers the 64 order bytes; change output has no memo; tx still verifies.
- **TS** `send-memo.test.ts`: `buildSendTransaction` threads the memo into the `buildAndSign` `SignRequest` (and leaves it undefined for a plain send), via the `setSigner` seam.

A live WASM-artifact ↔ live-watcher integration test needs the testnet bridge (per the issue's acceptance note); that remains the live-infra step. Everything above is real and green locally.

## Privacy

No weakening. The memo rides the existing per-output encrypted `e_memo` (one-time, recipient-only, ADR 0004 hybrid outputs). When no memo is provided the recipient output is byte-identical to today (no `e_memo`), so ordinary transfers are unchanged. Malformed (non-64-byte) memos are a hard error rather than a silently truncated/padded memo.

## Changes

- `transaction/clsag/src/lib.rs`: add `MemoPayload::destination_bytes`; unit test.
- `web/packages/wasm-signer/src/core.rs`: `SignRequest.memo` (optional hex), `parse_memo`, embed on recipient output; embed / no-regression / malformed tests.
- `web/packages/wasm-signer/src/index.ts`, `src/send.ts`: `memo` on `SignRequest` / `BuildSendParams`, threaded into `buildAndSign`.
- `web/packages/web-wallet/src/contexts/wallet.tsx`: un-ignore `send(to, amount, memo)`.
- `web/packages/wasm-signer/test/send-memo.test.ts`: new.
- `bridge/service/src/bth_scan.rs`: watcher-match proof test.

| `buildSendTransaction` attaches the destination memo in the exact format `bth_scan.rs` decrypts | ✅ | `destination_bytes` → `new_hybrid_to_address` on recipient output; `bth_scan` test recovers the order UUID |
| Round-trip: wallet-built deposit matched + bound to the order by the watcher | ✅ | `scan_matches_wallet_built_deposit_memo_to_order` (real `BridgeOrder`, same construction path) |
| Normal (no-memo) sends unaffected | ✅ | `absent_or_empty_memo_produces_no_output_memo`; empty ⇒ no `e_memo` |
| `wallet.send()` threads the memo (un-ignore `_memo`) | ✅ | `wallet.tsx` diff + `send-memo.test.ts` |
| Reuse chain memo/payment-id format, no new format, no privacy weakening | ✅ | Reuses `MemoPayload` + `e_memo` encryption; change output memo-free |

## Test Plan

- `cargo test -p bth-transaction-clsag -p bth-wasm-signer -p bth-bridge-service` — green
- `cargo clippy` (touched crates, `--all-targets`) + `cargo fmt --check` — clean
- `pnpm --filter @botho/wasm-signer build:wasm` — builds
- `pnpm -r typecheck`, `pnpm build:web`, `pnpm test:run` (961 passed / 6 skipped) — green

Closes #1037
Part of epic #1029

### DIFF
--- a/botho/src/consensus/validation.rs
+++ b/botho/src/consensus/validation.rs
@@ -1077,6 +1077,7 @@ mod tests {
                 amount: 4_000_000_000,
                 fee: MIN_TX_FEE,
                 created_at_height: 1000,
+                bridge_deposit_memo: None,
             };
 
             // Build + CLSAG-sign exactly as the browser does.
--- a/bridge/service/src/bth_scan.rs
+++ b/bridge/service/src/bth_scan.rs
@@ -537,6 +537,63 @@ mod tests {
         assert_eq!(tx.inputs.len(), 1);
     }
 
+    /// #1037 round-trip proof: a deposit built the way the WASM signer builds it
+    /// — a hybrid output to the reserve carrying the order memo via
+    /// `MemoPayload::destination_bytes` (the exact call
+    /// `bth_wasm_signer::core` makes) — is view-key-matched by the watcher back
+    /// to its order UUID.
+    ///
+    /// The memo is the *real* wire memo `BridgeOrder::generate_memo` produces
+    /// (UUID in the first 16 bytes), so this asserts the wallet embed format
+    /// and the watcher read format agree end to end: `scan_deposit_output`
+    /// decrypts the memo and `BridgeOrder::order_id_from_memo` recovers the
+    /// order's id.
+    #[test]
+    fn scan_matches_wallet_built_deposit_memo_to_order() {
+        use bth_bridge_core::{BridgeOrder, Chain};
+
+        let mut rng = StdRng::from_seed([37u8; 32]);
+        let reserve = AccountKey::random(&mut rng);
+        let kem = MlKem768KeyPair::from_seed(&[0x51u8; 32]);
+
+        // A real mint order; its generated memo carries the order UUID.
+        let mut order = BridgeOrder::new_mint(
+            Chain::Ethereum,
+            1_000_000_000_000,
+            1_000_000_000,
+            "bth_deposit_addr".to_string(),
+            "0x1234567890abcdef1234567890abcdef12345678".to_string(),
+        );
+        let order_memo: [u8; 64] = order.generate_memo();
+
+        // The wallet embeds the 64-byte memo via `destination_bytes` (the exact
+        // WASM construction path) on a hybrid deposit output to the reserve.
+        let deposit = TxOutput::new_hybrid(
+            1_000_000_000_000,
+            &reserve.default_subaddress(),
+            kem.public_key(),
+            0,
+            Some(MemoPayload::destination_bytes(&order_memo)),
+            ClusterTagVector::empty(),
+        );
+        let rpc = rpc_output_from_txoutput(&deposit, "0xdeposit", 0);
+
+        // The watcher scans it with the reserve account + KEM secret.
+        let scanned = scan_deposit_output(&rpc, &reserve, Some(&kem))
+            .unwrap()
+            .expect("reserve detects its hybrid deposit");
+        let memo = scanned.memo.expect("the deposit carries a decodable memo");
+        assert_eq!(memo, order_memo, "decrypted memo must equal the order memo");
+
+        // The watcher recovers the order UUID and it matches the order.
+        let recovered =
+            BridgeOrder::order_id_from_memo(&memo).expect("watcher recovers a UUID from the memo");
+        assert_eq!(
+            recovered, order.id,
+            "the recovered order id must match the order the wallet was told to reference"
+        );
+    }
+
     #[test]
     fn scan_detects_owned_output_and_decrypts_memo() {
         let mut rng = StdRng::from_seed([3u8; 32]);
--- a/mobile/rust-bridge/src/lib.rs
+++ b/mobile/rust-bridge/src/lib.rs
@@ -891,6 +891,7 @@ mod send_acceptance_tests {
             amount: send_amount,
             fee: MIN_TX_FEE,
             created_at_height: 1000,
+            bridge_deposit_memo: None,
         };
 
         let tx = build_and_sign_with_rng(&request, &mut rng).expect("mobile build+sign");
--- a/mobile/rust-bridge/src/wallet_ops.rs
+++ b/mobile/rust-bridge/src/wallet_ops.rs
@@ -282,6 +282,10 @@ pub async fn send(
         amount,
         fee,
         created_at_height: height,
+        // The mobile wallet has no bridge-deposit flow, so it never sets the
+        // dedicated bridge order-memo channel (#1037). An ordinary send carries
+        // no on-chain memo — byte-identical to before.
+        bridge_deposit_memo: None,
     };
 
     let tx_hex = build_and_sign_inner(&request).map_err(SendError::Other)?;
--- a/mobile/rust-bridge/tests/wallet_core_offline.rs
+++ b/mobile/rust-bridge/tests/wallet_core_offline.rs
@@ -289,6 +289,7 @@ fn build_and_sign_produces_node_verifiable_tx() {
         amount,
         fee: MIN_TX_FEE,
         created_at_height: 1,
+        bridge_deposit_memo: None,
     })
     // `build_and_sign_inner` self-verifies structure + ring signatures + the
     // balance equation under the node's own verifier before returning, so a
@@ -346,6 +347,7 @@ fn build_and_sign_rejects_insufficient_inputs() {
         amount: 50_000_000_000,
         fee: MIN_TX_FEE,
         created_at_height: 1,
+        bridge_deposit_memo: None,
     });
 
     assert!(
--- a/transaction/clsag/src/lib.rs
+++ b/transaction/clsag/src/lib.rs
@@ -256,6 +256,28 @@ impl MemoPayload {
         Self { data }
     }
 
+    /// Create a destination memo carrying raw 64-byte binary data.
+    ///
+    /// Unlike [`MemoPayload::destination`] (which takes a UTF-8 `&str` and
+    /// truncates/pads it), this preserves an arbitrary 64-byte payload exactly.
+    /// It is used to embed a bridge deposit's **order memo** — a 64-byte value
+    /// whose first 16 bytes are the mint-order UUID
+    /// (`BridgeOrder::generate_memo` in `bridge/core`) — because those
+    /// bytes are raw binary (a UUID), not UTF-8. The memo `type` is set to
+    /// `[0x02, 0x00]` (Destination) so the output is not treated as
+    /// "unused" ([`MemoPayload::is_unused`]), and
+    /// [`MemoPayload::memo_data`] returns the 64 bytes byte-for-byte — exactly
+    /// what the bridge watcher reads before recovering the order UUID
+    /// (`bridge/service/src/bth_scan.rs`).
+    pub fn destination_bytes(data: &[u8; 64]) -> Self {
+        let mut buf = [0u8; ENCRYPTED_MEMO_SIZE];
+        // Type: 0x0200 = Destination memo (non-zero => not "unused").
+        buf[0] = 0x02;
+        buf[1] = 0x00;
+        buf[2..ENCRYPTED_MEMO_SIZE].copy_from_slice(data);
+        Self { data: buf }
+    }
+
     /// Get the memo type bytes (first 2 bytes)
     pub fn memo_type(&self) -> [u8; 2] {
         [self.data[0], self.data[1]]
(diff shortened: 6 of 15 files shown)


rjwalters:
## Loom Judge review — REQUEST CHANGES

Thorough review of the crypto / consensus / privacy surface. The core signer work is **correct and well-tested** — but there are **two blocking defects** (one CI-failing build break, one user-facing regression) that must be fixed before merge.

### Verdict: REQUEST CHANGES

---

### BLOCKER 1 — Workspace build is broken: `botho-mobile` never sets the new `SignRequest.memo` field
`cargo build --workspace (Linux)` **fails** on this PR:
```
error[E0063]: missing field `memo` in initializer of `SignRequest`
  --> could not compile `botho-mobile` (lib)
```
`#[serde(default)]` only covers deserialization — it does **not** satisfy Rust struct-literal construction. Four sites construct `SignRequest { .. }` and were not updated:
- `mobile/rust-bridge/src/wallet_ops.rs:273` (the lib build the CI caught)
- `mobile/rust-bridge/src/lib.rs:879`
- `mobile/rust-bridge/tests/wallet_core_offline.rs:277` and `:330`

Add `memo: None` at each (or gate `SignRequest` construction through a builder/`..Default::default()`). The wasm-signer's own tests were updated with `memo: None`, but the mobile crate was missed.

### HIGH 2 — Free-text user memos now hard-fail ordinary sends (no-regression claim is only true for the empty case)
`wallet.send(to, amount, memo)` is already fed by **human free-text** in two existing flows, not just the bridge order memo:
- **Send modal** (`web/packages/features/src/wallet/components/send-modal.tsx:376-386`): a `Memo (optional) — "Add a note..."` text input → `SendFormData.memo` → `wallet.tsx:357 send(data.recipient, data.amount, data.memo)`.
- **Pay / payment-request** (`RequestModal.tsx` free-text memo → `/pay#` URL → `pay.tsx:271 send(request.to, amount, request.memo)`, rendered as a human "note", `pay.tsx:418-427`).

Before this PR `_memo` was dropped, so these sends succeeded. Now the memo is threaded into `parse_memo`, which **hard-errors** on anything that isn't exactly 64 bytes of hex:
- `"lunch"` → `hex::decode` fails → `"memo: invalid hex"`
- `"abcd"` → decodes to 2 bytes → `"expected a 64-byte memo (128 hex chars), got 2 bytes"`

So any user who types a note in the Send modal (or pays a request link carrying a note) now gets their transaction **rejected**. The PR's "normal (no-memo) sends unaffected" only holds for absent/empty memos.

The root cause is overloading one `memo: string` channel with two incompatible meanings (human note vs. 64-byte-hex bridge order id). Recommended fix: give the bridge deposit its own typed channel (e.g. `trade.tsx` passes an `orderMemo`/dedicated param that flows to the on-chain memo) and keep the free-text UI note out of the signer (leave it local/cosmetic as before), **or** at minimum stop routing the send-modal/pay free-text field into `parse_memo`. A green TS test that a free-text note still sends would lock this down.

---

### What is correct (verified, no action needed)

- **Privacy — no leak / no regression.** The memo rides the existing per-output `e_memo`: AES-256-CTR keyed by HKDF-SHA512 over the classical view-key DH secret (`apply_memo_keystream`, `new_with_memo`/`assemble_hybrid` encrypt to `create_shared_secret(view_pub, tx_priv)`), readable **only** by the recipient (bridge reserve) via `decrypt_memo`. No plaintext UUID/destination on chain. The change output stays memo-free (`core.rs` passes `None` for the index-1 change output — confirmed unchanged by the diff), and `absent_or_empty_memo_produces_no_output_memo` proves absent/empty ⇒ byte-identical (no `e_memo`). The only observable distinguisher is the *presence* of a recipient `e_memo`, which is the unavoidable minimum for on-chain matching and does not reveal the recipient (stealth addressing) or the order.
- **`destination_bytes` / type `0x0200`.** No new tag and **no collision** — it reuses the existing Destination type (same as `destination(&str)`), writes the 64 bytes into `buf[2..66]` of the 66-byte payload, and `memo_data()` returns them byte-for-byte. Round-trips exactly through encrypt → output → `decrypt_memo`, and matches the watcher read (`bth_scan.rs:150-155` `decrypt_memo → !is_unused → *memo_data`, then `order_id_from_memo` reads the first 16 bytes as the UUID).
- **Watcher-match proof is sound, not circular.** `bth_scan::scan_matches_wallet_built_deposit_memo_to_order` recovers a **real** `BridgeOrder`'s UUID from a deposit built via `destination_bytes`, and `bridge_order_memo_is_embedded_on_recipient_output` drives the full `build_and_sign` path (recipient index 0 carries it, change carries none, tx verifies). The format agreement between the two halves is established transitively. The single-process WASM-artifact ↔ live-watcher test is appropriately deferred to live infra.
- **Consensus / tx-format safe.** No new wire variant: `e_memo` is already `Option<EncryptedMemo>` and the memo is opaque bytes to validation. Serialization determinism and backward compatibility with existing outputs are unaffected.

### Informational (non-blocking)
- Memo confidentiality is **classical-only** (the `e_memo` DH is Ristretto, not ML-KEM), so a future quantum "harvest-now-decrypt-later" adversary could recover the order UUID. This is the pre-existing ADR-0004 `e_memo` design (not introduced here), and the order UUID is a low-sensitivity bridge-internal reference already known to the depositor — worth a one-line note, no change required.
- `scan_matches_wallet_built_deposit_memo_to_order` uses `TxOutput::new_hybrid` while the WASM path uses `new_hybrid_to_address`; both funnel through `assemble_hybrid` so the memo encryption is identical, but calling the exact `new_hybrid_to_address` would make the "same construction path" claim literal.

### CI
`cargo build --workspace (Linux)` = **fail** (Blocker 1). `cargo test (bridge crates)`, `typecheck + vitest`, `rustfmt`, `cargo deny` = pass; benchmarks/fuzz-guard still pending. Merge state UNSTABLE. No secrets introduced.

Fix Blocker 1 (unblocks CI) and HIGH 2 (restores free-text sends), then re-request review.

rjwalters:
## Both blockers addressed — pushed `4274e126`

Thanks for the thorough review. Both defects are fixed; the crypto / privacy / watcher-match surface you verified is untouched.

### BLOCKER 1 — workspace build (`E0063`) — FIXED
Renamed the field `SignRequest.memo` → `bridge_deposit_memo` and set it (`None`) at **every** struct-literal construction site. Your four sites, plus one more the compiler flagged that the review missed:
- `mobile/rust-bridge/src/wallet_ops.rs` (production send)
- `mobile/rust-bridge/src/lib.rs` (test)
- `mobile/rust-bridge/tests/wallet_core_offline.rs` (×2)
- `botho/src/consensus/validation.rs` — the `browser_built_transfer_passes_node_kem_gate` parity test (also constructs the wasm-signer `SignRequest`)

`cargo build --workspace` is now green.

### HIGH 2 — free-text notes hard-failing sends — FIXED (design correction)
Root cause was exactly as you diagnosed: one `memo: string` channel overloaded with two incompatible meanings. Split into two typed channels end-to-end:

- **WASM signer:** `SignRequest.memo` → `bridgeDepositMemo` (Rust `bridge_deposit_memo`); `BuildSendParams.memo` → `bridgeDepositMemo`. The strict 64-byte-hex `parse_memo` validator now applies to the **bridge channel only**.
- **Web `send()`:** now `send(to, amount, opts?: SendOptions)` where `SendOptions = { note?: string; bridgeDepositMemo?: string }`.
  - `note` (human free-text) is **cosmetic** — dropped, never routed into the signer (restores the pre-#1043 behavior, so `"lunch"` sends fine).
  - `bridgeDepositMemo` (64-byte hex) is the only value embedded on-chain.
- **Callers:** Send-modal (`pages/wallet.tsx`) and Pay (`pay.tsx`) pass `{ note }`; the bridge **ExportPanel** (`trade.tsx`) is the sole caller passing `{ bridgeDepositMemo }`. The bridge `submitDeposit({ memo })` order-id still flows to the on-chain channel and matches `bth_scan.rs` unchanged.

### Tests
- New TS lock-down in `contexts/wallet.test.tsx`: a free-text note `"lunch"` sends successfully and reaches the signer with `bridgeDepositMemo === undefined`; a 64-byte-hex bridge memo IS threaded.
- `send-memo.test.ts` + the Rust `core.rs` embed / no-regression / malformed / `bth_scan` watcher-match tests updated to the renamed channel (all still assert the same on-chain behavior).

### Verification
- `cargo build --workspace` — green
- `cargo clippy` (edited crates, `-D warnings`) — clean (the `botho-wallet` `should_implement_trait`/`question_mark` lints are pre-existing, not touched here)
- `cargo fmt --all --check` — clean
- `cargo test` — clsag / wasm-signer / bridge-service / mobile / consensus all pass
- `pnpm -r typecheck` + `pnpm build:web` + `pnpm test:run` — green (**963 passed / 6 skipped**)

Re-requesting review.

rjwalters:
## Judge re-review — commit `4274e126` delta (APPROVE)

Focused re-review of the loom-doctor fix over the already-approved crypto core. The approved privacy/consensus pieces (encrypted `e_memo`, clsag `destination_bytes` type `0x0200`, `bth_scan.rs` watcher-match) were **not** re-litigated — verified untouched by this commit. Only the two-blocker delta was reviewed.

### BLOCKER 1 — workspace build (E0063) — FIXED
`SignRequest.memo` → `bridge_deposit_memo`. All 13 `SignRequest { … }` construction sites now set the field:
- `core.rs` (helper + 8 test sites, incl. the `Some(memo_hex)` embed test)
- `mobile/rust-bridge/src/wallet_ops.rs` (real send path, `None`), `src/lib.rs` (acceptance test), `tests/wallet_core_offline.rs` (×2)
- `botho/src/consensus/validation.rs:1061` (the browser-parity test — the non-obvious one)

No stale bare `memo:` field remains (the surviving `memo:` greps are unrelated `e_memo: None` TxOutput fields and the `parse_memo` param name). **`cargo build --workspace (Linux)` CI = PASS.**

### BLOCKER 2 — free-text note regression — FIXED
Channels split end-to-end:
- Signer strict 64-byte-hex `parse_memo` now scoped to `req.bridge_deposit_memo` only (`build_and_sign_with_rng`, field label `"bridgeDepositMemo"`).
- `BuildSendParams.memo` → `bridgeDepositMemo`; `send.ts` threads it (and only it) into the `SignRequest`.
- `wallet.send(to, amount, opts?: SendOptions)` with `SendOptions = { note?; bridgeDepositMemo? }`. `send()` reads **only** `options?.bridgeDepositMemo`; `note` is never referenced past the destructure → provably cannot reach the signer.

Caller audit — every `send()`/`buildSendTransaction` caller on the new signature, zero positional-string callers:
- `pages/wallet.tsx:361` → `{ note: data.memo }` (Send-modal note, cosmetic)
- `pages/pay.tsx:275` → `{ note: request.memo }` (payment-request note, cosmetic)
- `pages/trade.tsx:67` → `{ bridgeDepositMemo: memo }` — **sole** on-chain caller (ExportPanel), threads the 64-byte order memo unchanged into the exact `e_memo` format the watcher matches
- `PaymentRequest.memo` remains a free-text wire note, correctly routed to `{ note }`

### Tests — genuinely lock the behavior
- `wallet.test.tsx` new suite: `send(addr, amt, { note: 'lunch' })` succeeds (`txHash === '0xhash'`) and asserts `buildSendTransaction` received `bridgeDepositMemo === undefined` — the regression is provably gone. Companion test asserts a 64-byte order memo IS threaded.
- `send-memo.test.ts` updated to assert `req.bridgeDepositMemo` threading / undefined-for-ordinary.
- **Web CI (`typecheck + vitest + worker dry-run`) = PASS.**

### Scope / privacy
Pure plumbing rename. On-chain behavior for a bridge memo is byte-identical to the already-approved path; free-text notes produce the same output as pre-PR (dropped). No privacy regression, no scope creep.

### CI
Green: cargo build --workspace, web typecheck+vitest, cargo test (bridge crates), rustfmt, cargo deny, benchmarks. Only non-blocking Fuzz Build Guard / Node Benchmarks still finishing (unrelated to this plumbing; benchmarks already trending pass).

**Verdict: APPROVE.** Both blockers truly resolved; nothing else broke.

PR #1047, opened 2026-07-17
feat(bridge): global order-create cap + expired-order pruning, strict BTH address validation, release-intent audit parity

### LINKED ISSUE #1042: bridge(public-api): hardening follow-ups from #1041 review (global cap, order pruning, BTH address validation)
Non-blocking hardening the #1041 security judge flagged on the public order API (`bridge/service/src/public_api.rs`). None gate the merge; all worth doing before the public API is enabled on a live (esp. mainnet) node.

- [ ] **(Low) Global order-create cap + expired-order pruning.** The per-IP fixed-window rate limiter bounds per-client rate but there's no global cap on pending orders and no DB pruning of expired/abandoned `AwaitingDeposit` orders → slow unbounded DB growth under distributed spam. Add a global create ceiling and/or a periodic prune of expired orders.
- [ ] **(Low) Strengthen `ChainAddress::validate` for BTH** in shared core — currently weak; tighten so an invalid BTH destination is rejected at order-create rather than later.
- [ ] **(Info) Release-intent info-leak** — the release-intent GET correlates on already-public on-chain data; benign, but document/scope what it returns so it doesn't grow into leaking non-public state.
- [ ] **(Nit) Audit-log symmetry** — mint create logs an audit line; release-intent create doesn't. Add parity.

Context: the public API is DISABLED by default (empty `[public_api].listen`); these harden it for when an operator opts in. Part of epic #1029; gates nothing currently.

rjwalters:
## Summary

Implements all four hardening follow-ups the #1041 security judge flagged on the public order API (`bridge/service/src/public_api.rs`), so the surface is safe to enable on a live node. Part of epic #1029.

## Changes

**1. Global order-create cap + expired-order pruning (Low)**
- New `public_api.max_open_orders` setting (default 10,000; 0 = unlimited): `POST /api/bridge/orders` is refused with 503 once the count of mint orders still awaiting their deposit reaches the cap; `POST /api/bridge/release-orders` independently caps on unexpired release intents. This backstops the per-IP limiter against distributed spam (which only bounds per-client rate).
- New DB prunes wired into the engine's 10s processing loop (right after `expire_stale_orders`): `prune_expired_mint_orders` deletes expired mint orders where no deposit ever arrived (`status='expired' AND order_type='mint' AND source_tx IS NULL`), and `prune_expired_release_intents` deletes long-expired intents — both after a 7-day retention window (`EXPIRED_ORDER_RETENTION_SECS`) so wallets can still poll the terminal `expired` status. Orders where funds ever moved are NEVER pruned.

**2. Strict BTH address validation in shared core (Low)**
- `ChainAddress::validate(Chain::Bth)` (in `bridge/core/src/chains/mod.rs`) previously only rejected empty strings. It now does a real structural check mirroring the service's `decode_recipient_address`: accepts `botho://2/<base58>` / `tbotho://2/<base58>` URIs whose body decodes to a plausible key payload (>= 64 bytes), or a legacy bare base58 body decoding to exactly 64 bytes (view||spend). Rejects non-base58 junk, retired v1/quantum prefixes (ADR 0008), unknown URI schemes, whitespace/control chars, and oversized input (> 8 KiB). Invalid BTH destinations now fail at order-create with a 400.
- `bs58` (already in the workspace lockfile) added to `bth-bridge-core` for the decode.

**3. Release-intent GET information-exposure scope documented (Info)**
- New "Information-exposure scope" section in the module header + a scope contract on the `get_release_order` handler doc: the endpoints must only ever return caller-supplied data or data derivable from public on-chain state, and must never grow to surface operator/ops state, internal errors, or per-user data. Docs only — no behavior change, per the issue.

**4. Audit-log symmetry (Nit)**
- `create_release_order` now writes a `release_intent_created` audit row, mirroring mint create's `order_created`. Intents are not `bridge_orders` rows, so the intent UUID travels in the details field rather than the FK-referencing `order_id` column.

| Global create ceiling + expired-order DB pruning | ✅ | `test_global_create_cap_on_mint_orders`, `test_global_create_cap_on_release_intents`, `test_count_and_prune_expired_mint_orders`, `test_count_and_prune_expired_release_intents`; prunes wired into the engine loop |
| Strengthened `ChainAddress::validate` for BTH rejects invalid destinations at order-create | ✅ | `test_bth_address_validation_accepts_valid_shapes`, `test_bth_address_validation_rejects_junk` (core); `test_create_release_order_rejects_invalid_bth_address` (API-level 400) |
| Release-intent GET scope documented | ✅ | Module-header "Information-exposure scope (#1042)" section + `get_release_order` doc contract |
| Audit-log symmetry on release-intent create | ✅ | `test_release_intent_create_writes_audit_row` |

## Test Plan

- `cargo build -p bth-bridge-core -p bth-bridge-service` — green.
- `cargo clippy -p bth-bridge-core -p bth-bridge-service --all-targets -- -D warnings` — green.
- `cargo fmt --all --check` — green.
- `cargo test -p bth-bridge-core -p bth-bridge-service` — green (71 + 230 passing; 8 new tests, plus the pre-existing suite updated for the tightened validator).

Closes #1042

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2045,6 +2045,7 @@ dependencies = [
 name = "bth-bridge-core"
 version = "0.6.0"
 dependencies = [
+ "bs58 0.4.0",
  "chrono",
  "ed25519-dalek",
  "hex",
--- a/bridge/core/Cargo.toml
+++ b/bridge/core/Cargo.toml
@@ -14,6 +14,10 @@ std = []
 [dependencies]
 serde = { workspace = true, features = ["derive", "alloc"] }
 chrono = { workspace = true, features = ["serde", "std", "clock"] }
+# Structural BTH address validation (#1042): base58-decode the address body so
+# junk destinations are rejected at order-create. Tiny, already in the
+# workspace lockfile (the bridge service uses it for recipient decoding).
+bs58 = { workspace = true }
 uuid = { version = "1", features = ["v4", "serde"] }
 toml = { workspace = true }
 hex = { workspace = true }
--- a/bridge/core/src/chains/mod.rs
+++ b/bridge/core/src/chains/mod.rs
@@ -55,16 +55,14 @@ impl ChainAddress {
     }
 
     /// Validate the address format for the chain.
+    ///
+    /// This is a STRUCTURAL check (prefix, base58 alphabet, payload size)
+    /// meant to reject junk before it is persisted — e.g. at public
+    /// order-create (#1042). It does not verify key material; the service's
+    /// address codec does full validation when an address is actually used.
     pub fn validate(&self) -> Result<(), String> {
         match self.chain {
-            Chain::Bth => {
-                // BTH addresses are base58-encoded public addresses
-                if self.address.is_empty() {
-                    return Err("BTH address cannot be empty".to_string());
-                }
-                // Basic validation - actual validation would check base58 and structure
-                Ok(())
-            }
+            Chain::Bth => validate_bth_address(&self.address),
             Chain::Ethereum => {
                 // Ethereum addresses are 0x-prefixed 40-char hex strings
                 if !self.address.starts_with("0x") {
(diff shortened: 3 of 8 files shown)


rjwalters:
✅ **Approved!**

Reviewed against #1042 and the #1041 security-review context. All four scope items are implemented, and I specifically scrutinized the two security-sensitive pieces:

## Security scrutiny

**1. BTH address validation (`bridge/core/src/chains/mod.rs`) — correct, no false rejects.**
- The v2 check (`botho://2/` / `tbotho://2/`, base58 body, payload >= 64 bytes) is a strict *superset* of the canonical codec (`address-codec/src/lib.rs` requires those exact prefixes plus an exact `ADDRESS_BODY_LEN` ~3.2 KB), so no address `bth_address_codec::decode_address` accepts can be rejected at order-create.
- Retired-prefix list matches the codec's ADR 0008 constants; `botho-pq://` is broader than the codec's `botho-pq://1/`, which is strictly safe.
- Legacy bare base58 must decode to exactly 64 bytes, matching `decode_recipient_address`'s fallback. The validator is *stricter* than that fallback for odd version-less URI shapes (e.g. `botho://<body>`), but I verified the only production `.validate()` call sites are the two public create handlers (`public_api.rs:459,614`) — the watcher/burn/release pipeline never calls it, so no in-flight burn can be dropped by the tightening. Stricter-at-create is the right posture.
- Length cap (8 KiB), ASCII/whitespace/control rejection, and clear error messages are all sensible.

**2. Pruning can never orphan a legitimate deposit.**
- `record_deposit_detected` (db.rs:876) only matches `WHERE status = 'awaiting_deposit'` — once an order is Expired, a late deposit can never attach to it, pruned or not. I also searched for any reopen/revive path for Expired orders: none exists.
- The prune predicate (`status='expired' AND order_type='mint' AND source_tx IS NULL AND updated_at < now-7d`) can only ever delete rows that are already permanently unmatchable and where no funds ever moved. Burn/settled orders are untouched (verified by test).
- Release intents are UX correlation records never read by the release pipeline; pruning them is inert to releases.

**3. Global cap.** Count-then-insert is not atomic, so concurrent creates can overshoot the cap by a few rows — acceptable for a spam backstop (the per-IP limiter still bounds rate; the ceiling's purpose is bounding total DB growth, which it does). Expired intents stop counting immediately and awaiting-deposit orders expire, so a full backlog self-drains rather than wedging the API. Config default (10,000), 0-disables semantics, and legacy-TOML back-compat are all tested.

**4. Docs + audit parity.** The information-exposure scope statements are accurate (I checked what `get_release_order` returns: caller-echoed intent + burn status/tx hashes, all public on-chain). `release_intent_created` audit row with the UUID in details (intents are not `bridge_orders` rows, so no FK `order_id`) is the right call.

## Test Execution

Test plan from PR description, executed in the `issue-1042` worktree (head `f8e24d32`):
1. `cargo test -p bth-bridge-core` — ✅ 71 passed (incl. both new address-validation tests)
2. `cargo test -p bth-bridge-service` — ✅ 230 passed (incl. cap, prune, 400-rejection, and audit-row tests)
3. build/clippy/fmt — ✅ covered by CI (all checks green)

## CI / Merge state

- `gh pr checks 1047`: all required checks pass (workspace build, bridge tests, fuzz build guard, coverage, cargo deny, rustfmt)
- `mergeStateStatus`: CLEAN
- PR body uses `Closes #1042`

Nice, tightly-scoped hardening PR. Ready to merge.

PR #1070, opened 2026-07-17
fix(node): explicit-mode solo node arms minting since #770 sync gate

### LINKED ISSUE #1056: node: explicit-mode solo node never arms minting since the #770 startup sync gate
## Original Issue

Found while running the wasm-signer node harness during the #815 snap spike (PR #1055).

**Symptom**: a node configured with `[network.quorum] mode = "explicit"`, `threshold = 1`, `members = []` and zero reachable peers boots with `initial_sync_complete = false` (because `min_peers_to_wait` is hardcoded to 1 for Explicit mode in `botho/src/commands/run.rs`), and `ChainSyncManager::tick` requires at least one connected peer before it can ever reach `Synced`. Result: `syncStatus: "discovering"` forever, minting requested-but-deferred forever, chain stuck at height 0.

**Where**:
- `run.rs` `min_peers_to_wait = match quorum.mode { Explicit => 1, Recommended => min_peers }` → seeds `initial_sync_complete = min_peers_to_wait == 0`.
- `network/sync.rs` `tick()` Discovery arm: `has_connected_peer` must be true before the Synced/Downloading decision.

**Impact**: any explicit-quorum solo/bootstrap configuration (e.g. a fresh single-node testnet, or the wasm-signer node harness before PR #1055 switched it to `recommended`/`min_peers=0`) silently never mints. The #770 doc comment says a "genuine solo node" takes the seeded-synced carve-out, but the carve-out only triggers for Recommended mode with `min_peers = 0` — an explicit solo set (threshold 1, no members) is arguably just as genuinely solo.

**Suggested fix direction**: treat an explicit quorum with no members (or threshold satisfiable with zero peers) as solo for the seeding decision, or add a Discovery timeout that concludes initial sync when the node has expected peers but none exist and the local chain is at genesis. Needs care not to reintroduce the #766/#998 races the gate exists to prevent.

_Not fixed node-side in PR #1055 (out of spike scope); only the test harness config was changed._

---

### Problem Statement

A node configured for a genuinely solo `explicit` quorum — `[network.quorum] mode = "explicit"`, `threshold = 1`, `members = []`, with zero reachable peers — boots but **never arms minting** and stays pinned at height 0 with `syncStatus: "discovering"` indefinitely. This is a regression introduced by the #770 startup sync gate: before #770 an explicit solo node minted; after #770 the seeding logic only grants the solo carve-out to `Recommended` mode with `min_peers == 0`, silently excluding the explicit-solo configuration.

This matters because any fresh single-node explicit testnet, bootstrap node, or harness (e.g. the wasm-signer node harness before PR #1055 worked around it by switching to `recommended`/`min_peers=0`) silently produces no blocks. There is no error — the node logs "Minting will start when initial sync completes" and then waits forever.

### Root Cause

Two independently-correct pieces interact to produce a deadlock for the empty-member explicit config:

1. `botho/src/commands/run.rs` (~line 361) hardcodes the sync-gate seed for explicit mode:
   ```rust
   let min_peers_to_wait = match config.network.quorum.mode {
       QuorumMode::Explicit => 1, // In explicit mode, wait for at least one peer
       QuorumMode::Recommended => config.network.quorum.min_peers as usize,
   };
   ```
   Then (~line 811) `let mut initial_sync_complete = min_peers_to_wait == 0;` — so an explicit node **always** starts `initial_sync_complete = false`, regardless of whether it actually expects peers.

2. `botho/src/network/sync.rs` `tick()` Discovery arm (~line 1199): `let has_connected_peer = !connected_peers.is_empty();` — the `Synced` transition is gated on `has_connected_peer && all_compatible_have_status`. A node with no peers never satisfies `has_connected_peer`, so it can never leave `Discovery`, never reaches `Synced`, and never flips `initial_sync_complete` true (run.rs ~line 2067).

Meanwhile `Config::can_reach_quorum` (`botho/src/config.rs` ~line 598) reports the quorum **is** satisfiable: for explicit `threshold=1, members=[]`, `quorum_size = trusted_connected + 1 = 1 >= threshold(1)`, so `can_mint_now = true`. The result is `should_arm_minting(can_mint_now=true, initial_sync_complete=false) == false` (run.rs ~line 209/833) forever, and control falls into the `else if mint && can_mint_now && !initial_sync_complete` branch that prints "Minting will start when initial sync completes." — an event that can never occur for a peerless node.

The `#770` gate exists to prevent a restarted peer-expecting node from re-entering SCP at a stale slot while the fleet is ahead (the #766/#998/#1000 races). That rationale simply does not apply when the operator has declared an **empty** member set: there is no fleet to fall behind. The seeding discriminator is wrong, not the gate itself.

### Implementation Guidance

**Recommended (conservative) fix** — make the explicit-mode seed depend on whether the config actually expects peers, keying on an empty member set:

```rust
let min_peers_to_wait = match config.network.quorum.mode {
    // An explicit quorum with a non-empty member set genuinely expects
    // trusted peers and must keep the #770 gate (wait for >=1 peer, sync,
    // then mint). An explicit quorum with NO members is solo-by-design:
    // its quorum is satisfied by self alone, so there is no fleet to fall
    // behind and the gate would deadlock (issue #1056).
    QuorumMode::Explicit => {
        if config.network.quorum.members.is_empty() { 0 } else { 1 }
    }
    QuorumMode::Recommended => config.network.quorum.min_peers as usize,
};
```

With `min_peers_to_wait == 0`, `initial_sync_complete` seeds `true` (the existing carve-out at run.rs ~line 811), the discovery wait loop (~line 375) does not block, and `should_arm_minting` returns true at startup — restoring pre-#770 solo behavior for the empty-member config while leaving the gate fully intact for every explicit config that lists members.

**Why key on `members.is_empty()` and not `threshold <= 1`:** an explicit config with `threshold = 1` **and** a non-empty member set can mint solo (self satisfies the threshold) yet still has trusted peers it should sync from before minting — exactly the stale-slot race the gate prevents. Only the empty-member set is genuinely peerless by design, so it is the correct discriminator. Generalizing to `min_peers_to_wait = threshold.saturating_sub(1).min(members.len())` is a tempting broader fix but would drop the gate for the threshold-1-with-members case and risks reintroducing #766 — do **not** take that path without separate analysis.

**Alternative (defense-in-depth, optional):** additionally add a bounded Discovery timeout in `sync.rs` `tick()` that concludes initial sync when the node has waited `min_peers_to_wait` peers, none have appeared, and the local chain is at genesis — so a mis-seeded node self-heals. This is broader and touches the #766/#998 guard surface; if pursued, gate it strictly on `local_height == 0` and treat as a follow-up, not part of this fix.

### Acceptance Criteria

- [ ] A node with `mode = "explicit"`, `threshold = 1`, `members = []`, zero peers arms minting at startup and advances past height 0 (produces blocks solo).
- [ ] Its RPC `syncStatus` reports `"synced"` (not `"discovering"`) shortly after boot.
- [ ] An explicit node with a **non-empty** member set still starts `initial_sync_complete = false` and defers minting until it connects to a peer and reaches `Synced` (the #770 gate is unchanged for fleet configs).
- [ ] `Recommended` mode behavior (both `min_peers == 0` solo and `min_peers >= 1`) is unchanged.
- [ ] No regression of the #766/#998/#1000 races: a peer-expecting node that boots behind the tip does not mint at a stale slot.

### Affected Files

- `botho/src/commands/run.rs` — change the `min_peers_to_wait` match arm for `QuorumMode::Explicit` to seed `0` when `quorum.members` is empty (~line 361). No other logic changes required; the existing `initial_sync_complete = min_peers_to_wait == 0` seeding, discovery wait loop, and `should_arm_minting` flow all follow correctly from this.
- `botho/src/network/sync.rs` — reference only for the `has_connected_peer` Discovery gate (~line 1199) that makes the deadlock possible; no change needed under the recommended fix (change is optional/defense-in-depth only).
- `botho/src/config.rs` — reference only: `can_reach_quorum` (~line 598) confirms quorum is satisfiable with zero peers for the affected config; no change needed.

### Test Plan

- [ ] Automated (unit): add a test in `botho/src/commands/run.rs` (there is an existing test module around line 3284 covering the solo/`initial_sync_complete` seeding) asserting that an explicit config with `threshold=1, members=[]` yields `min_peers_to_wait == 0` and therefore seeds `initial_sync_complete == true`, while an explicit config with a non-empty `members` list yields `min_peers_to_wait == 1` / `initial_sync_complete == false`.
- [ ] Manual verification: launch a single node with `mode="explicit", threshold=1, members=[]`, no bootstrap peers, DNS seeds disabled, `mint=true`; confirm logs show "Starting minting" (not the deferral message), `network_getSyncStatus` returns `synced`, and block height increases past 0.
- [ ] Regression: launch an explicit node with a non-empty `members` list and no reachable peers; confirm it correctly defers minting (`initial_sync_complete=false`, `syncStatus=discovering`) — the #770 gate still holds.
- [ ] Edge cases: explicit `threshold=1, members=[<one peer>]` (should wait, per the design note above); `recommended, min_peers=0` (should still mint solo, unchanged).

rjwalters:
## Summary

Fixes a regression from the #770 startup sync gate: a node configured with
`[network.quorum] mode="explicit", threshold=1, members=[]` and zero reachable
peers never armed minting and stayed pinned at height 0 with
`syncStatus: "discovering"` forever.

**Root cause:** `min_peers_to_wait` was hardcoded to `1` for Explicit mode in
`botho/src/commands/run.rs`, so `initial_sync_complete` seeded `false`. The
Discovery arm of `ChainSyncManager::tick` (`sync.rs`) requires `has_connected_peer`
before it can reach `Synced`, which a peerless solo explicit node can never satisfy —
so `should_arm_minting` was `false` indefinitely even though `can_reach_quorum`
reports the (self-only) quorum satisfiable.

## Fix

Extracted the seed decision into a pure helper `min_peers_to_wait_for_startup()`
that returns `0` for an explicit quorum with an **empty** member set (solo-by-design),
restoring pre-#770 solo behavior. The gate is left fully intact for:

- Explicit configs that **list members** (still wait for >=1 peer, sync, then mint —
  the #766/#998/#1000 stale-slot protection), including the `threshold=1`-with-members
  case.
- `Recommended` mode (`min_peers==0` solo carve-out and `min_peers>=1` both unchanged).

Keying on `members.is_empty()` (not `threshold<=1`) is deliberate per the curation
note: only the empty-member set is genuinely peerless by design.

## Tests

Added three unit tests in the existing `commands::run::tests` module:
- `explicit_solo_empty_members_waits_for_zero_peers` — the fix (seeds synced, mints solo)
- `explicit_with_members_still_waits_for_a_peer` — #770 gate unchanged for fleets
- `recommended_mode_peer_wait_is_unchanged` — no regression to Recommended mode

`cargo build -p botho` passes; `cargo test -p botho --lib commands::run::` → 39 passed, 0 failed.

## Affected Files

- `botho/src/commands/run.rs` — new `min_peers_to_wait_for_startup()` helper + call site + tests.

Closes #1056

### DIFF
--- a/botho/src/commands/run.rs
+++ b/botho/src/commands/run.rs
@@ -210,6 +210,39 @@ fn should_arm_minting(quorum_ok: bool, initial_sync_complete: bool) -> bool {
     quorum_ok && initial_sync_complete
 }
 
+/// Decide how many peers the startup discovery/sync gate should wait for before
+/// a node is allowed to consider itself synced and arm minting.
+///
+/// This is the discriminator that seeds `initial_sync_complete` (via
+/// `min_peers_to_wait == 0`) and drives the discovery wait loop. It must return:
+///
+/// - `Recommended` mode: the configured `min_peers`. A `min_peers == 0` config
+///   is a genuine solo node and is seeded synced (unchanged pre/post #770).
+/// - `Explicit` mode with a **non-empty** member set: `1`. The operator has
+///   declared trusted peers, so the node genuinely expects a fleet and must keep
+///   the #770 startup gate — wait for >=1 peer, sync to the tip, then mint. This
+///   is what prevents a restarted node from re-entering SCP at a stale slot while
+///   the fleet is ahead (#766/#998/#1000).
+/// - `Explicit` mode with an **empty** member set: `0`. Such a quorum is
+///   solo-by-design — its threshold is satisfied by self alone with zero peers
+///   (`Config::can_reach_quorum` reports it minteable), so there is no fleet to
+///   fall behind. Keeping the peer-wait gate here would deadlock forever: the
+///   `has_connected_peer` gate in `sync.rs` `tick()` can never become true, so
+///   `initial_sync_complete` never flips and minting is deferred indefinitely.
+///   Seeding `0` restores pre-#770 solo behavior for this config (issue #1056).
+fn min_peers_to_wait_for_startup(quorum: &QuorumConfig) -> usize {
+    match quorum.mode {
+        QuorumMode::Explicit => {
+            if quorum.members.is_empty() {
+                0
+            } else {
+                1
+            }
+        }
+        QuorumMode::Recommended => quorum.min_peers as usize,
+    }
+}
+
 /// Build the consensus config, honoring an optional test-only fixed-timing
 /// override.
 ///
@@ -358,10 +391,7 @@ async fn run_async(mut config: Config, config_path: &Path, mint: bool) -> Result
     let mut swarm = discovery.start().await?;
 
     // Determine minimum peers to wait for based on quorum config
-    let min_peers_to_wait = match config.network.quorum.mode {
-        QuorumMode::Explicit => 1, // In explicit mode, wait for at least one peer
-        QuorumMode::Recommended => config.network.quorum.min_peers as usize,
-    };
+    let min_peers_to_wait = min_peers_to_wait_for_startup(&config.network.quorum);
 
     // Wait for peers with timeout
     info!(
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the #770 solo-node regression fix. The carve-out is correct and appropriately scoped.

**Correctness of the carve-out (the load-bearing question):**
- `min_peers_to_wait_for_startup()` keys on `quorum.members.is_empty()`, NOT `threshold <= 1`. This is the right discriminator: in Explicit mode a genuinely networked node *must* enumerate its trusted peers, so an empty member set is solo-by-design and has no fleet to fall behind. A multi-member fleet (even `threshold=1`) still returns `1` and keeps the full #770 gate.
- The seed derivation at `run.rs:841` (`initial_sync_complete = min_peers_to_wait == 0`) is unchanged, so the helper is the single source of truth for both the discovery wait loop and the sync-gate seed. No divergence.
- **(a) Cannot weaken the gate for networked/multi-member nodes:** confirmed — only the empty-member Explicit case flips to `0`; every members-listed config and all of Recommended mode are byte-for-byte unchanged.
- **(b) No stale/forked-chain mint risk:** the only newly-armed case is a peerless, member-less quorum that is its own sole authority — there is no peer set to fork from or fall behind, so the #766/#998/#1000 stale-slot hazard does not apply.

**Tests (verified locally):** `cargo test -p botho --lib commands::run::` → 39 passed, 0 failed, including the three new cases: solo-arms (empty members → 0), members-still-gated (threshold=1 *and* threshold=2 multi-member → 1), and Recommended-unchanged (min_peers 0 and 3).

**Fixed during evaluation:** the initial push failed `rustfmt --check (pinned nightly)` — the new doc comment exceeded the `wrap_comments` width. I re-wrapped it with the pinned `nightly-2025-12-03` toolchain (doc-comment lines only, zero logic change) and pushed. CI is now fully green:
- rustfmt --check (pinned nightly) — pass
- cargo build --workspace (Linux) — pass
- cargo deny — pass
- Compile Benchmarks / Node / Crypto / Transaction Benchmarks — pass
- Fuzz Build Guard — pass

Merge state: CLEAN. Scope is tight (one helper + call site + three tests), doc comment thoroughly explains the deadlock rationale. Nice work.

PR #1071, opened 2026-07-17
fix(bridge): u128-accumulate /api/bridge/stats volume to avoid i64/u64 overflow (#1059)

### LINKED ISSUE #1059: bridge: /api/bridge/stats all-time SUM(amount) can overflow i64 at ~9.2M BTH gross volume
## Context

While reviewing PR #1058 (public `GET /api/bridge/stats`, #1054): `Database::aggregate_order_activity` computes gross volume with SQLite `SUM(amount)` over `bridge_orders`. SQLite integer SUM raises an "integer overflow" error when the running total exceeds `i64::MAX`.

`i64::MAX` picocredits is approximately 9.22M BTH. With a ~611M BTH eventual supply, cumulative all-time gross bridge volume crossing that threshold is plausible on a mainnet horizon. When it happens, the aggregation query errors, and the handler maps that to a permanent 500 ("stats unavailable") — a safe failure (no wrong numbers served, per the #541 lesson), but the endpoint and the /network card would degrade permanently.

## Proposed fix options

- Compute per-status sums in Rust with u128 accumulation (stream rows instead of SQL SUM), or
- `SUM(CAST(amount AS REAL))`/`TOTAL(amount)` with documented precision loss (probably not acceptable for reporting), or
- widen the serialized volume to a u128-safe decimal string and sum via `CAST` to text-safe chunks.

Rust-side u128 accumulation is likely the simplest: the query already does a single pass, and the response already serializes volumes as decimal strings.

## Scope

- `bridge/service/src/db.rs` (`aggregate_order_activity`)
- Note: `amount` is stored as `i64` per row (pre-existing storage convention across the DB layer); per-order amounts are fine — only the aggregate sum is at risk.

Non-blocking follow-up from the PR #1058 review; failure mode today is safe-but-degraded.

---

### Problem Statement

`Database::aggregate_order_activity` (`bridge/service/src/db.rs:751`) is the sole data source for the public `GET /api/bridge/stats` endpoint (#1054, shipped in PR #1058). It computes gross wrap/unwrap volume with SQLite `COALESCE(SUM(amount), 0)` grouped by status:

```sql
SELECT status, COUNT(*), COALESCE(SUM(amount), 0)
FROM bridge_orders
WHERE order_type = ?1 AND created_at >= ?2
GROUP BY status
```

SQLite's integer `SUM()` returns an `INTEGER` (i64) and **raises an "integer overflow" runtime error** the moment the running total for any status group exceeds `i64::MAX`. Because amounts are picocredits and 1 BTH = 1e12 picocredits, `i64::MAX` ≈ **9.22M BTH per status bucket**. Against a ~611M BTH eventual supply, a single long-lived `completed` bucket crossing that threshold on a mainnet horizon is entirely plausible.

When the SUM overflows, `query_map` yields a row error, `aggregate_order_activity` returns `Err`, and `get_bridge_stats` (`bridge/service/src/public_api.rs:990-993`) maps it to a permanent `500 "stats unavailable"`. The failure is *safe* (no wrong numbers are served — consistent with the #541 "never fake observability" lesson) but *permanent and total*: every window (wraps/unwraps × 24h/all-time) 500s because one all-time bucket overflowed, and the /network bridge-activity card degrades permanently with no self-recovery.

### Root Cause

Two nested integer-width ceilings, both below mainnet-scale cumulative volume:

1. **SQL layer (the trigger, ~9.22M BTH):** SQLite integer `SUM(amount)` accumulates in i64 and hard-errors on overflow. This is what surfaces first.
2. **Rust layer (a latent second ceiling, ~18.4M BTH):** even if the SQL SUM is removed, the accumulator field `ActivityAggregate.volume` is `u64` (`db.rs:151`), and `db.rs:792` does `bucket.volume.saturating_add(volume.max(0) as u64)`. `u64::MAX` ≈ 18.4M BTH, so a Rust-side u64 accumulation would *silently saturate* (worse than erroring — it serves a wrong-but-plausible number) at roughly double the SQL ceiling.

A complete fix must widen **both** the accumulation and the stored aggregate to `u128`. `u128` picocredits covers ~3.4e26 BTH — far beyond any conceivable supply — so saturation becomes unreachable in practice. The JSON boundary already serializes volume as a decimal **string** (`StatsAggregate.volume: String`, `public_api.rs:341`, `volume: a.volume.to_string()` at `:348`), so widening to u128 needs **no JSON/client changes** — `u128::to_string()` drops in unchanged and the "u64-safe decimal string" doc comment simply becomes "u128-safe".

### Recommended Approach: Rust-side u128 accumulation (exact, no float loss)

Replace the SQL `SUM(amount)` with a streamed per-row fetch, accumulating in u128 in Rust. This is exact (no floating-point rounding, unlike the `CAST(amount AS REAL)`/`TOTAL()` option, which is unacceptable for financial reporting) and keeps the single-pass shape the query already has.

Concretely:

1. Change the query to select raw per-row amounts instead of an aggregate. Keep the `GROUP BY status` shape *or* select `status, amount` per row and bucket in Rust. Selecting per-row avoids the SQL SUM entirely:
   ```sql
   SELECT status, amount
   FROM bridge_orders
   WHERE order_type = ?1 AND created_at >= ?2
   ```
   Read `amount` as `i64` (unchanged storage convention), clamp negatives with `.max(0)`, and accumulate `count` (u64) and `volume` (u128) per status bucket in the existing match arms (`db.rs:784-792`).

2. Widen `ActivityAggregate.volume` from `u64` to `u128` (`db.rs:151`) and keep `saturating_add` for defense-in-depth (it will never actually saturate at u128 width).

3. No change required at `public_api.rs:348` — `a.volume.to_string()` already produces the correct decimal string for u128. Update the doc comments at `db.rs:150-151` and `public_api.rs:335-336` from "u64-safe" to "u128-safe".

Note on `count`: `COUNT(*)` cannot realistically overflow (u64 rows) and its accumulator is already `u64` (`db.rs:791`), so only `volume` needs widening.

Trade-off vs. streaming cost: switching from `GROUP BY status` aggregation to per-row streaming reads N rows into the process instead of ≤4 grouped rows. This is acceptable because the endpoint is cached (`STATS_CACHE_TTL`, `public_api.rs:947`) so the query runs at most once per TTL, and `prune_expired_mint_orders` (`db.rs:806`) bounds the abandoned-order residue. If per-row streaming is later found too costly at very large row counts, the alternative is to keep `GROUP BY status` but push the sum into u128 via SQLite text: `GROUP BY status` selecting `CAST(SUM(amount) AS TEXT)` still overflows i64 inside SUM, so that does not help — per-row streaming (or a windowed/chunked sum) is the exact path. Recommend per-row streaming for simplicity.

### Acceptance Criteria

- [ ] `aggregate_order_activity` no longer uses SQLite integer `SUM(amount)`; gross volume is accumulated in Rust as `u128`.
- [ ] `ActivityAggregate.volume` is `u128` (not `u64`); accumulation cannot silently saturate at mainnet-scale volumes.
- [ ] A regression test proves no overflow error and a correct total when a single status bucket's summed volume exceeds `i64::MAX` (~9.22M BTH in picocredits) — e.g. two rows of `i64::MAX` sum to `2 * i64::MAX` exactly.
- [ ] A regression test proves no silent saturation past `u64::MAX` (~18.4M BTH) — i.e. the u128 widening is actually exercised, not just the i64 ceiling.
- [ ] `GET /api/bridge/stats` returns `200` with correct decimal-string volumes at these volumes (no permanent 500; no wrong-but-plausible saturated number).
- [ ] Existing `aggregate_order_activity` unit tests (`db.rs:3641-3740`) still pass unchanged (the small-value assertions must continue to hold with u128 fields).
- [ ] Per-order behavior is unchanged; only the aggregate path is touched. No JSON schema change (volume stays a decimal string).
- [ ] Doc comments updated from "u64-safe" to "u128-safe" where they describe the volume width (`db.rs:150-151`, `public_api.rs:335-336`).

### Affected Files

- `bridge/service/src/db.rs` — `aggregate_order_activity` (`:751`): replace `COALESCE(SUM(amount), 0)` query with per-row `status, amount` fetch + Rust u128 accumulation (`:757-793`). Widen `ActivityAggregate.volume` from `u64` to `u128` (`:151`). Update doc comment (`:150-151`). Add/adjust regression tests in the existing `#[cfg(test)]` block (`:3641-3740`) for the >i64::MAX and >u64::MAX cases.
- `bridge/service/src/public_api.rs` — verify `StatsAggregate.volume: String` (`:341`) and `volume: a.volume.to_string()` (`:348`) still compile against a u128 field (they do). Update the "u64-safe decimal STRING" doc comment (`:335-336`) to "u128-safe". No functional change to the handler.

### Test Plan

- [ ] Automated (unit, `bridge/service/src/db.rs` test module): add a test that inserts two orders whose amounts sum past `i64::MAX` in one status bucket, calls `aggregate_order_activity`, and asserts the returned `volume` equals the exact sum (no `Err`, no truncation).
- [ ] Automated (unit): add a test that pushes a bucket past `u64::MAX` (e.g. multiple `i64::MAX` rows) and asserts the exact u128 total — proving no saturation.
- [ ] Automated (regression): confirm existing tests `test_aggregate_order_activity_buckets_mint_orders`, `..._buckets_burn_orders_separately`, `..._empty_db_is_all_zero` still pass with the widened field.
- [ ] Manual verification: with a seeded large-volume DB, `curl` `GET /api/bridge/stats` and confirm `200` with a correct all-time volume decimal string (not a 500, not a saturated value).
- [ ] Edge cases: exactly `i64::MAX` in a single row (per-order storage boundary — must not be clamped or errored); a bucket at exactly `u64::MAX + 1`; all-zero / empty DB (unchanged behavior); negative-amount defense (`.max(0)`) preserved.

### References

- Query + accumulation: `bridge/service/src/db.rs:751-795`
- Aggregate struct: `bridge/service/src/db.rs:145-168`
- Handler + error→500 mapping: `bridge/service/src/public_api.rs:926-995`
- JSON string serialization of volume: `bridge/service/src/public_api.rs:334-352`
- Origin: PR #1058 (#1054 public stats endpoint) review follow-up
- Related lesson: #541 (never serve fake/wrong observability numbers — informs the "exact, no float loss" requirement)

rjwalters:
## Summary

`Database::aggregate_order_activity` (`bridge/service/src/db.rs`) — the sole data source for the public `GET /api/bridge/stats` endpoint — computed gross wrap/unwrap volume with SQLite integer `SUM(amount)`. SQLite's integer SUM **hard-errors** ("integer overflow") once a status bucket's total exceeds `i64::MAX` (~9.22M BTH in picocredits), which `get_bridge_stats` maps to a **permanent 500**. Every window would 500 because one all-time bucket overflowed, and the /network bridge-activity card would degrade with no self-recovery.

Curation flagged a **second latent ceiling**: the Rust accumulator `ActivityAggregate.volume` was `u64` with `saturating_add`, so simply moving the sum into Rust at u64 width would trade a safe-but-permanent 500 for a *silent saturation* (~18.4M BTH) — strictly worse per the #541 "never serve wrong observability numbers" lesson.

## Fix

- Replace SQLite `SUM(amount)` with a streamed per-row `SELECT status, amount` fetch and accumulate volume in Rust as **`u128`** — exact, no float precision loss.
- Widen `ActivityAggregate.volume` from `u64` to `u128`. u128 picocredits covers ~3.4e26 BTH, so saturation is unreachable in practice; `saturating_add` kept as defense-in-depth.
- **No JSON/schema/client change**: the boundary already serializes volume as a decimal string (`StatsAggregate.volume: String`), and `u128::to_string()` drops in unchanged.
- Doc comments updated from "u64-safe" to "u128-safe".

## Tests

- `test_aggregate_order_activity_no_i64_overflow_past_i64_max` — two `i64::MAX` rows in one bucket sum to `2 * i64::MAX` with no error (old SQL path would have errored).
- `test_aggregate_order_activity_no_u64_saturation_past_u64_max` — three `i64::MAX` rows sum past `u64::MAX` to the **exact** u128 total (proves the widening is exercised, not saturated).
- `test_aggregate_order_activity_single_row_at_i64_max_boundary` — a single `i64::MAX`-amount row round-trips unchanged.
- Existing `aggregate_order_activity` unit tests pass unchanged with the widened field.

Full `bth-bridge-service` suite: **239 passed, 0 failed**.

## Scope

Only `bridge/service/src/db.rs` and `bridge/service/src/public_api.rs` touched. Per-order behavior and the JSON schema are unchanged.

Closes #1059

### DIFF
--- a/bridge/service/src/db.rs
+++ b/bridge/service/src/db.rs
@@ -147,8 +147,11 @@ pub struct ReleaseIntent {
 pub struct ActivityAggregate {
     /// Number of orders in the bucket.
     pub count: u64,
-    /// Sum of gross order amounts, picocredits.
-    pub volume: u64,
+    /// Sum of gross order amounts, picocredits. `u128` so cumulative all-time
+    /// volume cannot overflow at mainnet scale (#1059): a `u64` sum saturates
+    /// at ~18.4M BTH and a SQLite `SUM(amount)` errors at ~9.22M BTH, both
+    /// below plausible cumulative bridge volume. `u128` covers ~3.4e26 BTH.
+    pub volume: u128,
 }
 
 /// Wrap/unwrap activity aggregates for one order type over one window
@@ -733,8 +736,8 @@ impl Database {
     /// Aggregate wrap/unwrap activity for one order type (#1054).
     ///
     /// Backs the public `GET /api/bridge/stats` endpoint: counts and gross
-    /// BTH volumes (picocredits, `SUM(amount)`) over `bridge_orders` rows of
-    /// `order_type`, bucketed by outcome:
+    /// BTH volumes (picocredits) over `bridge_orders` rows of `order_type`,
+    /// bucketed by outcome:
     ///
     /// * `completed` — settled orders (`completed` for mints, `released` for
     ///   burns; both strings map here so either type aggregates correctly).
@@ -748,19 +751,28 @@ impl Database {
     ///
     /// AGGREGATES ONLY: no per-order field leaves this query, so the result
     /// is safe for the unauthenticated public surface (#1042 scope rules).
+    ///
+    /// Volume is summed in Rust as `u128` rather than via SQLite integer
+    /// `SUM(amount)` (#1059): SQLite's integer SUM hard-errors on i64 overflow
+    /// (~9.22M BTH/bucket) which would 500 the endpoint permanently, and a
+    /// `u64` accumulator would silently saturate at ~18.4M BTH. Streaming the
+    /// raw `status, amount` per row and accumulating in u128 is exact (no
+    /// float loss) and cannot overflow at any realistic supply.
     pub fn aggregate_order_activity(
         &self,
         order_type: OrderType,
         since: Option<i64>,
     ) -> Result<ActivityBreakdown, String> {
         let conn = self.conn.lock().map_err(|e| format!("Lock error: {}", e))?;
+        // Stream raw per-row (status, amount) instead of a SQLite
+        // `SUM(amount)`: integer SUM errors on i64 overflow (#1059). Volume is
+        // accumulated in Rust as u128 so it cannot overflow at mainnet scale.
         let mut stmt = conn
             .prepare(
                 r#"
-                SELECT status, COUNT(*), COALESCE(SUM(amount), 0)
+                SELECT status, amount
                 FROM bridge_orders
                 WHERE order_type = ?1 AND created_at >= ?2
-                GROUP BY status
                 "#,
             )
             .map_err(|e| format!("Prepare failed: {}", e))?;
@@ -768,19 +780,13 @@ impl Database {
         let rows = stmt
             .query_map(
                 params![order_type.to_string(), since.unwrap_or(i64::MIN)],
-                |row| {
-                    Ok((
-                        row.get::<_, String>(0)?,
-                        row.get::<_, i64>(1)?,
-                        row.get::<_, i64>(2)?,
-                    ))
-                },
+                |row| Ok((row.get::<_, String>(0)?, row.get::<_, i64>(1)?)),
             )
             .map_err(|e| format!("Query failed: {}", e))?;
 
         let mut breakdown = ActivityBreakdown::default();
         for row in rows {
-            let (status, count, volume) = row.map_err(|e| format!("Row failed: {}", e))?;
+            let (status, amount) = row.map_err(|e| format!("Row failed: {}", e))?;
             let bucket = match status.as_str() {
                 "completed" | "released" => &mut breakdown.completed,
                 "expired" => &mut breakdown.expired,
@@ -788,8 +794,10 @@ impl Database {
                 s if s.starts_with("failed") => &mut breakdown.failed,
                 _ => &mut breakdown.pending,
             };
-            bucket.count = bucket.count.saturating_add(count.max(0) as u64);
-            bucket.volume = bucket.volume.saturating_add(volume.max(0) as u64);
+            bucket.count = bucket.count.saturating_add(1);
+            // Clamp defensively (`.max(0)`); `saturating_add` never triggers at
+            // u128 width but is kept as defense-in-depth.
+            bucket.volume = bucket.volume.saturating_add(amount.max(0) as u128);
         }
         Ok(breakdown)
     }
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Solid, well-scoped fix for the `/api/bridge/stats` gross-volume overflow (#1059). Verified against all four review dimensions:

**(a) Accumulation is exact and non-overflowing.** Volume is summed in Rust as `u128` over per-row `i64` amounts (each ≤ `i64::MAX`). Overflowing `u128` would require ~2^65 max-value rows — unreachable at any realistic supply. Pure integer addition, so no float precision loss. `saturating_add` retained as defense-in-depth is a reasonable belt-and-suspenders choice.

**(b) JSON contract preserved.** `StatsAggregate.volume` is a decimal `String` and the boundary serializes via `a.volume.to_string()` (public_api.rs:348); `u128: Display` drops in unchanged. The existing round-trip test asserting string volumes (`"1000"`, `"5000"`, etc.) still passes under `cargo test (bridge crates)`, confirming no schema drift.

**(c) No material performance regression.** The old `SUM(amount) ... GROUP BY status` already required a full scan of the `order_type`/`created_at`-filtered rows; the new path scans the same rows and materializes them to Rust instead of aggregating in SQLite. The only added cost is per-row marshaling rather than SQLite-side reduction — marginal for a runtime read endpoint, and this is not on any build-time/deploy critical path. Acceptable tradeoff for exactness.

**(d) Regression tests genuinely cover both ceilings.**
- `..._no_i64_overflow_past_i64_max`: 2×`i64::MAX` crosses the old SQLite integer-SUM error ceiling (asserts `expected > i64::MAX`).
- `..._no_u64_saturation_past_u64_max`: 3×`i64::MAX` ≈ 2.77e19 crosses `u64::MAX` and asserts the **exact** total, proving the widening is exercised, not saturated (asserts `expected > u64::MAX`).
- `..._single_row_at_i64_max_boundary`: per-order storage boundary round-trips unchanged.

All three pass locally (6/6 in the `aggregate_order_activity` group), and the fix correctly avoids trading a safe-but-permanent 500 for the strictly-worse silent-saturation failure mode called out in the #541 "never serve wrong observability numbers" lesson.

**Fixed during evaluation:** the pinned `nightly-2025-12-03` rustfmt gate was failing on a one-line `assert!` at db.rs:3773 (stable rustfmt formats it differently, so the builder's local `cargo fmt` missed it). I ran the pinned nightly rustfmt, which split it across lines, and pushed the fix (commit f8eb9d39). CI is now fully green.

## Test Execution
- `cargo test -p bth-bridge-service aggregate_order_activity` — ✅ 6 passed, 0 failed (incl. the 3 new regression tests)
- CI `cargo test (bridge crates)` — ✅ pass
- CI `cargo build --workspace (Linux)` — ✅ pass
- CI `rustfmt --check (pinned nightly)` — ✅ pass (after fix)
- CI `cargo deny`, `Fuzz Build Guard (PR)` — ✅ pass
- Merge state: CLEAN

PR description uses `Closes #1059`. Diff is confined to `bridge/service/src/db.rs` and `bridge/service/src/public_api.rs` as claimed.

Issue #1050, opened 2026-07-17
bridge(federation): replicate order records across federation members (deterministic burn-order ids + order sync)


rjwalters:
## Status — re-curated 2026-07-18 (code-layer scope)

> **Phase 1 is DONE. This issue now tracks Phase 2 only.**

**Phase 1 (deterministic burn-order ids) shipped in merged PR #1072.** Verified on `origin/main`:
- `bridge/core/src/order.rs:52` — `derive_burn_order_uuid(source_chain, tx_hash, ordinal) -> Uuid` exists, domain-separated by `BURN_ORDER_ID_DOMAIN_TAG` (`order.rs:21`, SHA256 + `Uuid::from_bytes`, matching the existing `derive_order_id` style).
- `order.rs:382` — `new_burn` now derives the id via `derive_burn_order_uuid` (`:392`); the random `Uuid::new_v4()` is gone from the burn path (it remains only for `new_mint`, `:353`, which is correct — see below).
- Both watchers feed the deterministic tuple: `watchers/ethereum.rs` (`Chain::Ethereum` + tx_hash + ordinal) and `watchers/solana.rs` (`Chain::Solana` + signature + ordinal).
- Stability/collision unit tests present (`order.rs:618-641`).

The **release/burn leg** therefore already aggregates across separate-DB members with zero cross-member trust. **Remaining open work = Phase 2: mint order-record replication** (the mint leg), described in detail below.

### Re-scope per maintainer direction #1060 (ratified 2026-07-17)

The bridge federation will **not** mirror the SCP quorum graph. Plan of record is a **small elected multisig of ≤5 machines** with periodic elections + scheduled rotation outages (see #1060, ADR 0010). Phase 2 order-record replication therefore only needs to work for **≤5 members** — a fixed, small, fully-connected peer set — **not** a consensus-scale gossip mesh. This *simplifies* Phase 2: broadcast-to-all-peers-on-create over the existing `PeerBroadcaster` fan-out is sufficient; no epidemic/anti-entropy gossip protocol is warranted at this size. The `AwaitingDeposit`-shell + independent-confirmation trust boundary (below) is unchanged and remains the safety-critical part.

### Scope guard for this issue (Phase 2, code-layer only)

[...]

PR #1072, opened 2026-07-17
feat(bridge): deterministic burn-order ids (#1050 Phase 1)

rjwalters:
## Part of #1050 (Phase 1: deterministic burn-order ids)

Implements **Phase 1 only** of the curated 2-phase decomposition of #1050. Phase 2 (mint order-record replication / signed order-gossip / `POST /api/attest` changes) is intentionally **out of scope** and remains open.

### Problem
Each federation member's Ethereum/Solana burn watcher independently observes the *same* finalized on-chain burn and calls `BridgeOrder::new_burn`, which assigned a **random** `Uuid::new_v4()`. Two members therefore minted two different order ids for one burn, so their attestation envelopes — bound via `order_id_bytes` → `derive_order_id` — never aggregated to threshold on a genuinely distributed (separate-DB) federation.

### Change
Derive the burn-order UUID **deterministically** from the stable burn-source tuple `(source_chain, tx_hash, ordinal)` — exactly the watchers' existing `burn_source_key` (`"<tx_hash>#<ordinal>"`) plus the chain. Two independent instances observing the same finalized burn now derive the **identical** order id and aggregate attestations with zero cross-member trust and no shared store.

- **`bridge/core/src/order.rs`**: new `derive_burn_order_uuid()` — domain-separated SHA256 over the length-prefixed source tuple, RFC-4122 v5 version/variant bits (mirrors the existing `derive_order_id` pattern, no new Cargo feature). `new_burn()` gains an `ordinal` param and derives the id internally.
- **`bridge/service/src/watchers/ethereum.rs` / `watchers/solana.rs`**: pass the per-tx `ordinal` (already computed for `burn_source_key`).
- **`bridge/service/src/db.rs`**: `insert_burn_order` now inserts the order row with `INSERT OR IGNORE`, so a cursor replay or a concurrent watcher presenting the same deterministic id is a clean no-op instead of a `UNIQUE constraint` hard-error. `processed_burns` (unique `source_key`) remains the exactly-once arbiter; FK ordering (order row before `processed_burns`) is preserved.
- Updated all other `new_burn` callers for the new signature. Tests that previously relied on two same-shape orders having *different random ids* now set a distinct id explicitly to exercise cross-order rejection in isolation.

### Tests
New in `order.rs`:
- `test_burn_order_id_is_deterministic_across_independent_instances` — two independent `new_burn` calls for the same tuple (even with different local fee) yield the **same** id, equal to the pure helper.
- `test_burn_order_id_differs_for_distinct_source_tuples` — distinct ordinal / tx-hash / chain each yield distinct ids.
- `test_burn_order_uuid_is_well_formed_v5` — version + variant bits.

`cargo build`, `cargo test -p bth-bridge-core -p bth-bridge-service` (74 + 239 pass), and `cargo clippy` on both crates are green.

### Scope confirmation
Phase 2 (signed mint order-gossip endpoint, trust-boundary hardening) is **NOT** included here — it is a separate, security-sensitive follow-up under #1050. This PR uses "Part of #1050" (not "Closes") so the issue stays open for Phase 2.

### DIFF
--- a/bridge/core/src/adversarial_tests.rs
+++ b/bridge/core/src/adversarial_tests.rs
@@ -61,6 +61,7 @@ fn burn_order() -> BridgeOrder {
         "0x1234567890abcdef1234567890abcdef12345678".to_string(),
         "bth_user_stealth_addr".to_string(),
         "0xburntx".to_string(),
+        0,
     )
 }
 
--- a/bridge/core/src/attestation.rs
+++ b/bridge/core/src/attestation.rs
@@ -1638,6 +1638,7 @@ mod tests {
             "0x1234567890abcdef1234567890abcdef12345678".to_string(),
             "bth_user_stealth_addr".to_string(),
             "0xburntx".to_string(),
+            0,
         )
     }
 
@@ -1978,13 +1979,18 @@ mod tests {
     fn test_order_binding_rejects_cross_order_reuse() {
         let sk = signing_key(1);
         let order_a = burn_order_from_eth();
-        let order_b = burn_order_from_eth(); // fresh UUID, same shape
+        // Same shape but a DIFFERENT id. Since #1050 made burn ids
+        // deterministic over the source tuple, a second order with an
+        // identical source tuple would derive the same id; we set a distinct
+        // id explicitly to exercise the id-mismatch branch in isolation.
+        let mut order_b = burn_order_from_eth();
+        order_b.id = Uuid::from_u128(order_b.id.as_u128() ^ 1);
 
         let env = sign(&release_kind(&order_a), &sk, "aa");
         let parsed = env.verify_and_parse_ed25519(&sk.verifying_key()).unwrap();
 
-        // Bound to order A: order B is rejected even though every other
-        // field (amount, recipient, chains, source tx) matches.
+        // Bound to order A: order B is rejected on the id mismatch even though
+        // every other field (amount, recipient, chains, source tx) matches.
         assert!(check_order_binding(&parsed, &order_a).is_ok());
         assert!(matches!(
             check_order_binding(&parsed, &order_b),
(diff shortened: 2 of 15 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the deterministic burn-order id change. This is a clean, correct, standalone-valuable Phase 1.

## Correctness / security verification

**(a) Determinism + collision resistance** — `derive_burn_order_uuid` is a pure function of `(source_chain, tx_hash, ordinal)`: domain-separated SHA256 (`BURN_ORDER_ID_DOMAIN_TAG`), with the chain tag and tx hash **length-prefixed** (`u32` LE) and the fixed-width `u32` ordinal last. No concatenation/field-boundary ambiguity is possible, and every input is a stable byte encoding, so the value is identical across processes and platforms. Truncation to 128 bits with v5/variant bits set leaves ~122 bits — collision is cryptographically negligible. Mirrors the established `derive_order_id` pattern. `test_burn_order_uuid_is_well_formed_v5` confirms the version/variant bits.

**(b) Ordinal is the right disambiguator and consistently sourced** — the `ordinal` fed to `new_burn` is the *exact same* value already feeding `burn_source_key` in both watchers:
- Ethereum: `with_tx_ordinals` sorts by `(block, log_index)` then assigns per-tx counters, so the intra-tx ordinal is stable across reorgs and identical for every member (relative log order within a tx is fixed).
- Solana: `enumerate()` over the tx's decoded burns, whose instruction order is fixed.

Both `source_key = burn_source_key(tx, ordinal)` and `new_burn(..., ordinal)` use one `ordinal` in each watcher, so the id and the exactly-once key can never diverge.

**(c) `INSERT OR IGNORE` cannot mask a real problem** — it is applied **only** on the burn path (`insert_burn_order`); the mint path still uses plain `INSERT` and hard-errors on duplicate id. Within the burn transaction, `processed_burns` (unique `source_key`) remains the exactly-once arbiter: `changed == 0 → Ok(false)` drops the whole transaction, rolling back any order row. A distinct order colliding on id would require a SHA256 collision (negligible), and a same-tuple replay derives the same id by construction — exactly the idempotent no-op intended. FK ordering (order row before `processed_burns`) is preserved.

**(d) Exactly-once invariant preserved** — `chaos_tests::load_test_concurrent_orders_exactly_once` and `adversarial_tests::prop_reorg_finality_fuzz_is_exactly_once` pass. The updated `insert_burn_order` replay test now asserts `dup.id == first.id` and that exactly one unchanged row survives.

**(e) Cross-order rejection coverage not reduced** — the tests that previously leaned on two random UUIDs being distinct now set an explicit distinct id (`Uuid::from_u128(id ^ 1)`). This makes the id-mismatch branch (`test_order_binding_rejects_cross_order_reuse`, threshold-set insert rejection, federation 404) exercised *intentionally* rather than incidentally — coverage is preserved and arguably clearer.

## Scope
Phase 2 (signed mint order-gossip / trust-boundary hardening) is correctly scoped out; the PR uses "Part of #1050" (not "Closes") so the issue stays open — appropriate, since Phase 1 stands on its own.

## Test Execution
- `cargo test -p bth-bridge-core -p bth-bridge-service` — ✅ 74 + 239 pass locally
- CI: all checks green (cargo build --workspace, cargo test bridge crates, cargo deny, rustfmt, hardhat, Fuzz Build Guard) — merge state CLEAN

Excellent work — the design rationale in the code comments is exemplary.

Issue #1066, opened 2026-07-17
bridge(build): adopt v2 term-document schema in the rotation driver (replace v1 mock; select-then-keygen sealing; resolve #1064 schema nits)


rjwalters:
Adopt the v2 term-document schema (`docs/bridge/election-dynamics.md` §5.2) in the bridge rotation driver, replacing the v1 mock format the drill currently emits.

## Context
ADR 0010 ratified option C. The #1063 rotation drill's `rotate-elect` step currently emits a v1 mock `term-<K>.json`; the election-dynamics doc defines the canonical **v2** schema that the real election will produce. This issue makes the drill consume/emit v2 so the drill and the eventual live election share one artifact format.

## Scope
- In `scripts/bridge-testnet-federation.sh`, update the `rotate-elect` step (and everything downstream that reads `term-<K>.json`) to the v2 schema: `termNumber`, `electionKind` (`scheduled`/`emergency`), electorate snapshot reference (`curationDocHash`), `status` (`elected` → `sealed`), elected members with per-member fresh-key submission (`keySubmissionSig`), per-chain `executionIntents` (Safe swaps / Squads config / BTH reserve re-key), `validity` window incl. `handoverDeadline`, and the signatures section (`tallyAttestations[]` + outgoing-set counter-signature at threshold k).
- Implement the **select-then-keygen** lifecycle: the mock election pins membership (`status: elected`); a subsequent seal step collects fresh per-term keys signed by each winner's long-lived identity key and flips to `status: sealed`.
- Resolve the three schema nits the #1064 judge flagged: (a) an `if/then` conditional making per-term keys REQUIRED when `status == "sealed"`; (b) document exactly which lifecycle snapshot each `tallyAttestations[].sig` covers; (c) replace look-alike example addresses in §5.3 with canonical/clearly-fake ones.
- Ship a JSON Schema file (e.g. `docs/bridge/schemas/term-document.v2.schema.json`) and validate the worked example + the drill's emitted document against it in a test.
- Update runbook Phase D to reference the v2 document.

## Acceptance
- The drill emits a v2 term document that validates against the committed JSON Schema; the elected→sealed transition (with key submission) is exercised; the three nits are resolved. Docs + driver + a validation test.

Part of #1060. Refs ADR 0010, #1062/PR #1064, #1061/PR #1063.

PR #1063, opened 2026-07-17
feat(bridge): e2e key-rotation drill — mock same-set election, full re-key handover

rjwalters:
## Summary

Adds the end-to-end **rotation drill** (#1061) to the testnet federation driver, serving the #1060 elected-multisig direction: a MOCK election re-elects the SAME member set — membership-stable, keys-fresh — and the entire handover machinery then runs with fresh keys on every custody surface, with **old-keys-are-powerless assertions gating the resume**. The drill was executed LIVE against the Phase C topology (3 instances, threshold 2, live Sepolia Safe) as part of this PR; the drill log is in the runbook.

## Changes

- **`scripts/bridge-testnet-federation.sh`**: new `rotate` phase (plus granular `rotate-elect|pause|drain|keys|safe|solana|bth|restart|verify|resume|attest`):
  - Mock same-set election emitting `federation/election/term-<K>.json` — the single seam a real #1060 election replaces (same schema, nothing else changes).
  - Breaker pause with public-surface verification; drain with hard (value-in-motion, never overridable) vs soft (parked-retryable, explicit recorded `BRIDGE_ROTATE_ACK_PARKED`) state discipline.
  - Fresh Ed25519 federation keys + attest bearer token + Safe owner secp256k1 keys; old material archived under `federation/retired/term-<K-1>/` (all writes under the `require_gitignored`-guarded secrets dir; LP relayer key deliberately not rotated — no roles, ADR 0002).
  - Live Sepolia Safe handover: `swapOwner(prev, old, new)` per member via 2-of-3 `execTransaction`, with the signing set migrating as swaps land.
  - Solana `SetAuthority` rotation (gated with exact operator commands when CLI tooling is absent) and BTH reserve wallet re-key (funds sweep operator-gated on #1051, per the issue).
  - `rotate-verify` — the resume gate: a **validly-signed** old-key attestation (canonical envelope + domain separation reproduced outside the Rust codebase via python+openssl) must be `refused:unknown_signer` on ALL instances; old key under a new signer id must be `refused:bad_signature`; old bearer token 401; old Safe owners `isOwner == false` AND an `execTransaction` eth_call simulation signed by old owners reverts GS026 while new owners simulate true.
  - `rotate-attest`: post-rotation threshold attestation round (audit `attestation_authorized` after resume) + proof-of-reserves drift exactly at the pre-pause baseline.
[...]

Issue #1060, opened 2026-07-17
architecture: bridge custody = small elected multisig (<5 machines), decoupled from SCP quorum structure, rotated via periodic elections


rjwalters:
## Direction (maintainer decision, 2026-07-17)

The bridge federation must NOT mirror the SCP consensus federation in the long run. SCP membership is a graph of overlapping quorum slices with no flat, globally-agreed signer set — there is no way to sign with that structure on any of our custody surfaces (Ethereum Safe owners, Solana Squads members, BTH reserve threshold attestation are all flat k-of-n lists).

**Plan of record:**

- The bridge is operated by a **small multisig of fewer than five machines** (e.g. 2-of-3 or 3-of-5), deliberately decoupled from SCP quorum structure.
- Bridge operators are chosen by **periodic elections among the consensus nodes** (cadence on the order of every few months).
- A **scheduled bridge outage** during each rotation is explicitly acceptable. The rotation doubles as **regular bridge key rotation** — a standing security win, not just a membership change.

## Why this direction

1. **Structural mismatch**: quorum slices ≠ signer lists. Keeping a bridge signer set synchronized with a drifting trust graph is a standing operational hazard; a small elected set is expressible on every chain we custody on.
2. **Dissolves #1050**: the "federation order records don't replicate" finding assumed the federation scales with the consensus set. With ≤5 elected machines, order-state replication is a small-cluster problem — or the shared-store topology the #1053 drill already validated.
3. **Key rotation by construction**: every election is a full re-key on all three chains. No separate rotation ceremony to schedule and forget.

## Design work this proposal needs (Architect scope)

- [ ] **Election mechanism + integrity**: electorate = the operator-curated consensus node set (P4.4 / #709 quorum curation is the natural sybil-resistance anchor). Define: ballot format, who tallies, where the result is recorded (on-chain vs operator-signed doc), tie/failure handling, and what happens if an elected node declines or dies mid-term.
- [ ] **Handover protocol (the hard part — must be atomic-ish)**: pause via the existing breaker → drain/settle all in-flight orders → rotate on all three chains — Safe owner swap (Ethereum), Squads member swap (Solana), and **BTH reserve re-key, which is a funds-moving transaction re-keying the reserve UTXOs to the new signer set** → verify old set has zero residual authority on every chain → resume. A voted-out quorum must be provably powerless before the pause lifts.
[...]

PR #1073, opened 2026-07-17
feat(bridge): on-chain ballot + deterministic tally tooling for elections (#1067)

### LINKED ISSUE #1067: bridge(build): on-chain ballot + deterministic tally tooling for bridge elections (option C / A1 memo convention)
Implement the on-chain ballot + deterministic tally tooling for bridge-multisig elections (option C / A1 memo convention).

## Context
ADR 0010 ratified option C with sub-variant **A1**: ballots are **memo-convention transactions on the Botho chain** (no protocol change, no new tx type), tallied by a deterministic off-consensus pure function at a published cutoff height over a snapshot of the P4.4 operator-curated node set (#709). This issue builds that mechanism (the piece the #1063 drill stubs with a mock election).

## Scope
- **Self-nomination memo** (candidacy is opt-in per ADR 0010): define the memo-convention format for a curated node to stand for a term, valid only before `openHeight`. One nomination per curated identity.
- **Ballot memo**: define the ballot format — approval voting (voter approves a subset of candidates), one-node-one-vote, signed by the node's long-lived curated identity key, valid within the voting window (`openHeight`..`closeHeight`).
- **Deterministic tally**: a pure function (library + CLI) that, given the ledger up to `closeHeight` and the curation snapshot at a pinned height, returns the top-N winners under the ratified rule (`approval-top-N-v1`, N per the 3-of-5 shape). Must be reproducible by anyone replaying the ledger — no wall-clock, no randomness, deterministic tie-break (specify it; ADR 0010/§6.3 references a tie-break — implement that rule).
- **Term-document assembly**: emit the elected-status v2 term document (membership only) from the tally result, ready for the seal/keygen step (that step may live in the schema-adoption issue — coordinate; this issue owns producing the `elected` document).
- Validation: unit tests over synthetic ledgers (clean win, tie, no-quorum, ineligible/duplicate ballots rejected, nomination window enforcement, curation-snapshot binding). No live chain needed.

## Non-goals
- Not the A2 dedicated-tx-type (explicitly deferred in ADR 0010).
- Not the cross-chain execution (that's the #1063 handover machinery).
- Not the seal/key-submission step if it lands in the schema-adoption issue — but the `elected` document this produces must be its input.

## Acceptance
- Given a synthetic ledger + curation snapshot, the tally CLI deterministically produces the correct top-N winners and an `elected`-status v2 term document; full failure-mode test coverage (tie, no-quorum, ineligible ballots, window enforcement).

Part of #1060. Refs ADR 0010, #1062/PR #1064, #709 (P4.4 curated set), #1061 (consumer of the output).

rjwalters:
Implements the ADR 0010 **option C / sub-variant A1** election mechanism: ballots are memo-convention transactions on the Botho chain (no protocol change, no new tx type), tallied off-consensus by a **deterministic pure function** at a published cutoff height over a pinned snapshot of the P4.4 operator-curated node set (#709).

## What landed

New `bth_bridge_core::election` module (all additive, pure — no async/IO):

- **`memo`** — the two memo-convention formats: self-nomination (`bridge.nominate`, valid only before `openHeight`, one per curated identity) and approval ballot (`bridge.ballot`, one-node-one-vote, valid within `openHeight..=closeHeight`). Canonical sorted-key/whitespace-free JSON, signed by the node's long-lived curated identity key under a dedicated domain (`botho.bridge.election.v1:`). Unknown/duplicate keys and float-where-int rejected fail-closed, mirroring the existing attestation-envelope discipline.
- **`snapshot`** — the pinned P4.4 curated electorate the tally binds to (`curationDocHash` + `snapshotHeight` + per-node identity pubkeys). One node, one vote; the tally reads the snapshot, never the live curation.
- **`tally`** — the `approval-top-N-v1` **pure function**. No wall-clock, no randomness; **input-order-independent** (memos sorted by `(height, txid)`); **deterministic, grind-free tie-break** = ascending lexicographic `nodeId` (ADR 0010 §6.3, explicitly *not* ledger-derived entropy); **full ranking** (not just top-N) for §6.2 exclusion-and-retry; quorum = strict majority of the electorate; failure modes (`NoQuorum`, `InsufficientCandidates`) carried as status, not errors. Emits a `resultHash` over the counted-ballot transcript so anyone replaying the ledger recomputes the identical result.
- **`term_doc`** — assembly of the **`elected`-status v2 term document (membership only)** — the input to the seal/keygen step.

## The #1066 seam (kept clean)

This PR ends at emitting the `elected` document. Member `keys`, `execution` intents, `keySubmissionSig`, and the `signatures` block are **deliberately absent** — they belong to the seal step (elected → sealed), which is #1066's scope per ADR 0010 §5.1. `assemble_elected_term_doc` refuses to produce a document from a non-`Elected` tally.

## CLI

`bridge-tally` (thin wrapper over the pure library):
- `bridge-tally tally <input.json>` → prints the elected term doc; exit `0` elected / `3` no-quorum / `4` insufficient candidates.
- `bridge-tally build-nomination` / `build-ballot` → canonical (optionally signed) memo builders for the #1063 drill and operators.

## Tests

- **20 unit tests** over synthetic ledgers: clean win, **tie-break** + tie-break symmetry under reordering, **reproducibility** (identical `resultHash`/winners under every ledger rotation + byte-identical re-run), no-quorum, insufficient candidates, ineligible/duplicate/forged ballots, nomination + ballot window enforcement, duplicate nomination, curation-snapshot binding, non-candidate-approval drop, and memo parse/round-trip/signature.
- **2 integration tests** driving the compiled `bridge-tally` binary end-to-end (elected doc output + no-quorum exit code).

`cargo test -p bth-bridge-core`: 91 lib + 2 integration pass. `cargo clippy --all-targets` clean. `cargo fmt --check` clean.

Closes #1067
Part of #1060

### DIFF
--- a/bridge/core/Cargo.toml
+++ b/bridge/core/Cargo.toml
@@ -29,5 +29,11 @@ sha2 = { workspace = true, features = ["std"] }
 ed25519-dalek = { workspace = true, features = ["std"] }
 serde_json = { workspace = true, features = ["std"] }
 
+# The deterministic bridge-election tally CLI (ADR 0010 option C / A1). A thin,
+# dependency-free wrapper over the pure `election` module.
+[[bin]]
+name = "bridge-tally"
+path = "src/bin/bridge-tally.rs"
+
 [dev-dependencies]
 tempfile = { workspace = true }
--- a/bridge/core/src/bin/bridge-tally.rs
+++ b/bridge/core/src/bin/bridge-tally.rs
@@ -1,0 +1,245 @@
+// Copyright (c) 2024 The Botho Foundation
+
+//! `bridge-tally` — the deterministic bridge-election tally CLI
+//! (ADR 0010 option C / A1).
+//!
+//! A thin, dependency-free wrapper over the pure [`bth_bridge_core::election`]
+//! library. Given a JSON bundle of `{ params, snapshot, ledger }`, it runs the
+//! `approval-top-N-v1` tally and prints the `elected`-status v2 term document
+//! (or the failure mode). It also builds/signs the two memo formats, so the
+//! #1063 drill and operators can produce ballots without re-implementing the
+//! canonical encoding.
+//!
+//! Everything is deterministic: same input → same output, byte for byte.
+//!
+//! ```text
+//! bridge-tally tally <input.json> [--ranking] [--out <file>]
+//! bridge-tally build-nomination --node-id <id> --term <n> [--secret-key-hex <hex>]
+//! bridge-tally build-ballot --node-id <id> --term <n> --approve <a,b,c> [--secret-key-hex <hex>]
+//! ```
+//!
+//! Exit codes for `tally`: 0 elected, 3 no-quorum, 4 insufficient candidates,
+//! 1 usage/parse error.
+
+use std::process::ExitCode;
+
+use bth_bridge_core::election::{
+    assemble_elected_term_doc, canonical_ballot_memo, canonical_nomination_memo,
+    sign_election_memo_ed25519, tally, CurationSnapshot, ElectionParams, MemoTransaction,
+    TallyResult, TallyStatus,
+};
+use ed25519_dalek::SigningKey;
+use serde::Deserialize;
+
+/// The JSON input bundle for the `tally` subcommand.
+#[derive(Debug, Deserialize)]
+struct TallyInput {
+    params: ElectionParams,
+    snapshot: CurationSnapshot,
+    ledger: Vec<MemoTransaction>,
+}
+
+fn main() -> ExitCode {
+    let args: Vec<String> = std::env::args().skip(1).collect();
+    if args.is_empty() {
+        eprintln!("{}", USAGE);
+        return ExitCode::from(1);
+    }
+
+    let result = match args[0].as_str() {
+        "tally" => cmd_tally(&args[1..]),
+        "build-nomination" => cmd_build_nomination(&args[1..]),
+        "build-ballot" => cmd_build_ballot(&args[1..]),
+        "-h" | "--help" | "help" => {
+            println!("{USAGE}");
+            return ExitCode::SUCCESS;
+        }
+        other => Err(format!("unknown subcommand `{other}`\n\n{USAGE}")),
+    };
+
+    match result {
+        Ok(code) => code,
+        Err(e) => {
+            eprintln!("error: {e}");
+            ExitCode::from(1)
+        }
+    }
+}
+
+const USAGE: &str = "\
+bridge-tally — deterministic bridge-election tally (ADR 0010 option C / A1)
+
+USAGE:
+  bridge-tally tally <input.json> [--ranking] [--out <file>]
+  bridge-tally build-nomination --node-id <id> --term <n> [--secret-key-hex <hex>]
+  bridge-tally build-ballot --node-id <id> --term <n> --approve <a,b,c> [--secret-key-hex <hex>]
+
+The `tally` input is a JSON object { \"params\": {...}, \"snapshot\": {...}, \"ledger\": [...] }.
+On an elected result it prints the elected-status v2 term document to stdout.
+
+EXIT CODES (tally): 0 elected, 3 no-quorum, 4 insufficient candidates, 1 error.";
+
+fn cmd_tally(args: &[String]) -> Result<ExitCode, String> {
+    let mut input_path: Option<String> = None;
+    let mut out_path: Option<String> = None;
+    let mut show_ranking = false;
+
+    let mut i = 0;
+    while i < args.len() {
+        match args[i].as_str() {
+            "--ranking" => show_ranking = true,
+            "--out" => {
+                i += 1;
+                out_path = Some(args.get(i).ok_or("--out needs a value")?.clone());
+            }
+            other if other.starts_with("--") => return Err(format!("unknown flag `{other}`")),
+            other => {
+                if input_path.is_some() {
+                    return Err(format!("unexpected argument `{other}`"));
+                }
+                input_path = Some(other.to_string());
+            }
+        }
+        i += 1;
+    }
+
+    let path = input_path.ok_or("tally needs an <input.json> path")?;
+    let raw = std::fs::read_to_string(&path).map_err(|e| format!("cannot read {path}: {e}"))?;
+    let input: TallyInput =
+        serde_json::from_str(&raw).map_err(|e| format!("invalid tally input JSON: {e}"))?;
+
+    let result = tally(&input.snapshot, &input.params, &input.ledger)?;
+
+    if show_ranking {
+        eprintln!("ranking:");
+        for c in &result.ranking {
+            eprintln!("  {:>2}. {} ({} approvals)", c.rank, c.node_id, c.approvals);
+        }
+        eprintln!(
+            "turnout {}/{} (quorum {}), status {:?}, resultHash {}",
+            result.turnout,
+            result.eligible,
+            result.quorum_required,
+            result.status,
+            result.result_hash
+        );
+    }
+
+    match result.status {
+        TallyStatus::Elected => {
+            let doc = assemble_elected_term_doc(&input.snapshot, &input.params, &result)?;
+            let json = doc.to_json_pretty();
+            emit(&json, out_path.as_deref())?;
+            Ok(ExitCode::SUCCESS)
+        }
+        TallyStatus::NoQuorum => {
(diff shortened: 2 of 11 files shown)


rjwalters:
❌ **Changes Requested**

Deep review of the election module. The tally, memo parsing, and signature handling are genuinely high quality — but the PR's central "clean seam with #1074" claim does **not** hold: the `elected` term document this PR emits **fails validation against the v2 schema #1074 just merged to `main`** (`docs/bridge/schemas/term-document.v2.schema.json`). That is a blocking integration break.

## Blocker: emitted `elected` doc is not v2-schema-valid

`ElectedTermDoc` (`bridge/core/src/election/term_doc.rs`) serialises these top-level keys:
`v, term, electionKind, status, electorate, tally, threshold, members, validity` — and **deliberately omits `execution` and `signatures`**.

But the v2 schema's top-level `required` (on `main`, from #1074) is:
```
["v","term","electionKind","status","electorate","tally","threshold",
 "members","execution","validity","signatures"]
```
The `allOf`/`if status==sealed` block only *tightens* member-level `keys`/`keySubmissionSig` and `signatures.outgoing.minItems`. It does **not** relax the top-level `required` for `elected` docs. So `execution` and `signatures` are required on **every** document, including `elected`.

Verified empirically with a Draft 2020-12 validator against a representative `elected` doc (exactly the shape `assemble_elected_term_doc` produces):
```
ERROR: [] - 'execution' is a required property
ERROR: [] - 'signatures' is a required property
```

Because #1074's seal step is documented as *consuming* this `elected` document, a seal step that validates its input against the schema on `main` will reject every document this PR produces. The PR description and the "builders reported a clean seam" note are both inaccurate on this point.

**CI is green only because nothing tests the seam.** No test in this PR loads `docs/bridge/schemas/term-document.v2.schema.json` and validates the emitted doc against it, so the divergence sailed through. That missing test is the root cause this wasn't caught.

### Resolution (pick one, then lock it with a test)
1. **Make #1073 conform** — emit `execution` and a `signatures` block (with `tallyAttestations` from the tally-attesting nodes and an empty `outgoing: []`) at `elected` status. Execution *intents* (`swapOwner` / member-swap / `reserveSweepFactor1`) are derivable from the membership diff, so this is arguably where they belong. This contradicts the current "execution/signatures are seal-only" framing, so it's a design call.
2. **Relax the schema** — move `execution` and `signatures` out of the top-level `required` into the `status==sealed` `then` branch (a companion change to #1074's schema on `main`). This matches ADR 0010 §5.1 ("elected pins membership only") and this PR's stated intent, and is likely the cleaner fix — but it must actually land, not be assumed.

Either way, **add a test** that deserialises `docs/bridge/schemas/term-document.v2.schema.json` and asserts `assemble_elected_term_doc(...).to_json()` validates against it. That is the guard that makes the "clean seam" real instead of asserted.

## What's solid (no changes needed here)

- **Determinism (tally.rs)** — genuinely pure: no wall-clock/RNG, `validity` is an input; input-order-independence via `sort_by (height, txid)`; grind-free tie-break = approvals desc then ascending `nodeId` (ADR 0010 §6.3); full ranking retained for §6.2 retry; `resultHash` is domain-separated (`botho.bridge.tally.v1:`) and length-prefixed over sorted counted txids + ranking → reproducible on replay. Failure modes carried as `TallyStatus`, not errors. This is exactly right.
- **Fail-closed memo parsing (memo.rs)** — rejects unknown keys, duplicate keys recursively at any nesting, float-where-int (`as_u64`), non-v1 versions, and non-allowlisted `kind`; trailing data caught by `serde_json::from_str`. One-nomination/one-vote enforced in the tally passes (`candidates.contains` / `voted` set).
- **Signature verification** — `verify_strict` (rejects malleable/low-order), 64-byte enforcement, correct domain separation under `botho.bridge.election.v1:`, distinct from tally/attestation/term domains.
- Merge state is CLEAN, no conflict with `main` post-#1074, all CI checks pass.

## Minor (non-blocking, address if convenient)

- `ElectionParams::validate` bounds `seats >= 2` but not `seats <= 5`; the schema caps `members` at `maxItems: 5`. `seats > 5` with enough candidates would emit a doc that also fails the schema. The ratified shape is 3-of-5, so low risk, but a `seats <= 5` check would make it defensive.

Fix the schema seam (and add the validation test), and I'll re-review promptly. Nice work on the deterministic core.

rjwalters:
Addressed the Judge's blocking finding and both notes.

**Blocking (schema seam):** relaxed `docs/bridge/schemas/term-document.v2.schema.json` per the recommended fix — moved `execution` and `signatures` out of the unconditional top-level `required` and into the `if status == sealed then required:[...]` conditional (alongside the existing member-key tightening). An `elected` (pre-keygen, pre-execution) document is now membership-only-valid; `sealed` documents still require `execution` + `signatures`. The sealed example fixture (`term-3.sealed.example.json`) still validates.

**Seam guard (the missing test):** added two tests in `bridge/core/src/election/tests.rs` that read the on-disk v2 schema and validate against it:
- `emitted_elected_doc_validates_against_v2_schema` — validates the exact JSON `assemble_elected_term_doc(...).to_json()` emits.
- `sealed_example_fixture_validates_and_execution_is_seal_only` — validates the sealed fixture and asserts (negative controls) that a sealed doc missing `execution`/`signatures` fails, and an elected doc missing `members` fails.

No new dependency: since no pure-Rust JSON-schema validator is in the workspace lockfile, the tests use a compact offline Draft-2020-12 validator covering exactly the keyword subset this schema uses (`type`/`const`/`enum`/`minimum`/`required`/`properties`/`items`/`minItems`/`maxItems`/`allOf`/`if`/`then`). Verified it has teeth: against the *un-relaxed* schema the elected test fails with exactly the errors you cited (`'execution' is a required property`, `'signatures' is a required property`); against the fixed schema it passes.

**Minor:** `ElectionParams::validate` now also bounds `seats <= 5` (mirrors the schema's `members.maxItems: 5`, the ratified 3-of-5 shape), with a test case added.

Branch rebased onto current `main` (post-#1074) — clean. `cargo test -p bth-bridge-core` = 96 + 2 CLI passing; `cargo fmt --check` and `cargo clippy --all-targets` clean. Back to you for re-review.

rjwalters:
✅ **Approved!** (re-review after Doctor pass — the single Doctor cycle)

My prior blocking finding — the emitted `elected` term document failed #1074's v2 schema because `execution` and `signatures` were unconditionally required at the top level — is fully resolved. Verified all four points:

**(a) Schema relaxation is sound.** `docs/bridge/schemas/term-document.v2.schema.json` top-level `required` (lines 7–8) no longer lists `execution`/`signatures`; they moved into the `if status == sealed then required:["execution","signatures"]` conditional (lines 128–146), alongside the existing member-key tightening and `outgoing.minItems: 1`. Sealed strictness is preserved: an elected doc (membership-only) is legal, a sealed doc still must carry both. The sealed example fixture `term-3.sealed.example.json` still validates.

**(b) The seam guard has real teeth.** `emitted_elected_doc_validates_against_v2_schema` deserialises the on-disk schema and validates the exact `assemble_elected_term_doc(...).to_json()` output against it, then asserts the elected doc carries no `execution`/`signatures`. `sealed_example_fixture_validates_and_execution_is_seal_only` adds two negative controls: (1) a sealed doc stripped of `execution`/`signatures` must fail, and (2) an elected doc stripped of `members` must fail — proving the validator enforces `required` rather than rubber-stamping. The offline Draft-2020-12 validator's `if`/`then` handling is correct (elected status produces a const-mismatch in the `if` probe → `then` skipped; sealed status passes the probe → execution/signatures required). Doctor's report that the elected test fails against the un-relaxed schema with exactly the two originally-cited errors confirms the guard would catch a seam regression.

**(c) No regression in the approved core.** The fix commit (`cc91bbf2`) touches only the schema, `tests.rs`, and 6 lines in `tally.rs` (the `seats <= 5` bound). The tally determinism, fail-closed memo parsing, and `verify_strict` signature verification I approved previously are untouched.

**(d) CI green + merge CLEAN.** cargo build --workspace, cargo test (bridge crates), cargo deny, rustfmt, and Fuzz Build Guard (native-smokes the bridge harnesses) all pass; `mergeStateStatus == CLEAN`, rebased on post-#1074 main. Local `cargo test -p bth-bridge-core` = 96 lib + 2 CLI passing.

Minor note (a) from my first pass is also addressed: `ElectionParams::validate` now bounds `seats <= 5` (tally.rs:111) to mirror the schema's `members.maxItems: 5`.

The #1067 → #1066 seam is now enforced by a test instead of asserted. Nice work.

Issue #1057, opened 2026-07-17
bth-transaction-core tests don't compile under --all-features (u64 vs u128 wealth in cluster_fee tests)


rjwalters:
## Bug Description

While reviewing PR #1055, a workspace-wide `cargo check --workspace --all-targets --all-features` failed with pre-existing errors unrelated to that PR (confirmed identical on `main`):

```
error[E0308]: mismatched types
  --> transaction/core/src/validation/cluster_fee.rs:691:77
   |
691 | let min_fee = fee_config.compute_fee(TransactionType::Hidden, 4000, effective_wealth, 0);
   |                                                                      ^^^^^^^^^^^^^^^^ expected `u128`, found `u64`
```

Same E0308 at cluster_fee.rs:781, :931, :1090 (and possibly more) — test code passes `u64` wealth where `compute_fee` / `compute_fee_with_outputs` now take `u128` (likely fallout from the u128 wealth widening) under some non-default feature combination.

## Reproduction

```bash
cargo check -p bth-transaction-core --all-targets --all-features
```

## Impact

- **Severity**: Low — CI compile-checks all test targets with default features only, so this is invisible to CI and to `cargo build --workspace`.
- It blocks anyone running `--all-features` locally (e.g., feature-matrix auditing) and hides whatever tests those cfg branches gate.

## Suggested fix

Cast/declare the test wealth values as `u128` (or `.into()`) in the affected `cluster_fee.rs` test functions; consider adding an `--all-features` check to CI once green.

Discovered during code review of PR #1055.

---

### Problem Statement

The `cluster_fee` validation module is compiled only when the non-default `cluster-tax` feature is enabled (`transaction/core/src/validation/mod.rs:9` — `#[cfg(feature = "cluster-tax")] mod cluster_fee;`; `transaction/core/Cargo.toml:11` has `default = []`). CI and `cargo build --workspace` use default features, so this module and its `#[cfg(test)]` block (starts at `cluster_fee.rs:556`) are never compiled there — the type error is completely invisible to the normal pipeline.

Under `--all-features` the module compiles, and four test-side call sites pass a `u64` cluster-wealth value into fee methods whose `cluster_wealth` parameter was widened to `u128`. Result: four `E0308: mismatched types (expected u128, found u64)` and `cargo check -p bth-transaction-core --all-targets --all-features` fails, which also prevents any feature-gated `cluster-tax` test from running.

### Root Cause

[...]

PR #1080, opened 2026-07-18
test(cluster-fee): fix stale fee-gaming test expectations under --all-features (#1068)

### LINKED ISSUE #1068: Two cluster-tax fee-gaming tests fail at runtime under --all-features (validate_ring_cluster_fee)
## Context

Follow-up from #1057 (which fixed the E0308 *compile* errors so `cargo check -p bth-transaction-core --all-targets --all-features` is clean). With the module now compiling, `cargo test -p bth-transaction-core --all-features` reveals **two pre-existing latent runtime failures** in the `cluster-tax`-gated test module. These tests had never actually run before because the module never compiled under `--all-features`, so this is newly-surfaced behavior, not a regression introduced by #1057.

## Failing tests

- `validation::cluster_fee::tests::test_validate_ring_cluster_fee_insufficient` — panic at `cluster_fee.rs:1010`: expected `Err(InsufficientClusterFee)` but got `Ok`.
- `validation::cluster_fee::tests::test_ring_fee_gaming_prevention` — panic at `cluster_fee.rs:1052`: "Attacker should not be able to pay reduced fee using poor decoys".

Result: `22 passed; 2 failed`.

## Why this is out of scope for #1057

- Neither failing test contains any of the four `as u128` cast sites fixed in #1057 (691, 781, 931, 1090 live in `test_cluster_fee_sufficient`, `validate_cluster_fee_from_tags`, `test_validate_ring_cluster_fee_sufficient`, `test_ring_fee_with_dynamic_base` — all of which pass).
- Both failing tests exercise `validate_ring_cluster_fee`, whose internal `max_cluster_wealth as u128` cast (`cluster_fee.rs:489`) is production code #1057 did not touch. The casts are value-preserving u64->u128 widenings, so any valid compile fix would produce these same two failures.
- #1057's acceptance criteria explicitly forbade changing production logic, helper return types, or test expectations ("the fix is limited to `u128` casts at the four call sites").

## What to investigate

Both tests assert that a transaction paying a fee computed for a *poor* cluster must be rejected when the ring contains a *rich* cluster (fee-evasion / gaming prevention). The tests now observe the required minimum fee coming out too low, so the poor fee is accepted. Likely candidates:

- Fee-curve semantics changed by the u128 wealth widening in `cluster-tax/src/fee_curve.rs` (e.g. `compute_fee_with_outputs` producing a smaller-than-expected fee for large wealth).
- `compute_single_utxo_cluster_wealth` / `compute_ring_max_cluster_wealth` effective-wealth weighting.
- Or stale test expectations that predate the u128 widening.

This matters because it is a fee-evasion / gaming-prevention path — determine whether it is a genuine security regression in the fee curve or merely outdated test expectations before deciding the fix.

## Reproduction

```bash
cargo test -p bth-transaction-core --all-features cluster_fee
```

## Suggested follow-up (separate)

Once green, add an `--all-features` job to CI so this class of breakage (both compile and runtime) can't silently recur — as noted in #1057's out-of-scope section.

rjwalters:
Closes #1068

## Verdict: (B) STALE TEST EXPECTATIONS — not a security regression

The two `validate_ring_cluster_fee` tests failed at runtime under `--all-features` (newly surfaced once #1057 made the `cluster-tax`-gated module compile). Tracing the actual numbers shows the fee curve is behaving **correctly**; the test expectations predate the #626 picocredit calibration.

### The evidence (traced numerically)

Cluster wealth is denominated in **picocredits**, and the fee curve's progressivity midpoint is `W_MID = 100,000 BTH = 1e17 pico` (`ClusterFactorCurve::W_MID_PICO`, #626). The failing tests used bare integers as "wealth":

| Test "wealth" (pico) | In BTH | `factor()` | `compute_fee_with_outputs` |
|---|---|---|---|
| 10,000 | 1e-8 BTH | 1000 (1x) | 16000 |
| 100,000 | 1e-7 BTH | 1000 (1x) | 16000 |
| 100,000,000 | 1e-4 BTH | 1000 (1x) | 16000 |

Every value is orders of magnitude below the 100,000-BTH tax threshold, so all collapse to the **flat 1x floor** — identical fee (16000). "Rich" and "poor" clusters produced the *same* fee, so `poor_fee == minimum_fee` and paying it was correctly accepted. There was genuinely nothing to game: neither cluster is wealthy enough to be taxed above 1x.

The `max_cluster_wealth as u128` widening from #1057 is a value-preserving u64→u128 cast and changes none of this. Production logic (`compute_ring_max_cluster_wealth`, `validate_ring_cluster_fee`) is untouched and correct.

For contrast, with realistic pico-scale wealth the curve is properly progressive:

| Wealth | `factor()` | fee |
|---|---|---|
| 100 BTH (1e14 pico) | 1050 (~1.05x) | 16800 |
| 1,000,000 BTH (1e18 pico) | 5093 (~5.1x) | ~81k |
| 10,000,000 BTH (1e19 pico) | 5745 (~5.7x) | 91920 |

## The fix

Update the two stale tests to use realistic picocredit wealth that **straddles the tax threshold** (poor = 100 BTH → ~1.05x, rich = 1,000,000 BTH → ~5.1x). Now the rich/poor fee gap is real, and the tests prove the gaming-prevention invariant holds:

- `compute_ring_max_cluster_wealth` selects the **rich** member's high wealth (the max across the ring),
- `validate_ring_cluster_fee` computes the higher required fee from it,
- the attacker's poor-decoy fee is **rejected** (`poor_fee < minimum_fee`), and only the rich fee passes.

Added `assert!(rich_fee > poor_fee)` / `assert!(poor_fee < rich_fee)` guards so these tests can **never again pass vacuously** on a flat curve. No production code changed — gaming prevention was already intact; the tests simply weren't exercising it.

## Gaming-prevention: preserved ✅

The conservative max-cluster-wealth rule is the security mechanism, and it is untouched. The attacker cannot pay less by mixing rich coins with poor decoys because the ring's fee is computed from the **maximum** effective cluster wealth among all members. The updated tests now actually demonstrate this (previously they asserted it against a flat curve where the claim was trivially — and misleadingly — satisfied).

## CI guard (issue's suggested follow-up)

Added a `cargo test -p bth-transaction-core --all-features` step to `workspace-build.yml`. The `cluster-tax`-gated code only compiles and runs under `--all-features`; default CI never exercised it, so both the compile break (#1057) and these runtime failures (#1068) sat green. `cargo check --all-features` alone was insufficient (the module compiled but its tests failed), so the step **runs** the crate's tests. Scoped to the one crate to stay fast (~7s) and deterministic.

## Verification

```
cargo test -p bth-transaction-core --all-features cluster_fee   # 24 passed; 0 failed
cargo test -p bth-transaction-core --all-features               # 120 passed; 0 failed
cargo fmt -p bth-transaction-core -- --check                    # clean
```

### DIFF
--- a/.github/workflows/workspace-build.yml
+++ b/.github/workflows/workspace-build.yml
@@ -84,3 +84,14 @@ jobs:
       # fast (seconds) and deterministic, so run them for real.
       - name: Run doctests
         run: cargo test --workspace --doc
+
+      # The `cluster-tax`-gated code in `bth-transaction-core` only compiles —
+      # and its tests only run — under `--all-features`. The default workspace
+      # build/test above never exercises that path, so both a COMPILE break
+      # (#1057) and two latent RUNTIME fee-gaming test failures (#1068) sat
+      # green for a long time. `cargo check --all-features` alone was not
+      # enough to catch #1068 (the module compiled but its tests failed), so we
+      # RUN the crate's `--all-features` tests here. Scoped to the one crate to
+      # stay fast and deterministic (~seconds).
+      - name: Run bth-transaction-core --all-features tests (cluster-tax path)
+        run: cargo test -p bth-transaction-core --all-features
--- a/transaction/core/src/validation/cluster_fee.rs
+++ b/transaction/core/src/validation/cluster_fee.rs
@@ -965,6 +965,15 @@ mod tests {
 
     #[test]
     fn test_validate_ring_cluster_fee_insufficient() {
+        // Cluster wealth is denominated in PICOCREDITS and the fee curve pins
+        // its progressivity midpoint at W_MID = 100,000 BTH = 1e17 pico (#626).
+        // Wealth values must therefore straddle that threshold to exercise the
+        // curve's dynamic range: a poor cluster below the midpoint pays ~1x, a
+        // rich cluster above it pays a materially higher multiplier. (Bare
+        // integers like 100_000 are ~1e-7 BTH — dust — and collapse to the flat
+        // 1x floor, which is why the original expectations were vacuous.)
+        const PICO: u64 = 1_000_000_000_000;
+
         // Ring with one rich member
         let tags_rich = make_tag_vector(&[(1, TAG_WEIGHT_SCALE)]);
         let tags_poor = make_tag_vector(&[(2, TAG_WEIGHT_SCALE)]);
@@ -972,9 +981,12 @@ mod tests {
         let ring_tags: Vec<&ClusterTagVector> = vec![&tags_rich, &tags_poor];
         let ring_values = vec![1_000_000u64, 1_000_000];
 
+        let poor_wealth = 100 * PICO; // 100 BTH: below midpoint, ~1x
+        let rich_wealth = 1_000_000 * PICO; // 1,000,000 BTH: above midpoint
+
         let mut wealth_map = ClusterWealthMap::new();
-        wealth_map.set(ClusterId(1), 100_000_000); // Very rich
-        wealth_map.set(ClusterId(2), 100_000);
+        wealth_map.set(ClusterId(1), rich_wealth); // Very rich
+        wealth_map.set(ClusterId(2), poor_wealth);
 
         let fee_config = FeeConfig::default();
 
@@ -998,9 +1010,22 @@ mod tests {
         let poor_fee = fee_config.compute_fee_with_outputs(
             TransactionType::Hidden,
             4000,
-            100_000, // Poor cluster wealth
+            poor_wealth as u128, // Poor cluster wealth
             2,
             0,
+        );
+        // Sanity: the conservative (max) fee genuinely exceeds the poor fee, so
+        // the rejection below is meaningful and not a flat-curve coincidence.
+        let rich_fee = fee_config.compute_fee_with_outputs(
+            TransactionType::Hidden,
+            4000,
+            rich_wealth as u128,
+            2,
+            0,
+        );
+        assert!(
+            rich_fee > poor_fee,
+            "rich cluster must incur a strictly higher fee than the poor cluster"
         );
         let result = validate_ring_cluster_fee(
             poor_fee,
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!** — determination (B) STALE TEST EXPECTATIONS is sound; gaming-prevention holds.

I did not rubber-stamp this. I independently re-derived the fee curve and confirmed every load-bearing claim:

**(a) Old tests were genuinely vacuous — verified empirically.** I ran a throwaway dump of `ClusterFactorCurve::default().factor(w)` for the exact old wealth values (then reverted it):
| wealth (pico) | factor() |
|---|---|
| 10,000 | 1000 (1x) |
| 100,000 | 1000 (1x) |
| 100,000,000 | 1000 (1x) |

All three collapse to the flat 1x floor because `W_MID_PICO = 100,000 BTH = 1e17 pico` (#626, const-asserted at `fee_curve.rs:705`) sits ~30 octaves above them, driving the log-domain sigmoid to 0. "Rich" and "poor" produced an identical fee → the old rejection was trivially satisfied. Confirmed vacuous.

**(b) No production/fee-curve logic changed.** `gh pr diff --name-only` = exactly two files: the test module + `workspace-build.yml`. The security mechanism is untouched: `compute_ring_max_cluster_wealth` still takes `.max()` across ring members (`cluster_fee.rs:416-421`) and `validate_ring_cluster_fee` still computes `minimum_fee` from that max (`:482-492`). (Note: the `max_wealth as u128` widening is #1057, already in origin/main — not this PR.)

**(c) New values straddle the threshold and are rejected by the real path.** Empirically: poor 100 BTH → factor 1050 (~1.05x), rich 1e6 BTH → factor 5093 (~5.1x). The `assert!(rich_fee > poor_fee)` / `assert!(poor_fee < rich_fee)` guards pass, and the poor-decoy fee is rejected by `validate_ring_cluster_fee` (which derives `minimum_fee` from the rich max). These guards make future flat-curve vacuous passes impossible.

**(d) No hidden (A) regression.** There is no wealth regime where an attacker pays a poor-cluster fee using rich decoys: the attacker's real (rich) input is always a ring member, so `.max()` always surfaces its high effective wealth and charges the higher fee. Gaming stays counter-productive. The mechanism was always correct; only the tests were stale.

**(e) CI addition is correct.** `cargo test -p bth-transaction-core --all-features` (not merely `check`) is appended cleanly after the doctests step in the `workspace-build.yml` job, scoped to one crate. It ran green in real CI (`cargo build --workspace (Linux)` — 13m30s pass), closing the coverage gap where `--all-features`-gated cluster-tax code sat unexercised.

**Local verification:** `cargo test -p bth-transaction-core --all-features` → 120 passed / 0 failed; the two fixed tests (`test_validate_ring_cluster_fee_insufficient`, `test_ring_fee_gaming_prevention`) pass. All CI checks green; merge state CLEAN. Description uses `Closes #1068`.

Issue #1092, opened 2026-07-20
snap(phase2): transaction history view


rjwalters:
## Original Issue

Decomposed from #1089. Add a transaction-history surface to the Snap. **Depends on** the persisted-state/windowed-scanning child (#1091) — history needs persisted scan state. Ref #1089.

---

## Curator Enhancement (2026-07-19)

Grounded against `web/packages/snap` on `origin/main`, which now **includes the merged dependency #1091** (PR #1098, "persist scan state + windowed incremental scanning via `snap_manageState`"). The dependency is satisfied — this issue is unblocked and buildable.

### What #1091 already gave us (read `src/state.ts`)

#1091 introduced the Snap's first persisted surface. The persisted blob is namespaced `{ version, scan }` (`SnapState`), and `scan.ownedOutputs` is an array of `PersistedOwnedOutput`, each of which **already carries the two facts a history view needs**:

- `blockHeight` — receive block height (comment in `state.ts`: *"Carried so #1092 renders history with no rescan."*)
- `txHash` — creating tx hash (same comment)
- plus `amount` (decimal-string picocredits), `subaddressIndex`, `targetKey`.

So **transaction history is a pure projection over already-persisted state** — no new chain fetch, no rescan, no schema migration, no manifest change (`snap_manageState` + `endowment:network-access` are already granted). This is exactly the extension point #1091 designed for.

### How this EXTENDS the #1091 schema (no migration)

- **Do NOT bump `STATE_VERSION`** and do NOT add a new persisted key. History is *derived*, not stored. Read `scan.ownedOutputs` via `readState()`/`usableScanState()` and project.
- Keep the `incrementalScanBalance` write path's `...(persisted ?? {})` spread intact so any sibling namespace is preserved — history introduces no new one.
- The receive side of history == the persisted owned-output set. The **spent/outgoing signal is recomputed live** (same as the balance path) via `chain_areKeyImagesSpent` / `spendableOwnedOutputs` — an owned output that is now spent is rendered as value that has left the wallet. Spent status stays **un-persisted**, consistent with #1091's design.

### Recommended approach

[...]

Issue #1091, opened 2026-07-20
snap(phase2): persisted state + windowed scanning via snap_manageState


rjwalters:
Decomposed from #1089 (MetaMask Snap Phase 2). The shipped MVP (`web/packages/snap`) uses **no** `snap_manageState` — it has zero persisted state. This is the **foundational** Phase-2 piece: add persisted Snap state and windowed UTXO/output scanning so the Snap remembers scan progress across sessions. Transaction history (#1092) and contacts (#1093) both build on the state store introduced here. Ref #1089.

---

## Curator Enhancement (2026-07-19)

Grounded against `web/packages/snap` on `origin/main` (Phase-1 MVP, PR #1075) and the `@botho/wasm-signer` scan/history orchestrators (`web/packages/wasm-signer/src/send.ts`).

### Problem / current behavior (confirmed)

The MVP has **no persisted state** — `git grep snap_manageState web/packages/snap/src` returns nothing; the only mentions are in `README.md` (deferred discussion). Every `botho_getBalance` / `botho_showBalance` call re-scans the **entire chain from genesis**:

- `index.ts` → `spendableBalance(wallet.keys, makeSendRpc(call))`
- `wasm-signer/src/send.ts:139` → `const candidates = await rpc.getOutputs(0, height)` — a full `[0, tip]` fetch + `scanOwnedOutputs` on **every** call.

This is O(chain height) per balance read, does not survive a session, and forces every future consumer (history, contacts) to re-derive owned outputs from scratch. This issue introduces the first persisted-state surface and makes scanning **incremental + resumable**.

### Scope (this issue only)

1. Add `snap_manageState` and persist a versioned, network-bound **scan-state blob**: last-scanned checkpoint height + the wallet's discovered owned outputs (with the metadata history needs).
2. Replace the full-genesis rescan with a **windowed incremental scan** that resumes from the persisted checkpoint and only fetches `(lastScannedHeight, tip]` in fixed-size windows.
3. Cover it with `@metamask/snaps-jest` tests against the existing mocked node (`test/mock-node.ts`) — no live betanet.

**Design the schema for the sibling consumers** (#1092 history, #1093 contacts): namespace the top-level state so contacts can live alongside scan state without collision, and persist enough per-output metadata (block height + tx hash) that #1092 can render history **without a full rescan**.

### Persisted-state schema (recommended)

[...]

PR #1106, opened 2026-07-20
ci(web): run the MetaMask Snap suite (@metamask/snaps-jest) in web-ci

rjwalters:
## Summary
The `@botho/snap` package's tests are `*.snap.ts` files run by `@metamask/snaps-jest` (`test:snap`), not `*.test.ts`. The root vitest run in `web-ci.yml` (`pnpm test:run`) only collects `packages/**/*.test.ts(x)`, so the entire 76-test Snap suite had **zero CI coverage** — every Snap PR in the Phase-2 sweep was validated only by the reviewer locally.

This wires the Snap suite into the existing path-filtered `web-ci` job so a Snap regression turns CI red.

## Changes
- `.github/workflows/web-ci.yml`:
  - Add a wasm Rust toolchain step (`dtolnay/rust-toolchain@master`, `nightly-2025-12-03` + `wasm32-unknown-unknown`). The pin matches `rust-toolchain.toml`; that file does **not** list the wasm32 target, so it is requested explicitly.
  - Add `Swatinem/rust-cache@v2` (key `web-ci-wasm-signer`) so the crypto-stack compile is cached across runs.
  - Add `jetli/wasm-pack-action@v0.4.0` (pinned `v0.13.1`) — stock ubuntu-latest has neither wasm-pack nor the wasm32 target.
  - Build the wasm-signer bundler artifact (`pnpm --filter @botho/wasm-signer build:wasm:bundler`) that `mm-snap build` inlines into the snap bundle (`src/signer.ts` imports `@botho/wasm-signer/pkg-bundler/bth_wasm_signer.js`).
  - Run the suite: `pnpm --filter @botho/snap test:snap`.
  - Bump `timeout-minutes` 15 -> 25 to cover the first (uncached) Rust compile; update the job name and header comment.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| `web-ci.yml` runs the Snap `*.snap.ts` suite on PRs touching `web/**` | ✅ | New `Run Snap tests` step; job stays under the existing `web/**` path filter |
| The wasm bundler prereq is built in-job (SES harness needs it) | ✅ | `build:wasm:bundler` step emits `pkg-bundler/` before `test:snap`; confirmed the snap imports `@botho/wasm-signer/pkg-bundler/bth_wasm_signer.js` |
[...]

PR #1107, opened 2026-07-20
fix(snap): prune reorged-out owned outputs by reconciling the rescanned range

### LINKED ISSUE #1099: snap: prune-on-rescan / finality-depth for mergeOwnedOutputs (reorg staleness)
## Problem

`mergeOwnedOutputs` in `web/packages/snap/src/state.ts` (#1091) is **monotonic — add-only, never prunes**. A betanet reorg that drops an already-persisted owned output leaves that output stale in the persisted `scan.ownedOutputs` set. Because both the **balance** path (`incrementalScanBalance`) and the **history** path (`deriveHistory`, #1092) read that set, both can over-report a receive that no longer exists on-chain.

This was flagged by the #1091 Judge and explicitly deferred out of #1092 (the transaction-history view) because a destructive change to the merge is **balance-critical** — it changes the shipped #1091 sum path — and deserves its own issue + test matrix rather than riding a read-only UI feature.

## What #1092 already did (non-destructive half)

#1092 surfaces per-entry **`confirmations = max(0, tip - blockHeight)`** in the history view, so a shallow, reorg-prone receive near the tip is *visibly* flagged to the user — without mutating persisted state. That is the read-layer half of the concern. This issue is the **write-layer** half.

## Scope of this issue (destructive — needs care)

- Decide a prune policy for `mergeOwnedOutputs` / the incremental scan: e.g. re-scanning a window should be able to **drop** a previously-persisted owned output whose creating output is no longer present in the re-fetched window (within the reorg buffer), and/or a finality-depth rule that only trusts outputs below a confirmation threshold.
- Preserve the target-key dedupe / idempotence guarantees for the non-reorg case.
- Test matrix: shallow reorg within `REORG_BUFFER` drops a stale output from both balance and history; deep/finalized outputs are never pruned; the happy-path incremental resume is unchanged.

## Why not in #1092

Balance correctness is the shipped, load-bearing behaviour; a UI feature PR is the wrong place to change it. Filed as a dedicated follow-up per the #1092 curator's explicit scope decision.

## References

- #1091 (persisted scan state; introduced `mergeOwnedOutputs`) / PR #1098
- #1092 (transaction history view; surfaces the non-destructive `confirmations` half)

---

## Curator Enhancement (2026-07-20)

Grounded against `origin/main` `web/packages/snap/src/state.ts` (evolved through #1091–#1096). The write-layer half of the #1091 Judge's reorg-staleness flag. This is a clean, provably-safe single buildable unit (~1 new pure function + 1 call-site rewire + pure-logic tests; ~2 files, well under limits).

### Root cause (current code)

`mergeOwnedOutputs(existing, discovered)` is **union with existing-wins**:

```ts
for (const o of existing) byKey.set(o.targetKey, o);
for (const o of discovered) if (!byKey.has(o.targetKey)) byKey.set(o.targetKey, o);
```

`incrementalScan` re-fetches the trailing `[scanStartHeight(lastScannedHeight), tip]` range each read (`start = max(0, lastScannedHeight - REORG_BUFFER)`, `REORG_BUFFER = 10`) and merges. Because the merge only *adds*, a previously-persisted owned output whose creating block was **reorged out** of that re-fetched range is never removed. It lingers in `scan.ownedOutputs`, and since **both** the balance sum (`incrementalScanBalance` → `spendable`) and the history projection (`deriveHistory`) read the same persisted set, both **over-report** a receive that no longer exists on-chain.

### The fix: prune-on-rescan, scoped strictly to the re-fetched range

Replace the add-only merge with a **reconcile** step. `REORG_BUFFER` already *is* the de-facto finality depth (outputs older than `lastScannedHeight - REORG_BUFFER` are never re-fetched, i.e. treated as final); the fix makes the re-fetched window **authoritative** — its persisted contents are *replaced* by what the node currently reports, instead of only being added to. This subsumes the "finality-depth policy" alternative: the finality depth is exactly `REORG_BUFFER`, and it can be widened later independently of this change without touching the reconcile logic.

Recommended pure function (accumulate all discovered outputs across the scan's windows, then reconcile **once** against the actually-fetched range `[start, tip]`):

```ts
/**
 * Reconcile the persisted owned set against a rescanned height range.
 * Retains every existing output BELOW the rescanned range and AT/BELOW the tip
 * (assumed final, not re-verified); replaces the rescanned range wholesale with
 * what the node currently reports (`discovered`). This DROPS a reorged-out
 * output (no longer returned in its window) and an output stranded ABOVE the
 * current tip (a chain that shrank), while never dropping a still-returned one.
 */
export function reconcileOwnedOutputs(
  existing: PersistedOwnedOutput[],
  discovered: PersistedOwnedOutput[],
  start: number,   // scanStartHeight(lastScannedHeight)
  tip: number,     // node's reported chain tip this scan
): PersistedOwnedOutput[] {
  const byKey = new Map<string, PersistedOwnedOutput>();
  // Retain only outputs strictly BELOW the re-fetched range AND at/below tip.
  for (const o of existing) {
    if (o.blockHeight < start && o.blockHeight <= tip) byKey.set(o.targetKey, o);
  }
  // The re-fetched range [start, tip] is replaced wholesale by node truth.
  for (const o of discovered) byKey.set(o.targetKey, o);
  return Array.from(byKey.values());
}
```

Call-site change in `incrementalScan`: accumulate `persistedDiscovered` across **all** windows into one array, then call `reconcileOwnedOutputs(ownedOutputs, allDiscovered, start, tip)` **after** the window loop (not a per-window merge — reconciling once against the whole `[start, tip]` avoids a window-boundary output being pruned before a later window rediscovers it; keying by `blockHeight`-in-range makes this exact). `mergeOwnedOutputs` may be retained only for its within-`discovered` dedupe, or folded into the reconcile.

Why `blockHeight <= tip` in the retain guard: on a chain that shrank so `tip < start` (deep reorg beyond the buffer), `windows(start, tip)` is empty and `discovered` is empty; the guard still prunes every persisted output **above the reported tip** (provably gone — the node says the chain is only `tip` high) while retaining those at/below `tip`. This is authoritative from the tip claim and cannot under-report.

### CRITICAL SAFETY INVARIANT (must hold after every scan)

Let `R = [start, tip]` be the height range actually re-fetched this scan (`start = max(0, lastScannedHeight - REORG_BUFFER)`; `R = ∅` when `tip < start`). After the scan the persisted `scan.ownedOutputs` set must satisfy, **exactly**:

1. **No phantom.** No persisted output with `blockHeight ∈ R` survives unless the node returned it in this scan's fetch of `R`. (Also: no persisted output with `blockHeight > tip` survives — provably reorged out.)
2. **No dropped.** Every owned output the node returned in `R` is present.
3. **No collateral prune.** Every persisted output with `blockHeight < start` (and `≤ tip`) is retained **unchanged**. A range that was NOT re-fetched is NEVER pruned.

Equivalently: `persisted ∩ R  ==  node-reported-owned ∩ R` (exactly — no phantom, no dropped), and `persisted \ R` is untouched. Pruning is **bounded to a range the node was authoritatively queried for** (plus the provably-empty region above the tip), which is what guarantees a still-valid output is never dropped — the dangerous under-report direction that would make funds appear lost.

### Interaction with the live spent-check (must be spelled out and tested)

Pruning is driven **purely by presence in `chain_getOutputs`** (existence on-chain), NOT by spent status. Spending an output does NOT remove it from `chain_getOutputs` — a spent output remains a mined output on-chain (the wallet already models spent status separately, live, via `chain_areKeyImagesSpent`, because "an output can be spent later"). Therefore:

- A **spent-but-still-mined** output is re-returned by the node in its window → present in `discovered` → **retained** in `ownedOutputs`, then filtered out of `spendable` by the live spent-check and rendered `direction: 'spent'` by `deriveHistory`. It is **never** pruned.
- Only a **reorged-out** output (no longer mined, absent from `chain_getOutputs`, or stranded above the tip) is pruned.

This decoupling is load-bearing: keying prune on the live *spent* status instead of on-chain *presence* would wrongly delete every spent output from history and could corrupt balance. The test matrix pins it.

### Recommended approach summary

Prune-on-rescan, scoped strictly to the re-fetched `[start, tip]` range, with `REORG_BUFFER` as the (already-existing) finality depth. No new RPCs, no schema change, no `STATE_VERSION` bump (persisted shape is unchanged — same `PersistedOwnedOutput` fields; only the *reconciliation rule* changes). The non-destructive read-layer half (per-entry `confirmations`) shipped in #1092 and stays as the user-facing staleness signal.

## Acceptance Criteria

- [ ] A new pure `reconcileOwnedOutputs` (or an equivalent reconcile inside `incrementalScan`) replaces the add-only `mergeOwnedOutputs` at the scan call-site; discovered outputs are accumulated across all windows and reconciled **once** against `[start, tip]`.
- [ ] After a rescan, a previously-persisted owned output with `blockHeight ∈ [start, tip]` that the node no longer returns is **removed** from `scan.ownedOutputs` — and consequently drops out of both `incrementalScanBalance` and `deriveHistory`.
- [ ] A persisted output with `blockHeight > tip` (chain shrank below its receive height) is pruned.
- [ ] A still-returned owned output in the rescanned range is **retained** (and not duplicated) — the safety invariant's "No dropped" / "No collateral prune" clauses hold.
- [ ] A persisted output with `blockHeight < start` (below the reorg buffer / re-fetched range) is **never** pruned, even if absent from `discovered`.
- [ ] When the re-fetched range is empty (`tip < start`, deep shrink), outputs at/below `tip` are retained and only those above `tip` are pruned; no range that was not queried is pruned.
- [ ] A **spent-but-still-mined** output (returned by `chain_getOutputs`, marked spent by the live `chain_areKeyImagesSpent`) is retained in `ownedOutputs` and rendered `direction: 'spent'` — never pruned.
- [ ] Happy-path incremental resume is unchanged: a second read at the same tip with no reorg yields the identical owned set and balance (idempotent), and still fetches only the reorg-buffer window.
- [ ] The safety invariant is documented in a code comment on `reconcileOwnedOutputs` and in `state.ts`'s module header (reorg-safety paragraph updated from "merges dedupe" to "reconciles the re-fetched range").

## Affected Files

- `web/packages/snap/src/state.ts` — Add `reconcileOwnedOutputs` (pure, exported, invariant documented); rewire `incrementalScan` to accumulate discovered across windows and reconcile once against `[start, tip]` instead of the per-window add-only merge; update the module-header "Reorg safety" paragraph.
- `web/packages/snap/test/state.snap.ts` — Add pure-logic reconcile tests (the deterministic, wasm-free proof of the invariant — this is the primary test layer). Optionally extend the behavioural `outputWindows` assertions.
- `web/packages/snap/test/mock-node.ts` — (Optional) Only if a behavioural reorg test is added; the mock's `chain_getOutputs` already honours `[start_height, end_height]`, so a reorg is simulated either via a custom `handlers['chain_getOutputs']` closure that mutates its returned set, or by restarting the mock (same `network`, new port) with a reduced output set between two reads on the same Snap install (state is network-bound, not URL-bound).

## Test Plan

Primary (mandatory) — **pure-logic** tests in `state.snap.ts`, mirroring the existing `mergeOwnedOutputs`/`deriveHistory` unit layer (no `installSnap`, no wasm — this is where the invariant is proven deterministically, since behavioural tests cannot cheaply mint a wasm-recognised owned output):

- [ ] **Reorg prunes a stale output:** existing `[A@h95, B@h50]`, rescan range `[90, 100]`, `discovered = []` (A reorged out). Assert result === `[B]` (A pruned, B — below start — retained). Then assert `incrementalScanBalance`-style sum and `deriveHistory` over the result both exclude A.
- [ ] **Still-valid output is NEVER pruned:** existing `[A@h95]`, rescan `[90,100]`, `discovered = [A]`. Assert result === `[A]` (retained, not duplicated).
- [ ] **Below-range output untouched:** existing `[B@h50]`, rescan `[90,100]`, `discovered = []`. Assert B retained (not in the re-fetched range → never pruned).
- [ ] **Chain-shrink above tip:** existing `[A@h98, B@h50]`, `start=90`, `tip=93`, `discovered=[]` (or `windows` empty). Assert A (`>tip`) pruned, B retained.
- [ ] **Deep-shrink safety (no un-queried prune):** existing outputs at `h=5,6,50,95`, `start=90`, `tip=5`, `discovered=[]`. Assert only `h=5` retained (all `>tip` pruned; none in the un-fetched `(5,90)` band is spared by accident of *not* being pruned — they are `>tip` so correctly gone), and assert nothing below/at tip is dropped.
- [ ] **Spent-but-present retained:** existing `[A@h95]`, rescan `[90,100]`, `discovered=[A]`, and a `spendable` set that EXCLUDES A (live-spent). Assert A stays in the reconciled `ownedOutputs`, `spentTargetKeys` includes A, and `deriveHistory` renders A as `direction: 'spent'` (not pruned).
- [ ] **Idempotent happy path:** existing `[A@h95, B@h50]`, rescan `[90,100]`, `discovered=[A]` (A re-returned unchanged). Assert result set equals input set (order-insensitive), no duplicates.
- [ ] **Multi-window reconcile:** discovered accumulated across ≥2 windows spanning `[start, tip]` reconciles correctly (an output at a low window boundary is not pruned by a later window's reconcile).

Optional (nice-to-have) — behavioural through the SES harness:

- [ ] Extend an existing empty-chain behavioural test to assert the resume still fetches only the reorg-buffer window after the reconcile rewire (guards against an accidental full-rescan regression).
- [ ] (If feasible) A two-read reorg-simulation via a mutable `chain_getOutputs` handler; acceptable to skip if a wasm-recognised owned-output fixture is impractical (the pure-logic layer already proves correctness).

Edge cases:

- [ ] `blockHeight` meta fallback: `incrementalScan` sets `blockHeight: meta?.height ?? 0`. Confirm the reconcile does not mis-place a height-0 fallback below the prune window in a way that strands a phantom; note in review whether a missing-height output should be treated as always-reconciled. (Data-quality; document if left out of scope.)
- [ ] `REORG_BUFFER` boundary: an output exactly at `blockHeight === start` is in-range (reconciled); one at `start - 1` is out-of-range (retained).

## References

- #1091 (persisted scan state; introduced `mergeOwnedOutputs`) / PR #1098
- #1092 (transaction history; shipped the non-destructive `confirmations` half)
- `web/packages/snap/src/state.ts` — `mergeOwnedOutputs`, `incrementalScan`, `scanStartHeight`, `REORG_BUFFER`
- `web/packages/snap/test/state.snap.ts` — pure-logic + behavioural test patterns to mirror

rjwalters:
## Summary

`mergeOwnedOutputs` in `web/packages/snap/src/state.ts` was **add-only** (existing-wins union). `incrementalScan` re-fetches the trailing reorg-buffer range `[start, tip]` (`start = max(0, lastScannedHeight - REORG_BUFFER)`) on every read and only *added* what it found, so an owned output reorged out of that window was **never removed** — over-reporting both the balance sum (`incrementalScanBalance` → `spendable`) and the transaction history (`deriveHistory`). This is the write-layer half of the #1091 Judge's reorg-staleness flag (the non-destructive `confirmations` read-layer half shipped in #1092).

Balance-critical, so implemented exactly to the curator's prune-on-rescan design with the funds-safety invariant preserved.

## Changes

- **New pure `reconcileOwnedOutputs(existing, discovered, start, tip)`** replaces the add-only merge at the scan call-site. It treats the re-fetched range as authoritative:
  - Retains every existing output **strictly below `start`** (never re-fetched → assumed final) **and at/below `tip`**.
  - Replaces `[start, tip]` **wholesale** with what the node currently reports (`discovered`).
  - Prunes anything **stranded above `tip`** (a chain that shrank below an output's receive height).
- **`incrementalScan` rewired** to accumulate `discovered` across **all** windows into one array and reconcile **once** against `[start, tip]` (not a per-window merge) — so a low-window-boundary output is never pruned before a later window rediscovers it.
- **Prune keys on on-chain presence (`chain_getOutputs`), NOT live spent status.** A spent-but-still-mined output is re-returned by the node → retained → split out of `spendable` by the live `chain_areKeyImagesSpent` check → rendered `direction: 'spent'`. Only a genuinely reorged-out output is pruned.
- `mergeOwnedOutputs` kept (no longer on the scan path) for its dedupe unit contract; documented as add-only/monotonic.
- Module-header "Reorg safety" paragraph updated from "merges dedupe" to "reconciles the re-fetched range"; full safety invariant documented on `reconcileOwnedOutputs`.
- No `STATE_VERSION` bump, no schema change, no new RPC (persisted `PersistedOwnedOutput` shape unchanged — only the reconciliation rule changed).
- `snap.manifest.json` shasum updated by `mm-snap build` (bundle changed), per repo convention for snap-src PRs.

### Safety invariant (holds after every scan, `R = [start, tip]`, `R = ∅` when `tip < start`)

`persisted ∩ R == node-reported-owned ∩ R` exactly (no phantom, no dropped); `persisted \ R` untouched; nothing above `tip` survives. Pruning is bounded to a range the node was authoritatively queried for, so a still-valid output is **never** dropped — the dangerous under-report direction that would make funds appear lost.

| New pure reconcile replaces add-only merge at call-site; discovered accumulated across all windows, reconciled once against `[start, tip]` | ✅ | `reconcileOwnedOutputs` + rewired `incrementalScan` (single `allDiscovered` array, one reconcile after the window loop) |
| Reorged-out output in `[start, tip]` removed → drops from balance and history | ✅ | Test "prunes a stale (reorged-out) output…" asserts removal + excluded from sum + `deriveHistory` |
| Output with `blockHeight > tip` (chain shrank) pruned | ✅ | Test "prunes an output stranded above the tip…" |
| Still-returned output retained, not duplicated (No dropped / No collateral prune) | ✅ | Test "NEVER prunes a still-returned output… (funds-safety)" |
| Output `blockHeight < start` never pruned even if absent from discovered | ✅ | Test "never prunes an output below the re-fetched range…" |
| Empty range (`tip < start`, deep shrink): at/below tip retained, only above-tip pruned, nothing un-queried pruned | ✅ | Test "deep shrink (tip < start)…" (`windows(90,5) === []`) |
| Spent-but-still-mined output retained + `direction:'spent'`, never pruned | ✅ | Test "retains a spent-but-still-mined output…" (in `discovered`, excluded from `spendable`, `spentTargetKeys` has it, history `direction:'spent'`) |
| Happy-path incremental resume unchanged (idempotent; still fetches only reorg-buffer window) | ✅ | Test "is idempotent on the happy path…"; behavioural "resumes from the checkpoint…" still green |
| Invariant documented on the fn + module header | ✅ | Doc comment on `reconcileOwnedOutputs` + updated "Reorg safety" paragraph |

## Test Plan

`pnpm typecheck && pnpm build:snap && pnpm test:snap` — all green (**85/85**, exit 0; wasm bundler artifact built first). New pure-logic matrix in `test/state.snap.ts` (8 reconcile cases + boundary case), all passing, including the two funds-safety guarantees:
- **still-valid never pruned** (retained, not duplicated, balance intact)
- **spent-but-present retained** (retained despite live-spent, rendered `direction:'spent'`)

Closes #1099

### DIFF
--- a/web/packages/snap/snap.manifest.json
+++ b/web/packages/snap/snap.manifest.json
@@ -3,7 +3,7 @@
   "description": "Manage a privacy-by-default Botho wallet from inside MetaMask. Keys are derived from your MetaMask Secret Recovery Phrase and never leave the Snap sandbox.",
   "proposedName": "Botho Wallet",
   "source": {
-    "shasum": "YkefLNaYb+2JC9kP21c9IPU9yweFTbRSbaMlE9mp+nM=",
+    "shasum": "8Iva7ztrgdLBlQ5RbwCA9P8GlgAEkJbXu8A5VMZKW/8=",
     "location": {
       "npm": {
         "filePath": "dist/bundle.js",
--- a/web/packages/snap/src/state.ts
+++ b/web/packages/snap/src/state.ts
@@ -36,8 +36,15 @@
  *
  *  - **Reorg safety.** betanet can reorg. On resume we conservatively re-scan a
  *    small trailing `REORG_BUFFER` of blocks below the checkpoint so a shallow
- *    reorg near the tip is picked up; merges dedupe by one-time target key so the
- *    overlap never double-counts.
+ *    reorg near the tip is picked up. Rather than only *adding* what the re-scan
+ *    finds (which can never remove a receive that was reorged out), we
+ *    **reconcile** the re-fetched range: the persisted contents of `[start, tip]`
+ *    are REPLACED wholesale by what the node currently reports, while everything
+ *    strictly below `start` (treated as final — never re-fetched) is retained
+ *    unchanged, and anything stranded above the reported `tip` is pruned. Pruning
+ *    is bounded to the range the node was authoritatively queried for, so a
+ *    still-returned output is never dropped (the dangerous under-report
+ *    direction). See {@link reconcileOwnedOutputs} for the safety invariant.
  */
 
 import type {
@@ -240,6 +247,12 @@ export function toOwnedOutput(p: PersistedOwnedOutput): OwnedOutput {
  * one-time target key (which is globally unique per output). Existing records
  * win, so a re-scanned reorg-buffer block never appends a duplicate. Returns a
  * new array (does not mutate `existing`).
+ *
+ * NOTE: this is **add-only** (monotonic) — it can never PRUNE a persisted output
+ * that has since been reorged out. The incremental scan therefore does NOT use it
+ * as the persist step; it reconciles the re-fetched range via
+ * {@link reconcileOwnedOutputs}. Retained only for its within-`discovered` /
+ * against-`existing` dedupe semantics and its historical unit contract.
  */
 export function mergeOwnedOutputs(
   existing: PersistedOwnedOutput[],
@@ -248,6 +261,63 @@ export function mergeOwnedOutputs(
   const byKey = new Map<string, PersistedOwnedOutput>();
   for (const o of existing) byKey.set(o.targetKey, o);
   for (const o of discovered) if (!byKey.has(o.targetKey)) byKey.set(o.targetKey, o);
+  return Array.from(byKey.values());
+}
+
+/**
+ * Reconcile the persisted owned set against a **rescanned** height range, so a
+ * reorged-out output is PRUNED instead of lingering forever (the add-only
+ * {@link mergeOwnedOutputs} can never remove one). Returns a new array (does not
+ * mutate `existing`).
+ *
+ * `start` is `scanStartHeight(lastScannedHeight)` and `tip` is the node's
+ * reported chain tip this scan; `[start, tip]` is exactly the range the scan
+ * re-fetched (window by window) and `discovered` is every owned output the node
+ * returned across ALL those windows. The re-fetched range is treated as
+ * AUTHORITATIVE: its persisted contents are replaced wholesale by `discovered`,
+ * while everything strictly below `start` (never re-fetched — assumed final) is
+ * retained unchanged. Anything stranded ABOVE the reported `tip` (a chain that
+ * shrank below an output's receive height) is provably gone and is pruned too.
+ *
+ * SAFETY INVARIANT — after every scan, for the re-fetched range `R = [start, tip]`
+ * (`R = ∅` when `tip < start`):
+ *
+ *   1. No phantom:  no persisted output with `blockHeight ∈ R` survives unless the
+ *      node returned it this scan; nor does any with `blockHeight > tip`.
+ *   2. No dropped:  every owned output the node returned in `R` is present.
+ *   3. No collateral prune:  every persisted output with `blockHeight < start`
+ *      (and `≤ tip`) is retained unchanged — a range NOT re-fetched is NEVER pruned.
+ *
+ * Equivalently `persisted ∩ R == node-reported-owned ∩ R` (exactly) and
+ * `persisted \ R` is untouched. Because pruning is bounded to a range the node
+ * was authoritatively queried for (plus the provably-empty region above the tip),
+ * a still-returned output is NEVER dropped — the dangerous under-report direction
+ * that would make funds appear lost.
+ *
+ * Prune keys on ON-CHAIN PRESENCE (membership in the node's `chain_getOutputs`
+ * response, i.e. `discovered`), NOT on live spent status. Spending an output does
+ * not remove it from `chain_getOutputs` — a spent output stays a mined output, so
+ * it is re-returned, RETAINED here, then split out of `spendable` by the live
+ * `chain_areKeyImagesSpent` check and rendered `direction: 'spent'`. Only a
+ * genuinely reorged-out output (absent from `discovered`, or above `tip`) is
+ * pruned.
+ */
+export function reconcileOwnedOutputs(
+  existing: PersistedOwnedOutput[],
+  discovered: PersistedOwnedOutput[],
+  start: number,
+  tip: number,
+): PersistedOwnedOutput[] {
+  const byKey = new Map<string, PersistedOwnedOutput>();
+  // Retain only outputs strictly BELOW the re-fetched range AND at/below the tip.
+  // `< start`  : not re-fetched, treated as final -> never pruned.
+  // `<= tip`   : guards the chain-shrank case (tip < start): anything above the
+  //              node's reported tip is provably gone and must not survive.
+  for (const o of existing) {
+    if (o.blockHeight < start && o.blockHeight <= tip) byKey.set(o.targetKey, o);
+  }
+  // The re-fetched range [start, tip] is replaced wholesale by node truth.
+  for (const o of discovered) byKey.set(o.targetKey, o);
   return Array.from(byKey.values());
 }
 
@@ -301,9 +371,15 @@ export async function incrementalScan(args: IncrementalScanArgs): Promise<Increm
 
   const persisted = await readState();
   const usable = usableScanState(persisted, network);
-  let ownedOutputs = usable ? usable.ownedOutputs : [];
+  const priorOwned = usable ? usable.ownedOutputs : [];
   const start = usable ? scanStartHeight(usable.lastScannedHeight) : 0;
 
+  // Accumulate every owned output the node reports across ALL windows of this
+  // scan's re-fetched range, then reconcile ONCE against `[start, tip]`. A single
+  // whole-range reconcile (rather than a per-window merge) is what makes the prune
+  // exact: an output at a low window boundary is never dropped by a later window's
+  // reconcile, because the whole `[start, tip]` set is compared together.
+  const allDiscovered: PersistedOwnedOutput[] = [];
   for (const [lo, hi] of windows(start, tip)) {
     const raw = await fetchWindow(lo, hi);
     if (raw.length === 0) continue;
@@ -324,15 +400,21 @@ export async function incrementalScan(args: IncrementalScanArgs): Promise<Increm
       })),
     });
 
-    const persistedDiscovered = discovered.map((o) => {
+    for (const o of discovered) {
       const meta = metaByTargetKey.get(o.targetKey);
-      return toPersistedOwnedOutput(o, {
-        blockHeight: meta?.height ?? 0,
-        txHash: meta?.txHash ?? o.targetKey,
-      });
-    });
-    ownedOutputs = mergeOwnedOutputs(ownedOutputs, persistedDiscovered);
+      allDiscovered.push(
+        toPersistedOwnedOutput(o, {
+          blockHeight: meta?.height ?? 0,
+          txHash: meta?.txHash ?? o.targetKey,
+        }),
+      );
+    }
   }
+
+  // Reconcile: replace the re-fetched `[start, tip]` range wholesale with node
+  // truth (dropping reorged-out receives), retaining everything below `start` and
+  // pruning anything stranded above `tip`. See `reconcileOwnedOutputs`.
+  const ownedOutputs = reconcileOwnedOutputs(priorOwned, allDiscovered, start, tip);
 
   await writeState({
     version: STATE_VERSION,
--- a/web/packages/snap/test/state.snap.ts
+++ b/web/packages/snap/test/state.snap.ts
@@ -31,6 +31,9 @@ import {
   scanStartHeight,
   windows,
   mergeOwnedOutputs,
+  reconcileOwnedOutputs,
+  spentTargetKeys,
+  deriveHistory,
   toPersistedOwnedOutput,
   toOwnedOutput,
   type PersistedOwnedOutput,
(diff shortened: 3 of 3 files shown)


rjwalters:
## Judge review — APPROVED (balance-critical, reviewed with maximum rigor)

Reviewed `reconcileOwnedOutputs` and its `incrementalScan` call site against the actual diff, built the wasm bundler + ran `pnpm test:snap` locally (**85/85 pass, exit 0**), and confirmed the PR's CI run genuinely executed the Snap suite.

### The under-report / fund-loss direction is provably safe

Retain rule: `existing` kept iff `blockHeight < start && blockHeight <= tip`, then `discovered` overlaid. Case analysis for a still-valid owned output at height `h`:
- `h < start` → retained directly (range never re-fetched → treated as final).
- `start <= h <= tip` → the node was **authoritatively re-queried** for this range, so if it's still on-chain it is in `discovered` and re-added; it is dropped **only** when the node did not return it → genuinely reorged out. Correct.
- `h > tip` → chain shrank below the receive height → provably gone → pruned. Correct.

So pruning is bounded exactly to `[start, tip]` (re-queried) plus the provably-empty above-tip region. A still-returned output is never dropped.

**(a) Bounded pruning** — confirmed above.
**(b) Accumulate-across-windows** — load-bearing and correct: `allDiscovered` accumulates every window's owned outputs and `reconcileOwnedOutputs` is called **once** over the whole `[start, tip]`. `windows(start, tip)` fully tiles the inclusive range; each output falls in exactly one window; dedupe is by `targetKey`. A low/high window-boundary output can't be pruned by a window that didn't cover it. Guarded by the multi-window test.
**(c) Empty `discovered` from an RPC hiccup — the case I most wanted to break — is SAFE.** `makeNodeCall` throws on HTTP `!res.ok` and on any JSON-RPC `error`; `getOutputsWithMeta` (the `fetchWindow`) does not catch; `incrementalScan` has no try/catch and `writeState` runs only **after** the window loop completes. Any window RPC failure therefore aborts the entire scan **before any persist** — no prune on error, and no partial prune if window N fails after window N-1 succeeded. `raw.length === 0 → continue` is only reachable for a genuine HTTP-200 empty-result window, which is a correct prune. A failed `getOutputs` can never be misread as "range is empty → prune everything."

I also checked the lagging-node case (a node reporting a lower tip prunes outputs above it): this **self-heals** on the next good scan because `start` is anchored to `checkpoint - REORG_BUFFER` and the full range is re-queried, and both balance and history always re-derive from node truth. No path to permanent loss.

### Other directions

- **Spent-but-still-mined retained** — prune keys on on-chain presence (`chain_getOutputs` → `discovered`), NOT spent status. A spent-but-mined output is re-returned → retained → split out of `spendable` by the live `chain_areKeyImagesSpent` check → rendered `direction: 'spent'`. Test asserts retained + in spent set + `direction === 'spent'`. Correct.
- **Over-report (the original #1099 bug)** — a reorged-out output is now pruned and drops from both the balance sum and `deriveHistory`. Test asserts. Fixed.
- **Boundaries** — `[start, tip]` inclusive both ends; `h === start` in-range (pruned unless rediscovered), `h === start-1` retained, `h === tip` requires rediscovery; `tip < start` → `windows() === []` → only `h <= tip` survive (deep-shrink test). All boundary tests present and green.
- **No schema churn** — `STATE_VERSION` still 1, `PersistedOwnedOutput` shape unchanged, `deriveHistory` consistent. `mergeOwnedOutputs` correctly kept off the scan path for its dedupe unit contract. Committed `snap.manifest.json` shasum matches the freshly built bundle (no git drift after `mm-snap build`).

### CI / merge

- `cargo deny` — pass. `web-ci` — pass; verified the run's steps include **Build wasm-signer bundler artifact** + **Run Snap tests (@metamask/snaps-jest)** (the #1106 coverage), so the snap suite really ran in CI.
- Merge state **CLEAN**, mergeable. Local `pnpm test:snap` **85/85**.

The dangerous under-report direction is bounded and, critically, an RPC failure throws before any persist so it can never be mistaken for an empty range. LGTM — approving.

Issue #1114, opened 2026-07-21
ops(testnet): eu/ap regional relays stranded on the pre-6.0.0 chain (v0.3.2, height 3233, 0 peers)


rjwalters:
## Observed (live, 2026-07-20)

Polling `node_getStatus` on all five ingress nodes:

| Node | Version | Height | peerCount | scpPeerCount | synced |
|---|---|---|---|---|---|
| seed / seed2 / faucet (US validators) | **0.6.0** | **202** | 2 | 2 | ✓ |
| eu.seed (Frankfurt relay) | **0.3.2** | **3233** | **0** | **0** | ✓ |
| ap.seed (Singapore relay) | **0.3.2** | **3233** | **0** | **0** | ✓ |

The three US validators are a healthy quorum on the current chain (v0.6.0 / protocol 6.0.0, height frozen at 202 = faucet 10k-BTH minting-cap pause). The **two regional relays were never brought across the 2026-07-16 v0.6.0 reset** — they are the "zombie old-protocol nodes" `infra/seed/TESTNET_RESET.md` warns about, still serving the *old* pre-reset chain (height 3233), isolated (0 peers), on stale software (0.3.2).

## Impact

- Anyone selecting **EU** or **AP** as their wallet ingress talks to a dead, forked chain.
- Both report `synced: true` (synced to their own singleton view) so they render green in the network dashboard — a false-healthy signal. (The dashboard-side hardening is #1115 ; this issue is the actual fix.)

## Fix (operator — needs SSH + AWS/DNS creds; cannot be automated)

Per `infra/seed/TESTNET_RESET.md` + `infra/seed/deploy-botho.sh` / `reset-to-testnet.sh`:

1. Deploy the v0.6.0 release binary to `eu.seed.botho.io` and `ap.seed.botho.io` (`deploy-botho.sh ubuntu@<host>`, `RELEASE_TAG` = the 0.6.0 tag the validators run).
2. Wipe the old-chain data + install the current testnet systemd unit (`reset-to-testnet.sh` on-host).
3. Re-bootstrap each relay against the live seeds so it syncs the 6.0.0 chain to height 202 and establishes peers.
4. Verify: `node_getStatus` on each relay shows `nodeVersion 0.6.0`, `chainHeight 202`, `peerCount > 0`, matching `tipHash`.

[...]

PR #1116, opened 2026-07-21
fix(web/network): don't let peer-isolated nodes poison fleet consensus height (#1115)

### LINKED ISSUE #1115: web(network): peer-isolated/forked relay poisons fleet consensus height and renders green
## Bug

`deriveFleetSummary` (`web/packages/features/src/network/fleet.ts`) sets `consensusHeight = Math.max(...allReachableHeights)`. A reachable-but-isolated node stuck on a stale/forked chain at a far-higher height therefore **wins the max** and becomes the reported fleet consensus tip.

Concretely, with the live eu/ap relay drift (see #1114): two isolated relays at height 3233 (0 peers, old chain) + three connected validators at 202 →
- **Consensus height renders as 3233** (the dead fork), not 202.
- `nodesInSync` counts only the two relays → the three *real* validators are painted as **3031 blocks behind** with a warning border. Exactly backwards.
- Both relays report `synced: true` (singleton view) so their cards render green — a false-healthy signal (the #541 hardcoded-observability lesson).

## Fix

A node with `peerCount === 0` is not participating in the mesh; its height is a singleton view and must not define or count toward consensus.

- `deriveFleetSummary`: derive `consensusHeight`/`nodesInSync` from **connected** nodes (`peerCount !== 0`); fall back to all-reachable only when none are connected (lone dev node). `undefined` peerCount is given the benefit of the doubt (back-compat). Add `nodesIsolated` to `FleetSummary`.
- `NodeCard`: render an explicit "isolated — 0 peers, off consensus" badge for reachable nodes with 0 peers, and suppress the misleading "N blocks behind" for them.
- `FleetSummaryStrip`: surface the isolated count ("N isolated") alongside unreachable, and warn when any node is isolated.

## Acceptance
- Isolated relays at a higher stale height do NOT change `consensusHeight` (unit test with the 202 vs 3233 scenario).
- Isolated nodes show the isolation badge, not "green/synced" and not "behind".
- Existing behavior preserved when `peerCount` is absent.

Files: `web/packages/features/src/network/{fleet.ts,types.ts,components/node-card.tsx,components/fleet-summary.tsx}` + tests.

rjwalters:
Closes #1115. Related: #1114 (the live operator fix for the eu/ap relays that surfaced this).

## Problem

`deriveFleetSummary` computed `consensusHeight = Math.max(...allReachableHeights)`. A reachable-but-**isolated** node (0 peers) stuck on a stale/forked chain at a far-higher height wins that max and becomes the reported fleet tip.

Live trigger: the eu/ap regional relays are stranded on the pre-6.0.0 chain (v0.3.2, **height 3233, 0 peers**) while the three validators are on v0.6.0 at **height 202**. Result on the network tab:
- Consensus height shows **3233** (the dead fork), not 202.
- The three *real* validators render as **3031 blocks behind** (warning border) — backwards.
- Both relays report `synced: true` (their own singleton view) → **green cards** (the #541 hardcoded-observability trap).

## Fix

A node with `peerCount === 0` isn't in the mesh; its height is a singleton view and must not define or count toward consensus.

- **`deriveFleetSummary`** — consensus/in-sync derived from connected nodes (`peerCount !== 0`); fall back to all-reachable only when none are connected (lone dev node); `undefined` peerCount kept (back-compat). New `nodesIsolated` field.
- **`NodeCard`** — explicit `isolated — 0 peers, off consensus` badge for reachable 0-peer nodes; the misleading "N blocks behind" is suppressed for them.
- **`FleetSummaryStrip`** — surfaces the isolated count ("N isolated") and warns when any node is isolated.

The discriminator is **peer connectivity, not network id** — all five nodes report `network: "botho-testnet"`, so the name can't distinguish the fork; `peerCount` can.

## Tests
`pnpm vitest run` on the two touched suites: **22 pass** (4 new: the 202-vs-3233 poisoning case, lone-dev fallback, undefined-peerCount back-compat, and a dashboard render test asserting the isolated badge + un-poisoned consensus). `@botho/features` typecheck clean.

### DIFF
--- a/web/packages/features/src/network/components/fleet-summary.tsx
+++ b/web/packages/features/src/network/components/fleet-summary.tsx
@@ -33,12 +33,12 @@ export function FleetSummaryStrip({
           }
           label="Nodes in sync"
           value={`${summary.nodesInSync}/${summary.nodesTotal}`}
-          sub={
-            summary.nodesReachable < summary.nodesTotal
-              ? `${summary.nodesTotal - summary.nodesReachable} unreachable`
-              : undefined
+          sub={syncSub(summary)}
+          warn={
+            summary.nodesInSync < summary.nodesReachable ||
+            summary.nodesIsolated > 0 ||
+            summary.anySlotStalled
           }
-          warn={summary.nodesInSync < summary.nodesReachable || summary.anySlotStalled}
         />
         <Metric
           icon={<Layers className="h-4 w-4" />}
@@ -59,6 +59,19 @@ export function FleetSummaryStrip({
       </CardContent>
     </Card>
   )
+}
+
+/**
+ * Sub-label for the "Nodes in sync" metric: call out unreachable and
+ * peer-isolated (off-consensus) nodes so a stale/forked relay is visible
+ * rather than silently green.
+ */
+function syncSub(summary: FleetSummary): string | undefined {
+  const parts: string[] = []
+  const unreachable = summary.nodesTotal - summary.nodesReachable
+  if (unreachable > 0) parts.push(`${unreachable} unreachable`)
+  if (summary.nodesIsolated > 0) parts.push(`${summary.nodesIsolated} isolated`)
+  return parts.length > 0 ? parts.join(', ') : undefined
 }
 
 /**
--- a/web/packages/features/src/network/components/network-dashboard.test.tsx
+++ b/web/packages/features/src/network/components/network-dashboard.test.tsx
@@ -76,6 +76,23 @@ describe('NetworkDashboard', () => {
     expect(screen.getByText('2/3')).toBeDefined()
   })
 
+  it('flags peer-isolated relays and does not let their stale height poison consensus', () => {
+    // The live eu/ap drift: one connected validator at 202, two isolated
+    // relays stuck on the old pre-reset chain at 3233 with zero peers.
+    renderDash({
+      seed: live('seed', { chainHeight: 202, peerCount: 2 }),
+      eu: live('eu', { chainHeight: 3233, peerCount: 0 }),
+      ap: live('ap', { chainHeight: 3233, peerCount: 0 }),
+    })
+    // Consensus is the connected validator's height, not the isolated relays'.
+    expect(screen.getAllByText('202')).toHaveLength(2) // summary + seed card
+    // Both isolated relays are called out; the validator is NOT "behind".
+    expect(screen.getAllByText(/isolated — 0 peers/)).toHaveLength(2)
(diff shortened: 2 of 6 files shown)


rjwalters:
## ✅ Approved

Thorough Judge review complete. This is a clean, well-scoped fix for a genuinely dangerous observability bug (the #541 class: a dead node rendering green while poisoning the reported fleet tip).

### Correctness — connectivity logic and fallbacks verified

Walked every branch of `deriveFleetSummary`:

- **`peerCount === 0` (isolated):** excluded from `participating`, counted in `nodesIsolated`. Correct — a 0-peer node is a singleton view.
- **`peerCount === undefined` (older node):** `s.peerCount !== 0` is `true`, so it is kept as participating. Back-compat preserved as documented.
- **All nodes isolated (lone dev node):** `participating.length === 0` → `heightPool` falls back to `reachable`, so the strip shows a height instead of an em-dash. `nodesIsolated` still reflects the count. Good.
- **Nothing reachable:** `reachable=[]` → `heightPool=[]` → `heights=[]` → `consensusHeight=null`, `nodesInSync=0`. Correct (existing test still green).
- **`nodesInSync`** correctly derives from `heightPool` (not `reachable`), so isolated nodes never inflate the in-sync count — the exact 3-of-5 case is pinned at 3.

### Tests pin the behavior

`npx vitest run` on both suites: **22 passed** (12 fleet + 10 dashboard). The 4 new tests are the right ones: the live 202-vs-3233 poisoning case (unit + full dashboard render), lone-dev fallback, and undefined-peerCount back-compat. The dashboard test asserts both the isolated badge and the un-poisoned consensus (202, not 3233) and the absence of a "blocks behind" badge on the real validator — it would fail against the old `Math.max`-over-all-reachable code.

### No consumer regressions

The new required `nodesIsolated` field is only produced by `deriveFleetSummary` (which sets it) and consumed by `FleetSummaryStrip`. The operator `trust-dashboard.tsx` uses a *different* type (`deriveTrustSummary`/`TrustSummary`), so it is unaffected. `@botho/features` typecheck is clean.

### UX

The three states are now distinct and honest: **unreachable** (danger border + WifiOff), **isolated** (danger border + "isolated — 0 peers, off consensus", peers stat warned, no misleading "N behind"), and **lagging** (warning border + "N blocks behind"). Suppressing the lag badge for isolated nodes is the right call. The strip surfaces "N isolated" alongside "N unreachable" and warns when any node is isolated.

### CI
- `typecheck + vitest + snap + worker dry-run` — pass
- `cargo deny` — pass
- mergeStateStatus: CLEAN

PR uses `Closes #1115` correctly. Solid work.

PR #1121, opened 2026-07-21
fix(rpc): report node isolation instead of latched synced:true at 0 peers

### LINKED ISSUE #1118: node(rpc): node_getStatus reports synced:true while peer-isolated on a stale singleton fork
## Context

Surfaced during the 2026-07-20 eu/ap relay outage (#1114). Two regional relays were stranded on a dead pre-reset chain — **0 peers**, on the wrong tip at height 3233 (validators were at 202), with `lastExternalizedSecondsAgo` ≈ 4.8–7.8 **days** — yet `node_getStatus` reported `synced: true` / `syncStatus: "synced"`. From the node's own singleton view it *is* at its own tip, so it self-certifies as synced.

This is distinct from the already-fixed #541 (which removed the *hardcoded* always-'synced' value by wiring `ChainSyncManager`). Here the sync machinery is wired, but "synced" ignores whether the node is actually connected to / keeping up with a live network. It forced the client-side workaround in #1116 (the network dashboard now excludes `peerCount === 0` nodes from consensus because it can't trust their `synced` flag).

Related: #541, #542, #543, #544 (the hardcoded-observability cluster, all closed).

## Repro

Point a node's `bootstrap_peers` at an unreachable/wrong set (or firewall its gossip port) so it runs at 0 peers on an old chain, then call `node_getStatus`: observe `synced: true` despite isolation and a days-stale last-externalized time.

## Suggested acceptance criteria
- [ ] `synced` / `syncStatus` reflect peer isolation and/or externalize-staleness — a 0-peer node, or one whose last externalize is far older than the slot cadence, is not reported as plain `synced`.
- [ ] Consider a distinct status (e.g. `isolated` / `stale`) or a boolean the dashboard can trust, so clients don't have to infer health from `peerCount` (removing the need for the #1116 workaround).
- [ ] Unit/RPC test pinning the isolated/stale case.

---

### Root Cause (verified against current `main`)

The bug is in `ChainSyncManager`'s state machine, not in `node_getStatus` itself — `node_getStatus` is faithfully reporting what the sync manager believes.

- `handle_node_status` (`botho/src/rpc/mod.rs:1179-1320`) reads `synced`/`syncStatus`/`syncProgress` straight from `state.sync_status_snapshot()` (`rpc/mod.rs:1194-1202`), which is a clone of whatever `ChainSyncManager::status_snapshot()` last published.
- `status_snapshot()` (`botho/src/network/sync.rs:625-648`) derives `synced` purely from `matches!(self.state, SyncState::Synced)` — it does not look at peer count at all.
- The `SyncState::Synced` arm of `tick()` (`botho/src/network/sync.rs:1288-1325`) only re-evaluates via `self.best_peer()`, which reads `self.peer_statuses`. But `on_peer_disconnected` (`sync.rs:1093-1113`) removes the disconnecting peer from `peer_statuses` on every disconnect and **only forces a state transition (`Synced`/`Downloading` → `Discovery`) when the disconnecting peer was the active download peer**. When a node in `SyncState::Synced` loses its *last* peer, `peer_statuses` becomes empty, `best_peer()` returns `None`, and the `Synced` tick arm's `if let Some((best_peer, status)) = self.best_peer()` guard simply never fires — nothing detects "we now have 0 peers." The node stays latched in `SyncState::Synced` indefinitely, however stale it gets.
- This is confirmed by the call site: `sync_manager.tick(&connected)` (`botho/src/commands/run.rs:2055-2056`) is fed the live libp2p `connected_peers` list every tick, and the exact same list backs `state.peer_count` (`peerCount` in the RPC response, mirrored at `commands/run.rs:1080-1093`). So `peerCount == 0` and "the sync manager has no way to notice isolation" are the same condition, just never cross-checked.

This is **not** a re-run of the hardcoded-field bug class (#541-#544) — the field is genuinely wired to live state, as the comments at `sync.rs:440` and `rpc/mod.rs:1187-1196` note. It's a state-machine gap: `Synced` is a fixed point with no isolation escape hatch, and the RPC layer doesn't independently cross-check `synced` against `peerCount`/staleness even though both signals are already flowing through the same response (`peerCount` at `rpc/mod.rs:1266`, `lastExternalizedSecondsAgo`/`effectiveSlotDurationSecs` from the SCP-slot snapshot at `rpc/mod.rs:1316-1317`, sourced from `consensus/service.rs` `SLOT_STALL_THRESHOLD_MULTIPLIER` = 3× the effective slot duration).

Also note `SyncStatusSnapshot::progress_percent()` (`sync.rs:464-...`) returns `Some(100.0)` whenever `synced == true`, so `syncProgress` also lies for an isolated node — worth fixing in the same PR.

Scope check: `ChainSyncManager::is_synced()` has exactly one other live consumer, `commands/run.rs:2098` (`if !initial_sync_complete && sync_manager.is_synced() { ... }`), which is a one-shot latch for "have we ever completed initial catch-up" — it is not re-evaluated as a live minting/propose gate, so this is an **observability bug, not a consensus-safety bug**. That keeps the fix scoped to reporting, not state-machine semantics that other subsystems depend on.

### Implementation Guidance

**Recommended: fix at the RPC layer (`handle_node_status`), not inside `ChainSyncManager`.**

Rationale: `ChainSyncManager` already has ~20 internal unit tests asserting `is_synced()` transitions, and `is_synced()`'s only other consumer (`run.rs:2098`) is a one-shot latch that shouldn't flap based on transient peer loss. Teaching the state machine itself about isolation (e.g. a new `SyncState::Isolated` variant) would touch every exhaustive match on `SyncState` in `sync.rs` and is a materially bigger, riskier change for what is currently an observability-only defect.

In `handle_node_status` (`botho/src/rpc/mod.rs`, around line 1194-1202, before the `synced`/`sync_status`/`sync_progress` locals are computed):

```rust
let sync_snapshot = state.sync_status_snapshot();
let raw_synced = sync_snapshot.as_ref().map(|s| s.synced).unwrap_or(true);
let raw_status: &str = sync_snapshot.as_ref().map(|s| s.status).unwrap_or("synced");

// A node with 0 peers has no live signal to justify claiming it is caught up
// with the network — `ChainSyncManager` has no isolation escape hatch (it
// simply latches in `Synced` once `best_peer()` has no candidates to
// disagree with; see sync.rs's `Synced` tick() arm). Override the raw
// snapshot rather than trust it when isolated (anti-#541-#544: still
// grounded in the real, live `peers` count, never a fabricated constant).
let isolated = peers == 0;
let synced = raw_synced && !isolated;
let sync_status: &str = if isolated { "isolated" } else { raw_status };
let sync_progress: f64 = if isolated {
    // Do not claim 100% progress for a node with no peer to measure
    // progress against.
    sync_snapshot.as_ref().and_then(|s| s.target_height).map(|_| 0.0).unwrap_or(0.0)
} else {
    match sync_snapshot.as_ref() {
        Some(s) => s.progress_percent().unwrap_or(100.0),
        None => 100.0,
    }
};
```

(`peers` is already computed one line above at `rpc/mod.rs:1184`, so no new plumbing is required — reorder if needed so `peers` is available before this block.)

Notes / open decisions for the Builder:
- The exact string value (`"isolated"` vs. `"stale"` vs. reusing `"stalled"`) is a naming choice; the issue's own suggested AC proposes `isolated`/`stale`. Pick one and use it consistently — check `SyncStatusSnapshot::status`'s existing four values (`"discovering"`, `"syncing"`, `"synced"`, `"stalled"`) don't already imply a meaning you'd collide with (`"stalled"` currently means `SyncState::Failed`, a *download* failure — distinct from peer isolation, so don't reuse it).
- Consider whether "0 peers" should also account for the *SCP*-specific peer count (`scpPeerCount`) vs. the general libp2p `peerCount` — in the reported repro they're identical (relay/discovery peer count == SCP peer count per `commands/run.rs:493`'s comment), but double-check that assumption doesn't drift for validator vs relay node roles.
- Out of scope for this issue (call out as a natural follow-up, do not implement here): once `node_getStatus` itself reports isolation honestly, the client-side `peerCount === 0` filter added in `web/packages/features/src/network/fleet.ts:75` (PR #1116, issue #1115) becomes redundant and could be simplified to trust `synced`/`syncStatus` directly. Leave that as a suggested fast-follow, not part of this PR's scope, to keep the change small and single-purpose.

### Acceptance Criteria

- [ ] `node_getStatus` reports `synced: false` when `peerCount == 0`, regardless of the underlying `ChainSyncManager`/`SyncStatusSnapshot` state (including when the manager is latched in `SyncState::Synced`).
- [ ] `node_getStatus`'s `syncStatus` reports a distinct value (e.g. `"isolated"`) — not `"synced"` — when `peerCount == 0`, distinguishable from the existing `"discovering"` / `"syncing"` / `"synced"` / `"stalled"` values.
- [ ] `syncProgress` does not report `100.0` for an isolated (0-peer) node.
- [ ] A node with `peerCount > 0` and a genuinely `Synced` `ChainSyncManager` state is unaffected — no regression to the healthy-multi-peer case.
- [ ] Unit/RPC test pins the isolated case: an `RpcState` wired with a `SyncStatusSnapshot { synced: true, status: "synced", .. }` sync handle and `peer_count` set to `0` yields `synced: false` / `syncStatus: "isolated"` from `handle_node_status`.
- [ ] Existing `ChainSyncManager` unit tests in `botho/src/network/sync.rs` and existing `node_getStatus` tests in `botho/tests/rpc_integration.rs` continue to pass unmodified (this is an RPC-layer-only fix; no `SyncState` semantics change).

### Affected Files

- `botho/src/rpc/mod.rs` — `handle_node_status` (~lines 1179-1320): cross-check the raw sync snapshot against `peers == 0` before computing `synced`/`syncStatus`/`syncProgress`; add/extend a unit test near the existing `handle_node_status` test at line ~4638.
- `botho/tests/rpc_integration.rs` — extend or add an integration test covering the 0-peer isolated case (existing `node_getStatus` calls at lines 121, 928, 1063-1064, 1325 show the test harness pattern).
- `botho/src/network/sync.rs` — reference only (root-cause location: `status_snapshot()` at lines 625-648, `Synced` tick arm at lines 1288-1325, `on_peer_disconnected` at lines 1093-1113); not expected to change under the recommended (RPC-layer) fix, but a Builder choosing the deeper state-machine fix (see Implementation Guidance) would touch this file plus its ~20 internal tests that assert on `SyncState`/`is_synced()`.

### Test Plan

- [ ] Manual verification: spin up a node with 0 connected peers (or firewall its gossip port per the issue's repro) after it has reached `SyncState::Synced`, call `node_getStatus`, confirm `synced: false` and `syncStatus: "isolated"` (or chosen name) instead of `synced: true`.
- [ ] Automated tests: new unit test in `botho/src/rpc/mod.rs`'s test module (mirroring the pattern at line ~4638) constructing an `RpcState` with `peer_count = 0` and a `SyncStatusSnapshot { synced: true, .. }`, asserting the RPC response overrides to isolated.
- [ ] Regression: run `cargo test -p botho --lib rpc::` and `cargo test -p botho --test rpc_integration` to confirm no existing `node_getStatus` assertions (e.g. `test_operator_*` tests that call `node_getStatus` as a smoke check, `rpc_integration.rs:1325`) break.
- [ ] Edge case: a node with `peerCount == 0` that has *never* connected to any peer (single-node/genesis/test setups, where `sync_status_snapshot()` returns `None` and the current fallback is "assume caught up") — confirm intended behavior for this case explicitly (likely still fine to report synced, since a lone dev/test node genuinely has nothing to sync against and no live network to be isolated *from*; the bug is specifically about a node that *was* connected and lost all peers). Document the distinction in the fix's code comments so it doesn't get miscategorized as isolation.

### Related

- #1114 (closed) — the live-testnet healing incident that surfaced this.
- #1115 / PR #1116 — the client-side dashboard workaround (`peerCount === 0` filter in `web/packages/features/src/network/fleet.ts`) that this issue's fix would make redundant (not removed here — flagged as a suggested fast-follow above).
- #541-#544 — the hardcoded-observability cluster (closed); this issue is explicitly *not* that pattern (the field is live-wired), see Root Cause above.
- #538, #653 — prior precedent for deriving *derived* health verdicts (`minerStalled`, `slotStalled`) from live snapshots rather than raw state, which this fix follows.

rjwalters:
## Summary

`node_getStatus` reported `synced: true` for a node stranded on a stale singleton fork with **0 peers** (the #1114 eu/ap relay outage: relays days behind the live chain still self-certifying as synced). The root cause is a `ChainSyncManager` state-machine gap: once it reaches `SyncState::Synced` it stays latched there even after its last peer disconnects — `best_peer()` returns `None` and the `Synced` tick arm never re-evaluates, so nothing detects "we now have 0 peers." `node_getStatus` faithfully reported the latched state.

Per the curated guidance, this is fixed at the **RPC layer** (an observability-only defect), not by rewriting the `SyncState` machine that ~20 sync unit tests and the one-shot initial-sync latch depend on.

## Changes

- `handle_node_status` (`botho/src/rpc/mod.rs`) now cross-checks the published sync snapshot against the live `peers` count:
  - `synced: false` when `peers == 0` (even if the manager is latched in `SyncState::Synced`)
  - `syncStatus: "isolated"` — a new value distinct from the existing `discovering` / `syncing` / `synced` / `stalled` (note: `stalled` means a *download* failure, not peer isolation, so it is not reused)
  - `syncProgress: 0.0` instead of the latched `100.0`
  - Gated on a snapshot having been published (`sync_snapshot.is_some()`): a lone single-node / genesis / test node that never ran a sync loop returns `None` and still reports caught-up — it has no live network to be isolated *from*. The bug is specifically a node that *was* connected and then lost every peer.
- Grounded entirely in the real, live `peers` count — never a fabricated constant (anti-#541–#544). No `SyncState` semantics change.
- Tests: new `test_node_status_reports_isolated_when_no_peers`; updated `test_node_status_reports_real_sync_state` to set a realistic peer count (a downloading/synced node has peers).

| `synced: false` when `peerCount == 0` even if latched `Synced` | ✅ | New test asserts it against a `synced:true` snapshot |
| `syncStatus` distinct value (`isolated`) when isolated | ✅ | New test; value distinct from discovering/syncing/synced/stalled |
| `syncProgress` not `100.0` for isolated node | ✅ | New test asserts `0.0` |
| `peerCount > 0` + genuinely `Synced` unaffected | ✅ | Regression guard in new test (same snapshot, 3 peers → synced/100) |
| Unit/RPC test pins the isolated case | ✅ | `test_node_status_reports_isolated_when_no_peers` |
| Existing sync.rs unit + rpc_integration tests pass unmodified | ✅ | 81 sync unit + 46 integration tests green, untouched |

## Test Plan

- `cargo build -p botho` — clean (pre-existing warnings only)
- `cargo test -p botho --lib rpc::` — 123 passed / 0 failed
- `cargo test -p botho --lib network::sync::` — 81 passed / 0 failed
- `cargo test -p botho --test rpc_integration` — 46 passed / 0 failed

## Follow-up (out of scope)

Once `node_getStatus` reports isolation honestly, the client-side `peerCount === 0` filter added in `web/packages/features/src/network/fleet.ts` (PR #1116 / #1115) becomes redundant and could be simplified to trust `synced`/`syncStatus` directly. Left as a suggested fast-follow to keep this PR single-purpose.

Closes #1118

### DIFF
--- a/botho/src/rpc/mod.rs
+++ b/botho/src/rpc/mod.rs
@@ -1192,13 +1192,40 @@ async fn handle_node_status(id: Value, state: &RpcState) -> JsonRpcResponse {
     // loop has not yet published a snapshot, fall back to the caught-up
     // assumption — a lone node with no peers has nothing to sync against.
     let sync_snapshot = state.sync_status_snapshot();
-    let synced = sync_snapshot.as_ref().map(|s| s.synced).unwrap_or(true);
-    let sync_status: &str = sync_snapshot.as_ref().map(|s| s.status).unwrap_or("synced");
+    let raw_synced = sync_snapshot.as_ref().map(|s| s.synced).unwrap_or(true);
+    let raw_status: &str = sync_snapshot.as_ref().map(|s| s.status).unwrap_or("synced");
+
+    // Peer-isolation cross-check (#1118). `ChainSyncManager` has no isolation
+    // escape hatch: once it reaches `SyncState::Synced` it stays latched there
+    // even after its last peer disconnects (`best_peer()` then returns `None`
+    // and the `Synced` tick arm never re-evaluates), so a node stranded on a
+    // stale singleton fork keeps self-certifying as `synced` — the #1114 relay
+    // outage, where two 0-peer relays reported `synced: true` while days behind
+    // the live chain. Cross-check the raw snapshot against the live `peers`
+    // count here rather than trust it.
+    //
+    // Gated on a snapshot having been published (`sync_snapshot.is_some()`): a
+    // lone dev / single-node / genesis node that never wired or ran a sync loop
+    // returns `None` and is legitimately caught up — it has no live network to
+    // be isolated *from*. The bug is specifically a node that *was* connected,
+    // published sync state, then lost every peer. This stays grounded entirely
+    // in the real, live `peers` count — never a fabricated constant
+    // (anti-#541–#544).
+    let isolated = peers == 0 && sync_snapshot.is_some();
+    let synced = raw_synced && !isolated;
+    let sync_status: &str = if isolated { "isolated" } else { raw_status };
     // Real percentage when a best-known tip is available; 100.0 when synced or
-    // when we have no peer to compare against (nothing to catch up to).
-    let sync_progress: f64 = match sync_snapshot.as_ref() {
-        Some(s) => s.progress_percent().unwrap_or(100.0),
-        None => 100.0,
+    // when we have no peer to compare against (nothing to catch up to). An
+    // isolated node reports 0.0 — it has no peer to measure progress against, so
+    // claiming 100% (which `progress_percent()` returns whenever `synced`)
+    // would be the same lie as the latched `synced` flag.
+    let sync_progress: f64 = if isolated {
+        0.0
+    } else {
+        match sync_snapshot.as_ref() {
+            Some(s) => s.progress_percent().unwrap_or(100.0),
+            None => 100.0,
+        }
     };
 
     // Byzantine-fault-tolerance posture (#509). Participating node count
@@ -4806,6 +4833,12 @@ mod tests {
         )
         .with_sync_status(sync_handle.clone());
 
+        // A node that is downloading from / caught up with the network has live
+        // peers; set a realistic count so the #1118 isolation cross-check does
+        // not fire (that path is exercised separately in
+        // `test_node_status_reports_isolated_when_no_peers`).
+        *state.peer_count.write().unwrap() = 3;
+
         // --- Behind: downloading, local 50 of best-known 100. ---
         *sync_handle.write().unwrap() = Some(SyncStatusSnapshot {
             synced: false,
@@ -4856,6 +4889,72 @@ mod tests {
             Arc::new(WsBroadcaster::new(16)),
         );
 
+        let resp = handle_node_status(json!(1), &state).await;
+        let result = resp.result.unwrap();
+        assert_eq!(result["synced"], json!(true));
+        assert_eq!(result["syncStatus"], json!("synced"));
+        assert_eq!(result["syncProgress"].as_f64().unwrap(), 100.0);
+    }
+
+    /// #1118: a node whose `ChainSyncManager` is latched in `SyncState::Synced`
+    /// but has lost every peer must NOT keep reporting `synced: true`. This is
+    /// the #1114 relay-outage failure: a node stranded on a stale singleton
+    /// fork self-certifies as synced because the state machine has no
+    /// isolation escape hatch. `handle_node_status` cross-checks the
+    /// published snapshot against the live peer count and reports `synced:
+    /// false` / `syncStatus: "isolated"` / `syncProgress: 0` instead. A
+    /// node with the SAME snapshot but live peers is unaffected.
+    #[tokio::test]
+    async fn test_node_status_reports_isolated_when_no_peers() {
+        use crate::{ledger::Ledger, mempool::Mempool, network::SyncStatusSnapshot};
+
+        let dir = tempfile::tempdir().unwrap();
+        let ledger = Ledger::open(dir.path()).unwrap();
+
+        let sync_handle = Arc::new(RwLock::new(None));
+        let state = RpcState::new(
+            ledger,
+            Mempool::new(),
+            Network::Testnet,
+            None,
+            None,
+            vec![],
+            Arc::new(WsBroadcaster::new(16)),
+        )
+        .with_sync_status(sync_handle.clone());
+
+        // The sync manager is latched in Synced (it published a synced snapshot
+        // before its peers dropped) and claims a full 100% progress.
+        *sync_handle.write().unwrap() = Some(SyncStatusSnapshot {
+            synced: true,
+            status: "synced",
+            local_height: 3233,
+            target_height: Some(3233),
+        });
+
+        // --- Isolated: 0 peers overrides the latched `synced` snapshot. ---
+        *state.peer_count.write().unwrap() = 0;
+        let resp = handle_node_status(json!(1), &state).await;
+        let result = resp.result.unwrap();
+        assert_eq!(
+            result["synced"],
+            json!(false),
+            "a 0-peer node must not report synced even when latched in SyncState::Synced"
+        );
+        assert_eq!(
+            result["syncStatus"],
+            json!("isolated"),
+            "syncStatus must be a distinct `isolated` value, not `synced`"
+        );
+        assert_eq!(
+            result["syncProgress"].as_f64().unwrap(),
+            0.0,
+            "an isolated node has no peer to measure progress against — must not claim 100%"
+        );
+        assert_eq!(result["peerCount"], json!(0));
+
+        // --- Regression guard: the SAME snapshot with live peers is healthy. ---
+        *state.peer_count.write().unwrap() = 3;
         let resp = handle_node_status(json!(1), &state).await;
         let result = resp.result.unwrap();
         assert_eq!(result["synced"], json!(true));


rjwalters:
✅ **Approved**

Faithful, well-scoped implementation of the curated RPC-layer fix for #1118. Reviewed the diff, ran the suites, and verified CI.

## What I verified by execution (in `.loom/worktrees/issue-1118`)
- `cargo build -p botho` — clean (pre-existing warnings only)
- `cargo test -p botho --lib rpc::` — **123 passed / 0 failed**
- `cargo test -p botho --lib network::sync::` — **81 passed / 0 failed** (unmodified; no `SyncState` semantics changed)
- `cargo test -p botho --test rpc_integration` — **46 passed / 0 failed**
- Named tests confirmed green: `test_node_status_reports_isolated_when_no_peers` (new) and `test_node_status_reports_real_sync_state` (edited)
- CI: all required checks pass (`cargo build --workspace (Linux)` 22m, benchmarks, `cargo deny`, `rustfmt`); merge state **CLEAN**

## Scrutiny of the flagged concerns
- **`is_some()` gating** — `isolated = peers == 0 && sync_snapshot.is_some()`. A never-connected single/genesis/test node returns a `None` snapshot, so `isolated` is false and `synced` falls back to the caught-up assumption (`unwrap_or(true)`). Correct — a lone node has no live network to be isolated *from*. Covered by the pre-existing `test_node_status_sync_fallback_when_no_handle` (still green).
- **`"isolated"` is a genuinely new status value** — distinct from the four existing (`discovering`/`syncing`/`synced`/`stalled`); `stalled` (download failure) is correctly not reused. No misparse risk for consumers: the client dashboard already excludes 0-peer nodes from consensus (#1116), and the safety property (never report `synced` at 0 peers) holds regardless of how the string is consumed.
- **`syncProgress` → 0.0** — sensible; `progress_percent()` returns `Some(100.0)` whenever `synced`, which would be the same lie. Zero is the honest value with no peer to measure against.
- **No consensus/sync state-machine change** — confirmed RPC-observability-only; `ChainSyncManager` untouched, `is_synced()`'s one-shot initial-sync latch consumer unaffected. This is not a consensus-safety change.
- **Edit to `test_node_status_reports_real_sync_state` (peer_count = 3)** — justified and does not mask a regression. That test previously left `peer_count` at its default 0, which the new cross-check would now flag as `isolated` and override the snapshot-driven assertions. Setting a realistic peer count restores the test's actual precondition (a downloading/synced node has peers); all its original assertions (behind/downloading/synced progress values) are preserved. The isolation path is pinned separately in the new test with a proper regression guard (same snapshot + 3 peers → `synced`/100).

Grounded entirely in the live `peers` count (anti-#541–#544), correctly narrows the earlier client-side workaround into an honest server-side signal, and leaves the `fleet.ts` simplification as an explicit out-of-scope follow-up. PR body uses `Closes #1118`.

LGTM.

Issue #852, opened 2026-07-13
ci(contracts): compile + test contracts/ethereum in CI, and pin the npm lockfile


rjwalters:
Follow-up from the PR #851 review (issue #826).

Two related gaps:

1. **No CI coverage for `contracts/ethereum/`.** No workflow compiles or tests the Solidity side; the #851 review verified the 30-test Hardhat suite locally, but nothing prevents a future PR from silently breaking the contract or the ABI-pinning tests (`bridgeMint` selector + `BridgeMint`/`BridgeBurn` topic layout that `bridge/service/src/mint/ethereum.rs` and `watchers/ethereum.rs` bind). A small workflow triggered on `contracts/ethereum/**` (and ideally also on `bridge/service/src/{mint,watchers}/ethereum.rs`) running `npm ci && npx hardhat test` closes this.

2. **OpenZeppelin version is not pinned in-repo.** `contracts/ethereum/package.json` uses caret ranges (`@openzeppelin/contracts: ^5.0.0`) and the `.gitignore` added in #851 excludes `package-lock.json`. For a security-critical bridge contract, builds should be reproducible: commit the lockfile (required for `npm ci` in the workflow anyway) or pin exact dependency versions, so the audited OZ version is the deployed one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1185, opened 2026-08-11
build(contracts/ethereum): migrate to the hardhat 3 toolchain (toolbox-mocha-ethers 3, typescript 7)

### LINKED ISSUE #1174: Ethereum contracts toolchain upgrade: hardhat 3 + hardhat-toolbox 7 + typescript 7
Three interlocked Dependabot majors against `contracts/ethereum` fail its `hardhat compile + test` CI and cannot merge independently:
- hardhat 2.28 → 3.x (#1131): ESM-first, new config format, plugin system rework
- @nomicfoundation/hardhat-toolbox 4 → 7 (#1127): requires hardhat 3
- typescript 5.9 → 7 (#1125): fails against the hardhat-2 toolchain's ts integration

**Task**: one deliberate migration of `contracts/ethereum` to the hardhat-3 toolchain:
- Port hardhat.config.ts to the v3 format (ESM, `plugins` array), update toolbox to 7, then take typescript 7
- Keep the existing test suite green (`hardhat compile + test` CI leg) — the wBTH contract tests and the live-defi-roundtrip scripts must run unchanged
- Check ethers/viem compatibility choices the toolbox-7 line imposes before starting

The three Dependabot PRs are closed in favor of this issue; ignore rules in .github/dependabot.yml suppress these majors so they don't reopen (minors/patches still flow).

rjwalters:
## Summary

Migrates `contracts/ethereum` off hardhat 2.29 / hardhat-toolbox 4 / typescript 5.9 onto the hardhat 3 toolchain in one deliberate move, per #1174. This unblocks the three interlocked Dependabot majors (#1131 hardhat 3, #1127 toolbox 7, #1125 typescript 7) that could not merge individually, and removes the dependabot.yml ignore block that was suppressing them.

**ethers/viem choice**: `@nomicfoundation/hardhat-toolbox@7` turned out to be a tombstone package that supports *neither* HH2 nor HH3 (its README says so explicitly). The real HH3 successors are `hardhat-toolbox-mocha-ethers` (keeps mocha+chai+ethers) and `hardhat-toolbox-viem` (viem + node:test — a full test rewrite). Took the least-invasive path: **`@nomicfoundation/hardhat-toolbox-mocha-ethers@^3`**, so the existing 39-test wBTH suite keeps its structure, matchers and ethers API.

## Changes

- **package.json**: `"type": "module"` (ESM-first), hardhat `^3.12.0`, typescript `^7.0.2`, toolbox-mocha-ethers `^3.0.7` + explicit ethers/chai/mocha/@types peers; `package-lock.json` regenerated (still committed per #852).
- **hardhat.config.ts** (HH3 format): explicit `plugins` array; per-network `type` (`edr-simulated` / `http`); `configVariable()` fallbacks for unset RPC URLs (HH3 rejects empty-string URLs at config load); etherscan key moved under `verify.etherscan`; **fixed `gas: 16_000_000` on the simulated network** — HH3's minimal-gas estimation starves the inner `bridgeMint` inside `SafeStub.execTransaction` (the outer tx "succeeds" and emits `ExecutionFailure`, silently skipping the mint), which HH2 never hit because it sent gas = blockGasLimit; 16M stays under EDR's EIP-7825 per-tx cap of 16,777,216.
- **Tests**: `network.getOrCreate("hardhat")` connection API (the no-arg default is an implicit `default` network that ignores `networks.hardhat` config; `getOrCreate` is the non-deprecated replacement for `connect()` and shares one in-process node across test files like HH2); `loadFixture`/`time` from `connection.networkHelpers`; deprecated `.to.be.reverted` → `.to.be.revert(ethers)` (chai-matchers v3). Test bodies otherwise untouched — including the ABI/selector/EIP-712 pins that guard the Rust bindings.
- **Scripts**: `deploy.ts`/`deploy-safe.ts`/`deploy-all.ts` use the connection API (`--network` flag drives the no-arg default — verified empirically); CJS `require.main === module` guard in deploy-safe.ts replaced with an argv/`import.meta.url` check; `__dirname` → `import.meta.dirname` across hl-*/live scripts (undefined under ESM); stale `ts-node --compiler-options '{"module":"commonjs"}'` run instructions → `npx tsx`; `./hl-lib` imports get explicit `.js` extensions; bigint annotations in live-defi-roundtrip.ts so `tsc --noEmit` is clean under typescript 7 (latent `any`-arithmetic holes surfaced by the bump).
- **.gitignore** (contracts/ethereum): HH3 typechain generates `types/` (was `typechain-types/`).
- **.github/dependabot.yml**: removed the hardhat/toolbox/typescript major-ignore block that was waiting on this migration (minors/patches were already flowing; majors for the retired `@nomicfoundation/hardhat-toolbox` can no longer reopen since the dep is gone from the manifest).

- [x] hardhat.config.ts ported to v3 format (ESM, `plugins` array) — compiles and loads
- [x] Toolchain on hardhat 3 line / toolbox HH3 successor / typescript 7 (`npx tsc --version` → 7.0.2; `npx hardhat --version` → 3.12.0)
- [x] `hardhat compile + test` CI leg green: `npm ci && npx hardhat compile && npx hardhat test` → **39 passing, 0 failing** from a clean checkout state (exact CI commands from `.github/workflows/contracts-ci.yml`, Node 22-compatible — verified locally on Node 26)
- [x] wBTH contract tests run unchanged in substance (ABI pins, EIP-712 cross-language vector, replay/rate-limit fuzz all intact)
- [x] live-defi-roundtrip + deploy scripts still run: full `deploy.ts` executed against the simulated network (WrappedBTH deployed); live scripts load under `hardhat run` and fail only at their expected env/secret checks; `hl-lib-selftest.ts` passes offline via `npx tsx` (all golden vectors)
- [x] Dependabot ignore block for this ecosystem removed
- [x] No changes to contracts/solana or anything outside contracts/ethereum + dependabot.yml (CI workflow file needed no changes)

## Test Plan

- `cd contracts/ethereum && npm ci` — clean install from the regenerated lockfile (mirrors CI)
- `npx hardhat compile` — 3 Solidity files compile (solc 0.8.20)
- `npx hardhat test` — 39 passing / 0 failing, no deprecation warnings
- `npx tsc --noEmit` — clean under typescript 7
- `npx tsx scripts/hl-lib-selftest.ts` — all signing self-tests pass (offline)
- `WBTH_*_SAFE=<addrs> npx hardhat run scripts/deploy.ts --network hardhat` — real deploy succeeds
- TDD: N/A — toolchain migration with no intended behavior change; the pre-existing 39-test suite is the regression net (red under HH3 before the config/test/gas fixes, green after; the Safe gas-starvation failure mode was diagnosed test-first via an isolated repro before the config fix)

Closes #1174

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -79,16 +79,6 @@ updates:
     schedule:
       interval: "weekly"
     labels: ["dependencies"]
-    ignore:
-      # hardhat-3 toolchain moves as one migration (#1174): hardhat 3 (ESM,
-      # new config) + toolbox 7 (requires hardhat 3) + typescript 7 (breaks
-      # the hardhat-2 ts integration). Individual majors are guaranteed-red.
-      - dependency-name: "hardhat"
-        update-types: ["version-update:semver-major"]
-      - dependency-name: "@nomicfoundation/hardhat-toolbox"
-        update-types: ["version-update:semver-major"]
-      - dependency-name: "typescript"
-        update-types: ["version-update:semver-major"]
     groups:
       eth-minor-patch:
         patterns: ["*"]
--- a/contracts/ethereum/.gitignore
+++ b/contracts/ethereum/.gitignore
@@ -2,6 +2,8 @@ node_modules/
 artifacts/
 cache/
 typechain-types/
+# hardhat-typechain (hardhat 3) generates into types/
+types/
 coverage/
 coverage.json
 .env
--- a/contracts/ethereum/hardhat.config.ts
+++ b/contracts/ethereum/hardhat.config.ts
@@ -3,10 +3,13 @@
 // addresses resolve without a manual `source .env` (#1011). Secrets are only
 // read from this git-ignored file — never printed, never committed.
 import "dotenv/config";
-import { HardhatUserConfig } from "hardhat/config";
-import "@nomicfoundation/hardhat-toolbox";
+import { configVariable, type HardhatUserConfig } from "hardhat/config";
+import hardhatToolboxMochaEthers from "@nomicfoundation/hardhat-toolbox-mocha-ethers";
 
+// Hardhat 3 config (#1174): ESM-first, explicit `plugins` array, and every
+// network declares its `type` ("edr-simulated" in-process vs "http" remote).
 const config: HardhatUserConfig = {
+  plugins: [hardhatToolboxMochaEthers],
   solidity: {
     version: "0.8.20",
     settings: {
@@ -18,33 +21,51 @@ const config: HardhatUserConfig = {
   },
   networks: {
     hardhat: {
+      type: "edr-simulated",
       chainId: 31337,
+      // Hardhat 2 sent gas = blockGasLimit on the in-process network; Hardhat
+      // 3 estimates instead. For SafeStub.execTransaction the outer tx
+      // "succeeds" (ExecutionFailure event) even when the inner call runs out
+      // of gas, so minimal-gas estimation silently starves the inner
+      // bridgeMint. Pin a fixed gas allowance like HH2 (kept under EDR's
+      // 16,777,216 per-transaction gas cap, EIP-7825).
+      gas: 16_000_000,
     },
     localhost: {
+      type: "http",
       url: "http://127.0.0.1:8545",
     },
     // Local Sepolia FORK for dry-running deploys with no real testnet ETH
     // (#1011/#992). Point BRIDGE_FORK_RPC_URL at an `anvil --fork-url <sepolia>`
     // node; the deployer is any key funded on the fork via anvil_setBalance.
     fork: {
+      type: "http",
       url: process.env.BRIDGE_FORK_RPC_URL || "http://127.0.0.1:8545",
       accounts: process.env.PRIVATE_KEY ? [process.env.PRIVATE_KEY] : [],
     },
+    // Hardhat 3 rejects empty-string URLs at config-load time, so unset RPC
+    // URLs fall back to configVariable(): resolution is lazy and only fails
+    // when the network is actually connected to.
     sepolia: {
-      url: process.env.SEPOLIA_RPC_URL || "",
+      type: "http",
+      url: process.env.SEPOLIA_RPC_URL || configVariable("SEPOLIA_RPC_URL"),
       accounts: process.env.PRIVATE_KEY ? [process.env.PRIVATE_KEY] : [],
     },
     mainnet: {
-      url: process.env.MAINNET_RPC_URL || "",
+      type: "http",
+      url: process.env.MAINNET_RPC_URL || configVariable("MAINNET_RPC_URL"),
       accounts: process.env.PRIVATE_KEY ? [process.env.PRIVATE_KEY] : [],
     },
   },
   // Etherscan source verification (#1013). Key is read from the git-ignored
-  // .env (ETHERSCAN_API_KEY) via dotenv above — never committed.
-  etherscan: {
-    // Etherscan API V2 — a single multichain key (chainid selects the
-    // explorer). V1 per-network keys are deprecated.
-    apiKey: process.env.ETHERSCAN_API_KEY || "",
+  // .env (ETHERSCAN_API_KEY) via dotenv above — never committed. hardhat-verify
+  // v3 nests the etherscan block under `verify`.
+  verify: {
+    etherscan: {
+      // Etherscan API V2 — a single multichain key (chainid selects the
+      // explorer). V1 per-network keys are deprecated.
+      apiKey: process.env.ETHERSCAN_API_KEY || "",
+    },
   },
 };
 
--- a/contracts/ethereum/package-lock.json
+++ b/contracts/ethereum/package-lock.json
@@ -8,11 +8,17 @@
       "name": "wbth-ethereum",
       "version": "0.6.0",
       "devDependencies": {
-        "@nomicfoundation/hardhat-toolbox": "^4.0.0",
+        "@nomicfoundation/hardhat-toolbox-mocha-ethers": "^3.0.7",
         "@openzeppelin/contracts": "^5.0.0",
+        "@types/chai": "^5.2.0",
+        "@types/mocha": "^10.0.10",
+        "@types/node": "^22.8.5",
+        "chai": "^5.1.2",
         "dotenv": "^17.4.2",
-        "hardhat": "^2.29.0",
-        "typescript": "^5.3.0"
+        "ethers": "^6.14.0",
+        "hardhat": "^3.12.0",
+        "mocha": "^11.0.0",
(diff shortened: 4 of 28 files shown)


rjwalters:
## Judge review: APPROVED

Independently verified every claim from a clean state in `.loom/worktrees/issue-1174` (`rm -rf node_modules artifacts cache types` first, then the exact commands from `.github/workflows/contracts-ci.yml`).

### Verification results

| Check | Result |
|---|---|
| `npm ci` | clean from the committed lockfile |
| `npx hardhat compile` | 3 Solidity files, solc 0.8.20 (evm target: shanghai) |
| `npx hardhat test` | **39 passing, 0 failing** |
| `npx tsc --noEmit` | exit 0 |
| `npx tsc --version` / `npx hardhat --version` | 7.0.2 / 3.12.0 |
| `npx tsx scripts/hl-lib-selftest.ts` | all golden vectors pass offline |
| `hardhat run scripts/deploy.ts --network hardhat` | WrappedBTH deployed, no roles to deployer |
| CI | `hardhat compile + test` pass, `cargo deny` pass |
| Merge state | CLEAN, 0 commits behind main |

### The gas pin is load-bearing, and the diagnosis is correct

I did not take this on faith. Removing `gas: 16_000_000` from the simulated network and re-running produced **exactly 5 failures, all and only the SafeStub-mediated tests**, with the failure mode the PR describes:

```
1) Bridge flow through the Safe (happy paths, #828)
     BTH -> wBTH: threshold mint through the Safe
       mints with 2-of-3 owner signatures submitted by a plain relayer:
   AssertionError: Expected event "BridgeMint" to be emitted, but it wasn't
```

The mechanism checks out against the contract: `SafeStub.execTransaction` does `(success, ) = to.call{value: value}(data)` with no explicit gas forwarding and then `emit ExecutionFailure(txHash)` on failure rather than bubbling the revert. Because the *outer* tx succeeds either way, `eth_estimateGas`'s binary search converges on the smallest gas at which the outer tx succeeds — which under EIP-150's 63/64 rule is precisely the value where the inner `bridgeMint` OOGs. That is the textbook estimateGas pitfall for catch-and-swallow contracts. HH2 never hit it because `networks.hardhat.gas` defaulted to the block gas limit; HH3 estimates. Pinning `gas` restores HH2 semantics with a one-line, commented change rather than touching 5 test bodies — correct call for a migration whose stated contract is no behavior change.

Both numeric claims in the comment are verifiable in hardhat's own source:
- `network-manager/edr/utils/convert-to-edr.js:284` → `return 16777216n; // EIP-7825 transaction gas cap`. 16,000,000 < 16,777,216. ✅
- `edr-constants.js:4` → `DEFAULT_EDR_NETWORK_BLOCK_GAS_LIMIT = 60000000n`, so a 16M tx sits comfortably inside a block. ✅

Critically, the pin does **not** mask anything: the tests it rescues are the *positive-path* ones that assert a mint actually happened (`BridgeMint` emitted, balances/supply updated). The negative test at `BridgeFlow.test.ts:262` still asserts `ExecutionFailure` **and** `.not.to.emit(wbth, "BridgeMint")` **and** `balanceOf == PICO` — its failure is caused by the replay guard, not by gas. No test in either file makes a gas assertion, so the pin has no other reach.

### No test was weakened

- 39 `it()` blocks on main, 39 on the branch — nothing deleted. No `.skip`/`.only`/`xit` anywhere.
- Test diff is +21/-9, entirely import plumbing plus **two** assertion lines.
- The only assertion change, `.to.be.reverted` → `.to.be.revert(ethers)`, is semantically equivalent and strictly *safer*: a stale `.reverted` **property** on an unregistered matcher passes vacuously, whereas `.revert(ethers)` is a call that throws `TypeError` if unregistered. Both sites are correctly `await expect(...)`ed. The 26 other `revertedWith`/`revertedWithCustomError` assertions and the ABI/selector/EIP-712 cross-language pins are untouched.
- **Zero `.sol` files changed** (`git diff --name-only | grep -c '\.sol$'` → 0), so the Rust binding guardrails in `bridge/service/src/mint/ethereum.rs` are pinned against the same bytecode.

### Other spot checks

- **dependabot.yml**: removal is scoped exactly to the `/contracts/ethereum` block's `ignore:` list. The `groups.eth-minor-patch` block, the `/web` block, the `/contracts/solana` block, and every cargo block are byte-identical.
- **`invokedDirectly` guard** (the trickiest ESM port — `require.main === module` has no ESM equivalent): verified empirically **both ways**. Run directly, the stack shows `at <anonymous> (deploy-safe.ts:207)` — guard fired. Run via `deploy-all.ts`, the stack shows `at main (deploy-all.ts:27)` with no `:207` frame — imported without double-executing. Correct.
- **`verify.etherscan` nesting** confirmed against `@nomicfoundation/hardhat-verify`'s own `type-extensions.d.ts` (`HardhatUserConfig.verify?: VerificationProvidersUserConfig` → `etherscan?`). Worth knowing: HH3 does *not* reject unknown top-level config keys (I injected a bogus one and `compile` accepted it), so the real guard here is the `HardhatUserConfig` excess-property check under a green `tsc --noEmit` — which holds.
- **`import.meta.dirname`** requires Node ≥ 20.11; CI pins Node 22 and HH3 itself requires ≥22.10. Fine. All 12 replacements are 1:1 mechanical.
- **`ART` path** `../artifacts/contracts/WbthPeerToken.sol/WbthPeerToken.json` still resolves under HH3's artifact layout — verified post-compile.
- **`npx tsx`** resolves from the local tree (transitive dep of hardhat 3), so the updated run instructions are not aspirational.
- **Lockfile**: every `resolved` URL is `registry.npmjs.org`; no non-registry sources. `npm audit` reports 1 high (`serialize-javascript@6.0.2`, a mocha transitive) — **pre-existing and byte-identical on main**, dev-only, not a regression from this PR.
- **Worktree clean** after all of the above: HH3's `types/` output is correctly gitignored, nothing generated leaked into the diff.

### Non-blocking notes (no action required)

1. The network-level `gas` pin means the suite no longer exercises gas estimation at all. That is precisely the HH2 property being restored, so it is not a regression — but if a future test ever wants to assert genuine out-of-gas behavior it will need a per-call `gasLimit` override.
2. `hardhat coverage` sets `transactionGasCap: false` to let instrumented bytecode exceed the cap, but an explicit `networks.hardhat.gas` is preserved as a user value and would still cap each tx at 16M. There is no coverage step in `contracts-ci.yml` and no `coverage` script in `package.json`, so this is latent, not live.
3. `configVariable("SEPOLIA_RPC_URL")` falls back to the same env var name the `||` just found empty, so it is effectively a lazy "not configured" error. That is the intended HH3 idiom and produces a better message than the old `url: ""`.

Clean, well-scoped toolchain migration with an unusually honest and *correct* root-cause writeup on the one non-obvious behavior change. The hard part — proving the gas pin is a genuine HH2-parity fix rather than a green-washing hack — holds up under adversarial testing.

Issue #1187, opened 2026-08-11
bug: test_compact_block_estimated_size_accuracy fails on main (estimate 568 vs actual 1609 for 10 txs)


rjwalters:
`botho/tests/compact_block_integration.rs::test_compact_block_estimated_size_accuracy` fails on unmodified `main` (verified at 5cc9a715, 2026-08-11):

```
Estimated size (568) should be roughly accurate vs actual (1609) for 10 txs
```

`CompactBlock::estimated_size()` (botho/src/network/compact_block.rs, ~line 258) uses hard-coded per-component estimates (header ~200 B, MintingTx ~300 B, 6 B per short id). The actual bincode-serialized size of a 10-tx compact block is now 1609 bytes, so the estimate ratio (0.35) falls outside the test's accepted 0.5–2.0 band. The estimate has drifted as `BlockHeader`/`MintingTx` grew (RandomX PoW fields, ML-DSA minting signatures, lottery summary, etc.).

This is NOT caught by per-PR CI because `workspace-build.yml` only runs `cargo test --workspace --no-run` (compile-only) plus doc tests; the integration test only executes in fuller local/scheduled runs.

Fix options:
- Recompute the constants from current serialized sizes (cheap, will drift again), or
- Derive the estimate from actual component serialization (e.g. serialize header+minting_tx once), or
- Loosen/delete the accuracy band if `estimated_size()` is only used for logging/heuristics.

Found while running the full workspace test suite for the dalek-5 migration (#1153); the failure reproduces identically with and without that migration.

_(See Curator Enhancement comment below for verified root-cause math, acceptance criteria, affected files, and test plan.)_

rjwalters:
## Curator Enhancement

Verified against `origin/main` @ `bcfc46c0`, 2026-08-17 — root cause and byte math re-derived from current struct definitions (no changes to `compact_block.rs`/`block.rs` since the 5cc9a715 repro cited in the issue body, other than an unrelated `rand`-migration import touch in `b95804b3`).

### Confirmed root cause

`CompactBlock::estimated_size()` (`botho/src/network/compact_block.rs:258-269`) budgets `MintingTx` at a flat **~300 bytes**:

```rust
let base_size = 200 + 300 + 8; // header + MintingTx + nonce
```

[...]

Issue #1154, opened 2026-08-11
Rand-family migration: rand_core 0.6 → 0.10 line (rand / rand_chacha / rand_hc / rand_distr / getrandom in lockstep)


rjwalters:
**Rescoped after investigation** (was: "Migrate rand_hc 0.3 → 0.5"): rand_hc 0.5 requires **`rand_core ^0.10`**, while the workspace pins the rand-0.8 family across the board (Cargo.toml ~lines 102, 138-142: rand 0.8, rand_chacha 0.3, rand_core 0.6, rand_distr 0.4, getrandom 0.2). Every RNG consumer in the workspace moves together or not at all — this is the workspace rand-family migration:

- `rand_core` 0.6 → 0.10 line (SeedableRng API changes; the original Dependabot breakage was `Hc128Rng::from_seed` not found)
- `rand` 0.8 → matching release, `rand_chacha`, `rand_distr`, `getrandom` in lockstep
- `rand_hc` 0.3 → 0.5

**Risk**: `Hc128Rng` seeds deterministic key derivation (`account-keys`, `crypto/keys`, `crypto/ring-signature/signer`, `crypto/multisig`). Derivation must be BIT-IDENTICAL after the migration — existing golden/derivation tests must pass unchanged; do NOT regenerate fixtures. Audit any `gen_range`/distribution call sites for the rand-0.9 semantics changes.

Dependabot PR #1151 cannot merge standalone and is closed in favor of this issue; a Dependabot `ignore` rule for the family now prevents re-opening (see .github/dependabot.yml).

Sibling (independent): #1153 dalek-5 family. #1155 primitive-types is NOT gated and is being migrated directly.

PR #1184, opened 2026-08-11
feat(mobile): upgrade Expo SDK 50 -> 57 with lockstep package set + mobile CI

rjwalters:
## Summary

Deliberate Expo SDK 50 → 57 upgrade for `mobile/app`, done the supported way (`expo@~57.0.12` then `npx expo install --fix`), so every expo-* / react-native-* package lands on the SDK-57-matched version set. Supersedes the 5 unmergeable single-package Dependabot bumps (#1139, #1140, #1157, #1159, #1161). Also adds the mobile CI workflow the issue called for (mobile/ previously had **no CI**) and removes the now-obsolete Dependabot ignore entries that were waiting on this migration.

`mobile/rust-bridge` (the `botho-mobile` crate) is untouched — no expo-modules-core codegen config change was required.

## Changes

**SDK upgrade (`mobile/app/package.json` + `pnpm-lock.yaml`)**
- `expo` 50 → 57.0.12 with the SDK-57-paired set: `react` 19.2.3, `react-native` 0.86.2, `expo-router` 57, `react-native-screens` 4.26, `react-native-safe-area-context` 5.7, `babel-preset-expo` 57, `jest-expo` 57, `eslint-config-expo` 57, `@types/react` 19.2 — all chosen by `expo install --fix`, not hand-picked
- `expo-modules-core` removed as a direct dependency (expo-doctor flags direct installs; `requireOptionalNativeModule` is re-exported from `expo` — `src/native/walletModule.ts` updated accordingly)
- `expo-constants` + `expo-linking` added: required `expo-router` peers in SDK 57
- `typescript` moved to `~6.0.3`, the SDK 57 expected line (`expo install --fix` selects it; expo-doctor flags 7.x as a version mismatch). The TS-7-era tsconfig cleanups are kept: dropped deprecated `baseUrl` (hard error TS5101 under TS6), made the `@/*` path mapping relative, added explicit `types: ["jest"]` (jest globals are no longer auto-included)
- `app.json`: removed the SDK-57-removed top-level `splash` block and the `icon`/`adaptiveIcon` entries referencing `./assets/*` files that have never existed in-repo (all are now hard schema failures in expo-doctor)
- `app/unlock.tsx`: moved the mount effect below `handleBiometricAuth` — eslint-config-expo 57's react-hooks plugin makes access-before-declaration a lint **error** (`react-hooks/immutability`)
- jest `transformIgnorePatterns`: allow `standard-navigation` (untranspiled ESM dependency of expo-router 57)
[...]

PR #1188, opened 2026-08-11
chore(deps): migrate dalek family to curve25519-dalek 5 / ed25519-dalek 3 / ed25519 3 / signature 3 / x25519-dalek 3

### LINKED ISSUE #1153: Dalek-5 family migration: curve25519-dalek 5 / ed25519-dalek 3 / ed25519 3 / signature 3
**Rescoped after investigation** (was: "Migrate ed25519 crate 2.2 → 3.0"): ed25519 3 requires `signature ^3`, and the only ed25519-dalek line compatible with that (3.0.0) requires **`curve25519-dalek ^5`**. The workspace pins curve25519-dalek 4 and uses it in `crypto/digestible`, `crypto/ring-signature` (CLSAG), `crypto/keys`, `transaction/core`, and `account-keys` — so this is one lockstep migration, not a crate bump:

- `curve25519-dalek` 4 → 5 (check the `[profile.release.package.curve25519-dalek]` constant-time limb exemption at workspace Cargo.toml ~line 240 still applies)
- `ed25519-dalek` 2 → 3
- `ed25519` 2 → 3 (original Dependabot breakage: `signature::Error` lost its `From<Box<dyn Error + Send + Sync>>` conversions used in `crypto/keys/src/ed25519.rs` / `lib.rs`)
- `signature` 2 → 3
- `x25519-dalek` 2 → whatever pairs with curve25519-5

**Risk**: CLSAG ring signatures and key derivation are consensus-critical. Acceptance requires byte-identical signatures/derivations against existing golden vectors (no fixture regeneration), plus a review of constant-time guarantees across the curve25519-5 changelog.

Dependabot PR #1152 cannot merge standalone (two ed25519/signature majors in-tree with incompatible trait impls) and is closed in favor of this issue; a Dependabot `ignore` rule for the family now prevents re-opening (see .github/dependabot.yml).

Sibling (independent): #1154 rand-family. #1155 primitive-types is NOT gated and is being migrated directly.

rjwalters:
## Summary

One lockstep migration of the dalek-5 family — curve25519-dalek 4→5, ed25519-dalek 2→3, ed25519 2→3, signature 2→3, x25519-dalek 2→3 — with a hard bit-identical-behavior bar: **no golden vectors, fixtures, or snapshot values were regenerated, updated, or deleted**, and every affected test suite passes unchanged.

The family cannot move one crate at a time (ed25519-dalek 3 requires curve25519-dalek ^5 and signature ^3), which is why Dependabot PR #1152 was closed in favor of this issue.

## Changes

**Workspace manifest (`Cargo.toml`)**
- Bump the five family crates; add `zeroize` to curve25519-dalek features (previously enabled transitively by dalek-4-era dependents).
- Add two narrow bridge deps: `curve25519-dalek-v4` (renamed v4 pin, bulletproofs boundary only) and `rand-core-v010` / `digest-v011` (trait-generation bridges, crypto/keys only).
- The `[profile.release.package.curve25519-dalek]` overflow-checks exemption is **unchanged and still correct for v5**: the default serial/simd backends keep the constant-time limb arithmetic in-crate; `fiat-crypto` only compiles under an explicit `curve25519_dalek_backend="fiat"` cfg override (verified against the v5 build.rs and CHANGELOG).

**Byte-level compat shims (the core of the migration).** dalek 5 moved its convenience constructors behind rand_core 0.10 / digest 0.11 traits; the workspace stays on rand_core 0.6 (rand family is #1154) and digest 0.10. Each call site now uses the exact upstream algorithm at the byte level (verified against the vendored dalek 4 and 5 sources — both fill 64/32 bytes via `fill_bytes` and reduce/map identically):
- `bth-crypto-ring-signature::compat` — `scalar_from_hash`, `point_from_hash`, `random_scalar`, `random_point`; used by CLSAG/MLSAG signing+challenge code, one-time keys, generators, transaction/core, transaction/types.
- Local equivalents in `core/subaddress.rs`, `account-keys/burn_address.rs`, `cluster-tax` (crates that do not depend on ring-signature).
- `crypto/keys`: byte-draw `SigningKey`/scalar/point generation; `RngCompat` (rand_core 0.6→0.10 pass-through) for x25519 secrets (`EphemeralSecret` has no byte-level constructor by design); `DigestCompat` (digest 0.10→0.11 forwarding wrapper) for Ed25519ph prehashed sign/verify; the `signature`-2-shaped `DigestSigner`/`DigestVerifier` traits are now defined locally in `traits.rs` (signature 3 changed them to closure-based APIs over digest 0.11) — same shape, so downstream callers are unchanged.
- `SchnorrkelPublic::from_point` → `from_bytes` over the canonical compressed encoding (schnorrkel-og is still on dalek 4; same point, one extra decompression per schnorrkel verify).

**bulletproofs-og boundary (`transaction/core/src/range_proofs`)**: no bulletproofs release supports dalek 5 (upstream max 5.0.0 pins dalek ^4.1.1), so scalars, compressed points, and Pedersen generators cross the dalek 4/5 boundary via canonical 32-byte encodings — bit-exact, infallible for values originating from valid points/scalars. The module's public API keeps dalek-5 types.

**Dependabot (`.github/dependabot.yml`)**: removed the now-obsolete family ignores for `ed25519`, `ed25519-dalek`, `signature`, `x25519-dalek`. The `curve25519-dalek` major ignore is deliberately **kept** with an updated comment: the workspace now carries an intentional v4 pin for the bulletproofs boundary, and a Dependabot major bump of it would be guaranteed-red until bulletproofs supports dalek 5. Rand-family ignores untouched (#1154).

**Misc**: ed25519-dalek 3 dropped its `std` feature — botho/bridge manifests now use default/`alloc`; `botho/src/operator_key.rs` key generation byte-draws from `thread_rng` (identical to the old `generate`); fuzz crate bumped to dalek 5 and its `fuzz_clsag` harness updated.

- [x] Lockstep bump of all five family crates in the workspace `Cargo.toml` — no crate left on a mixed major within the workspace's own dependency edges (remaining v4/v2 lock entries belong to third-party crates: bulletproofs-og/schnorrkel-og/snow/libp2p-identity, which never exchange dalek types with our code except at the two bridged boundaries above).
- [x] Constant-time limb exemption reviewed against the v5 backend selection — still applies as written (fiat backend is opt-in via cfg, not default).
- [x] Byte-identical signatures/derivations against existing golden vectors — **no fixture regeneration**; see Test Plan. The seeded-RNG doctests (e.g. the x25519 fingerprint goldens over `Hc128Rng::seed_from_u64(0)`) prove the RNG byte-stream consumption is unchanged.
- [x] Dependabot ignore updates (see above for the one deliberate retention).

## Test Plan

All run locally in the worktree; **no fixtures/goldens/snapshots were modified** (diff touches only `.rs`, `.toml`, `.yml`, `Cargo.lock`).

- `cargo build --workspace` — passes.
- `cargo check --workspace` — passes (no new warnings from this change).
- Hard-bar suites, all green:
  - `bth-crypto-keys`: 30 passed (+2 ignored openssl-nightly) + 12 doctests (incl. seeded x25519 fingerprint goldens)
  - `bth-crypto-ring-signature`: 51 passed (CLSAG/MLSAG sign+verify, one-time keys)
  - `bth-crypto-digestible`: 10 passed (tests/basic.rs)
  - `bth-account-keys`: 69 passed (with dev-deps run) / 9 lib
  - `bth-transaction-core`: 99 passed; `--all-features`: 120 passed (CI-parity job)
  - `bth-transaction-types`: 69 passed
- Consensus + neighbors: `bth-consensus-scp` 101, `bth-core` 24, `bth-core-types` 46, `bth-crypto-box` 30, `bth-crypto-multisig` 8, `bth-crypto-sig` 8, `bth-crypto-hashes` ok.
- `bth-cluster-tax`: 504 passed + 5 doctests.
- `cargo test --workspace --exclude bth-cluster-tax --no-fail-fast`: 136 suites ok; 4 failing targets (`compact_block_integration`, `e2e_consensus_integration`, `e2e_progressive_fees`, `e2e_transfer_patterns`) were **each re-run against unmodified main (5cc9a715) and fail identically there** — pre-existing, not a regression (matches the known stale fee-e2e history, #1080). The compact-block estimate drift is now tracked as #1187.
- `cargo deny check` — advisories/bans/licenses/sources ok (`multiple-versions = "warn"`).
- `cargo fmt --check` — clean (workspace + fuzz crate); fuzz crate resolves and `fuzz_clsag` type-checks on the pinned nightly.

TDD: N/A — dependency migration with a bit-identical-behavior requirement; correctness is gated on the pre-existing golden/seeded suites passing unchanged rather than new tests.

Closes #1153

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,16 +33,13 @@ updates:
       # bumps are guaranteed-red (two trait-incompatible majors in-tree) and
       # each family upgrade is a deliberate migration through
       # consensus-critical code, tracked as an issue:
-      # dalek-5 family (#1153): CLSAG / key code
-      - dependency-name: "ed25519"
-        update-types: ["version-update:semver-major"]
-      - dependency-name: "ed25519-dalek"
-        update-types: ["version-update:semver-major"]
+      # The dalek-5 family migrated in #1153, but the workspace keeps a
+      # deliberate curve25519-dalek 4 pin (`curve25519-dalek-v4`) for the
+      # bulletproofs-og interop boundary in transaction/core/src/range_proofs.
+      # A major bump of that pin is guaranteed-red until a bulletproofs
+      # release supports curve25519-dalek 5; remove this ignore when the
+      # v4 pin is retired.
       - dependency-name: "curve25519-dalek"
-        update-types: ["version-update:semver-major"]
-      - dependency-name: "signature"
-        update-types: ["version-update:semver-major"]
-      - dependency-name: "x25519-dalek"
         update-types: ["version-update:semver-major"]
       # rand family (#1154): deterministic key derivation. Pre-1.0, so 0.x
       # "minors" are the breaking axis — ignore those too; patches flow.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -292,7 +292,7 @@ dependencies = [
  "either",
  "serde",
  "serde_with",
- "sha2",
+ "sha2 0.10.9",
 ]
 
 [[package]]
@@ -1503,6 +1503,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
 
 [[package]]
+name = "base16ct"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd307490d624467aa6f74b0eabb77633d1f758a7b25f12bceb0b22e08d9726f6"
+
+[[package]]
 name = "base256emoji"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1770,7 +1776,7 @@ dependencies = [
  "dashmap",
  "digest 0.10.7",
  "dirs 5.0.1",
- "ed25519-dalek",
+ "ed25519-dalek 3.0.0",
  "futures",
  "getrandom 0.2.17",
  "heed",
@@ -1805,7 +1811,7 @@ dependencies = [
  "serde_json",
  "serial_test",
  "sha1",
- "sha2",
+ "sha2 0.10.9",
  "siphasher",
  "statrs",
  "tempfile",
@@ -1947,7 +1953,7 @@ dependencies = [
  "rpassword",
  "serde",
  "serde_json",
- "sha2",
+ "sha2 0.10.9",
  "tempfile",
  "tiny-bip39",
  "tokio",
@@ -2009,7 +2015,7 @@ dependencies = [
  "bth-util-serial",
  "bth-util-test-helper",
  "criterion",
- "curve25519-dalek",
+ "curve25519-dalek 5.0.0",
  "displaydoc",
  "hkdf",
  "prost 0.12.6",
@@ -2017,7 +2023,7 @@ dependencies = [
  "rand_core 0.6.4",
  "rand_hc",
  "serde",
- "sha2",
+ "sha2 0.10.9",
  "tiny-bip39",
  "zeroize",
 ]
@@ -2048,11 +2054,11 @@ version = "0.6.0"
 dependencies = [
  "bs58 0.4.0",
  "chrono",
- "ed25519-dalek",
+ "ed25519-dalek 3.0.0",
  "hex",
  "serde",
  "serde_json",
- "sha2",
+ "sha2 0.10.9",
  "tempfile",
  "toml 0.9.12+spec-1.1.0",
  "uuid",
@@ -2078,7 +2084,7 @@ dependencies = [
  "bth-transaction-types",
  "chrono",
  "clap",
- "ed25519-dalek",
+ "ed25519-dalek 3.0.0",
  "hex",
  "proptest",
  "rand 0.8.7",
@@ -2087,7 +2093,7 @@ dependencies = [
  "rusqlite",
  "serde",
  "serde_json",
- "sha2",
+ "sha2 0.10.9",
  "tempfile",
  "tokio",
  "tracing",
@@ -2103,7 +2109,7 @@ dependencies = [
  "bth-transaction-types",
  "clap",
  "criterion",
- "curve25519-dalek",
+ "curve25519-dalek 5.0.0",
  "indicatif",
  "rand 0.8.7",
  "rand_chacha 0.3.1",
@@ -2111,7 +2117,7 @@ dependencies = [
  "rand_distr",
  "rayon",
  "serde",
- "sha2",
+ "sha2 0.10.9",
 ]
 
 [[package]]
@@ -2205,13 +2211,13 @@ dependencies = [
  "bth-crypto-keys",
  "bth-util-from-random",
  "clap",
- "curve25519-dalek",
+ "curve25519-dalek 5.0.0",
  "hex",
  "hkdf",
  "rand_core 0.6.4",
  "serde",
  "serde_json",
- "sha2",
+ "sha2 0.10.9",
  "slip10_ed25519",
  "tiny-bip39",
  "zeroize",
@@ -2224,7 +2230,7 @@ dependencies = [
  "bth-crypto-keys",
  "bth-util-from-random",
  "bth-util-test-helper",
- "curve25519-dalek",
+ "curve25519-dalek 5.0.0",
  "prost 0.12.6",
  "rand_core 0.6.4",
  "rand_hc",
@@ -2255,11 +2261,11 @@ version = "7.1.0"
 dependencies = [
  "bth-crypto-digestible-derive",
  "cfg-if",
- "curve25519-dalek",
- "ed25519-dalek",
+ "curve25519-dalek 5.0.0",
+ "ed25519-dalek 3.0.0",
  "generic-array",
  "merlin",
- "x25519-dalek",
+ "x25519-dalek 3.0.0",
 ]
 
 [[package]]
@@ -2276,7 +2282,7 @@ name = "bth-crypto-digestible-signature"
 version = "7.1.0"
 dependencies = [
  "bth-crypto-digestible",
- "signature 2.2.0",
+ "signature 3.0.0",
 ]
 
 [[package]]
@@ -2300,26 +2306,28 @@ dependencies = [
  "bth-util-repr-bytes",
  "bth-util-serial",
  "bth-util-test-helper",
- "curve25519-dalek",
+ "curve25519-dalek 5.0.0",
  "digest 0.10.7",
+ "digest 0.11.3",
  "displaydoc",
- "ed25519",
- "ed25519-dalek",
+ "ed25519 3.0.0",
+ "ed25519-dalek 3.0.0",
  "hex",
  "hex_fmt",
  "pem",
+ "rand_core 0.10.1",
  "rand_core 0.6.4",
  "rand_hc",
  "schnorrkel-og",
  "semver 1.0.28",
  "serde",
  "serde_json",
- "sha2",
- "signature 2.2.0",
+ "sha2 0.10.9",
+ "signature 3.0.0",
  "static_assertions",
  "subtle",
  "tempfile",
- "x25519-dalek",
+ "x25519-dalek 3.0.0",
  "zeroize",
 ]
 
@@ -2353,7 +2361,7 @@ dependencies = [
  "rand_chacha 0.3.1",
  "rand_core 0.6.4",
  "serde",
- "sha2",
+ "sha2 0.10.9",
  "thiserror 2.0.20",
  "zeroize",
 ]
@@ -2373,7 +2381,8 @@ dependencies = [
  "bth-util-serial",
  "bth-util-test-helper",
  "criterion",
- "curve25519-dalek",
+ "curve25519-dalek 5.0.0",
+ "digest 0.10.7",
  "displaydoc",
  "proptest",
  "prost 0.12.6",
@@ -2410,7 +2419,7 @@ dependencies = [
  "hmac",
  "k256",
  "rand 0.8.7",
- "sha2",
+ "sha2 0.10.9",
  "sha3 0.10.9",
  "thiserror 2.0.20",
  "tiny-bip39",
@@ -2502,7 +2511,7 @@ dependencies = [
  "rand_core 0.6.4",
  "serde",
  "serde_json",
- "sha2",
+ "sha2 0.10.9",
 ]
 
 [[package]]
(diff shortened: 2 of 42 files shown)


rjwalters:
## Judge review — APPROVED (consensus-critical, deep pass)

Reviewed at head `96f186c7`, in the `issue-1153` worktree. I treated this as a
consensus-critical migration and independently re-derived every bit-exactness
claim from the vendored dalek sources rather than taking the PR body's word for
it. **Everything checks out.**

### 1. The dalek-4 ↔ dalek-5 bridge (the central risk) — sound

`transaction/core/src/range_proofs/mod.rs` is the only place the two majors meet,
and every crossing is a **canonical byte-level** crossing, never a transmute or a
reinterpretation of internal representation:

| Crossing | Mechanism | Verdict |
|---|---|---|
| commitments in (`check_range_proofs`) | `CompressedRistrettoV4(c.to_bytes())` | raw 32-byte passthrough. `CompressedRistretto(pub [u8;32])` in **both** 4.1.3 and 5.0.0 (verified in `src/ristretto.rs` line 219 / 223). No canonicalization on either side, so an attacker-supplied non-canonical encoding still fails inside bulletproofs exactly as before — acceptance behavior is *identical*. |
| commitments out (`generate_range_proofs`) | `CompressedRistretto(c.to_bytes())` | same, reverse direction. |
| generators (`convert_gens` → `point_to_v4`) | `compress()` → v4 `decompress()` | Ristretto255 compression is canonical-by-spec and both majors implement the same RFC map; round-trip is exact. Inputs are `generators(token_id)` / `B_BLINDING`, always valid ⇒ the `expect` is unreachable. |
| blindings (`scalar_to_v4`) | `to_bytes()` → v4 `from_canonical_bytes()` | exact, and the `expect` is unreachable — see below. |

**Non-canonical / unreduced scalar analysis (the thing that could have made
`scalar_to_v4`'s `expect` a panic vector):** dalek 5's `Scalar` doc invariant #2
states *"It is not possible to construct an unreduced `Scalar` from the public API
unless `legacy_compatibility` is enabled."* I grepped the workspace: nothing calls
`Scalar::from_bits` (the only mention is a comment in `account-keys/src/identity.rs`
explaining they deliberately use `from_canonical_bytes` instead). Every wire-facing
scalar decode goes through `CurveScalar::from_bytes_mod_order` (always reduced) or
dalek's serde impl (which rejects non-canonical). Additionally I traced call sites:
`generate_range_proofs` (the only user of `scalar_to_v4`) is reachable **only from
the signing path** (`rct_bulletproofs::sign`) plus tests/benches — verification uses
`check_range_proofs`, which touches no scalar conversion at all. So neither `expect`
is attacker-reachable.

The v4 pin's blast radius is correctly contained: `bth-transaction-core` is the
**only** workspace crate on `curve25519-dalek 4.1.3` (confirmed by walking
`Cargo.lock` reverse-deps; the other 4.1.3 consumers are `bulletproofs-og`,
`schnorrkel-og`, `snow`, `x25519-dalek 2`, `ed25519-dalek 2` — all third-party).

The `crypto/keys` schnorrkel crossing (`SchnorrkelPublic::from_point` →
`from_bytes(&self.to_bytes())`) is likewise exact: schnorrkel's `RistrettoBoth`
holds `{compressed, point}` and both constructors populate the same pair for a
valid encoding. `sign_schnorrkel` already went through bytes and is untouched.

### 2. Replaced convenience calls — byte-level equivalence verified against sources

I diffed the actual crates in `~/.cargo/registry`, not the docs:

- **`Scalar::random`** — 4.1.3 `scalar.rs:597` and 5.0.0 `scalar.rs:589` are
  character-identical bodies: `[0u8;64]` → `fill_bytes` → `from_bytes_mod_order_wide`.
  Matches `compat::random_scalar` exactly (`as_rngcore().fill_bytes` on a rand_core-0.6
  CSPRNG is the same byte draw).
- **`Scalar::from_hash`** — 4.1.3 `:671` vs 5.0.0 `:663`, identical bodies. Matches
  `compat::scalar_from_hash`.
- **`RistrettoPoint::random` / `from_hash` / `from_uniform_bytes`** — `from_uniform_bytes`
  bodies identical in both majors (same elligator one-way map, same RFC citation);
  dalek 5's `random` just delegates to `try_random`, which is `[0u8;64]` →
  `try_fill_bytes` → `from_uniform_bytes`. Matches `compat::random_point`.
- **`SigningKey::generate`** — ed25519-dalek 2.2.0 `signing.rs:202` and 3.0.0 `:213`
  are *identical*: `SecretKey::default(); csprng.fill_bytes(&mut secret); Self::from_bytes(&secret)`,
  with `pub type SecretKey = [u8; 32]` in both. The PR's inlined 32-byte draw is exact.
  `ExpandedSecretKey::from(&SecretKey)` (the actual key-derivation step) diffs **clean**
  between 2.2.0 and 3.0.0.
- **`x25519 EphemeralSecret/StaticSecret::random_from_rng`** — 2.0.1 and 3.0.0 both do
  `[0u8;32]` → `fill_bytes`. The `RngCompat` bridge implements rand_core 0.10's
  `TryRng`/`TryCryptoRng` with `Error = Infallible`, and rand_core 0.10's blanket
  `impl<R: TryRng<Error = Infallible>> Rng for R` routes `fill_bytes` → `try_fill_bytes`
  → the workspace RNG's `fill_bytes`. Same single 32-byte draw, same stream.
  **This one is also empirically pinned**: `crypto/keys/src/x25519.rs` carries seeded
  `Hc128Rng::seed_from_u64(0)` doctests with hardcoded public-key fingerprints
  (`0b:5e:58:...`), unchanged by this PR and passing — a real cross-version golden.

**Bonus check the PR didn't claim (and the one I'd have blocked on):** I diffed
ed25519-dalek's *verification* internals 2.2.0 → 3.0.0 to make sure signature
acceptance didn't silently tighten or loosen. `raw_sign_prehashed` diffs clean;
`RCompute::new/finish` (the dom2 prefix `"SigEd25519 no Ed25519 collisions"`, flag
`[1]`, context-length byte, R‖A absorption, `from_hash` challenge) is
character-identical; `check_scalar` diffs only by a removed `#[allow(deprecated)]`.
The only `raw_verify`/`raw_verify_prehashed` change is `&[u8]` → `&[&[u8]]` for the
new `MultipartVerifier` shape, and `RCompute::compute` just iterates the slices into
the same `update` calls. **Ed25519 / Ed25519ph acceptance semantics are unchanged** —
no signature that verified before is rejected now, and vice versa.

The `DigestCompat` digest-0.10 → 0.11 shim only forwards `update` and copies the
finalized 64 bytes, so the same digest output reaches `sign_prehashed`/`verify_prehashed`.

### 3. No fixtures/goldens/tests weakened — verified mechanically

`git diff --stat` vs merge base `5cc9a715` touches 42 files, **all** `.rs` /
`Cargo.toml` / `Cargo.lock` / `dependabot.yml` — zero fixture, golden, snapshot or
test-data files. Counts across the whole tree, base vs head:

| Marker | base | head |
|---|---|---|
| `#[test]` | 3389 | **3389** |
| `#[tokio::test]` | 269 | **269** |
| `proptest!` | 12 | **12** |
| `#[ignore` | 40 | **40** |

Nothing deleted, nothing newly ignored.

### 4. Independently run locally (aarch64-darwin, nightly-2025-12-03)

- `cargo build --workspace` → **EXIT=0**, 0 errors (10m59s).
- `cargo test -p bth-crypto-keys -p bth-crypto-ring-signature -p bth-account-keys -p bth-transaction-core -p bth-crypto-digestible -p bth-transaction-types -p bth-cluster-tax` → **EXIT=0, 789 passed / 0 failed**:
  crypto-keys 30 + **12 doctests** (incl. the seeded x25519 fingerprint goldens),
  ring-signature 51, digestible 10 (`tests/basic.rs`), transaction-core 99,
  transaction-types 69, cluster-tax 504 + 5 doctests, account-keys 9.
- `cargo test -p bth-account-keys --all-features` → 20 passed / 0 failed.
- `cargo deny check` → **`advisories ok, bans ok, licenses ok, sources ok`** (only
  pre-existing `advisory-not-detected` warnings). `[bans] multiple-versions = "warn"`
  plus the existing `skip-tree` on `bulletproofs-og`/`schnorrkel-og` absorbs the
  deliberate v4/v5 coexistence without a new exception.

### 5. Profile exemption still valid

`curve25519-dalek 5.0.0`'s `Cargo.toml` still gates fiat-crypto behind
`[target.'cfg(curve25519_dalek_backend = "fiat")']`, and `build.rs` only selects
`"fiat"` from an explicit env/cfg override — the default remains serial/simd. So
`[profile.release.package.curve25519-dalek] overflow-checks = false` still covers the
same audited constant-time limb arithmetic the #663 comment describes, and now
applies to both in-tree majors (bare package name matches all versions), which is the
desired outcome.

Also correct and easy to miss: the workspace dep gained `features = ["zeroize"]`.
Under dalek 4 that feature arrived by unification from `bulletproofs-og`/`schnorrkel-og`;
dalek 5 is a *separate package* so that unification no longer applies, and
`RistrettoPrivate`'s `#[derive(Zeroize)]` needs it. Good catch.

### 6. Mergeability / CI

- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE` against current `main`
  (`07b3282c`, i.e. after #1185 and #1184).
- `git merge-tree --write-tree HEAD origin/main` → rc=0, no conflicts. I rendered the
  merged `.github/dependabot.yml` and read it: the cargo `ignore:` block is coherent —
  the kept `curve25519-dalek` major ignore sits under its new rationale comment, the
  four retired dalek ignores are gone, and the npm blocks from #1185/#1184 are intact.
- `gh pr checks 1188`: **12 pass, 0 fail**, 5 `skipping` (the fuzz-run jobs, expected
  on PRs; `Fuzz Build Guard (PR)` itself passes).

### Non-blocking nits (no action required before merge)

1. **PR body test count**: "account-keys 69" is wrong — account-keys is 9 (20 with
   `--all-features`); 69 is the transaction-types number, listed twice. Purely a
   description typo; the global `#[test]` census above proves nothing was lost.
2. **`random_scalar`/`scalar_from_hash` is now implemented four times**
   (`crypto/ring-signature/src/compat.rs`, `cluster-tax/src/crypto/dalek_compat.rs`,
   `crypto/keys/src/ristretto.rs`, `core/src/subaddress.rs`, plus the bench copy).
   Defensible — collapsing them would force `cluster-tax`/`core` to take new
   inter-crate deps — but it is four copies of consensus-relevant key-derivation
   code, so worth a note for whoever does #1154.
3. **dependabot scope**: retiring the `ed25519` / `ed25519-dalek` / `signature` /
   `x25519-dalek` major ignores means a future major of any of those will be proposed
   alone and be guaranteed-red until `curve25519-dalek` (still ignored) can follow.
   Realistically years out and the kept comment is honest about the pin, so not worth
   holding the PR — just flagging that the family can now drift in dependabot's view.

### Verdict

The bit-exactness claim is the whole ballgame here, and it holds under
source-level scrutiny at every point I could find: the two bridges are canonical
byte round-trips, every replaced convenience constructor is character-identical to
what it replaced, Ed25519(ph) acceptance semantics are unchanged, and the seeded
x25519 goldens that would catch a byte-stream regression are untouched and green.
Build, targeted tests, and `cargo deny` all pass locally; CI is fully green; the
branch merges clean. **Approved.**

PR #1193, opened 2026-08-11
chore(deps): migrate rand family to the rand_core 0.10 line (rand 0.10 / rand_chacha 0.10 / rand_hc 0.5 / rand_distr 0.6 / getrandom 0.4)

### LINKED ISSUE #1154: Rand-family migration: rand_core 0.6 → 0.10 line (rand / rand_chacha / rand_hc / rand_distr / getrandom in lockstep)

rjwalters:
## Summary

Migrates the entire workspace off the rand-0.8 / rand_core-0.6 generation to the newest coherent rand-family set, in lockstep, with deterministic key derivation verified **bit-identical** (no fixture or golden was regenerated).

**Resolved version set** (newest coherent family on crates.io; note the issue's hypothetical "rand 0.9 / rand_core 0.9 vs rand_hc 0.5 / rand_core 0.10" conflict does not arise — a full 0.10 line exists):

| crate | before | after |
|---|---|---|
| rand | 0.8.5 | **0.10.2** (requires rand_core ^0.10) |
| rand_core | 0.6.4 | **0.10.1** |
| rand_chacha | 0.3.1 | **0.10.0** |
| rand_hc | 0.3.2 | **0.5.1** (requires rand_core ^0.10) |
| rand_distr | 0.4.3 | **0.6.0** (requires rand ^0.10) |
| getrandom | 0.2 | **0.4.3** |

## Changes

- **Workspace `Cargo.toml`**: family bumped as above. The dalek-5-era `rand-core-v010` bridge alias is retired; a `rand-core-v06` alias replaces it for the opposite direction (externals still on 0.6).
- **Determinism-critical verification (HC-128)**: diffed rand_hc 0.3.2 vs 0.5.1 sources — the HC-128 core algorithm is byte-for-byte unchanged (only formatting + trait plumbing), the eSTREAM known-answer test vectors (including the `u64` and byte-stream vectors that pin word-combining/serialization order) are identical, and rand_core 0.10.1's `BlockRng` has identical `next_u32`/`next_u64` (lo|hi<<32)/`fill_bytes` (LE words, partial-word discard) semantics to 0.6.4's. Seed loading (`utils::read_words` vs `le::read_u32_into`) is identical LE.
- **Trait/API migration**: `RngCore + CryptoRng` bounds → `CryptoRng` (0.10's `CryptoRng` subsumes the infallible `Rng` methods); `CryptoRngCore` → `CryptoRng`; `.gen()/.gen_range()/.gen_bool()` → `.random()/.random_range()/.random_bool()` via `RngExt`; `thread_rng()` → `rand::rng()`; `rand::distributions` → `rand::distr` (`WeightedIndex` → `distr::weighted`); `SliceRandom::choose` → `IndexedRandom::choose`; `getrandom::getrandom()` → `getrandom::fill()`; `StandardUniform` no longer samples `usize` (two call sites rewritten).
- **OsRng replacement**: rand_core 0.10 has no `OsRng`. Added `bth_util_from_random::OsRng` — a zero-sized getrandom-backed infallible `CryptoRng` with the old panic-on-failure semantics — and migrated all former `rand_core::OsRng` / `rand::rngs::OsRng` users (argon2's own `password_hash::rand_core::OsRng` re-export untouched).
- **Shim collapse (per #1188 follow-up)**: `crypto/keys/src/x25519.rs`'s `RngCompat` (0.6→0.10 bridge) is deleted — x25519-dalek 3's `random_from_rng` now takes workspace rngs directly, drawing the same 32 bytes via `fill_bytes`.
- **New reverse shim for the dalek-4 externals**: `crypto/keys/src/compat.rs` adds `Rng06Compat` (0.10→0.6, direct method forwarding = identical byte stream) used at exactly two boundaries: schnorrkel-og `attach_rng` (deterministic `sign_schnorrkel`, seeded by the same merlin nonce → same Hc128 stream → same signatures) and bulletproofs-og `prove/verify_multiple_with_rng` in `transaction/core/src/range_proofs`.
- **Unchanged 0.6-line islands (deliberate)**: `crypto/pq` keeps its private rand_core 0.6 / rand_chacha 0.3 / rand 0.9 pins for the ml-kem/ml-dsa boundary; `mc-rand` (`common/hasher_builder.rs`) now imports `RngCore` from mc-rand's own re-export.
- **wasm**: `getrandom` 0.2 `"js"` feature sites → workspace 0.4 `"wasm_js"` (backend cfg already in `.cargo/config.toml`); wasm-signer keeps renamed getrandom 0.2/0.3 deps for crypto/pq's transitive generations. `cargo check -p bth-wasm-signer --target wasm32-unknown-unknown` and `-p bth-transaction-clsag` pass.
- **fuzz workspace** (excluded from main workspace, path-deps on migrated crates): bumped to rand 0.10 / rand_chacha 0.10; `cargo check --features native-smoke` green (all fuzz targets type-check).
- **`.github/dependabot.yml`**: removed the six rand-family ignore entries (rand / rand_core / rand_chacha / rand_hc / rand_distr / getrandom) that were waiting on this migration.

- [x] rand_core 0.6 → 0.10 line, with rand / rand_chacha / rand_hc 0.5 / rand_distr / getrandom in lockstep — see table above
- [x] Derivation BIT-IDENTICAL: **no golden/fixture regenerated or modified**. The seeded Hc128Rng x25519 fingerprint goldens (`0b:5e:58:…`) pass unchanged (doctests), as do all seeded derivation/signing suites
- [x] `gen_range`/distribution call sites audited for rand-0.9+ semantics (float/int ranges unchanged; the two `random::<usize>()` sites — wallet mnemonic word check, circuit selection — rewritten since `StandardUniform` dropped `usize`; both were non-deterministic-by-design uses)
- [x] Dependabot ignore block removed

## Test Plan

- [x] `cargo build --workspace` — green
- [x] `cargo check --workspace --all-targets` (incl. benches) — green, 0 errors; no new rand-related warnings (the unused-`Rng`-import cases my sweep introduced were cleaned up)
- [x] `cargo test -p bth-crypto-keys -p bth-crypto-ring-signature -p bth-account-keys -p bth-transaction-core -p bth-crypto-multisig -p bth-crypto-digestible` — **219 passed, 0 failed** (30+51+9+99+8+10 lib + 12 doctests)
- [x] `cargo test -p bth-transaction-core --all-features` (CI job) — 120 passed
- [x] `cargo test --workspace --no-run` and `cargo test --workspace --doc` (the two workspace-build.yml gates) — green (44 doc suites ok)
- [x] Broader `cargo test --workspace --no-fail-fast`: 44+ suites ok; the **only** failing targets are `compact_block_integration`, `e2e_consensus_integration`, `e2e_progressive_fees`, `e2e_transfer_patterns` — exactly the 4 targets documented as failing on unmodified main in #1188/#1187 (pre-existing, not a regression)
- [x] Remaining packages individually (`bth-common`, consensus, crypto, transaction, util, gossip, wasm-signer, …): 55 suites ok, 0 failed
- [x] `cargo test -p bth-cluster-tax` (long sim suite): **509 passed, 0 failed**, rc=0
- [x] `cargo deny check` — advisories ok, bans ok, licenses ok, sources ok
- [x] `cargo fmt --check` — clean (workspace + fuzz)
- [x] wasm32 target checks for `bth-wasm-signer` and `bth-transaction-clsag` — green

TDD: N/A — dependency migration with a bit-identical-behavior requirement; correctness is gated on the pre-existing golden/seeded suites passing unchanged (they do, none modified) rather than new tests.

Closes #1154


rjwalters:
## Judge review: APPROVED

Reviewed as a consensus/wallet-critical change. The load-bearing claim — **the seeded RNG byte stream is unchanged** — is not just asserted here, it is verifiable, and I verified it independently at the source level rather than relying on the green test run.

### 1. HC-128 byte-stream identity (the central risk) — CONFIRMED

Diffed the registry sources directly rather than trusting the changelog.

**`rand_hc` 0.3.2 → 0.5.1 (`src/hc128.rs`)** is formatting + trait plumbing only:
- `step_p` / `step_q` / `h1` / `h2` / `init` / `generate` are character-for-character identical modulo `rustfmt` reflow (`cc+0` → `cc + 0`) and one `wrapping_add` chain rejoin.
- Trait changes only: `BlockRngCore` → `Generator`, `RngCore` → `TryRng` (`next_u32`→`next_word`, `next_u64`→`next_u64_from_u32`), `CryptoRng` → `TryCryptoRng`, and `BlockRng::<Hc128Core>::from_seed(seed)` → the explicit `BlockRng::new(Hc128Core::from_seed(seed))` it always desugared to.
- The eSTREAM known-answer test module is unchanged apart from its `use` line.

**Seed decode** — `le::read_u32_into` (0.6) vs `utils::read_words` (0.10): both little-endian `u32`, same word order, same `chunks_exact`. Equivalent.

**`BlockRng` output plumbing, rand_core 0.6.4 vs 0.10.1** — checked all three of the semantics that could silently reorder the stream:
- `next_u32`/`next_word`: identical (regenerate at `index >= N`, take `results[index]`, advance one).
- `next_u64`/`next_u64_from_u32`: identical in **all three** index branches (`index < N-1`, `index == N-1`, `index >= N`) — same `lo`/`hi` word order, same `(hi << 32) | lo` assembly, same resulting index (2 / 1 / 2 respectively).
- `fill_bytes`: identical LE word emission, and identically consumes a **whole** word when the trailing word is only partially used (0.6 via `num_chunks = ceil(byte_len/size)`, 0.10 via the `index += 1` in the `dest_rem` arm). 0.10's "store the index in `results[0]`" trick never loses a live word, because `index == 0` is only ever transient *within* a single call — every call returns with `index >= 1`, at which point `results[0]` is already consumed.

**`SeedableRng::seed_from_u64`** (which is what actually picks the seed in the golden): the default impl moved files but the PCG32 constants are bit-identical — verified `0x5851_F42D_4C95_7F2D == 6364136223846793005` and `0xA176_54E4_6FBE_17F3 == 11634580027462260723`.

**Empirical pin.** The `0b:5e:58:80:…` fingerprint goldens in `crypto/keys/src/x25519.rs` (lines 325 and 347) are textually untouched by the diff — I confirmed against the merge base, not just by eye — and both doctests **ran and passed**:
```
test crypto/keys/src/x25519.rs - x25519::X25519Public::from (line 315) ... ok
test crypto/keys/src/x25519.rs - x25519::X25519Public::from (line 337) ... ok
```
Those derive an X25519 pubkey from `Hc128Rng::seed_from_u64(0)` via `fill_bytes`, so they are exactly the cross-version fixed point. Worth noting the net is much wider than those two: `bth_util_test_helper::get_seeded_rng()` is `Hc128Rng::from_seed([7u8; 32])`, so essentially the entire seeded test corpus would break on any stream drift.

### 2. The new `Rng06Compat` reverse bridge — CONFIRMED

`crypto/keys/src/compat.rs` is pure forwarding: `next_u32`/`next_u64`/`fill_bytes` delegate straight through, `try_fill_bytes` is `fill_bytes` + `Ok(())`. No internal buffer, no re-chunking, so draw order and word boundaries are preserved. Both call sites construct it inline so it cannot accumulate hidden state across calls:
- `ristretto.rs` — `attach_rng(t, Rng06Compat(&mut csprng))` for schnorrkel-og deterministic signing
- `range_proofs/mod.rs` — `&mut Rng06Compat(rng)` on both `prove` and `verify` for bulletproofs-og

The direction is also right: this is the mirror of the `RngCompat` bridge #1188 added, which correctly collapses now that x25519-dalek 3 takes workspace RNGs directly.

### 3. `OsRng` replacement — CONFIRMED

`util/from-random`'s new `OsRng` draws from `getrandom::fill` (OS entropy), not a seeded or thread RNG, and panics on source failure — matching the retired `rand_core 0.6` `OsRng` semantics exactly. I checked **every** replacement site in the diff: all ~25 are 1:1 (`rand_core::OsRng` / `rand::rngs::OsRng` → `bth_util_from_random::OsRng`). No site was silently downgraded to `rand::rng()` or a seeded RNG, and no new site adopted `OsRng` where it wasn't already used.

### 4. rand 0.9/0.10 renames — no semantic drift

I diffed the **multiset** of range expressions (count + text) between removed `gen_range(...)` and added `random_range(...)` lines. Exact match, so inclusivity is preserved at every site (`0..n` stayed exclusive, `0..=n` stayed inclusive). Two legitimate non-mechanical rewrites, both forced and both benign:
- `circuit.rs`: `rand::random::<usize>() % active_circuits.len()` → `rand::rng().random_range(0..len)`. Forced — rand 0.9 removed `usize` from `StandardUniform` ([#1487]). Also removes the modulo bias, so a strict improvement.
- `commands/init.rs`: `random::<usize>() % 24` → `random::<u32>() as usize % 24`. Same cause; it's the "type word #N of your seed phrase" UX quiz, not security randomness.

On the **`Uniform` value-stability break** (rand 0.9 switched to Canon/Lemire): I checked this does not touch anything consensus-agreed. Every `random_range` / `shuffle` / `choose` / `WeightedIndex` site is either `#[cfg(test)]`, an offline simulation (cluster-tax, quorum-sim), or driven by non-deterministic `rand::rng()`/`OsRng` (decoy selection, cover traffic, relay selection). The SCP shuffles in `consensus/scp/src/msg.rs` are test-only. Key derivation goes through `fill_bytes`, never `Uniform`, so it is untouched by that break.

One float note, non-blocking: rand 0.9 also allows `UniformFloat` to yield the `high` bound ([#1462]). The only production float range is `botho/src/network/privacy/selection.rs:381` (`random_range(0.0..total_weight)` in `weighted_random_select`), which already has a `candidates.last().copied()` fallback, so the newly-possible `value == total_weight` case is handled correctly. The adjacent comment `// Generate random value in [0, total_weight)` is now marginally stale — worth a one-line touch-up if the file is edited again, not worth blocking.

### 5. Tests / fixtures — nothing weakened

Census against the merge base is **byte-identical**: 3389 `#[test]`, 269 `#[tokio::test]`, 40 `#[ignore]`, 10321 `assert`. Zero diffs to any fixture/golden/snapshot file — the only non-source changes are `Cargo.lock` and `.github/dependabot.yml`.

### 6. Independently run (macOS, this worktree)

| Check | Result |
|---|---|
| `cargo build --workspace` | exit 0 |
| `cargo test` (crypto-keys, ring-signature, account-keys, transaction-core, multisig, digestible) | 111 passed / 0 failed, incl. 12 `bth_crypto_keys` doctests |
| + `-p bth-cluster-tax` | **728 passed / 0 failed**, exit 0 |
| `cargo deny check` | advisories/bans/licenses/sources ok |
| `cargo check -p bth-wasm-signer --target wasm32-unknown-unknown` | exit 0 |
| `gh pr checks 1193` | **13/13 pass** |
| `mergeStateStatus` | **CLEAN** / MERGEABLE |

Also verified the wasm story specifically, since getrandom 0.2→0.4 is the classic place this breaks: the `getrandom_backend="wasm_js"` cfg in `.cargo/config.toml` is paired with the `wasm_js` feature on the workspace `getrandom`, and the renamed `getrandom_02` dep preserves the 0.2 `js`-feature path for the ml-kem boundary. The wasm32 check confirms it links.

### 7. Rebase / lockfile hygiene

Branch is directly on top of current `main` (merge base == `origin/main` == `2e2322e2`, #1188), so there is no merge to reason about. `.github/dependabot.yml` removes only the 6 rand-family ignores and **preserves #1188's `curve25519-dalek` major-version block**. In the lockfile, `rand_hc` collapses to a single `0.5.1` — no split HC-128 in the graph, which matters here — and `rand_core 0.6.4` survives only for the compat bridge plus `bip39`. `crypto/pq`'s private `rand_core-0.6` pins and `mc-rand` are untouched, as claimed.

### Non-blocking observations (no action required for merge)

1. `selection.rs:381` — the `[0, total_weight)` comment is now marginally inaccurate under rand 0.9+ float semantics (behaviour is safe either way, see §4).
2. `util/test-helper::random_bytes_vec` has a **pre-existing** bug unrelated to this PR: `Vec::with_capacity(n)` followed by `fill_bytes(&mut result)` fills a zero-length slice, so it always returns an empty `Vec`. Carried over unchanged; flagging it only so it isn't mistaken for migration fallout.
3. `util/from-random` is `#![no_std]` but now links `getrandom` unconditionally. Fine for every target this workspace actually builds (native + wasm32, both verified above); only relevant if a backend-less bare-metal target is ever added.
4. The `unused import: RistrettoPoint` warning in `range_proofs/mod.rs` is pre-existing on `main` — not introduced here.

Careful, well-documented migration. The comments left at each compat seam explain *why* the bridge exists and which direction it points, which is exactly what the next person touching this needs. Approving.

